PR #448, opened 2015-01-15
Add bibliographic import (VOV-1639)

mbklein:


app/assets/javascripts/dynamic_fields.js.coffee
```diff
@@ -17,7 +17,10 @@
 $ ->
   add_button_html = '<div class="input-group-btn"><button type="button" class="add-dynamic-field btn btn-success"><span class="glyphicon glyphicon-plus"></span></button></div>'
   remove_button_html = '<div class="input-group-btn"><button type="button" class="remove-dynamic-field btn btn-success"><span class="glyphicon glyphicon-minus"></span></button></div>'
+  import_button_html = '<div class="input-group-btn"><button type="submit" name="media_object[import_bib_record]" class="btn btn-success" value="yes">Import</button></div>'
 
+  $('#media_object_bibliographic_id').closest('div').append(import_button_html)
+
```
cjcolvar: Could we move this somewhere else since it doesn't really relate to the dynamic fields functionality?  As it is now this file is generic and I'd like to keep it that way and not tied to the resource description page.

app/models/resource_description_step.rb
```diff
@@ -19,8 +19,13 @@ def initialize(step = 'resource-description', title = "Resource description", su
 
   def execute context
     mediaobject = context[:mediaobject]
-    mediaobject.permalink = context[:media_object].delete(:permalink)
-    mediaobject.update_datastream(:descMetadata, context[:media_object])
+    populate_from_catalog = context[:media_object][:import_bib_record].present?
+    if populate_from_catalog
+      mediaobject.descMetadata.populate_from_catalog!(context[:media_object][:bibliographic_id])
+    else
+      mediaobject.permalink = context[:media_object].delete(:permalink)
```
cjcolvar: Why is the permalink not set if also populating from the catalog?  Could there be a case where you want to import from the catalog but also specify a new or different permalink?

app/assets/javascripts/dynamic_fields.js.coffee
mbklein: Yes, I'll move that elsewhere.

app/models/resource_description_step.rb
mbklein: The permalink isn't stored in the MODS; it's stored elsewhere. It's also specific to the Avalon version of the resource, which doesn't exist yet at import time, so the catalog wouldn't have a reason to know about it.

cjcolvar: I think I get it...it is included in the else because the step is getting called two times: import and save.  So the else clause should run on the save?

cjcolvar:
It would be nice to have an example config in avalon.yml.example.

PR #465, opened 2015-01-29
Fix Comments form

bkeese:


app/controllers/comments_controller.rb
```diff
@@ -23,12 +23,13 @@ def index
 	def create
 	  @comment = Comment.new
 	  @comment.name = params[:comment][:name]
+          @comment.nickname = params[:comment][:nickname]
```
mbklein: I'm trying hard not to care about the over-indentation here. :)

PR #471, opened 2015-02-04
Import bib data on batch ingest [VOV-1640]

mbklein:


mbklein:
@cjcolvar I found an errant `.xlsx` file in `spec/fixtures` that was causing my local failures. The weird thing is that it wasn't a local untracked file. It was present upstream in `develop`. But it's working now, and the feature is working, and *quick merge it before something happens*. :wink:

PR #475, opened 2015-02-09
Fix missing pagination for new session.

bkeese:


app/controllers/catalog_controller.rb
```diff
@@ -214,8 +214,8 @@ def save_sticky_settings
   end
 
   def apply_sticky_settings(solr_parameters, user_parameters)
-    solr_parameters[:rows] = session[:per_page]
-    solr_parameters[:sort] = session[:sort]
+    solr_parameters[:rows] = session[:per_page] if session[:per_page].present?
+    solr_parameters[:sort] = session[:sort] if session[:per_page].present?
```
cjcolvar: I'm guessing you mean session[:sort] here.

bkeese:
Thanks. Fixed.

PR #481, opened 2015-02-12
Add option to bulk unhide to modal

lealeelu:


app/views/bookmarks/update_access_control.html.erb
```diff
@@ -32,64 +32,67 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-	    <legend>Discovery</legend>
-	    <%= label_tag :hidden do %>
-	      <%= hidden_field_tag :hidden %>
-	      <%= check_box_tag :hidden_display, '1', (@media_objects.all? {|doc| doc.hidden? == true})%>
-	      Hide this item from search results 
-	    <% end %>
-	  </fieldset>
+      <legend>Discovery</legend>
+      <%= label_tag :hidden do %>
+        <%= hidden_field_tag :hidden %>
```
bkeese: I don't think this hidden field is useful.

app/models/media_object.rb
```diff
@@ -387,7 +387,7 @@ def access_control_bulk documents, params
       successes = []
       documents.each do |id|
         media_object = self.find(id)
-        media_object.hidden = params[:hidden] == "true" if params[:hidden].present?
+        media_object.hidden = params[:hidden]
```
bkeese: What if params[:hidden] isn't present? I think it would be more clear if this had the same form as the next line (visibility).

PR #485, opened 2015-02-18
Update year in license header

lealeelu:
Used this method:
https://wiki.dlib.indiana.edu/display/VarVideo/Update+License+Headers


cjcolvar:
I don't think we should add headers wholesale to config or db/migrate since some of those files are not ours.

lealeelu:
Then should I remove all headers of our headers from config and db/migrate?

cjcolvar:
@mbklein what are your thoughts?  Is it worth putting the copyright header on config files, initializers, and db migrations?  We already have some on migrations but none in config.

lealeelu:
I removed the headers from the config files since they weren't there to begin with. With the db/migrate files it looks like only four of them didn't have it to begin with and I recognize a couple that are probably written by us.

cjcolvar:
I think all of the migration scripts that are having license headers added shouldn't.  So only migrations with existing license headers should be updated and none added.

lealeelu:
@cjcolvar sounds good to me.

cjcolvar:
One last thing: could you remove the test directory?  We don't need it and it's not our code anyways so it shouldn't have headers.

lealeelu:
Updated again. How's that?

cjcolvar:
:+1: thanks for doing this tedious work!

PR #488, opened 2015-02-19
Explicitly set defType to edismax in bookmark solr queries

bkeese:
Overriding blacklight's bookmarks#index.


cjcolvar:
If defType=lucene queries bomb all the time, them maybe we should make edismax the default in CatalogController which should be inherited by BookmarksController.  This would be a lot cleaner than overriding #index and should help ensure we don't run into this in the future.  That all said, I haven't been able to make that work locally.  We can take a look at it tomorrow.

cjcolvar:
I think this is fine for now since we're pinned to a specific Blacklight revision.  Whenever we upgrade Blacklight/tomcat/solr we should revisit this.

PR #493, opened 2015-02-27
Fix 1px desktop/mobile size gap

bkeese:


cjcolvar:
:+1 I think this only affects app/views/modules/_header.html.erb so it should be relatively low risk.

PR #499, opened 2015-03-13
Fix for VOV-3446

paclough:
Update CSS for selected facet labels: float the label left and the
remove button to the right, so the remove button stays aligned
correctly.


paclough:
Test on Windows/Linux/Android, especially IE on Windows

PR #503, opened 2015-03-17
Prevent an ad hoc group name with bad punctuation

paclough:
Add validations to prevent ad hoc group names from containing a period,
percent symbol, semicolon or forward slash and flash an alert to the
user.


cjcolvar:
:+1:

PR #504, opened 2015-03-18
Prevent second process from removing lock of first process

bkeese:


cjcolvar:
Looks good as far as I can tell.  @mbklein, your thoughts?

PR #505, opened 2015-03-19
Feature/hydra8

cjcolvar:
Upgrade to Hydra8-RC1 ... we'll need to fix the Gemfile after hydra8 is released.


cjcolvar:
@bkeese @mbklein I've upgraded to Hydra 8 and got the tests passing...looks like I missed something in HH 8.1 so I'll have to make a HH 8.1.1 or HH 8.2 and upgrade Avalon to that, but for now this should be good to merge.

PR #509, opened 2015-04-02
Fix for items with only one section which has structure

bkeese:


app/views/media_objects/_sections.html.erb
```diff
@@ -14,7 +14,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 ---  END LICENSE_HEADER BLOCK  ---
 %>
 <% show_progress = sections.any? { |s| not s.succeeded? } %>
-<% unless (sections.blank? or sections.length == 1) and not show_progress %>
+<% unless (sections.blank? or (sections.length == 1 && sections.first.get_structural_metadata.xpath('//Item').empty?)) and not show_progress %>
```
cjcolvar: This looks like a scary long line for in a view partial.  Could this be put in a helper so it could be tested?

cjcolvar:
:+1  Feel free to merge this once test pass

PR #515, opened 2015-04-13
Feature/vov 3585

bkeese:


lib/avalon/batch/entry.rb
```diff
@@ -110,6 +110,13 @@ def self.offset_valid?( offset )
           true
         end
 
+      def self.attach_structure_to_master_file( master_file, filename )
+          structural_file = "#{filename}.structure.xml"
+          if File::exists? structural_file
```
cjcolvar: Why File::exists? instead of File.exists?  Is the former required by the class context of the method?

PR #518, opened 2015-04-16
Add oembed endpoint

bkeese:


cjcolvar:
Seems like this might better live in a decorator or helper since it isn't really about modifying the model or reading back fields but rather about a specific presentation of the model.

cjcolvar:
I think moving the route to point to MasterFilesController works for now but we may need to generalize it later to handle different models (MediaObject, playlist, clip. etc)

app/controllers/master_files_controller.rb
```diff
@@ -47,6 +47,37 @@ def embed
     end
   end
 
+  def oembed
+    if params[:url].present?
+      pid = params[:url].split('?')[0].split('/').last
+      mf = MasterFile.find(pid)
+      if mf.present?
+        width = params[:maxwidth] || EMBED_SIZE[:medium]
```
cjcolvar: @bkeese EMBED_SIZE isn't defined in the controller context and is throwing an exception

app/controllers/master_files_controller.rb
```diff
@@ -47,6 +47,37 @@ def embed
...
+        height = mf.is_video? ? (width.to_f/mf.display_aspect_ratio.to_f).floor : AUDIO_HEIGHT
```
cjcolvar: Probably same as above for AUDIO_HEIGHT

PR #519, opened 2015-04-16
Refresh token every 5 minutes

mbklein:


cjcolvar:
@mbklein I think you only added this to the embedded player.  Can you make sure the JS is included on media_objects/_item_view.html.erb?

PR #523, opened 2015-04-22
Add feedback when invalid structure file is uploaded

bkeese:
The @masterfile.structuralMetadata.save in line 104 is what fixed it.

### DIFF
--- a/app/models/structural_metadata.rb
+++ b/app/models/structural_metadata.rb
@@ -14,9 +14,21 @@
 
 class StructuralMetadata < ActiveFedora::Datastream
 
+  def parse_xml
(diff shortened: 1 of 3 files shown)


app/models/structural_metadata.rb
```diff
@@ -14,9 +14,21 @@
 
 class StructuralMetadata < ActiveFedora::Datastream
```
cjcolvar: Maybe a lot of this would be handled if StructuralMetadata subclassed NokogiriDatastream instead of just plain Datastream.

cjcolvar:
Closing since fixed with #530

PR #526, opened 2015-04-27
Feature/vov 3595

bkeese:


app/assets/javascripts/dynamic_fields.js.coffee
```diff
@@ -48,13 +48,27 @@ $ ->
...
+          current_textarea.after(new_input_group)
+          new_input_group.after(new_textarea)
+        else
+          current_input_group.after(new_input_group)
         
       $(document).on 'click', '.remove-dynamic-field', (e) ->
         e.preventDefault()
-        $(this).closest('.input-group').remove()
+        current_input_group = $(this).closest('.input-group')
+        $('#'+current_input_group.data('textarea')).remove()
```
cjcolvar: This line looks like it will try to remove "#undefined" if there is no textarea associated with the input group.  I think we should protect this line so it will only be run if there is a textarea defined.

app/models/media_object.rb
```diff
@@ -375,6 +384,8 @@ def to_solr(solr_doc = Hash.new, opts = {})
     all_text_values << solr_doc["language_sim"]
     all_text_values << solr_doc["physical_description_si"]
     all_text_values << solr_doc["date_sim"]
+    all_text_values << solr_doc["notes_sim"]
+    all_text_values << solr_doc["table_of_contents_sim"]
```
cjcolvar: Could you please add a test to catalogcontroller for these two fields?

PR #537, opened 2015-05-15
Make Raw XML the default view for structure editor

bkeese:


cjcolvar:
This looks nice and clean...probably worth turning into a PR to upstream.

cjcolvar:
+1 to merging when the build passes

PR #539, opened 2015-06-01
Use ActiveEncode to create encoding jobs [VOV-3407]

mbklein:
Fix input file in #process for simple case
Upgrade to newer active-encode
Cleanup
Poll ActiveEncode for progress while MasterFile is not #finished_processing?
Hook up progress bar to active-encode
Use active-encode for destroy and other cleanup of destroy/delete
Remove unused route
Update token security to eliminate dependence on mediapackage_id
Fix failing tests
Use newline-delimited authorization responses instead of space-delimited
Use _ssi instead of _tesim for the indexed stream path.
Fix batch ingest tests
Update active-encode to version that prototypes multiple-quality skip transcoding
[VOV-3722] Store encoder class with MasterFile; infer it from workflow name if it doesn't exist
Add migration to change masterfile status SUCCEEDED to COMPLETED and back

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,7 +63,8 @@
   #gem 'devise-guests'
   gem 'haml'
 
-  gem 'rubyhorn', git: "https://github.com/avalonmediasystem/rubyhorn.git", tag: 'v3.2'
+  gem 'active-encode', git: "https://github.com/projecthydra-labs/active-encode.git"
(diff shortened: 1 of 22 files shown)


mbklein:
@cjcolvar I squashed the feature/VOV-3407 branch into this one. It's still there if we want to continue making atomic commits or need to revert something out.

PR #541, opened 2015-06-04
Add an outline to clickable elements when they are in :focus

paclough:
Fix the facet remove icon CSS
Fix-up formatting in CSS

### DIFF
--- a/app/assets/javascripts/avalon.js
+++ b/app/assets/javascripts/avalon.js
@@ -47,4 +47,6 @@ $(document).ready(function() {
     })
     $('input[readonly], textarea[readonly]').attr("readonly", false);
   }
+
+  $('a').has('img, ul').addClass('block');
 });
--- a/app/assets/stylesheets/avalon.css.scss
+++ b/app/assets/stylesheets/avalon.css.scss
@@ -1,18 +1,18 @@
-/* 
+/*
  * Copyright 2011-2015, The Trustees of Indiana University and Northwestern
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/avalon.js
```diff
@@ -47,4 +47,6 @@ $(document).ready(function() {
     })
     $('input[readonly], textarea[readonly]').attr("readonly", false);
   }
+
+  $('a').has('img, ul').addClass('block');
```
cjcolvar: I don't understand why you're adding the class here in the javascript versus adding the class in the markup.  I can think of pros and cons to each approach but would favor less magic so I would prefer it in the html.  What are other people's thoughts, @mbklein @lealeelu @bkeese ?

paclough: The idea for using JS was to avoid needing to remember to add the class for future  changes, for what is mostly a visual fix. No problem changing it if others prefer it in the HTML.

bkeese: Too bad it can't be done with just css.  I'm okay with using js for this.

paclough: Agreed. This is a fix for browser cross-compatibility - WebKit doesn't expand a link to wrap it's contained content without using `inline-block` or `block`. FF doesn't support `inline-block` the same way as WebKit and using `block` universally on links has poor (visual) consequences.

PR #544, opened 2015-06-08
Fix height/width calculation for oembed

bkeese:


app/controllers/master_files_controller.rb
```diff
@@ -55,8 +55,8 @@ def oembed
       if mf.present?
         width = params[:maxwidth] || MasterFile::EMBED_SIZE[:medium]
         height = mf.is_video? ? (width.to_f/mf.display_aspect_ratio.to_f).floor : MasterFile::AUDIO_HEIGHT
-        maxheight = params['maxheight']
-        if maxheight.present? && height>maxheight.to_f
+        maxheight = params['maxheight'].to_f
```
cjcolvar: is this safe when maxheight isn't included in the params?  It doesn't look like it.  Maybe tests would be good for this method.

bkeese: It returns 0.0. (nil.to_f == 0.0).

PR #545, opened 2015-06-15
Add skip-to-content link and JS work-around

paclough:
Add 'content' id to player div and update scripts

### DIFF
--- a/app/assets/javascripts/avalon.js
+++ b/app/assets/javascripts/avalon.js
@@ -49,4 +49,14 @@ $(document).ready(function() {
   }
 
   $('a').has('img, ul').addClass('block');
+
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/avalon.js
```diff
@@ -48,5 +48,15 @@ $(document).ready(function() {
     $('input[readonly], textarea[readonly]').attr("readonly", false);
   }
 
+  window.addEventListener("hashchange", function(event) {
```
paclough: This is a work-around for WebKit

cjcolvar:
:+1:  @bkeese @lealeelu @mbklein What do you think?

bkeese:
Looks good to me.

PR #549, opened 2015-06-17
Add a title attribute to the embed link

paclough:


app/models/master_file.rb
```diff
@@ -256,7 +256,7 @@ def embed_code(width, permalink_opts = {})
         url = embed_master_file_path(self.pid, only_path: false, protocol: '//')
       end
       height = is_video? ? (width/display_aspect_ratio.to_f).floor : AUDIO_HEIGHT
-      "<iframe src=\"#{url}\" width=\"#{width}\" height=\"#{height}\" frameborder=\"0\" webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>"
+      "<iframe title=\"#{self.mediaobject.title}\" src=\"#{url}\" width=\"#{width}\" height=\"#{height}\" frameborder=\"0\" webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>"
```
cjcolvar: Do we really want to display the MediaObject title?  Seems like we might want to make a method which joins the MediaObject title and the MasterFile label (or filename if label is not present).

PR #551, opened 2015-06-23
Fix batch ingest of single-value fields

bkeese:
We had previously decided to let it fail if an array of values was passed for a single-value field. Batch passes arrays for every field and my attempts to change that caused other errors. This fixes it while keeping the failure behavior we wanted.


app/models/media_object.rb
```diff
@@ -319,6 +319,8 @@ def update_attribute_in_metadata(attribute, value = [], attributes = [])
     else
       values = if self.class.multiple?(attribute)
         Array(value).select { |v| not v.blank? }
+      elsif Array(value).length==1
+        Array(value).first()
```
cjcolvar: It would be cleaner to drop the parans on #first.

cjcolvar:
This masks the ActiveFedora behavior to throw an error when attempting to set a single-valued field with a multiple values.  This might make sense here but we should be really cautious and think about how to get rid of these update_attribute methods.

@mbklein what are your thoughts on this?

bkeese:
This will still throw an error if more than value is sent.  It only converts arrays of length == 1.

cjcolvar:
You're right.  Nevermind.

PR #552, opened 2015-06-25
VOV-3707 & VOV-3794

mbklein:
This is a combined PR for the work on active-encode and the creation of a more flexible storage structure for derivatives.


mbklein:
Hold off; I just realized I need to add something.

app/models/derivative.rb
```diff
@@ -72,22 +69,28 @@ def self.from_output(output, opts={})
     derivative.encoding.video.video_codec = output[:video_codec]
     derivative.encoding.video.resolution = "#{output[:width]}x#{output[:height]}" if output[:width] && output[:height]
 
-    derivative.location_url = output[:url]
-    derivative.absolute_location = File.join(opts[:stream_base], Avalon::MatterhornRtmpUrl.parse(derivative.location_url).to_path) if opts[:stream_base]
+    derivative.hls_track_id = hls_output[:id]
+    derivative.hls_url = hls_output[:url]
```
bkeese: Is this safe? Will there always be an hls_output?

app/models/derivative.rb
```diff
@@ -55,12 +53,11 @@ class Derivative < ActiveFedora::Base
...
   end
 
-  def self.url_handler
-    url_handler_class = Avalon::Configuration.lookup('streaming.server').to_s.classify
-    @url_handler ||= UrlHandler.const_get(url_handler_class.to_sym)
-  end
+  def self.from_output(dists, opts={})
+    #output is an array of 1 or more distributions of the same derivative (e.g. file and HLS segmented file)
+    hls_output = dists.delete(dists.find {|o| o[:url].ends_with? "m3u8" })
+    output = dists.first || hls_output
```
bkeese: So, output will only be set to hls_output if no others exist. This seems like it might not be correct, based on the comment above.

db/migrate/20150430131441_r3_content_to_r4.rb
```diff
@@ -1,27 +0,0 @@
-class R3ContentToR4 < ActiveRecord::Migration
```
bkeese: Isn't this migration still required?

mbklein:
Sorry, @bkeese, I bullied @paclough into a merge before I saw your comments. ;-) Yes, the migration is still required; I changed the timestamp to get it to re-run with the new content instead of creating a new partial migration.

PR #554, opened 2015-06-29
Fix issues with derivative object generation after encode successfully finishes processing

cjcolvar:


app/models/master_file.rb
```diff
@@ -293,7 +293,7 @@ def update_progress_with_encode!(encode)
     self.duration = encode.tech_metadata[:duration] if encode.tech_metadata[:duration]
     self.file_checksum = encode.tech_metadata[:checksum] if encode.tech_metadata[:checksum]
     self.workflow_id = encode.id
-    self.workflow_name = encode.options[:preset]
+    #self.workflow_name = encode.options[:preset] #MH can switch to an error workflow
```
cjcolvar: I don't understand why github is showing this change here since it is already present in develop.

app/models/derivative.rb
```diff
@@ -69,8 +69,10 @@ def self.from_output(dists, opts={})
     derivative.encoding.video.video_codec = output[:video_codec]
     derivative.encoding.video.resolution = "#{output[:width]}x#{output[:height]}" if output[:width] && output[:height]
 
-    derivative.hls_track_id = hls_output[:id]
-    derivative.hls_url = hls_output[:url]
+    if hls_output
```
cjcolvar: @bkeese :+1:

PR #557, opened 2015-07-07
Feature/vov 2771 - Add LTI URL share tab using conditional partials

bkeese:


cjcolvar:
:+1:
@mbklein Could you give a look over this and merge if it looks good?

app/controllers/media_objects_controller.rb
```diff
@@ -17,11 +17,24 @@
 class MediaObjectsController < ApplicationController 
   include Avalon::Workflow::WorkflowControllerBehavior
   include Avalon::Controller::ControllerBehavior
+  include ConditionalPartials
 
 #  before_filter :enforce_access_controls
   before_filter :inject_workflow_steps, only: [:edit, :update]
   before_filter :load_player_context, only: [:show, :show_progress]
```
mbklein: Can the shared `if` proc be assigned to a lambda and reused below, both for clarity and maintainability in case the rules change?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -64,6 +64,8 @@ def find_user(auth_type)
 
     if request['media_object_id']
       redirect_to media_object_path(request['media_object_id'])
+    elsif request['section_id']
+      redirect_to masterfile_path(request['section_id'])
```
mbklein: What if we replace both `media_object_id` and `section_id` with `target_id` and figure out which class type to route to on dispatch? That might even let us create a nicer route (say, `/courselink/:target_id`, that routes to `users/omniauth_callbacks {:action=>"lti",:target_id=>target_id}`)

PR #558, opened 2015-07-07
Make section title a link instead of toggling structure

bkeese:


app/helpers/media_objects_helper.rb
```diff
@@ -178,8 +178,8 @@ def structure_html section, index, show_progress
        # If there are no subsections within the structure, return just the header with the single section
        else
          tracknumber = index
-         wrapperopen = "#{headeropen}<span><ul>"
-         wrapperclose = "</ul></span>#{headerclose}"
+         wrapperopen = "#{headeropen}<a href="#{share_link_for( section )}"><span><ul>"
```
cjcolvar: I think you need to escape the double quotes for the href.

PR #577, opened 2015-07-23
Fix for facets that are too long and unbroken to hyphenate on Chrome

mbklein:


cjcolvar:
@bkeese You should look this over since you've done a lot of work on this previously for avalon and in your PR to blacklight: https://github.com/projectblacklight/blacklight/pull/1224

bkeese:
Yes. I'm afraid this will break IE. I think the changes in my blacklight PR offer a good solution to this.

PR #579, opened 2015-07-28
Add outline switch to each element and ensure outline disappears on blur

paclough:


cjcolvar:
Could you please remove these now unused lines? https://github.com/avalonmediasystem/avalon/blob/bugfix/VOV-3941/app/assets/javascripts/avalon.js#L127-L133

paclough:
Whoops ... done

PR #586, opened 2015-07-29
Remove /courselink route

mbklein:
revert to /users/auth/lti/callback for all links


mbklein:
Thanks, @bkeese; I rebased, added that change, and force pushed.

PR #602, opened 2015-08-12
Set player outline width to 1px, point Gemfile to updated skin branch

paclough:


paclough:
Requires pull request from mediaelement-skin-avalon

PR #605, opened 2015-08-18
Subvert all normal callbacks the update the player UI until after play for Android devices

cjcolvar:
These changes override the methods used for updating the progress bar and current time until first play on Android devices.  This makes the UI look responsive to structure clicks and media fragment urls, but does not address slider clicks.  It also does not update the still image for the video but leaves the poster image in place.  This approach was taken because Android isn't properly handling seeks and seek events.


bkeese:
:+1:

mbklein:
Gross but :+1: :grinning:

PR #607, opened 2015-08-20
Don't show player until skin and media are initialized.

mbklein:
MediaElement.js doesn't have a `created` event, so I set up a timeout loop that watches for `player.created` to become true and triggers an event. That allowed me to initialize the player with `visibility: hidden` and show it after the initial heavy lifting is done.

### DIFF
--- a/app/assets/javascripts/switch_streams.js
+++ b/app/assets/javascripts/switch_streams.js
@@ -47,6 +47,17 @@
       }
     },
 
+    setupCreationTrigger: function(playerThing) {
+      var watchForCreation = function() {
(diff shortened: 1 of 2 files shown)


bkeese:
Do we still need the loadedmetadata handler here?
https://github.com/avalonmediasystem/avalon/blob/bugfix/hide-native-player/app/views/media_objects/_item_view.html.erb#L110

mbklein:
I don't think so, but I'm not 100% sure. In fact, the way it's coded, I'm not sure it's ever even firing.

mbklein:
Actually, after some testing, it seems we do still need the `loadedmetadata` event handler in the view.

bkeese:
:+1:

PR #610, opened 2015-08-24
Add Media Object migration to R4

bkeese:


db/migrate/20150824155143_r3_media_object_to_r4.rb
```diff
@@ -0,0 +1,27 @@
+class R3MediaObjectToR4 < ActiveRecord::Migration
+  def up
+    say_with_time("R3->R4") do
+      prefix = Avalon::Configuration.lookup('fedora.namespace')
+      ActiveFedora::Base.reindex_everything("pid~#{prefix}:*")
```
cjcolvar: Is this reindex really necessary?  What purpose does it serve?

bkeese: Because identifiers are supposed to be searchable.

bkeese: I'm not sure. I'll guess I was just following the example of a previous migration.

cjcolvar:
:+1:

PR #611, opened 2015-08-25
Don't prepend the prefix if it is already there

cjcolvar:
This fix was put in place on IU's production instance in May but the change never made it upstream.


bkeese:
:+1:

PR #612, opened 2015-08-26
Encapsulated Avalon player with leaner stream switching and seeking behavior

mbklein:


cjcolvar:
:+1: for merging now and doing another round of testing on all test machines

PR #616, opened 2015-09-04
Coveralls config

carrickr:
This adds in coveralls support and badging for the repo.  The badge is currently 'unknown' in that it is targeting stats for the master branch, which currently has no stats.

This also updates the travis.yml file to build against 2.1.7 in addition to the other ruby versions.


cjcolvar:
I think 2.1.7 should replace 2.1.5 in the travis config.  Also please point this PR at develop instead of master.

PR #621, opened 2015-09-30
Add rake task for set_streaming_locations

bkeese:


cjcolvar:
:+1:

@mbklein Could you review this?  I'm unsure if it is safe with respect to your work on storage locations?

PR #622, opened 2015-10-05
Remove bookmarks when user loses access or is destroyed

bkeese:


spec/controllers/media_objects_controller_spec.rb
```diff
@@ -417,6 +417,19 @@
...
+  describe "#save" do
+    it 'removes bookmarks that are no longer viewable' do
+      media_object = FactoryGirl.create(:published_media_object)
+      user = FactoryGirl.create(:public)
+      bookmark = Bookmark.create(document_id: media_object.pid, user_id: user.id) 
+      expect(Bookmark.exists? bookmark.id).to be_truthy
+      login_user media_object.collection.managers.first
+      request.env["HTTP_REFERER"] = '/'
+      get 'update_status', id: media_object.pid, status: 'unpublish'
+      expect(Bookmark.exists? bookmark.id).to be_falsey
```
cjcolvar: Looks like this works and would simplify the expects:

      expect {
        get 'update_status', id: media_object.pid, status: 'unpublish'
      }.to change { Bookmark.exists? bookmark }

app/models/media_object.rb
```diff
@@ -546,4 +548,9 @@ def update_permalink_and_dependents
       true
     end
   
+    def remove_bookmarks
+      Bookmark.where(document_id: self.pid).each do |b|
+        b.destroy if ( !User.exists? b.user_id ) or ( Ability.new( User.find b.user_id ).cannot? :read, self )
```
cjcolvar: I think the second clause here needs to be reworked since a user could have access to an item based upon their session info (virtual_groups, IP, etc.) that isn't being taken into account here.  The result could be you login via LTI and bookmark, the object gets saved and you lose your bookmark, you log back in via LTI and are missing your bookmark.

PR #623, opened 2015-10-07
Add style to mejs cannot play message

bkeese:


cjcolvar:
Does this require a new version of the mediaelement_rails gem in the Gemfile (or Gemfile.lock)?

PR #624, opened 2015-10-10
Quick-load MasterFiles and Derivatives

mbklein:
This change became more involved than I expected, because `ActiveFedora::SimpleDatastream` simply doesn't provide a mechanism to determine the correct solr field name for a given attribute field. I created a `CachingSimpleDatastream` wrapper that translates attribute names to solr field names, but stays completely out of the way otherwise. It works as a drop-in with existing data, using the field definition blocks from `has_metadata`, to gather its information, and there is no reindexing needed.


mbklein:
Closed in favor of #669

PR #628, opened 2015-11-09
Update MARC21slim2MODS3-5-avalon.xsl

jlhardes:
Added check that controlField008-35-37 variable is not set to 'N/A' from old cataloging practices

### DIFF
--- a/lib/avalon/bib_retriever/MARC21slim2MODS3-5-avalon.xsl
+++ b/lib/avalon/bib_retriever/MARC21slim2MODS3-5-avalon.xsl
@@ -1414,8 +1414,8 @@ Revision 1.02 - Added Log Comment  2003/03/24 19:37:42  ckeith
 
 		<!-- language 041 -->
 		<xsl:variable name="controlField008-35-37"
-			select="normalize-space(translate(substring($controlField008,36,3),'|#',''))"/>
-		<xsl:if test="$controlField008-35-37 != 'N/A'">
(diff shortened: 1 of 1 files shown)


lib/avalon/bib_retriever/MARC21slim2MODS3-5-avalon.xsl
```diff
@@ -1414,7 +1415,7 @@ Revision 1.02 - Added Log Comment  2003/03/24 19:37:42  ckeith
 		<!-- language 041 -->
 		<xsl:variable name="controlField008-35-37"
 			select="normalize-space(translate(substring($controlField008,36,3),'|#',''))"/>
-		<xsl:if test="$controlField008-35-37">
+		<xsl:if test="$controlField008-35-37 != 'N/A'">
```
bkeese: Should this test be in the the variable declaration? Something like this on line 1417:
<xsl:variable name="controlField008-35-37"
			select="normalize-space(translate(substring($controlField008,36,3),'|#','')) != 'N/A' "/>

I'm afraid that putting the test in lin 1418 will lead to false positives (when the field isn't found, field!='N/A' will be true).

jlhardes: Thank you, that is a better place for that test. I'll move it and commit again.

jlhardes:
This should be better now.

PR #632, opened 2015-11-24
Add ingest API get methods for media objects

bkeese:


app/controllers/media_objects_controller.rb
```diff
@@ -95,8 +95,13 @@ def custom_update
     flash[:notice] = @notice
   end
 
+  def index
+    respond_to do |format|
+      format.json   { render json: MediaObject.all.map{|mo| mo.to_json } }
```
cjcolvar: I think we need to introduce pagination here and maybe back on collection#index as well.

app/models/media_object.rb
```diff
@@ -417,8 +417,14 @@ def to_solr(solr_doc = Hash.new, opts = {})
...
     {
-      id: self.pid,
-      title: self.title
+      id: pid,
+      title: title,
+      collection: collection.name,
+      main_contributors: creator, 
+      publication_date: date_created, 
+      published_by: avalon_publisher,
+      published: !avalon_publisher.nil?,
```
cjcolvar: why not use #published? instead?

app/controllers/media_objects_controller.rb
bkeese: I kind of like this approach: http://metabates.com/2012/02/22/adding-pagination-to-an-api/ Or is there another way you had in mind?

app/models/media_object.rb
bkeese: Good point. Now why didn't I think of that?

app/controllers/media_objects_controller.rb
cjcolvar: I'm looking into this right now: http://jsonapi.org/format/#fetching-pagination

bkeese:
Rebased from develop to pull in ActiveFedora 8.1.0

PR #633, opened 2015-11-24
Initial transpec pass.

carrickr:
Conversion to Rspec 3 syntax, zero transpec warnings, all supposedly converted.


cjcolvar:
This looks good, but I don't like the its expansions.  I'll take a pass at rewriting those and add it to this PR.

PR #637, opened 2015-12-04
Add mediaobject post api

bkeese:


app/controllers/media_objects_controller.rb
```diff
@@ -61,6 +61,65 @@ def new
     redirect_to edit_media_object_path(@mediaobject)
   end
 
+  # POST /media_objects
+  def create
```
cjcolvar: Is #create and #json_update access controlled right now?  If not, that's fine but we'll need to remember that for VOV-4172.

PR #654, opened 2016-01-04
Allow API to override set_streaming_locations! and accept multiple quality level derivatives

bkeese:


cjcolvar:
How about changing external to managed and default it to true?  I think that might be closer semantically to how we'll use that boolean.  We'll also need a migration to populate this field for existing items.

PR #669, opened 2016-01-22
Quick-load MasterFiles and Derivatives

cjcolvar:


cjcolvar:
The main difference between this and #624 is adding an option to not reify from solr which was needed to avoid quirks when editing.

PR #670, opened 2016-01-22
WIP: add resource_types to displayMetadata [skip-ci]

bkeese:


bkeese:
I think this is done, with the possible exception that I'm running into the problem of solr not persisting when it should.

cjcolvar:
@jlhardes At sprint planning we talked about decoupling the resource type that is in the MODS and pulled in via bib import from the computed resource type that is used for faceting and thumbnail display.  The question I have is how decoupled?  The problem arose when the bib import overrode what Avalon had computed.  If we make them totally separate that could leave the displayed and editable metadata showing what came in through bib import which might be different and wrong compared with what Avalon computes.  Is that really just an edge case or do we even care?  I'm thinking that fully separating them is the best way to avoid problems in the future.  Thoughts?

jlhardes:
I think they should be fully separated/decoupled and Avalon should detect a computed resource type for faceting and thumbnail display on its own. These items are coming in with the best descriptive info we have, which sometimes might not match what Avalon detects. I think that is OK and that descriptive metadata should come in as-is. Avalon is going to be correct in what it detects so that should be used for functionality. The descriptive metadata is as accurate as we can make it. Sometimes it will not be correct but I think it’s cleaner to keep it as it is from bib import and clean it up later rather then trying to sort it out on initial ingest.

cjcolvar:
@jlhardes :+1:  Thanks!

cjcolvar:
Since we do want these totally separate, I think we need to use a different field name since batch ingest or the api might try to write to resource_type and cause problems.  Having a different name would safely avoid any potential collision.

PR #672, opened 2016-01-27
Preserve other identifiers across bib import and ensure that supplied…

cjcolvar:
… other identifiers are persisted with API calls

This might create duplicates in the case when bib import fails.


bkeese:
This looks good to me. I wonder if in the future we will wish we did a more comprehensive merge of supplied fields and bib imported fields.

cjcolvar:
I thought of that too, but the only I knew for certain were other identifiers.

PR #678, opened 2016-02-04
Fix: Whitespace in Solr query string will return nothing for Solr cloud 4.10.x

taichun:
Please ignore the Gemfile and Gemfile.lock changes as those are related to our specific environment (using MySQL and Wowza)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,6 @@
   source 'http://rubygems.org'
 
+  gem 'mysql2', '~>0.3.20'
   gem 'iconv'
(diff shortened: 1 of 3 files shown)


cjcolvar:
@taichun Could you please make a new branch that includes only the solr changes?  Also please point this PR at avalonmediasystem:develop instead of master.

taichun:
PR #681 created per request.

Also: The colons in the query string have to be escaped so the %{} would still be needed to bypass that.

cjcolvar:
Closed in favor of #681

PR #679, opened 2016-02-04
mediaobject#set_session_quality was returning 401 always

bkeese:
This enables quality level switching in avalon media player, but reveals a playability problem with MEJS switchStream that had been only visible in embedded players. Play vs Pause mode gets confused when switching quality levels. 2-3 presses of play/pause sorts out the problem. Problem still needs to be fixed in mediaelement_rails.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -19,7 +19,7 @@ class MediaObjectsController < ApplicationController
   include Avalon::Controller::ControllerBehavior
   include ConditionalPartials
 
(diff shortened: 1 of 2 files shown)


bkeese:
I can't see why this test fails for the PR, but not for the push. It passes locally for me too, so debugging is difficult.

cjcolvar:
I'm seeing that same failure on #681 so I think it is an issue in develop.  I think if we fix it there and rerun the build then it should pass.

bkeese:
This is passing now.

PR #680, opened 2016-02-04
Make imported bib records behave better

mbklein:
This tiny issue was causing almost all of the errors in imported metadata. It was hiding behind other validation errors, but they turned out to be red herrings.

### DIFF
--- a/app/models/mods_templates.rb
+++ b/app/models/mods_templates.rb
@@ -51,8 +51,8 @@ module ModsTemplates
       end
 
       def get_origin_info
-        node = find_by_terms(:origin_info)
-        if node.empty?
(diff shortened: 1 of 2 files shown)


bkeese:
Those look equivalent to me. What am I missing?

mbklein:
A `.first` and the change of `.empty?` to `.nil?`

bkeese:
Yes. I saw that, but aren't those functionally equivalent?

mbklein:
No. The original version might return a `NodeSet` instead of a single `Node`, which causes the addition of whatever child element is being added to fail.

bkeese:
OOOOH. I was just looking at the test, not the return value. Gotcha.

PR #681, opened 2016-02-09
Removed whitespace in Solr query to work with Solr cloud 4.10.x

taichun:
The space in the solr query will cause nothing to return from a Solr cloud built with 4.10.4 version


cjcolvar:
:+1: I'm going to wait until the Travis build passes before merging.

cjcolvar:
I tracked down why the build is failing.  It is because we have environment variables defined in travis.org which aren't available to pull requests from external repos.  I feel confident that this PR is safe so I'm going to merge and let travis build on develop.  We'll have to rework our test process a little bit so this doesn't block outside contributors.

PR #682, opened 2016-02-10
Make sure MediaObject#parts_with_order always returns all parts...

mbklein:
…and none that don't actually exist


cjcolvar:
How about putting all of the logic from #get_ordered_pids into #section_pid so all calls to #section_pid won't be out of sync with #parts?  We'd have to push the existing #section_pid into another method #_section_pid in order to do this.

Also some tests around #get_ordered_pids (or whatever it becomes) would be good to explain what it is trying to fix.

mbklein:
Sure.

PR #685, opened 2016-02-22
Convert uuuu dates to unknown/unknown

bkeese:


cjcolvar:
@jlhardes Does this look good to you?  What are your thoughts about putting this in the XSLT instead of somewhere else?

jlhardes:
Yes, that looks good to me and it makes sense to handle that change in the XSLT. Can that also be noted with a dated comment at the beginning of that section and a dated comment at the top where we're noting Avalon XSLT changes?

bkeese:
Good idea. I'll add those comments.

PR #688, opened 2016-02-25
Add validation errors to flash messages

cjcolvar:


cjcolvar:
@bkeese what do you think about these changes?  I didn't see any validate:false for collections so I thought testing valid? made more sense than saved?.  I changed the validation message so it reads better.

PR #692, opened 2016-03-04
Callback refactor

mbklein:


spec/controllers/media_objects_controller_spec.rb
```diff
@@ -782,6 +785,26 @@
       media_object.reload
       expect(media_object.descMetadata.media_type).to eq(["video/mp4"])
     end
+
+    context 'large objects' do
+      let!(:media_object) do
+        mo = FactoryGirl.create(:published_media_object)
+        10.times { FactoryGirl.create(:master_file_with_derivative, mediaobject: mo) }
```
mbklein: When I ran this test on my laptop, I did it with 50 MasterFiles. But this one test took more than 10 minutes to run because of all the instrumentation rspec injects, so I reduced it to 10, which is enough to get it to fail _sometimes_ under the old version of the code.

app/models/media_object.rb
```diff
@@ -569,7 +563,48 @@ def move_bulk documents, params
...
+      self.descMetadata.permalink = self.permalink
+    end
+  end
+  handle_asynchronously :_update_permalink_and_dependents
+  
+  def update_permalink_and_dependents
+    self._update_permalink_and_dependents
+  end
+
+  def _remove_bookmarks
```
cjcolvar: Were you intending to run this asynchronously as well?  I think it makes sense to do so.

cjcolvar: Wait, never mind.  The problem of doing this asynchronously (and the problem with this at all) is that session variables aren't making it into Ability and thus losing access allowed by external groups like LDAP.

app/models/media_object.rb
```diff
@@ -569,7 +563,48 @@ def move_bulk documents, params
 
   end
 
+  def _update_permalink_and_dependents
```
cjcolvar: Maybe this should be broken down to update permalink which should be synchronous on before_save and update dependents' permalinks which should be asynchronous on after_save.

app/models/media_object.rb
mbklein: Right. I had it running asynchronously, because if there are a lot of bookmarks it could take some time, but it was causing other tests to fail. I left it broken out in case we wanted to revisit it later, but I'm happy to remove the indirection if you think it's confusing.

app/models/media_object.rb
mbklein: :+1: I'll give it a shot. I'll also add a non-trivial permalink generator to the test.

cjcolvar:
:+1:

PR #694, opened 2016-03-04
reverting to last stable mejs

csyversen:


csyversen:
going to submit this pr to the vov-4387 branch instead!

PR #701, opened 2016-03-17
Remove MasterFile#update_mediaobject! callback in favor of MediaObject#update_dependent_properties!

mbklein:
During testing, I discovered a potential for a `409 Conflict` error in the `MasterFile#update_media_object!` callback. Reloading costs time, but it's the only way to avoid the conflict at this time. It's my hope that this is still an improvement over the old callback flow, even though it's not as fast as fully deferring the MediaObject's callback would be.

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -79,7 +79,6 @@ class MasterFile < ActiveFedora::Base
 
   has_model_version 'R3'
   before_save :update_stills_from_offset!
-  around_save :update_media_object!, if: Proc.new { |mf| mf.duration_changed? or mf.file_location_changed? or mf.file_format_changed? }
 
   define_hooks :after_processing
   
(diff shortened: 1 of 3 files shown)


mbklein:
Argh. I guess I'll have to take a look at this tomorrow. :rage:

app/models/master_file.rb
```diff
@@ -105,6 +105,7 @@ class MasterFile < ActiveFedora::Base
   def update_media_object!
     yield
     return if mediaobject.nil?
+    mediaobject.reload
     mediaobject.set_duration!
```
bkeese: What about using update_dependent_properties! here instead?

mbklein: Ugh. That callback was supposed to go away completely. I missed it in my commit.

app/models/media_object.rb
```diff
@@ -35,6 +35,7 @@ class MediaObject < ActiveFedora::Base
 
   after_create :after_create
   before_save :normalize_desc_metadata!
+  before_save :update_dependent_properties!
```
cjcolvar: Maybe making this after_save would solve the test failure you're getting.  Before your PR this was happening after the yield in an around_save.  Ah, but that would cause double-saving so maybe we just need to make the mods datastream autocreate like here: https://github.com/avalonmediasystem/avalon/blob/develop/app/models/admin/collection.rb#L35

mbklein: It was happening after the yield, but the yield was on the child object in `MasterFile.around_save`. I moved the callback to `MediaObject.before_save`. I'm still trying to reproduce this on my laptop.

mbklein:
@cjcolvar: The build failure was an environment thing. Rebuild has passed with no changes.

cjcolvar:
:+1:

mbklein:
:shipit:

PR #706, opened 2016-03-29
date digitized can now be edited

csyversen:
no date picker, but it works


cjcolvar:
Looking good.  We should probably add a tooltip though to explain what the date digitized really means.

csyversen:
sounds good, i'll add that in and update this PR

csyversen:
this is ready to be looked at again

PR #708, opened 2016-04-04
Feature/vov 4194

bkeese:


app/views/media_objects/_resource_description.html.erb
```diff
@@ -31,7 +31,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 
   <%= render partial: 'text_field',
              locals: {form: form, field: :creator,
-                      options: {display_label: 'Main contributor(s)', required: true, multivalued: true}} %>
```
cjcolvar: Should we move this field down below the date issued since it isn't required anymore?

PR #709, opened 2016-04-05
[skip ci] add label getters to models

bkeese:


cjcolvar:
Please add tests since you're adding new methods.

PR #713, opened 2016-04-07
add test for relevancy weighting

bkeese:
Adds boost to title field and adds boosts to fields in qf. This seems to fix the immediate weighting problem. We might want to consider adjusting weights for other fields too (like creator).


bkeese:
I added a test for the solr weighting and it passes. However, it must not be good enough since sorting fails on Lancelot.

cjcolvar:
I'm going to merge this but it is probably something we want to talk over with POs and others.

PR #723, opened 2016-04-25
Feature/vov 4688

bkeese:


cjcolvar:
Before merging I think we should write a test for Lease#lease_type.  I can take care of that.

mbklein:
Can we wait and see if @coveralls wants to post anything in this thread?

PR #725, opened 2016-04-29
config tweaks necessary to deploy

yulgit1:
per:
https://groups.google.com/forum/#!msg/fedora-community/Pz8sXYCiitA/oY6QWjRXEgAJ
and file naming (post install) in:
./fedora/default/server/xsd/.
./fedora/test/server/xsd/.


cjcolvar:
Avalon is currently installing with a 3.6.1 Fedora and we weren't planning on upgrading it in release 5 which is due out in a month or so.  After that we'll be migrating Avalon to Fedora 4 which won't need this change.  As it is just a config change, you can safely keep it in your local repo/branch.  Thanks for the diff and we'll share it with anyone else who is trying to run Avalon on Fedora 3.8.

PR #727, opened 2016-05-04
Refactor #update_dependent_permalinks so that only the object's pid is needed for the background job

cjcolvar:


cjcolvar:
@mbklein This should fix some of the errors we were seeing during demo on lancelot.

PR #728, opened 2016-05-11
Make AvalonAnnotation behave more like an ActiveRecord::Base

mbklein:
Don't override `#initialize` on an ActiveRecord model; use `after_initialize` callback instead. Treat `#master_file` like any a regular, non-persisted attribute.

### DIFF
--- a/app/models/avalon_annotation.rb
+++ b/app/models/avalon_annotation.rb
@@ -13,28 +13,24 @@ class AvalonAnnotation < ActiveAnnotations::Annotation
   alias_method :title, :label
   alias_method :title=, :label=
 
-  # Intialize an annotation with the start and end time set to the lengths of the master_file by default
-  # @param [MasterFile] :source the master file referenced by the annotation
-  def initialize(master_file: master_file)
-    super
-    @master_file = master_file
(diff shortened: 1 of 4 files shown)


mbklein:
The one failing test (on one version of Ruby) seems to have nothing to do at all with the AvalonAnnotation class or any of these changes. May I beg the indulgence of a merge even though Travis is complaining so we can move on with the playlist work?

bkeese:
Interesting:
"Pull requests that have a failing status can’t be merged from a phone."


Sent via the Samsung GALAXY S®4 Active™, an AT&T 4G LTE smartphone


-------- Original message --------
From: "Michael B. Klein"
Date:05/11/2016 1:01 PM (GMT-05:00)
To: avalonmediasystem/avalon
Subject: Re: [avalonmediasystem/avalon] Make AvalonAnnotation behave more like an ActiveRecord::Base (#728)


The one failing test (on one version of Ruby) seems to have nothing to do at all with the AvalonAnnotation class or any of these changes. May I beg the indulgence of a merge even though Travis is complaining so we can move on with the playlist work?

PR #735, opened 2016-05-12
Feature/vov 4680

lealeelu:
Add media element add-to-playlist feature


carrickr:
closed, want to run against a newer develop with a fix in it

PR #736, opened 2016-05-12
Bugfix/dupmigration

carrickr:
@mbklein 

I think we ended up with two migrations for annotation in the end?


mbklein:
Any thoughts on the failing test? I'm on sufia this week so I don't have time to look into the code, but I'd like to merge this asap.

carrickr:
@mbklein 

It is actually fixed by the work you did last week regarding annotations not saving, this branch was spawned off develop while that bug was still present

carrickr:
I've merged it and now it should pass on develop since current develop has the fix

PR #746, opened 2016-05-13
Add Playlists button to global_nav user_util_links

bkeese:


app/views/_user_util_links.html.erb
```diff
@@ -34,6 +34,11 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
     </li>
     <% end %>
+    <% if user_signed_in? %>
```
cjcolvar: I think this works for now, but we probably should change this to work off an ability:
if current_ability.can? :create, Playlist
Or something like that.

app/views/_user_util_links.html.erb
```diff
@@ -34,6 +34,11 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+    <li>
+      <%= link_to 'Playlists', playlists_path, id:'playlists_nav' %>
```
cjcolvar: I think we also need the conditional class that does the background highlight like the other global nav tabs:

<li class=<%= active_for_controller('playlists') %>><%= link_to 'Playlists', playlists_path, id: 'playlists_nav' %></li>

PR #748, opened 2016-05-14
add annotation to scrubber

bkeese:


app/controllers/media_objects_controller.rb
```diff
@@ -18,6 +18,7 @@ class MediaObjectsController < ApplicationController
   include Avalon::Workflow::WorkflowControllerBehavior
   include Avalon::Controller::ControllerBehavior
   include ConditionalPartials
+  include MediaObjectsHelper
```
cjcolvar: I don't think we should include the helper here, but should use a different approach.  See [http://stackoverflow.com/a/11161692]

cjcolvar:
Does this have any ill side effects on clicking current structure links?

PR #755, opened 2016-05-17
playlist controller tests for invalid params

bkeese:


app/controllers/playlist_items_controller.rb
```diff
@@ -4,12 +4,23 @@ class PlaylistItemsController < ApplicationController
 
   # POST /playlists/1/items
   def create
+    title =  playlist_item_params[:title]
+    comment = playlist_item_params[:comment]
+    start_time = time_str_to_milliseconds playlist_item_params[:start_time]
+    end_time = time_str_to_milliseconds playlist_item_params[:end_time]
+    if title.blank? or !numeric?(start_time) or !numeric?(end_time)
+      flash[:notice] = "Title not specified or times are invalid."
```
cjcolvar: I think it would be best if we could split these errors out so it could be more clear to the user what to fix.

cjcolvar: Maybe this should be flash[:alert] instead of flash[:notice]?  I forget how we do it other places.

PR #760, opened 2016-05-18
Pre Vov 4608

carrickr:
Pre swarm PR


carrickr:
Using a rebase pr instead

PR #763, opened 2016-05-18
better way to find related annotations on the same playlist

carrickr:


app/views/playlists/_related_items.html.erb
```diff
@@ -16,13 +16,15 @@ Unless required by applicable law or agreed to in writing, software distributed
 <div id="related_items">
   <ul>
   <% @related_annotations.sort_by { |hash| hash['start_time'].to_i}.each do |annotation| %>
-    <% position = PlaylistItem.where("playlist_id == :p_id AND annotation_id == :a_id", { p_id: @playlist.id, a_id: annotation.id}).first.position %>
+    <% position = playlist_position(@playlist.id) %>
```
cjcolvar: Is #playlist_position available here in the view?  I'm guessing this probably needs to be @playlist.playlist_position(@playlist.id)

PR #769, opened 2016-05-19
CSS formatting of the new playlist page

lealeelu:


bkeese:
I'm in the middle of some major changes to _form.html.erb. Let's hold off on merging this into develop until we can compare.

PR #779, opened 2016-05-20
Use numericality validator for all start/end time validations instead of throwing argument errors

cjcolvar:
I believe this should also fix the failing tests in develop


carrickr:
Updated the factory so it makes a default duration, looks like it was just the factory object failing in tests, no code issues.

PR #786, opened 2016-05-24
unlock to globe and other css adjustments

lealeelu:


carrickr:
PO's specifically asked for 0.3 for margin top, bottom, and side.  We should adjust to that for styling.

carrickr:
PO sign off on new design

PR #798, opened 2016-05-26
Playlist playthrough via page reloading

cjcolvar:


carrickr:
I don't really feel great about introducing a playthrough that runs via a brute force reload of the page this late in the release.  I feel as if we're missing other things the POs might want, such as stopping when the end of a clip is reached vs reloading and would like to revert this out until we consider flow and what exactly we want more @cjcolvar @csyversen

cjcolvar:
@carrickr OK, I agree ... merge pending PO approval

PR #802, opened 2016-05-31
Feature/playlist playthrough

cjcolvar:
Ok'd by POs.  I'd welcome any javascript improvements.

### DIFF
--- a/app/views/media_objects/_item_view.html.erb
+++ b/app/views/media_objects/_item_view.html.erb
@@ -88,7 +88,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   <% if @currentStream.present? and @currentStream.derivatives.present? %>
   <script>
     function advancePlaylist() {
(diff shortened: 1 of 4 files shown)


app/views/media_objects/_item_view.html.erb
```diff
@@ -87,6 +87,13 @@ Unless required by applicable law or agreed to in writing, software distributed
 
   <% if @currentStream.present? and @currentStream.derivatives.present? %>
   <script>
+    function advancePlaylist() {
+      nextUrl = $('a.playable:not(a.current-stream):first').attr('href')
```
mbklein: Won't this find the first playable stream that's not currently playing? e.g., if the first stream is current, it'll find the second one; otherwise, it'll find the first one.

cjcolvar: Good catch.  I dealt with this for playlists but I didn't for media objects.

PR #803, opened 2016-05-31
Bugfix/new playlist

cjcolvar:
Branched from VOV-4951 so merge after #801


cjcolvar:
Rebased since #801 was merged

PR #884, opened 2016-06-15
Update mediaelement_rails for captions button race condition work around

### LINKED ISSUE #880: Captions intermittenly don't show up in player
Steps:
1. Load an item with captions, ie http://lancelot.dlib.indiana.edu/media_objects/avalon:sample-beginning-responsibility-lunchroom
2. Try refreshing or logging out

Observed: CC button shows in player then quickly disappears. No captioning during playback

Bug does not seem to be happening with Wowza instead of Red5.

bkeese:
fixes #880


cjcolvar:
Fixes #880

PR #922, opened 2016-06-23
Marker test

anirudhpillai16:


cjcolvar:
The avalon_marker factory isn't being found.  Please move spec/fixtures/avalon_marker.rb to spec/factories/avalon_marker.rb.

spec/models/avalon_marker_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+require 'spec_helper'
+
+describe AvalonMarker, type: :model do
+  describe 'abilities' do
+    subject { ability }
+    let(:ability) { Ability.new(user) }
+    let(:user) { FactoryGirl.create(:user) }
+    let(:master_file) { FactoryGirl.create(:master_file) }
+    let(:avalon_marker) { FactoryGirl.create(:avalon_marker, playlist_item: playlist_item) }
```
bkeese: This refers to playlist_item, which is defined in the next line. The two lines should be flipped, I think.

cjcolvar: I don't think it matters since rspec lazy evaluates these let statements.  Swapping the order had no effect for me.

PR #925, opened 2016-06-23
Do away with therubyracer; rely on external node.js for asset compilation

mbklein:
`libv8` and `therubyracer` both fail to compile or load every time a new version of OS X or XCode is released. Since we're using `execjs` for asset pipeline management, we can simply rely on `NodeJS` to do the same work.


cjcolvar:
@mbklein This failed to deploy to lancelot with an error message about ExecJS not finding a runtime.

PR #1007, opened 2016-07-28
5.1.0 master

carrickr:


carrickr:
Wow we need to use git squash more!

PR #1023, opened 2016-08-03
Display a dash when the section's file location is nil

cjcolvar:


app/views/media_objects/_structure.html.erb
```diff
@@ -42,7 +42,11 @@ Unless required by applicable law or agreed to in writing, software distributed
 	  </li><li>
             <%= section.label.blank? ? '&#8212;'.html_safe : section.label %>
 	  </li><li>
-            <%= truncate_center(File.basename(section.file_location), 50, 20) %>
+            <% if part.file_location.present? %>
```
bkeese: This should be section.file_location, right?

PR #1059, opened 2016-08-12
the big rename

carrickr:
# fixes 1058


carrickr:
@cjcolvar @bkeese I did change the tests to use 'Sound Recording' and 'Moving Image' rather than 'sound recording 1','sound recording 2', etc.  Was there a reason we were sticking the number on there?

Issue #998, opened 2016-07-26
Refactor Avalon configuration


mbklein:
The following files need to become optional/obsolete:

* `authentication.yml`
* `avalon.yml`
* `database.yml` 
* `fedora.yml`
* `lti.yml`
* `matterhorn.yml`
* `role_map_ENVIRONMENT.yml`
* `solr.yml`

mbklein:
Most of them (`database`, `fedora`, `matterhorn`, `solr`) can be done simply with environment variable substitution. The others are more complicated and may require more sophisticated serialization or rethinking of how to store their values.

mbklein:
Paths pointing to executables should be replaced by symlinks into `[RAILS_ROOT]/bin`

mbklein:
This set of environment variables covers all of the current values in `avalon.yml`, but needs to be wired up:

```
APP_NAME=avalon
FEDORA_NAMESPACE=avalon
AVALON_URL=http://avalon:80/
DROPBOX_PATH=/srv/avalon/dropbox
DROPBOX_URI=sftp://localhost${DROPBOX_PATH}
MEDIAINFO_PATH=/usr/local/bin/mediainfo
STREAM_BASE=/streamfiles
STREAM_SERVER=wowza
STREAM_TOKEN_TTL=20
STREAM_RTMP_BASE=rtmp://streaming/avalon
STREAM_HTTP_BASE=http://streaming:1935/avalon/_definst_
STREAM_DEFAULT_QUALITY=low
SYSTEM_GROUPS=administrator,group_manager,manager
MASTER_FILE_STRATEGY=move
MASTER_FILE_PATH=/masterfile_storage
FFMPEG_PATH=/usr/bin/ffmpeg
CONTROLLED_VOCABULARY=config/controlled_vocabulary.yml
EMAIL_COMMENTS=avalon-comments@example.edu
EMAIL_NOTIFICATION=avalon-notifications@example.edu
EMAIL_SUPPORT=avalon-support@example.edu
SMTP_ADDRESS=smtp.example.edu
SMTP_PORT=587
SMTP_DOMAIN=example.edu
SMTP_USER_NAME=foo
SMTP_PASSWORD=blargh
SMTP_AUTHENTICATION=plain
SMTP_ENABLE_STARTTLS_AUTO=false
SMTP_OPENSSL_VERIFY_MODE=client_once
```

davidschober:
We would also want something that spits out all of these variables. Maybe a Rake task.

[...]

PR #1063, opened 2016-08-16
Full configuration refactor [#998]

mbklein:
Refactor Avalon configuration to load from environment or avalon.yml

Use MATTERHORN_URL, and don't attempt to migrate avalon.yml if it doesn't exist

Use rubyhorn master instead of a branch or tag

Initialize ActiveFedora using FEDORA_URL and SOLR_URL

Force Blacklight to use ActiveFedora's solr connection.

Make rugged a development dependency only

Use LTI_AUTH_KEY and LTI_AUTH_SECRET to initialize LTI authentication instead of having an :lti stanza in authentication.yml

Allow use of SECRET_KEY_BASE instead of AVALON_SECRET_KEY_BASE for consistency with other values

Add documentation for environment variable configuration


cjcolvar:
Is this really supposed to be in develop? Or is it supposed to go into hydra10-fresh?

Issue #1076, opened 2016-08-19
Add AVALON prefix to environment variables


phuongdh:
Some env var needs the prefix in order to not collide with other services

PR #1095, opened 2016-08-25
Copy ENV["AVALON_XXXXX"] environment variables to ENV["XXXXX"] before…

mbklein:
… app initialization

This will allow you to prefix any environment variable with `AVALON_` to keep it from clashing with other processes' variables, while keeping the option to leave them unprefixed for the sake of clarity if Avalon is the only Rails service on the box.


mbklein:
Addresses #1076

PR #1098, opened 2016-08-25
Add models for playlists, playlist_items, annotations, clips, markers

bkeese:


Gemfile
```diff
@@ -1,6 +1,8 @@
 source 'https://rubygems.org'
 
 gem 'hydra', git: 'https://github.com/projecthydra/hydra.git'
+gem 'active_annotations', git: 'https://github.com/avalonmediasystem/active_annotations.git', tag: 'avalon-r5'
```
cjcolvar: I was collecting our gem dependencies lower down in the Gemfile.  Could you please move these there?

app/models/ability.rb
```diff
@@ -57,7 +57,7 @@ def create_permissions(user=nil, session=nil)
   end
 
   def custom_permissions(user=nil, session=nil)
-    # playlist_permissions
+    playlist_permissions
```
cjcolvar: Let's move the additional permission methods into ability logic instead of calling them directly from custom_permissions.  See https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/lib/hydra/ability.rb#L9-L11

PR #1103, opened 2016-08-26
Set Travis to Use MySQL2 rather than SQLLite

carrickr:
Because SQLLite is broken for some reason


Gemfile
```diff
@@ -83,5 +83,6 @@ group :test do
   gem 'fakeweb'
   gem 'email_spec'
   gem 'capybara'
+  gem 'mysql2'
```
cjcolvar: Is there a way to scope this to a travis group instead of test?

Issue #1069, opened 2016-08-17
MediaObject class predicate review


jlhardes:
- [x] duration - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#duration (ebucore:duration)
- [x] identifier -  http://id.loc.gov/vocabulary/identifiers/local (loc:local)

Issue #1068, opened 2016-08-17
Permalink class predicate review


jlhardes:
- [x] hasPermalink -  http://purl.org/dc/elements/1.1/identifier (dc:identifier)

Issue #1055, opened 2016-08-11
Encoding class predicate review


jlhardes:
- [x] quality - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#encodingLevel (ebucore:encodingLevel)
- [x] mimeType - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasMimeType (ebucore:hasMimeType)
- [x] resolution - can this be broken up into width and height? if so, http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#width (ebucore:width) and http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#height (ebucore:height) can be used

EBUCore only supplies a single property for bitrate and for codec, so apply to video and keep audio properties as Avalon-specific:
- [x] videoCodec - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasCodec (ebucore:hasCodec)
- [x] videoBitrate - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#bitRate (ebucore:bitRate)

jlhardes:
- [ ] mimeType - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasMimeType (ebucore:hasMimeType)
- [ ] resolution - can this be broken up into width and height? if so, http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#width (ebucore:width) and http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#height (ebucore:height) can be used
- [ ] quality - Either http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasFormat (ebucore:hasFormat) or http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasGeneration (ebucore:hasGeneration)

EBUCore only supplies a single property for bitrate and for codec. If we can only apply to one type between audio and video, apply to video:
- [ ] videoCodec - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasCode (ebucore:hasCode)
- [ ] videoBitrate - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#bitRate (ebucore:bitRate)

Issue #1053, opened 2016-08-09
MasterFile class predicate review


jlhardes:
- [x] fileLocation - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#locator (ebucore:locator)
- [x] fileChecksum - http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#hashValue (nfo:hashValue)
- [x] fileSize - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#fileSize (ebucore:fileSize)
- [x] duration - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#duration (ebucore:duration)
- [x] displayAspectRatio - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#aspectRatio (ebucore:aspectRatio)
- [x] originalFrameSize - can this be broken up into width and height? if so, http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#width (ebucore:width) and http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#height (ebucore:height) can be used
- [x] fileFormat - http://id.loc.gov/ontologies/premis.html#hasFormatName (premis:hasFormatName)
- [x] dateDigitized - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#dateCreated (ebucore:dateCreated)
- [x] physicalDescription - http://purl.org/dc/elements/1.1/format (dc:format)
- [x] masterFile - http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#filename (ebucore:filename)
- [ ]  identifier - http://id.loc.gov/vocabulary/identifiers/local (loc:local)

PR #1105, opened 2016-08-26
Update predicates for Derivative, Encoding, MasterFile, MediaObject, …

mbklein:
…and Permalink properties

Addresses #1042, #1053, #1055, #1069, #1068

### DIFF
--- a/app/models/derivative.rb
+++ b/app/models/derivative.rb
@@ -15,36 +15,38 @@
 require 'avalon/stream_mapper'
 
 class Derivative < ActiveFedora::Base
+  include FrameSize
+  
   belongs_to :master_file, class_name: 'MasterFile', predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isDerivationOf
 
-  property :location_url, predicate: Avalon::RDFVocab::Derivative.locationURL, multiple: false do |index|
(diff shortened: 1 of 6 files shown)


app/models/derivative.rb
```diff
@@ -53,6 +55,31 @@ def initialize(*args)
     self.managed = true
   end
 
+  alias_method :_width=, :width=
```
cjcolvar: Since these appear in master file as well, should they be pulled out into a concern?

mbklein: I thought about making a concern out of it, but because one uses `#original_frame_size` and the other uses `#resolution`, I wasn't sure it was a good idea. I'll take a stab at it though.

cjcolvar: Ah, I didn't notice that.  Never mind then.

mbklein: @cjcolvar Done. Please see latest commit

carrickr:
@mbklein @cjcolvar All green, do we want to squash out some of the solr_wrapper and mysql stuff pre merge or just stick it in as is?

mbklein:
I'm gonna flatten and rebase. Give me a minute.

mbklein:
@cjcolvar @carrickr Rebased on develop, so the solr_wrapper and mysql stuff just fell away from this PR. No other changes.

PR #1129, opened 2016-09-07
Add more javascript to precompile list

lealeelu:


bkeese:
Also need: 
Rails.application.config.assets.precompile += %w( marker.js )

PR #1138, opened 2016-09-12
Capify

carrickr:


csyversen:
i think this pr is fine, but is there any reason we're sticking with capistrano 2? I've used 3.x on other projects and it has a few new nice features (like environment specific config support built in)

carrickr:
@csyversen I was mostly being lazy, err efficient, and just reusing our recipe for Lancelot which I know works under 2 (since we'll be totally changing the deploy method in Avalon 7 when we break the app up anyway).  I'm not against bringing 3 in though.

cjcolvar:
I'm going to merge this now, but I think it would be great to update this to Cap3.  @csyversen: PRs welcome!

PR #1144, opened 2016-09-15
Bypass the Hydra gem and update to ActiveFedora 11.

mbklein:
Fix failing tests.
Watch tests fly.


app/models/media_object.rb
```diff
@@ -436,6 +436,10 @@ def remove_bookmarks
     self._remove_bookmarks
   end
 
+  def policies_of_type(klass)
```
cjcolvar: Could this be rewritten as a scope?

Gemfile
```diff
@@ -1,6 +1,10 @@
 source 'https://rubygems.org'
 
-gem 'hydra', git: 'https://github.com/projecthydra/hydra.git'
+gem 'hydra-head', '~> 10.1'
+gem 'active-fedora', '~> 11.0'
+gem 'active_fedora-datastreams'
```
cjcolvar: Do we need active_fedora-datastreams and curation_concerns?

mbklein: `ActiveFedora::Datastream` and its descendants (`OmDatastream`, `NogogiriDatastream`) were pulled out of `active-fedora` into `active_fedora-datastreams`. We can't do MODS without it.

I only included `curation_concerns` because I thought it pulled in other stuff we needed. I'll comment it out and run the tests to see what happens.

app/models/media_object.rb
mbklein: Ooh, good call. I'll give it a shot.

mbklein:
Tests are all passing on my machine. I don't know what's going on with Travis.

PR #1163, opened 2016-09-22
add in a rake task and schedule to purge sessions more than 7 days old

carrickr:
Currently the ActiveRecord Sessions are never purged (IU found 5.3 million rows representing 2 years of sessions in our production db).  I've added in a rake task that runs every 6 hours and clears any session that hasn't been updated for a week.  We should probably also consider cherry-picking this into develop for a 6.0 feature.  Also in 6.0 we might want to make it more configurable, etc.

Also note if you have a large sessions table you'll want to do the first clear via SQL, ActiveRecord will probably timeout or otherwise dislike doing millions of rows in one pass.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -5,3 +5,7 @@ job_type :locking_rake, "cd :path && :environment_variable=:environment script/l
 every 1.minute do
   locking_rake "avalon:batch:ingest", :lock_name => "batch_ingest"
 end
+
+every 6.hours do
+  rake 'avalon:session_cleanup'
+end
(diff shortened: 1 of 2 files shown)


config/schedule.rb
```diff
@@ -5,3 +5,7 @@
 every 1.minute do
   locking_rake "avalon:batch:ingest", :lock_name => "batch_ingest"
 end
+
+ever 6.hours do
```
bkeese: typo 'ever'

carrickr:
@bkeese fixed (fix was squashed)

PR #1165, opened 2016-09-23
Add mime types from avalon 5.1 mime types initializer

cjcolvar:
This keeps the authorize route from bombing with a 500 NoMethodError.


mbklein:
@cjcolvar Sorry; I missed this PR while I was working on the `/authorize` fix. It was a super-small change so I just pushed it straight to develop. I think we should pull MIME types in as we need them instead of dragging in everything from 5.1. I don't think we ever used most of these.

cjcolvar:
No problem.  Thanks for handling this and the solr query in StreamToken that was the next thing I was going to work on but now I don't need to.

PR #1168, opened 2016-09-23
Fixes to allow extract_frame to run without errors

cjcolvar:


bmquinn:
Tested #extract_frame in the console and specs pass. Merging!

PR #1179, opened 2016-09-26
refactor #deliver_content to use datastream parameter

kdid:
test by clicking View Raw Metadata button in media object show view


cjcolvar:
Could you switch this back to file for the parameter and update what is calling it to use file?  "Datastreams" no longer exist in Fedora 4/Hydra 10 so I think it better to use the current nomenclature.

cjcolvar:
Looks good. 👍

Issue #850, opened 2016-06-09
Avalon does not throw an error when incorrect username is entered


davidschober:
Avalon is not currently throwing an error with incorrect credentials. 

This only manifests in stock Avalon. In reality, most institutions will want to integrate with their campus authentication.
Improving the out of the box experience.

### Done looks like
A nice error is thrown is thrown.

PR #1190, opened 2016-09-27
Wrap omniauth-identity forms in the app layout and add bootstrap classes

mbklein:
This takes the auto-generated login and registration forms from `OmniAuth::Identity` and renders them within the Avalon layout with bootstrap controls. The result is a more integrated look and feel, and login error flash messages render properly. This is potentially extensible to other login mechanisms that rely on app-driven forms (like LDAP, but not OAuth or CAS), but right now it only works for Identity.

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -1157,3 +1157,10 @@ h5.panel-title {
   bottom: 6pt;
   width: 1em !important;
 }
+
+.omniauth-form {
+  width: 320px;
+  border: 2px solid #eee;
+  margin: 0px auto;
+  padding: 8px;
+}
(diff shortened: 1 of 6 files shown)


mbklein:
Addresses #850

bkeese:
Can we get rid of this now? avalon/config/initializers/omniauth-identity.rb

mbklein:
I don't see an omniauth-identity.rb initializer, @bkeese. What are you referring to?

mbklein:
Ah, there it is. Yes we can ditch that initializer if we merge this. 😄

mbklein:
Ugh. Last commit included stuff I didn't mean for it to. Please hold.

cjcolvar:
Can you add some controller tests?

mbklein:
I spent a couple hours banging my head against how to write these as controller tests, because I really wanted to be able to `expect(response).to render_template(layout: "avalon")`. Unfortunately, because OmniAuth uses weird middleware routing, I couldn't figure out how to trap the call in a way that let me test it as a controller action. I ended up writing it as an integration test that looks for text from the layout.

Issue #1216, opened 2016-09-29
Playlist views return undefined method errors


bmquinn:
app/views/playlists/_current_item.html.erb
app/views/playlists/_edit_form.html.erb

kdid:
Fixed with issue #1217

PR #1217, opened 2016-09-29
update views to use media_object.id instead of media_object.pid

kdid:


kdid:
References issue #1216

PR #1222, opened 2016-09-30
Hydra10 css

lealeelu:


app/assets/stylesheets/avalon.scss
```diff
@@ -649,10 +649,6 @@ a[data-trigger='submit'] {
   margin-top: 0;
 }
 
-.form-control {
```
bkeese: This seems like a pretty big change, affecting all forms. Are you sure this is what you want to do?

PR #1237, opened 2016-10-07
Fix views for playlists that have references to deleted masterfiles

bkeese:


carrickr:
Test failures do not appear to be related to Brian's code changes, but rather our testing suite issues.

PR #1244, opened 2016-10-11
Asset pipeline

csyversen:


app/assets/javascripts/application.js
```diff
@@ -12,12 +12,16 @@
...
-//= require jquery.nestable
 
 // Required by Blacklight
 //= require jquery-ui
-//= require jquery.ui.nestedSortable
 //= require blacklight/blacklight
 //= require browse_everything
-//= require pop_help
-//= require input_form_attribute_support
+//= require mediaelement_rails
```
cjcolvar: Is this including more than we want since before we were only including mediaelement-and-player?

app/assets/javascripts/application.js
```diff
@@ -12,12 +12,16 @@
...
+//= require me-thumb-selector
+//= require me-add-to-playlist
```
cjcolvar: Should other mediaelement plugins be included here?  I didn't look through to see which we're really using but we have repos for mediaelement-qualityselector, mediaelement-skin-avalon, mediaelement-hd-toggle, mediaelement-title, mediaelement-logo-plugin.

PR #1264, opened 2016-10-13
Update rake for mediaelement_rails

### LINKED ISSUE #1155: Add Marker Icon Moves Outside Of Media Player
### Environment 
Firefox 48.0.1 on OS 10.11.6

### Steps to Reproduce
* Go to a Playlist item in a Playlist
* Size the browser window so that the media player is slightly less than maximum width
* Expand the "Source Item Details" section of the accordion element beneath the media player
* The "Add Marker" icon breaks out of the media player area and wraps down into the accordion div

### Example
GIF

bkeese:
Fixes #1155

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,6 +66,7 @@ gem 'whenever', git: "https://github.com/javan/whenever.git", require: false
 gem 'with_locking'
 
 #MediaElement.js related
+gem 'rake', '~>11.3'
 gem 'mediaelement_rails', git: 'https://github.com/avalonmediasystem/mediaelement_rails.git', branch: 'captions'
 gem 'mediaelement-qualityselector', git:'https://github.com/avalonmediasystem/mediaelement-qualityselector.git', tag: 'avalon-r4'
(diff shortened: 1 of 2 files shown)


cjcolvar:
@bkeese why does rake need to be constrained?  Is there a version that we're trying to avoid pulling in or does it just need to be greater than 11.3?

bkeese:
11.3 is the version bundler is looking for. So I looked at how we defined in in 5.x and did the same here.

carrickr:
why not just bundle update rake and commit the new gemfile.lock?

carrickr:
if it is in media-element rails we should fix it there probably

bkeese:
Further commits welcome...

cjcolvar:
I don't think specifying rake is the real fix...the underlying issue is that our capistrano recipe isn't bundle installing the versions of gems our Gemfile.lock is requiring.  I handled this on our demo server and we can open a ticket if it becomes a problem again.

PR #1288, opened 2016-10-18
Fix removal of read users and groups

### LINKED ISSUE #1228: [6.0 Specific]:- Removing item from either Avalon user, Avalon groups, External groups throws error.
Environment:- https://spruce.dlib.indiana.edu
All Devices/Browsers

**Steps to Reproduce**

1. Launch URL on any browser.
2. Login with valid credentials. 
3. Access "Manage Content" and access any collection.
4. Add either avalon user, avalon group or external group.
5. Try removing any of them.

Observe error message is displayed when user tries to remove user under avalon user or group under avalon group or external group.

Note:- This issue is not reproducible on mallorn.

**Done looks like**
User should be able to remove any item from any of the groups.

Please refer attached screenshots for details:-
remvoing_avalon_user
removing_group
removing_external_group

bkeese:
Fixes #1228 

ActiveTriple::Relation doesn't have -= method, and delete doesn't automatically trigger persists.

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -143,9 +143,11 @@ class Admin::CollectionsController < ApplicationController
 
         if params["remove_#{title}"].present?
           if ["group", "class", "ipaddress"].include? title
-            @collection.default_read_groups -= [params["remove_#{title}"]]
+            @collection.default_read_groups_will_change!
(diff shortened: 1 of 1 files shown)


cjcolvar:
I think this works for now, but we make a note to later figure out why the activemodel::dirty tracking isn't automatically working.

PR #1304, opened 2016-10-25
master_files_controller#embed uses embed layout which includes update…

csyversen:
…d embed stylesheet. 

This PR DRYs stuff up a bit. The embedded player is using the player partial and the avalon.scss layout with a small nested layout that uses css to remove a few elements that shouldn't show up in an embedded player page. We're not loading any duplicate js/css assets anymore and since we're reusing the player and avalon layout, we won't have to make changes in more than one place if we decide to change anything on the player. 

The main changes are here: `app/views/layouts/embed.html.erb` with the rest of the changes supporting this nested layout


bkeese:
But we don't want the embedded player to use the avalon layout, do we? It should basically just be the player without all the navigation, branding, etc.

csyversen:
the `embed.html.erb` layout and 'embed.scss' file remove that, if you look at an embedded player now the video takes up 100% of the iframe

bkeese:
I still don't get it. The embed layout renders the avalon layout which includes all the stuff we don't want here. Doesn't it?

bkeese:
Oh, now I kinda get it. The css has the display:none on a couple of things, but isn't there much more on the page to hide?

csyversen:
surprisingly, no! once we hid those two elements the player was taking up 100% of the iframe and there was nothing else to display

csyversen:
(i was mostly following this pattern to see how we could display our app differently: http://guides.rubyonrails.org/layouts_and_rendering.html#using-nested-layouts )

cjcolvar:
@csyversen Where is this at?  Can you rebase this?

csyversen:
@cjcolvar i didn't realize you commented on this a week ago! It must've gotten lost in my mailbox with the 1000 other github emails i get now. 

anyway we discussed this in standup today, i'm going to merge in the latest develop changes into here and update this pr

csyversen:
ok, this is ready

PR #1316, opened 2016-10-28
Titlize avalon_resource_type in index

### LINKED ISSUE #1232: [6.0 Specific] [Minor issue]:- Moving Image and Sound Recording under Format should have first letter capitalized.
Environment:- https://spruce.dlib.indiana.edu
Devices:- All Devices/Browsers

**Steps to Reproduce**
1. Launch URL on any browser.e.g:- Chrome.
2. Expand Format under Browse by on the left side.

Observe moving image and sound recording are in lower case.

Note:- Issue is not reproducible on mallorn. Both moving image and sound recording have first letters in upper case.

**Done looks like**
moving image and sound recording should have first letters in upper case.

Please refer attached screenshot for details:-
format_lower_case

lealeelu:
fixes #1232 
This needs more testing once some of the bugs are finished. We can wait to pull it in then, I just wanted to get this pull request started.

### DIFF
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -209,6 +209,7 @@ class MediaObject < ActiveFedora::Base
       solr_doc['section_id_ssim'] = master_file_ids
       solr_doc["section_label_tesim"] = section_labels
       solr_doc['section_physical_description_ssim'] = section_physical_descriptions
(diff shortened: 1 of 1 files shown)


lealeelu:
I also think there might be more indices we need to add back in, we should take a look at that.

PR #1348, opened 2016-11-04
add in a self we thought we did not need but we do

carrickr:


carrickr:
@cjcolvar this should solve the issue with it not being assigned and stored, do you want to review?  we'll need another reindex

cjcolvar:
:+1: just waiting for travis to finish running.

PR #1350, opened 2016-11-04
Resque UI exposed to admins only

### LINKED ISSUE #1317: ensure resque is exposed to admin only
currently any user can get to /jobs

This should be restricted to admin only.

phuongdh:
fixes #1317

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -149,13 +149,11 @@ Rails.application.routes.draw do
   get '/about(.:format)', to: redirect('/')
   get '/about/health(.:format)', to: redirect('/')
 
-  require 'resque/server'
-  mount Resque::Server, at: '/jobs'
-  
   constraints(route_can?(:manage, Resque)) do
     require 'resque/server'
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -136,8 +136,10 @@
 
   match "/oembed", to: 'master_files#oembed', via: [:get]
   
-  require 'resque/server'
-  mount Resque::Server, at: '/jobs'
+  authenticate :user, lambda {|u| Ability.new(u).can? :manage, Resque } do
```
cjcolvar: You should pass the session, if available, to the ability constructor since we grant/revoke abilities based on session information.

mbklein: @bmquinn and I came up with a good way to use this using `constraints`. Please see #1351 for an example that can be reused easily by this issue.

cjcolvar: @pdinh Heads up, I commented on #1351 which suggests a slight change.

mbklein:
Force-rebased onto develop now that #1351 was merged.

PR #1351, opened 2016-11-04
Add about_page and avalon-about gems

bmquinn:
Installs and configures the about_page and avalon-about gems with new support for Fedora 4. Test by running `bundle install` and navigating to `/about` and `/about/health`

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,4 @@
-  Rails.application.routes.draw do
+Rails.application.routes.draw do
 
   mount Blacklight::Engine => '/'
   root to: "catalog#index"
@@ -41,8 +41,7 @@
   end
 
   mount BrowseEverything::Engine => '/browse'
-  mount AboutPage::Engine => '/about(.:format)', :as => 'about_page'
-
+    
   # Avalon routes
   match "/authorize", to: 'derivatives#authorize', via: [:get, :post]
   match "/authorize/:path", to: 'derivatives#authorize', via: [:get, :post]
@@ -136,6 +135,18 @@
   end
 
   match "/oembed", to: 'master_files#oembed', via: [:get]
+
+  def can_see?(thing, scope=nil)
+    lambda do |request|
+      warden = request.env['warden']
+      warden.authenticate? && Ability.new(warden.user(scope), warden.session(scope)).can?(:manage, thing)
+    end
+  end
+
+  constraints(can_see?(Admin::Collection)) do
+    mount AboutPage::Engine => '/about(.:format)', as: 'about_page'
+  end
+  get '/about(.:format)', to: redirect('/')
 
   require 'resque/server'
   mount Resque::Server, at: '/jobs'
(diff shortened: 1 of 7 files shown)


cjcolvar:
What about resque and resque-scheduler?  We'll want those to be monitorable via health.yaml as well.

cjcolvar:
We also need to restrict these routes to admin only but still allow a token access for our Nagios checks.  @phuongdh 's work on the jobs routes should be a good pattern to follow: #1350

config/routes.rb
```diff
@@ -137,6 +136,18 @@
 
   match "/oembed", to: 'master_files#oembed', via: [:get]
 
+  def can_see?(thing, scope=nil)
```
cjcolvar: I think this is too narrow a scope.  I think we'd want to be able to declare new abilities and be able to use them here like:

def route_can?(action, thing, scope=nil)
constraints(route_can?(:read, :about_page))

spec/features/secure_routes_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+require 'rails_helper'
+
+describe "Secure Routes" do
```
cjcolvar: Could these be rewritten as route specs?  That seems a little more natural to me.

mbklein: I looked at route specs, but that would require reaching down into the engines we're mounting. I also couldn't figure out how to write routing tests that take advantage of the state/identity of the logged in user. There's no `login_as` available to routing specs.

config/routes.rb
mbklein: Can do.

spec/features/secure_routes_spec.rb
cjcolvar: ah, got it

config/routes.rb
```diff
@@ -136,14 +136,14 @@
...
+  def route_can?(action, thing, scope=nil)
     lambda do |request|
       warden = request.env['warden']
-      warden.authenticate? && Ability.new(warden.user(scope), warden.session(scope)).can?(:manage, thing)
+      warden.authenticate? && Ability.new(warden.user(scope), warden.session(scope)).can?(action, thing)
     end
   end
 
-  constraints(can_see?(Admin::Collection)) do
+  constraints(route_can?(:manage, Admin::Collection)) do
```
cjcolvar: I think we need a separate ability here like :read, :about_page which will grant access to administrators and a token (for heatlth.yaml checks by Nagios).

mbklein: OK. What are the constraints on that?

cjcolvar: Granting this to administrators is easy.  For Nagios, how about we write a HOWTO wiki page that says to create a user and assign it an api token then add that user to the administrators group?

bmquinn:
@cjcolvar @mbklein @kdid Is this PR ready? The last commit added Resque and Resque Scheduler functionality for the newly-tagged R6 avalon-about gem and with the new ability it looks like that's it for the requirements. Anyone up for testing and merging?

PR #1352, opened 2016-11-04
Upgrade to capistrano 3 and simplify

mbklein:
Capistrano 3.x does a lot of things for us that we were having to do manually (and clunkily) under Capistrano 2.x.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,7 +1,15 @@
-load 'deploy'
+# Load DSL and set up stages
+require "capistrano/setup"
+
+# Include default deployment tasks
+require "capistrano/deploy"
+
+require 'capistrano/rvm'
+require 'capistrano/bundler'
+require 'capistrano/rails/assets'
+require 'capistrano/rails/migrations'
+require 'capistrano/passenger'
 require 'capistrano-resque'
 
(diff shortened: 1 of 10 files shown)


config/browse_everything_providers.yml
```diff
@@ -0,0 +1,20 @@
+#
```
cjcolvar: Did you mean to include this in the PR?

cjcolvar:
With this deployed on spruce, I can't get to the rails console anymore.
@carrickr found this which might be helpful: https://stackoverflow.com/questions/39752623/accessing-rails-console-after-deploy-to-server

mbklein:
The problem is that the avalon user is defaulting to the system ruby, 1.8.7. You need to run `rvm default` before trying to load the console. I don't know why it's doing that; I switched to rbenv on my dev machine months ago.

config/browse_everything_providers.yml
mbklein: Nope. I'll remove it.

cjcolvar:
You're right, I see that I'm not currently on any of the rvm-listed rubies.

phuongdh:
I've put in the workaround to always run `rvm default` on login for avalon user.

PR #1371, opened 2016-11-11
Fix out-of-memory errors with variations bookmark imports

bkeese:


cjcolvar:
Looks fine enough to me. :+1:
@mcwhitaker you worked on this code before, right?  Could you please review and merge?

mcwhitaker:
@cjcolvar no, I did not work on this. I worked on the bookmark export code, which is not rails. Someone more knowledgeable than me should review.

PR #1403, opened 2016-11-22
add in rb-readline

carrickr:
This fixes an issue that emerged on some OS X systems in Oct 2016 where homebrew was not assigning readline to ruby.  Developers would see an error when bringing up the test or dev environments.  Devs either have the option to recompile ruby or else just add the rb-readline gem.  I propose just making rb-readline a dependency for all dev and test environments.


carrickr:
This can wait till post Turkey Day if we want @mbklein to review as well

PR #1425, opened 2016-11-29
Concatenate email from ccid when email is not presented in the shibboleth response. 

weiweishi:


weiweishi:
Sorry wrong repo...

Issue #1409, opened 2016-11-23
Develop proof of concept for track/section based scrubber


davidschober:
## Description
Create a POC for the second scrubber ahead of design work. The POC can be dead simple/minimal interface. A box that takes a number and moves the main scrubber. It should also prove out that we can grab a time offset from an individual track based on metadata. 

## Done

- [x] Some input moves the main scrubber
- [x] Time / Offset is shown for individual track/ section.

## Steps
- [x] Design html/css for scrubber (Brian)
- [x]  Use javascript to create and inject html (relative to existing player/scrubber $(div.mejs-controls) )
- [x]  Write handlers to update scrubber position on 'timeupdate' event ($(@player.media).on 'timeupdate', => @updateScrubber())
- [x]  Move code into new gem
- [x]  Handle mouse interactions (hover over scrubber and click scrubber to seek)
- [x]  Tweak styling
- [x] Add in images for zoom buttons
- [x] Add 'bubble' to regular scubber indicating current track

bkeese:
code for add the bubble: 
https://github.com/avalonmediasystem/avalon/blob/master/app/assets/javascripts/avalon_player.js.coffee#L81-L89

bkeese:
Mallorn now has a working track scrubber. It can be seen here: 
https://mallorn.dlib.indiana.edu/media_objects/avalon:3136

davidschober:
This looks great!

joncameron:
It does! PO approved. I'm moving this to done and will create an issue for further refinement on the track scrubber.

PR #1428, opened 2016-12-01
WIP: Feature/track scrubber

bkeese:
References #1409

A first run at adding a track-based scrubber.


cjcolvar:
I would rather keep this in a branch and switch mallorn to use that branch instead of merging WIP into a -stable branch.

PR #1435, opened 2016-12-02
quick fix to add/delete user rake tasks

csyversen:
This is just a quick fix to the add/delete avalon rake tasks so they work again


cjcolvar:
:+1: good to merge after travis finishes running.

PR #1438, opened 2016-12-06
update rubocop, we should really do this more often

carrickr:
we should probably update rubocop more often


csyversen:
the Gemfile.lock only shows a ldp git repo being added, a parser gem update, and a rdf-xml gem update. Are we missing a Gemfile commit too? @carrickr

carrickr:
@csyversen Looks like some weird case where my branch was behind on rubocop but dev was actually on the latest

PR #1442, opened 2016-12-07
R6 track scrubber

csyversen:
This is the r6 complement to #1439

The only differences between this PR and #1439 is how the js/css assets are handled and the branch of the mediaelement-track-scrubber that we're referencing

### DIFF
--- a/app/views/media_objects/_item_view.html.erb
+++ b/app/views/media_objects/_item_view.html.erb
@@ -53,7 +53,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
     </section>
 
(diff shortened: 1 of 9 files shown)


app/views/media_objects/_item_view.html.erb
```diff
@@ -53,6 +53,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
     </section>
 
+    <%= render file: '_track_scrubber.html.erb' if current_user.present? %>
```
csyversen: done, thanks!

carrickr:
@bkeese Do you want to give final sign off to this?

PR #1452, opened 2016-12-09
Staging

carrickr:
Prep scrubber for 5.x release


Gemfile
```diff
@@ -72,15 +73,15 @@
   end
 
   gem 'avalon-about', git: "https://github.com/avalonmediasystem/avalon-about.git", tag: 'avalon-r4'
-  gem 'about_page', git: "https://github.com/avalonmediasystem/about_page.git", tag: 'avalon-r4'
+  gem 'about_page', git: "https://github.com/avalonmediasystem/about_page.git"
```
cjcolvar: This eventually needs to have a tag because the latest about_page revs in master are for avalon-6 only.

Gemfile
```diff
@@ -72,15 +73,15 @@
...
 
   # You are free to implement your own User/Authentication solution in its place.
   gem 'devise', '~>3.2.0'
   #gem 'devise-guests'
   gem 'haml'
 
   gem 'active_encode', git: "https://github.com/projecthydra-labs/active_encode.git", tag: 'v0.0.3'
-  gem 'rubyhorn', git: "https://github.com/avalonmediasystem/rubyhorn.git", tag: 'avalon-r5'
+  gem 'rubyhorn', git: "https://github.com/avalonmediasystem/rubyhorn.git"
```
cjcolvar: Same is probably true here that it should be tied to a tag.

app/models/api_token.rb
```diff
@@ -0,0 +1,11 @@
+require 'securerandom'
+
+class ApiToken < ActiveRecord::Base
```
cjcolvar: Do we really want to pull in this class and all of the changes around this?  I think not, but I could be wrong.

config/application.rb
```diff
@@ -15,6 +15,7 @@ module Avalon
   class MissingUserId < Exception; end
 
   class Application < Rails::Application
+    require 'avalon/configuration'
```
cjcolvar: I don't think we need the ENV configuration for this release either.

mbklein: Letting ENV configuration merge and then testing it out with existing file configs is probably easier than excising it via cherry-pick. ENV configuration was written to be backward compatible with existing files.

PR #1532, opened 2017-01-13
MasterFile needs to be saved before creating directly contained files

### LINKED ISSUE #1531: Migration doesn't associate poster, thumbnail, or captions with master files
The migrator creates AF::File objects for a master file's poster, thumbnail, and captions datastreams but it does so before the master file has been saved and thus they become top-level objects instead of directly contained by the master file.

Done Looks Like
Migrated master files have poster, thumbnail, and caption directly contained files.

cjcolvar:
Fixes #1531

### DIFF
--- a/app/migration/fedora_migrate/master_file/object_mover.rb
+++ b/app/migration/fedora_migrate/master_file/object_mover.rb
@@ -12,9 +12,10 @@ module FedoraMigrate
       def migrate_datastreams
         migrate_dublin_core
         migrate_desc_metadata
+        migrate_transcoding_metadata
         migrate_title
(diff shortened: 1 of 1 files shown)


cjcolvar:
I see that I pushed another unrelated commit to this branch...I'll move that to a different branch to clean up this PR.

mbklein:
I added one more commit, but now that I read the conversation, it's clear that 1458451 should be its own PR, and f1f14cc and 0a3b455 should be a separate PR.

cjcolvar:
I moved my commit to another branch: f4_migration_db

PR #1554, opened 2017-01-24
Open related item URLs in new tab.

cwant:
Clicking on a related item URL can disrupt playback when that URL opens in the same tab.
This patch makes the related item link open in a new tab.

The display_metadata method needs a new keyword option (allow_target_attribute: true) for
this to work, otherwise the HTML sanitizing will scrub off the "target" attribute for an
anchor, which is what makes this work.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -73,9 +73,16 @@ module ApplicationHelper
     end
   end
 
-  def display_metadata(label, value, default=nil)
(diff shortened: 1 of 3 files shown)


carrickr:
@davidschober @joncameron (Avalon POs).  I know we talked about new window and new tab before and potentially changing the behavior.  What are your feelings on this?  

I am closing this, since we'd want to bring it in on 5.x-stable and then run it through the dev and test servers before pushing into master.

PR #1555, opened 2017-01-24
Open related item URLs in new tab.

cwant:
Clicking on a related item URL can disrupt playback when that URL opens in the same tab.
This patch makes the related item link open in a new tab.

The display_metadata method needs a new keyword option (allow_target_attribute: true) for
this to work, otherwise the HTML sanitizing will scrub off the "target" attribute for an
anchor, which is what makes this work.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -73,9 +73,16 @@ module ApplicationHelper
     end
   end
 
-  def display_metadata(label, value, default=nil)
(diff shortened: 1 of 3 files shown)


carrickr:
Merging in since we have known bugs with forked PRs.  I will revert if the result doesn't pass.

carrickr:
Fixes #1556 for 5.x

Issue #1421, opened 2016-11-29
Database Migration - SelectedItems


joncameron:
### Description
#1418 is the master issue for database migrations and contains the pattern or method to follow when creating each migration.
### Done Looks Like
Tools and/or instructions are written to migrate database tables for **SelectedItems**. This may include migrating identifiers from old avalon:id pids to newly assigned noids.

- [x] add their migration status to the reporting tool
- [x] write a rake task to delete ones that have failed migration
- [x] rename migration report view columns to be "Avalon 5 ID" and "Avalon 6 ID"

cjcolvar:
Dump, load, then migrate document_id  from fedora3 pid to new noid.

cjcolvar:
What should we do if item referenced didn't migrate successfully?  Maybe we should leave the old selected item in place with the fedora3 pid and warn.  This way the migration could be rerun until it succeeds.  If some are still not resolvable then provide a method for removal since they would probably break things if a user tries to go to their selected items when one still has a fedora3 pid.

cjcolvar:
Migrating a selected item should be as simple as:
ActiveFedora::Base.where("identifier_ssim:\"#{fedora3_pid}\"").id

PR #1578, opened 2017-02-08
Add bookmark_cleanup task

lealeelu:
refs #1421 

Adds the avalon:migrate:bookmark_cleanup task to run after all bookmarks have been moved over. It checks to make sure the MediaObject it references exists. If it doesn't exist, the Bookmark is destroyed.

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -19,7 +19,7 @@ namespace :avalon do
     require 'active_fedora/cleaner'
     ActiveFedora::Cleaner.clean!
   end
-  
+
   desc "Migrate Avalon 5.x to 6.x"
   task migrate: :environment do
     Rake::Task['avalon:migrate:repo'].invoke
@@ -38,7 +38,7 @@ namespace :avalon do
       migrator.migrate_objects
       migrator
     end
-    
+
     desc "Migrate my database"
     task db: :environment do
       Bookmark.all.each do |b|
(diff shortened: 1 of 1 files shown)


lib/tasks/avalon.rake
```diff
@@ -89,6 +89,23 @@ namespace :avalon do
         end
       end
     end
+
+    desc "Cleanup failed bookmarks"
+    task bookmark_cleanup: :environment do
+      Bookmark.all.each do |b|
+        if MediaObject.where(id: b.document_id).count > 0
```
cjcolvar: This seems a bit chatty.

lib/tasks/avalon.rake
```diff
@@ -89,6 +89,23 @@ namespace :avalon do
...
+          puts "Bookmark pid: #{b.id} exists"
+        else
+          puts "Bookmark pid: #{b.id} does not exist. Deleting...."
```
cjcolvar: How about moving this to after the delete and say deleted instead of deleting?

lib/tasks/avalon.rake
```diff
@@ -89,6 +89,23 @@ namespace :avalon do
...
+          begin
+            b.destroy
+          rescue Exception => e
+            puts "Failed to delete"
```
cjcolvar: Should these be puts or Rails.logger ?

lealeelu: Everything else was puts. I checked before deciding on puts. Now that we're getting to something that will be used in production, maybe we could change everything to logger?

lib/tasks/avalon.rake
lealeelu: can do

lib/tasks/avalon.rake
lealeelu: Do you mean the lines below this? I wanted to make sure it worked.

cjcolvar: I was meaning the existance message.

lib/tasks/avalon.rake
cjcolvar: Hmmm...let me see what is recommended for this context (rake).

bkeese: I recommend output like this:

error message 1
error message 2
...
z bookmarks failed to deleted
y bookmarks deleted
x total bookmarks inspected

cjcolvar:
:+1:  I'll wait for travis to merge

cjcolvar:
Fixes #1421

PR #1587, opened 2017-02-10
fix facet count width

### LINKED ISSUE #1586: Facet Counts Wrap Incorrectly
### Description
Facet counts break in the middle of long (3 digit and more) numbers, leading to unsightly spacing. This story will be timeboxed, and this issue will be worked on in a future point release if the scope is too broad.

### Example from Mallorn on 2/10/17, running Avalon 6.0
screen shot 2017-02-10 at 10 38 55 am

### Done Looks Like
There are no line breaks in facet counts.

bkeese:
Fixes #1586 

Fixes width of facet counts by overriding blacklight's javascript.

Stock blacklight creates an object used to set the width for the facet count column and attaches that object to the body node. It inherits css from body that causes the width to be incorrect. This change attaches the object to the facet count itself, so it inherits the correct css.

### DIFF
--- a/app/assets/javascripts/blacklight/facet_load.js
+++ b/app/assets/javascripts/blacklight/facet_load.js
@@ -1,0 +1,23 @@
+/*global Blacklight */
+
+(function($) {
+  'use strict';
+  
+  Blacklight.doResizeFacetLabelsAndCounts = function() {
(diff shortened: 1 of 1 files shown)


cjcolvar:
:+1: waiting for travis

PR #1589, opened 2017-02-10
fix fix

bkeese:
Fixes #1586 
Fixes #1025 (titles wrap relatively cleanly)
Fixes #1054

Fixes facet count width for facets that are closed when page is loaded.

PR #1590, opened 2017-02-10
add in no wrap for just the facet-value.  This will work up to five i…

### LINKED ISSUE #1025: Long titles and numbers wrapped improperly in facets

### LINKED ISSUE #1054: Facet counts shouldn't wrap
Expected:
browse - avalon media system search 2016-08-11 12-26-55

Observed:
browse - avalon media system search 2016-08-11 12-27-23

I believe adding
```css
.facet-count { white-space: nowrap; }
```
should do the trick.

### LINKED ISSUE #1586: Facet Counts Wrap Incorrectly

carrickr:
After speaking with POs, since this is a time boxed fix the best move was to set facet-count to no-wrap.  I believe we do not need to expand max-width for the sidebar style, but if desired for look we can.  

Fixes #1586 
Fixes #1025 (titles wrap relatively cleanly)
Fixes #1054 

This is an example of what we end up with:
screen shot 2017-02-10 at 3 23 27 pm

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -582,6 +582,7 @@ a[data-trigger='submit'] {
 
 .facet-values {
   display: table;
+  //white-space: nowrap; I don't think we want values to no wrap because collection names can be long
   table-layout: fixed;
   //set hyphen definitions at this (display:block) level
   -ms-word-break: keep-all;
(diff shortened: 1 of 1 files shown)


bkeese:
The other PR (https://github.com/avalonmediasystem/avalon/pull/1589) directly fixes the problem.

PR #1621, opened 2017-02-16
mbk and clr quick and dirty fix for a has many bug in governing policies

### LINKED ISSUE #1614: Leases are applied to all items, regardless of user type or collection
### Description
Leases created on a media object get copied to every media object in the Avalon instance, across collections and units. Also, newly created media objects immediately have pre-existing leases.

### Steps to Reproduce
1. Add a lease to a media object. Level of user and date range doesn't matter.
2. Look at any other item in the parent collection of the object to which the lease was added. Each item now has that exact lease.

### Done Looks Like
Leases only affect a single object from the standard access control edit page.

### Original Write-up from Julie
A lease created on https://mallorn.dlib.indiana.edu/media_objects/ng451h56v with a user who is a default manager is being applied to every item on mallorn in every collection. A lease created on https://mallorn.dlib.indiana.edu/media_objects/8g84mm35j with a user who does not have access to manage any collections is also being applied to all items on mallorn in every collection.

carrickr:
Fixes #1614 

Basically we uncovered an issue with the has many and belongs to association in ActiveFedora.  I have in a temporary fix that @mbklein suggested.  mbk will also put in a ticket and failing test case to ActiveFedora so that in the long run the bug can be fix.

### DIFF
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -286,9 +286,7 @@ class MediaObject < ActiveFedora::Base
   end
 
   def leases(scope=:all)
-    criteria = { has_model_ssim: 'Lease' }
-    criteria.merge!(lease_type_ssi: scope) unless scope == :all
-    governing_policies.where(criteria)
+    governing_policies.select { |gp| gp.is_a?(Lease) and (scope == :all or gp.lease_type == scope) }
   end
 
   private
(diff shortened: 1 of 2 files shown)


bkeese:
+1 pending travis

cjcolvar:
Can someone please explain why the tests were removed?

PR #1632, opened 2017-02-20
try calling in a virtual group over a lease when there is no external…

### LINKED ISSUE #1626: Leases Should Accomodate Lack of Start or End Date
### Description 
When creating a lease, the start or end date should be allowed to be unspecified. This is a requirement for feature parity with 5.x. 

### Done Looks Like
Leases can be created with:
- a start date and an end date
- a start date but no end date
- an end date but no start date

carrickr:
Fixes #1626 

Leases without start and end dates just end up as general governing policies that have type 'external', 'ip', etc.  They are returned in solarization of the record and thus display properly, but are no longer returned as a lease due to the code change in #1614.  This deals with removing those obsolete tests.  After merging this in 1614, PR #1621 should then be green to merge.

### DIFF
--- a/spec/models/media_object_spec.rb
+++ b/spec/models/media_object_spec.rb
@@ -485,17 +485,25 @@ describe MediaObject do
       expect(solr_doc['section_label_tesim']).to include('CD 1')
       expect(solr_doc['section_label_tesim']).to include('Test Label')
     end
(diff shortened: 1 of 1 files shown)


bkeese:
Did you mean to xit those tests?

carrickr:
@bkeese Yes, those are two are now just governing policies

PR #1639, opened 2017-02-22
Fix default active masterfile

bkeese:
Fixes default active masterfile for media_object show page.

Uses ordered_master_files instead of master_files to get first section.


cjcolvar:
It would be really nice to have a test for this.

PR #1650, opened 2017-02-27
Revert "try calling in a virtual group over a lease when there is no …

cjcolvar:
…external date (#1632)"

This reverts commit e3ed23be14fcce038e91efbbc7f24ba9cae68e34.

These tests cover proper indexing at the media object level of leases and should continue to exist.

### DIFF
--- a/spec/models/media_object_spec.rb
+++ b/spec/models/media_object_spec.rb
@@ -485,25 +485,17 @@ describe MediaObject do
       expect(solr_doc['section_label_tesim']).to include('CD 1')
       expect(solr_doc['section_label_tesim']).to include('Test Label')
     end
-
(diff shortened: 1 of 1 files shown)


carrickr:
Disagree, with the new model they're not leases anymore.  If we want them to be leases we need improve the entire leases model to handle leases without date ranges rather than the current governing policies + leases noise and that seems like a 6.1 improvement that this place.

cjcolvar:
Governing policies is how leases are exposed to gated discovery and enforced.  Leases without date ranges is a feature improvement in my opinion so would have to wait until 6.1.  So I don't see the problem with turning these tests back on since they are testing current expected behavior.

mbklein:
Merged after discussion

PR #1655, opened 2017-02-27
Fix deleting media objects with multiple sections

### LINKED ISSUE #1634: Error while deleting an Item with multiple sections
### Descriptive summary

Testing: Regression
Browser: Chrome
Issue: The delete operation on an item that contains multiple sections is resulting in a error page. If we search for the item, it is still present in the list, and accessing it again is navigating to a new error page.
This issue is not seen if the item contains only one file/section

1. Login to the spruce system
2. Create an item in  collection, the item should contain two files/sections added to it.
3. After the creation process is done, click on the item and perform the delete operation
4. It will navigate to an error page displaying "Ldp::Gone in MediaObjectsController#destroy"
5. Go back to the home page, and search for the deleted item. This returns the item, click on this item.
6. This navigates to another error page with "NoMethodError"

Note: I have created an item "test_n3" with two sections, for a one time use to reproduce the error.
### Related work

deleting object issue

error accessing deleted object

cjcolvar:
Fixes #1634 

I wrote the tests first and they passed but I still got the error when testing manually through the interface.  I copied how we are removing master files from media objects in the media object controller update action.  This appears to fix things and the new tests still pass.

### DIFF
--- a/spec/controllers/media_objects_controller_spec.rb
+++ b/spec/controllers/media_objects_controller_spec.rb
@@ -711,12 +711,23 @@ describe MediaObjectsController, type: :controller do
       login_user collection.managers.first
     end
 
-    it "should remove the MediaObject and MasterFiles from the system" do
+    it "should remove a MediaObject with a single MasterFiles" do
       media_object = FactoryGirl.create(:media_object, :with_master_file, collection: collection)
       delete :destroy, id: media_object.id
       expect(flash[:notice]).to include("success")
       expect(MediaObject.exists?(media_object.id)).to be_falsey
       expect(MasterFile.exists?(media_object.master_files.first.id)).to be_falsey
+    end
+
+    it "should remove a MediaObject with multiple MasterFiles" do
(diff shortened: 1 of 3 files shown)


spec/controllers/media_objects_controller_spec.rb
```diff
@@ -711,14 +711,25 @@
...
 
+    it "should remove a MediaObject with multiple MasterFiles" do
+      media_object = FactoryGirl.create(:media_object, :with_master_file, collection: collection)
+      FactoryGirl.create(:master_file, media_object: media_object)
+      master_file_ids = media_object.master_files.map(&:id)
+      media_object.reload
+      delete :destroy, id: media_object.id
+      expect(flash[:notice]).to include("success")
+      expect(MediaObject.exists?(media_object.id)).to be_falsey
+      master_file_ids.each { |mf_id| expect(MediaObject.exists?(mf_id)).to be_falsey }
```
bkeese: This should be MasterFile.exists?, right?

PR #1662, opened 2017-02-28
Bugfix/1646

### LINKED ISSUE #1646: Manager is unable to delete a collection
### Descriptive summary

Testing: Regression
Issue: A manager level user is unable to delete a collection from the system, it is redirecting to the home page with a message stating that user is unauthorized to perform the operation.

1. Login to spruce system, with a Manager credentials not an Administrator
2. Go to Manage Content page, which lists the collections that this user is a part of
3. Click on Delete button for any collection
collection delete button
4. The user is navigated to the home page with a message ribbon saying that the user is not authorised to perform the operation.
unauthorized manager to delete collection

### Related work

carrickr:
Fixes #1646 

Give managers the manage ability

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -54,7 +54,11 @@ class ApplicationController < ActionController::Base
 
   def get_user_collections
     if can? :manage, Admin::Collection
-      Admin::Collection.all
+      collections = []
+      Admin::Collection.all.each do |collection|
+        collections << collection if can? :manage, collection
+      end
+      collections
     else
       Admin::Collection.where("inheritable_edit_access_person_ssim" => user_key).to_a
     end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -67,15 +67,6 @@ class Ability
 
   def custom_permissions(user=nil, session=nil)
 
-    # unless full_login? and @user_groups.include? "manager"
-    #   can :manage, Admin::Collection do |collection|
-    #     @user.in?(collection.managers)
-    #   end
-    #   can :manage, MediaObject
-    #   can :manage, MasterFile
-    #   can :manage, Admin::Group
-    # end
-
     unless full_login? and @user_groups.include? "administrator"
       cannot :read, MediaObject do |media_object|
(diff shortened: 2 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -54,7 +54,11 @@ def handle_api_request
 
   def get_user_collections
     if can? :manage, Admin::Collection
-      Admin::Collection.all
```
cjcolvar: I think this could be left with Admin::Collection.all since the ability to manage Admin::Collection as a class instead of an instance of the class implies all instances.

app/models/ability.rb
```diff
@@ -55,6 +55,9 @@ def create_permissions(user=nil, session=nil)
 
       if @user_groups.include? "manager"
         can :create, Admin::Collection
+        can :manage, Admin::Collection do |collection|
```
cjcolvar: It looks like before we didn't grant blanket manage ability to managers and instead went piecemeal but I don't remember why and can't think of a reason right now why not to do it this way.  So I guess I'm just noting this for the record.

app/controllers/application_controller.rb
carrickr: If you do this, you get all collections (even collections you are not the manager of) on the index screen.

carrickr: @csyversen @mbklein

carrickr:
@cjcolvar Talked to the POs and they feel managers should only be able to see collections they can manage

cjcolvar:
Yes, I'm in agreement.  I think the code in question with the changes I proposed though will only display all collections for administrators and for managers/editors/depositors only collections that they have access to.  If not, then I think there is something wrong with our Ability.

carrickr:
@cjcolvar To clarify, I had tested the code you proposed on my dev branch before submitting the PR and I was seeing things I should not when logged in as a manager.  So something is wrong with the ability.  However this restriction does resolve it and this is the only function where we get can manage collections, so fixing it here does resolve the issue for 6.0 and we can look at a larger ability rewrite in 6.x.

PR #1664, opened 2017-03-01
reset the create collection button after a failure

### LINKED ISSUE #1643: After an error in create collection, the "Create Collection" button in pop-up is disabled for next attempt
### Descriptive summary

Testing: Regression
Issue: If the Name field is left empty while creating a collection and clicking on Create Collection button generates an error message to notify the user that Name field is empty. After this message is generated; for the next attempt to create collection the button is displaying text "Saving..." and is disabled even if the user enters text in Name field this time.

1. Log in to spruce system with manger or administrator level credentials.
2. Click on Manage Content option
3. Click on "Create Collection" button in the page, it opens a pop-up to enter the details
4. Do not enter any text in the Name field and click on "Create Collection" button
createcolletcion pop-up

5. It generates an error message
createcollection error msg

6. Click on the Create Collection button again to open the pop-up, the save button is disabled, and it is not re-enabling even though we enter the Name
createcollection save button disabled

### Related work

carrickr:
Fixes #1643 

I'm hoping Brian or Chris S might know a more elegant way to do this (reseting the whole form didn't do it), but if not, this works well enough for just enabling one button.

### DIFF
--- a/app/views/admin/collections/_form.html.erb
+++ b/app/views/admin/collections/_form.html.erb
@@ -48,10 +48,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       if (xhr.responseJSON.hasOwnProperty('errors')){
         $(this).closest('.modal').modal('hide')
         inject_alert(xhr.responseJSON['errors'].join('<br/>'));
-        button = document.getElementsByName('commit')[0]
-        button.value = 'Create Collection'
-        button.disabled = false;
-        button.classList.remove('disabled');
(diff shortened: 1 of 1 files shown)


csyversen:
as far as js asset handling goes i think this is fine for right now. We could move every bit of js into their own js files if we wanted to, but it'd require a bit of work to track them all down and make sure they work and i don't think we want to put the effort into that at the moment

cjcolvar:
+1 to trying @bkeese's suggestion

PR #1687, opened 2017-03-03
Remove form-control class so btn-primary and btn-warning styles aren't overridden

### LINKED ISSUE #1682: Style enhancements for delete collection page buttons
The text on the confirm and cancel buttons is black and hard to read.  Change the text to white to match media object confirm delete page.

cjcolvar:
Fixes #1682.

@bkeese can you verify that this doesn't break anything else?  I see that you added the form-control classes 2 years ago here: https://github.com/avalonmediasystem/avalon/commit/f5350652e2debb7f829c0b13e5a60544e2175eb3

### DIFF
--- a/app/views/admin/collections/remove.html.erb
+++ b/app/views/admin/collections/remove.html.erb
@@ -42,8 +42,8 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
       <div class="row-fluid form-inline">
 	<div class="form-group">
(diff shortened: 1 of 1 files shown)


bkeese:
If the buttons still work without the class, then I see no problem removing it.

cjcolvar:
Yup, they appear to when I remove that class using Chrome's developer tools.

bkeese:
We'll need to make sure the buttons still work when class is missing at page load (when handlers are set up) too.

Issue #1676, opened 2017-03-03
merge and test extreme performance


davidschober:
Run Regression Testing on extreme performance branch.

- [ ]  Run regression test and make final decision. (possibly covered by #1624 if stars align)

davidschober:
covered under #1444

PR #1690, opened 2017-03-08
Extreme performance

### LINKED ISSUE #1679: Final refactor of extreme performance
- [x] final changes pushed

mbklein:
Closes #1679 ; refs #1676 

Use Solr for access; fall back to Fedora only when absolutely necessary

This change overhauls the `MediaObjectsController#show` logic to take advantage of the [`speedy-af`](https://github.com/projecthydra-labs/speedy_af) gem and render output from solr.

Changes proposed in this pull request:
* Extensive reworking of `MediaObjectsController` and the `MasterFile`, `Derivative`, and `StructuralMetadata` modules to provide interface-compatible, SpeedyAF-ready access
* Add an `OPTIMIZED_DEV` environment variable that turns on caching and asset bundling while retaining other aspects of dev mode.

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -131,7 +131,7 @@ EOC
   end
 
   desc "Index MasterFiles and subresources to take advantage of SpeedyAF"
-  task :index_for_speed do
+  task index_for_speed: :environment do
     MasterFile.find_each do |mf|
       $stderr.print "m["
       mf.update_index;
@@ -143,7 +143,7 @@ EOC
     end
     $stderr.puts
   end
-  
+
   namespace :services do
     services = ["jetty", "felix", "delayed_job"]
     desc "Start Avalon's dependent services"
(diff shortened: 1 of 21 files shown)


lib/tasks/avalon.rake
```diff
@@ -129,6 +129,21 @@ EOC
...
   end
+
+  desc "Index MasterFiles and subresources to take advantage of SpeedyAF"
+  task index_for_speed: :environment do
+    MasterFile.find_each do |mf|
+      $stderr.print "m["
+      mf.update_index;
+      mf.declared_attached_files.each_pair do |name, file|
+        $stderr.print name.to_s[0]
+        file.update_external_index if file.respond_to?(:update_external_index)
```
cjcolvar: Does this indexing happen normally and this rake task is just for migration?

cjcolvar: Nevermind...found it: https://github.com/projecthydra-labs/speedy_af/blob/master/lib/speedy_af/indexed_content.rb#L6-L8

PR #1717, opened 2017-03-10
Blacklight helper fix for development environment.

cwant:
After code changes, the development environment reloads in a
way that it uses the wrong Blacklight helper definitions.
This change ensures it always uses our Blacklight helpers.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -23,6 +23,22 @@ class CatalogController < ApplicationController
 
   before_filter :save_sticky_settings
 
+  if Rails.env.development?
+    # Hot-reload of rails confuses which Blacklight helpers it wants to use,
+    # we beat it into submission so that it uses ours ...
+    include ActionView::Helpers::TagHelper
+    include ActionView::Context
+    include ActionView::Helpers::NumberHelper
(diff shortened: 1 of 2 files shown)


cjcolvar:
Is it possible to move this into a monkey-patch initializer?  It would make it easier to throw away and not clutter up the catalog controller.

cwant:
I did not have luck experimenting with an initializer. This might be because the root of the problem is that the app initializes fine, but the main app and the blacklight engine have problems reloading.

I have successfully extracted the patch into a controller concern -- would that work better? (So only one line added to the catalog controller, which can easily be commented out.)

mbklein:
Yeah, I think a controller concern would be cleaner.

cwant:
Update pushed with this fix as a concern.

cjcolvar:
I'm going to wait on this until we cut a release, hopefully today, and then I'll merge it.

carrickr:
I need to delete the old 5.x stable for our release so I have to close this, otherwise github won't let me delete.  @cwant in about 30 seconds there will be a 5.x-dev you can submit to.

cwant:
Done. Here is the new PR: https://github.com/avalonmediasystem/avalon/pull/1746

PR #1731, opened 2017-03-13
Use #id instead of removed #pid

### LINKED ISSUE #1730: Playlist is hanging at the end of first video in the list, and not giving access to remaining items
### Descriptive summary
Testing: Regression
Browser: Chrome(56.0.2924.87 (64-bit))
Operating System: Windows10
Issue: If the playlist contains more than one item, I cannot access the items other than the first item in the list. And player is hanging at the end after playing the first item

1. Login to the spruce system
2. Add two items to a new playlist
3. Try to access the items other than first item (not accessible)
4. Play the first item, and wait until it reaches the end of the timeline (player hangs at this point)

### Related work

cjcolvar:
Fixes #1730.

### DIFF
--- a/app/views/playlists/refresh_info.js.erb
+++ b/app/views/playlists/refresh_info.js.erb
@@ -21,10 +21,10 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% clip_start = @current_clip.start_time / 1000.0 %>
 <% clip_end = @current_clip.end_time / 1000.0 %>
 <% clip_frag = "?t=#{clip_start},#{clip_end}" %>
-<% link = pid_section_media_object_path(@current_mediaobject, @current_masterfile.pid) + clip_frag %>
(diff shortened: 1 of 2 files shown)


app/views/playlists/refresh_info.js.erb
```diff
@@ -21,10 +21,10 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% clip_start = @current_clip.start_time / 1000.0 %>
 <% clip_end = @current_clip.end_time / 1000.0 %>
 <% clip_frag = "?t=#{clip_start},#{clip_end}" %>
-<% link = pid_section_media_object_path(@current_mediaobject, @current_masterfile.pid) + clip_frag %>
+<% link = pid_section_media_object_path(@current_mediaobject, @current_masterfile.id) + clip_frag %>
```
bkeese: I think this needs to be id_section_media_object_path

cjcolvar: good catch

Issue #1339, opened 2016-11-04
Final 6.0 Code Merge


joncameron:
## Code 
- [x] Update license headers in files
- [x] Tag dependency repos
  - [x] Create release of fedora-migrate
- [x] Update gemfile to point to those tags
- [ ] 6.0 merged into master
- [ ] Create tag for v6.0.0

PR #1734, opened 2017-03-13
Tag gemfile dependencies for R6

lealeelu:
Task from #1339

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,15 +1,14 @@
 source 'https://rubygems.org'
-gem 'fedora-migrate', git: 'https://github.com/projecthydra-labs/fedora-migrate.git'
+gem 'fedora-migrate', '~> 0.5.0'
 gem 'rdf-rdfxml'
 
 gem 'hydra-head', '~> 10.3.4'
 gem 'active-fedora', '>= 10.3.0'
 gem 'active_fedora-datastreams'
 gem 'active_fedora-noid', '~> 2.0.0'
-gem 'speedy-af', '~> 0.1.0', git: 'https://github.com/projecthydra-labs/speedy_af.git'
+gem 'speedy-af', git: 'https://github.com/projecthydra-labs/speedy_af.git', tag: 'v0.1.0'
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -119,27 +120,20 @@ GIT
...
       activesupport
 
 GIT
   remote: https://github.com/projecthydra/ldp.git
-  revision: a988e5fa5d94eb0f3f0b3a27d1190f632048736b
+  revision: b528e6fd3079cc1058bdeb3a09aa5dc81486b85d
+  tag: v0.6.4
   specs:
-    ldp (0.6.2)
+    ldp (0.6.4)
```
lealeelu: Is it ok that we're going to 0.6.4? That's the latest from 20 days ago.

Gemfile
```diff
@@ -1,15 +1,15 @@
...
 gem 'hydra-head', '~> 10.3.4'
 gem 'active-fedora', '>= 10.3.0'
 gem 'active_fedora-datastreams'
 gem 'active_fedora-noid', '~> 2.0.0'
-gem 'speedy-af', '~> 0.1.0', git: 'https://github.com/projecthydra-labs/speedy_af.git'
+gem 'speedy-af', git: 'https://github.com/projecthydra-labs/speedy_af.git', tag: 'v0.1.0'
 gem 'blacklight', '~> 6.6'
 gem 'rdf', '~> 2.1.1'
-gem 'ldp', git:'https://github.com/projecthydra/ldp.git'
+gem 'ldp', git:'https://github.com/projecthydra/ldp.git', tag: 'v0.6.4'
```
cjcolvar: Looks like we pegged this to master to get a bug patch, but using the 0.6.4 released gem should be fine or even taking this line out entirely might work.

lealeelu:
Removed ldp line of Gemfile. Looks like it shows back up in the dependencies just fine.

mbklein:
:+1: Looks good to me.

cjcolvar:
👍 Looks good to me too.

PR #1747, opened 2017-03-14
Cwant/dev reload fix for avalon upstream

cwant:
Third time's a charm, see previous PRs, particularly: https://github.com/avalonmediasystem/avalon/pull/1717

After code changes, the development environment reloads in a
way that it uses the wrong Blacklight helper definitions.
This change ensures it always uses our Blacklight helpers.


cwant:
I guess the irony is that since travis runs the test suite (I presume), this code will never get touched since it only runs in the dev environment.

PR #1790, opened 2017-03-27
Convert Comment to a hash before passing to mailer, and then to an OpenStruct for processing

### LINKED ISSUE #1789: Fix comment submission error
When submitting a comment on PawPaw, ActiveJob can't serialize the Comment class, so it crashes with a "We're sorry, but something went wrong".

Here's the error from the production log:
https://gist.github.com/lealeelu/3a9042f3d04f1baa9ebce0ecb0fba466

http://api.rubyonrails.org/classes/ActiveJob/SerializationError.html
There are limited types that ActiveJob can take, and our Comment Class is not one of them. If we could convert it to json and pass that through instead, it would probably work.

I think this is a problem with 6.0 but haven't verified in other places yet.

mbklein:
Fixes #1789 

ActiveJob can't serialize a `Comment`, but it can serialize a `Hash`. By leveraging `OpenStruct`, we can convert the `Comment` to a `Hash` for serializing, and then back to something that quacks like a `Comment` for rendering.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -31,16 +31,16 @@ class CommentsController < ApplicationController
 
     if (@comment.valid?)
       begin
-	CommentsMailer.contact_email(@comment).deliver_later
+        CommentsMailer.contact_email(@comment.to_h).deliver_later
       rescue Errno::ECONNRESET => e
-	logger.warn "The mail server does not appear to be responding \n #{e}"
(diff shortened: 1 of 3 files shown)


cjcolvar:
I believe these tests are failing because the branch was cut from master instead of develop.  I cherry picked this commit to a branch cut from develop and created #1791.  After it is merged to develop, we should merge it to staging and then to master.

cjcolvar:
Closing due to #1791 being merged.

PR #1795, opened 2017-03-29
Update develop to latest master

phuongdh:
Looks like someone committed directly into master


cjcolvar:
Let's wait on this a second until master gets sorted out.

PR #1812, opened 2017-04-06
Make predicate method Avalon::Batch::Entry#valid? return a boolean.

cwant:
The predicate method Avalon::Batch::Entry#valid? doesn't return anything (it returns nil).
It should return a boolean value.


cwant:
I have added some tests for this method that cover it's behaviour well.

cjcolvar:
Looks great! 💯

PR #1815, opened 2017-04-11
Update mediaelement_rails gem with bugfix

### LINKED ISSUE #1798: Chrome Flash audio embeded player doesn't load flash automatically
Embeded audio that requires flash won't load automatically because Chrome blocks the plugin. It doesn't happen to embeded video.

Might be related to #1647, or maybe both can be fixed.

lealeelu:
fixes #1798 
Updated the gemfile to pull the latest mediaelement_rails.


bkeese:
Shouldn't this change be made here and mediaelement_rails rebuild from that: https://github.com/avalonmediasystem/mediaelement/blob/captions/src/js/me-shim.js#L434

lealeelu:
@bkeese I'll try that out and see if I get anything different.

PR #1833, opened 2017-04-14
Allow multiple queries to be configured for SRU bib import

bkeese:


spec/lib/avalon/bib_retriever_spec.rb
```diff
@@ -77,6 +77,28 @@
         expect(Nokogiri::XML(response)).to be_equivalent_to(mods)
       end
     end
+
+    describe 'multiple configured queries' do
+      let(:sru_url) { "http://zgate.example.edu:9000/db?version=1.1&operation=searchRetrieve&maximumRecords=1&recordSchema=marcxml&query=cql.serverChoice=%27%5EC%5E%257763100%27" }
```
cjcolvar: I think it would be clearer if you swap sru_url and sru_url2 since sru_url2 is what is queried first according to the order in the Avalon::Configuration.

spec/lib/avalon/bib_retriever_spec.rb
```diff
@@ -77,6 +77,28 @@
...
+        Avalon::Configuration['bib_retriever'] = { 'protocol' => 'sru', 'url' => 'http://zgate.example.edu:9000/db', 'query' => ["rec.id='not_a_real_id'","cql.serverChoice='^C%{bib_id}'"] }
+        FakeWeb.register_uri :get, sru_url, body: sru_response
+        FakeWeb.register_uri :get, sru_url_2, body: ''
+      end
+
+      after :each do
+        FakeWeb.clean_registry
+      end
+
+      it 'retrieves proper MODS' do
```
cjcolvar: I think this is good, but I think you also need to test that both queries are being executed.  Maybe you could do that with:
expect(RestClient).to receive(:get).with(sru_url)
expect(RestClient).to receive(:get).with(sru_url2)

PR #1836, opened 2017-04-19
Strip whitespace from query string for bib import

bkeese:


spec/models/media_object_spec.rb
```diff
@@ -596,19 +596,37 @@
...
+        expect { media_object.descMetadata.populate_from_catalog!(bib_id, 'local') }.to change { media_object.title }.to "245 A : B F G K N P S"
+      end
     end
-    it 'should not override resource_type' do
-      expect { media_object.descMetadata.populate_from_catalog!(bib_id, 'local') }.to_not change { media_object.resource_type }
+    describe 'should strip whitespace from bib_id parameter' do
+      let(:sru_url) { "http://zgate.example.edu:9000/db?version=1.1&operation=searchRetrieve&maximumRecords=1&recordSchema=marcxml&query=rec.id=#{bib_id}" }
+      let(:sru_response) { File.read(File.expand_path("../../fixtures/#{bib_id}.xml",__FILE__)) }
+      let!(:request) { stub_request(:get, sru_url).to_return(body: sru_response) }
+      before do
```
cjcolvar: Do you need this before block to override the parent block?  If so, why not push the parent before into the describe blocks that use it instead of overriding it here?  If not, please remove it.

PR #1870, opened 2017-05-03
Update README.md

carrickr:
remove unneeded line


cjcolvar:
I think this is a bug and not a problem with the instructions.  I think we need to revert back to the secrets.yml.example and cut a new release when it makes sense since this file shouldn't have been committed in the first place or at least not with secrets in it.

cjcolvar:
I take back my previous comment and think that this change is fine after reviewing other Hydra apps in github.

Issue #1583, opened 2017-02-09
Ensure migration is re- runnable


davidschober:
## DONE

Migration is re- runnable without having to clear system first (e.g. proxies are duped, etc).

- [ ] investigate why migration is not rerunnable 
- [ ] Figure out the dupes

joncameron:
This is more about picking up when it left off, and returning to get items that were not covered.

cjcolvar:
And failed items as well.

Issue #927, opened 2016-06-24
Create multi-threaded migration script for Fedora and Solr


phuongdh:
### Done Looks Like

Once the single-threaded migration is done, @carrickr can talk through the parallel gem.  
- [ ] create documentation for users on how to tune so they don't blow up machine by spawning Fedora and SOLR threads. 
- [x] Re-indexing is multi-threaded
- [x] The script contains the ability to specify the number of threads to use

davidschober:
@carrickr to talk through it at a swarm, then decided how to attack it.

joncameron:
@carrickr Has any work been done on this yet? Or should this issue be mainly for doing the testing and documentation?

davidschober:
@cjcolvar you grabbed this as part of sprint 126 it looks like. I'll add it to the milestone.

cjcolvar:
:+1:

davidschober:
@cjcolvar it looks like this got merged in, this is done, correct?

cjcolvar:
I was leaving this open because documentation hasn't been written yet.

davidschober:
Sounds good.

PR #1890, opened 2017-05-06
Run migration in parallel threads

cjcolvar:
This PR runs the migration in parallel threads defaulting to number of processors - 2 if not specified in the rake command.  This PR also attempts to handle rerunning issues by deleting the object if the migration fails or before rerunning the migration of an object.


cjcolvar:
Fixes #1583.  This PR does not overwrite previously migrated records that may have changed since the last run.

cjcolvar:
Fixes #927.

PR #1897, opened 2017-05-09
Eliminate empty strings when updating attributes

### LINKED ISSUE #1667: R6 Notes displayed for empty table of contents
## Description

Seen on spruce and mallorn, for some items without any defined notes or table of contents, the item view page displays a header for Notes/CONTENTS even when no value is present.

## Done looks like

The header should not be displayed if no notes or table of content is present.

- [ ] View partial is updated to handle empty list OR different solution is found

bkeese:
Fixes #1667 

Eliminate empty strings when updating attributes.

When the UI form is submitted, empty strings are submitted for those fields without values. This PR removes empty strings from multi-value fields.

### DIFF
--- a/app/models/concerns/media_object_mods.rb
+++ b/app/models/concerns/media_object_mods.rb
@@ -86,7 +86,7 @@ module MediaObjectMods
 
   def alternative_title=(value)
     delete_all_values(:alternative_title)
-    Array(value).each { |val| descMetadata.add_alternative_title(val) } if value.present?
+    Array(value).each { |val| descMetadata.add_alternative_title(val) if val.present? }
   end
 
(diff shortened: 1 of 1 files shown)


app/models/concerns/media_object_mods.rb
```diff
@@ -336,7 +336,7 @@ def table_of_contents
   end
   def table_of_contents=(value)
     delete_all_values(:table_of_contents)
-    descMetadata.table_of_contents = value if value.present?
+    descMetadata.table_of_contents = Array(value).compact if value.present?
```
cjcolvar: Looks like `.compact` only removes nil values and will leave empty strings.

PR #1908, opened 2017-05-11
Remove redundant override

cjcolvar:
This Blacklight override is already happening in app/assets/javascripts/blacklight/facet_load.js

### DIFF
--- a/app/assets/javascripts/avalon.js
+++ b/app/assets/javascripts/avalon.js
@@ -17,17 +17,6 @@
 // Empty file for future js
 /* Override the search_context so it stops POSTing links which confuses
  * Rails and causes it to redirect to the wrong place. */
-$(document).ready(function() {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/avalon.js
```diff
@@ -17,17 +17,6 @@
 // Empty file for future js
 /* Override the search_context so it stops POSTing links which confuses
  * Rails and causes it to redirect to the wrong place. */
-$(document).ready(function() {
```
bkeese: Need to keep this line.

app/assets/javascripts/avalon.js
```diff
@@ -17,17 +17,6 @@
...
-  Blacklight.do_search_context_behavior = function() {}
```
bkeese: Need to keep this line?

PR #1919, opened 2017-05-15
Allow using file of pids to migrate along with comma delineated list …

cjcolvar:
…of pids on commandline


lib/tasks/avalon.rake
```diff
@@ -40,6 +40,7 @@ EOC
         exit 1
       end
       ids = ENV['pids'].split(',') unless ENV['pids'].nil?
+      ids = Array(ids) + File.readlines(ENV['pidfile']).map(&:strip) unless ENV['pidfile'].nil?
```
bkeese: How about using | instead of +, to eliminate duplicates.

PR #1921, opened 2017-05-15
Silence deprecation warnings in production to keep the logs cleaner

cjcolvar:


cjcolvar:
The failed travis builds are because I cancelled them since they are pointless as the only change is to the production environment config file.

PR #1923, opened 2017-05-15
Secrets example

bkeese:


cjcolvar:
Looking at other Hydra apps up on github it looks like many commit the generated secrets.yml that we currently have checked in.  So I'm going to close this and look #1870 again.

PR #1928, opened 2017-05-16
Use bookmark count as solr :rows parameter instead of default

### LINKED ISSUE #1927: Bulk actions are working only on first 10 items irrespective of number of items selected
### Descriptive summary
Server: spruce.indiana.edu
Testing: Exploratory
Browser: Chrome(58.0.3029.110 (64-bit))
Flash version: 24.0.0.221
Operating System: Windows10
Issue: Bulk actions Delete/Move/UpdateAccessControls/AddToPlaylist are working only on first 10 items irrespective of number of items selected

1. Login to spruce
2. Select more than 10 items, and navigate to Managed Selected Objects page.
3. Initially only 10 items will be displayed in the page, change items per page to 100.
selected items 100

4. Make sure that all the items in that page are selected.
5. Click on Delete button, which displays a pop-up with all the items to be deleted, confirm it.
delete 100

6. It reloads the page and shows only first 10 items again, also the operation is performed on those 10 items and not on all the selected items.
delete 10

### Related work

### Tasks
- [ ] track down where we need to reconfigure Blacklight/rsolr defaults
- [ ] reconfigure them
- [ ] write a test to prove that it is fixed

bkeese:
Fixes #1927 

Overrides blacklight default for bulk action bookmark count.

### DIFF
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -35,6 +35,15 @@ describe BookmarksController, type: :controller do
       expect(flash[:success]).to eq(I18n.t("blacklight.delete.success", count: 3))
       media_objects.each {|mo| expect(MediaObject.exists?(mo.id)).to be_falsey }
     end
+    it "should remove more than the blacklight default number of items (>10)" do
+      8.times do
+        media_objects << mo = FactoryGirl.create(:media_object, collection: collection)
+        post :create, id: mo.id
+      end
+      post :delete
+      expect(flash[:success]).to eq(I18n.t("blacklight.delete.success", count: 11))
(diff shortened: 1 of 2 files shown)


spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -35,6 +35,15 @@
       expect(flash[:success]).to eq(I18n.t("blacklight.delete.success", count: 3))
       media_objects.each {|mo| expect(MediaObject.exists?(mo.id)).to be_falsey }
     end
+    it "should remove more than the blacklight default number of items (>10)" do
+      8.times do
```
cjcolvar: Why does 8 times give you 11 total deletions?  Could you put a comment to explain this?

bkeese: There are already 3, this adds 8 more.

PR #1944, opened 2017-05-22
Speed up wipeout using ActiveFedora::Cleaner and delete_all

cjcolvar:
We ran into performance issues with this when trying to wipeout a failed production migration.  This works much better.


phuongdh:
This fails on large collection. Possible solution: when it fails, there's instruction to manually wipeout Fedora.

PR #1955, opened 2017-05-25
Fix UTF8 conversion issue

cjcolvar:
This fixes a corner case bug where content being copied from the masterFile datastream into the new RDF property threw an error dealing with converting between ASCII-UTF8 and UTF8.


cjcolvar:
Now that this is in I'm not sure if this is a safe change.  @mbklein, can you look over this and propose a better solution if there is one?

Issue #1924, opened 2017-05-16
Move noid minter to DB


cjcolvar:
We use active_fedora-noid to mint identifiers for our Admin::Collection, MediaObject, and MasterFiles.  This by default writes the minter's state to a file at /tmp/minter-state.  The file's location is configurable but a more robust solution is to use the DB-backed minter.  Currently this would require a customization (and not just configuration).

Tasks
- [x] Run generator: `rails generate active_fedora:noid:install`
- [x] Run `rake db:migrate db:test:prepare`
- [x] Configure AF::Noid to use the DB minter
- [x] Add tweaks to rails_helper.rb for tests
- [x] Write documentation including upgrade instructions using `rake active_fedora:noid:migrate:file_to_database`

cjcolvar:
Documentation included on wiki page under development:
https://wiki.dlib.indiana.edu/display/VarVideo/Upgrading+Avalon+6.0+to+Avalon+6.1

PR #1966, opened 2017-05-30
Use DB-backed noid minter by default

cjcolvar:
This is a change to use the recommended database backed minter for noids instead of the minter state yaml file.  There is a rake task to copy the state from the file into the database: `rake active_fedora:noid:migrate:file_to_database`.  This should hopefully clear up errors we are seeing running our migration with multiple processes where we get Ldp::Conflict about every 1000 objects.


mcwhitaker:
@cjcolvar Is this related to #1924 ? Should you update that ticket if so?

PR #1997, opened 2017-06-05
Resolve derivative resolution to width/height

### LINKED ISSUE #1992: Migration is not setting ebucore:width and ebucore:height on derivatives
The migration is looking in the wrong place for width and height in the encoding datastream for derivatives.  Instead of trying to set width and height separatly it is possible to read the value of `//video/resolution` from the datastream and set it on the derivative using `.resolution`.

bkeese:
Fixes #1992 

Checks derivatives resolution and height/width for migration. Raise error if they both exist but do not match.

### DIFF
--- a/app/migration/fedora_migrate/derivative/encoding_datastream_mover.rb
+++ b/app/migration/fedora_migrate/derivative/encoding_datastream_mover.rb
@@ -28,9 +28,15 @@ module FedoraMigrate
         target.audio_codec = present_or_nil(encodingProfile.xpath('//audio/codec').text)
         target.video_bitrate = present_or_nil(encodingProfile.xpath('//video/bitrate').text)
         target.video_codec = present_or_nil(encodingProfile.xpath('//video/codec').text)
-        target.width = width = present_or_nil(encodingProfile.xpath('//video/resolution/width').text)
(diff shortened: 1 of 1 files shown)


app/migration/fedora_migrate/derivative/encoding_datastream_mover.rb
```diff
@@ -28,9 +28,16 @@ def migrate
...
+        resolution = present_or_nil(encodingProfile.xpath('//video/resolution').text)
+        res_width,res_height = resolution.present? resolution.split('x') : [nil,nil]
+        width = present_or_nil(encodingProfile.xpath('//video/resolution/width').text)
+        height = present_or_nil(encodingProfile.xpath('//video/resolution/height').text)
+        if (res_width.present? and width.present? and res_width!=width) or (res_height.present? and height.present? and res_height!=height)
+          raise FedoraMigrate::Errors::MigrationError, "Derivative resolution and height/width do not match" 
+        end
+        target.width = width || res_width
+        target.height = height || res_height
+        target.resolution = "#{target.width}x#{target.height}" if target.width.present? && target.height.present?
```
cjcolvar: `.resolution` ends up setting `.width` and `.height` and vice versa so I don't think we need to call it as well.  You should be fine with lines 38-39 and skipping this one.

PR #2001, opened 2017-06-06
Better migration report

cjcolvar:
This PR adds a lot more details to the detail view page including the full contents of Fedora 3 datastreams as well as full contents of Fedora 4 fcr:metadata node and the fcr:metadata nodes of attached files.  These are accordianed closed by default to save screen real estate.  The migrated to id was changed to an actionable link to see the item in the current avalon instance.  

The Fedora3 and Fedora4 content is pulled in from the servers to expose content that might not be available to metadata reviewers or other non-system administrators. The Fedora3 content will go away once the Fedora3 instance is taken offline.  There are probably some error cases which aren't being handled right now but since this is a one-time admin interface the corner cases can probably be worked around.

Screenshot:

screenshot 20170605

@joncameron @davidschober


joncameron:
Awesome! Thanks Chris, this looks great. I'd love to have this in the demo for Friday.

PR #2008, opened 2017-06-07
Move drag and drop css out of blacklight_folders

bkeese:
We were using css styling from blacklight_folders/_nestable.scss. Move this out of blacklight_folder namespace and remove all blacklight_folder css.


cjcolvar:
:+1:

PR #2029, opened 2017-06-07
Move MediaObject delete to background job, fixes #1965

### LINKED ISSUE #1965: User cannot delete a 100 section item from preview page
### Descriptive summary
Server: mallorn.dlib.indiana.edu
Testing: Performance
Browser: Chrome(58.0.3029.110 (64-bit))
Flash version: 24.0.0.221
Operating System: Windows10
Issue: When tried to delete a 100 section item using the delete button present on the item preview page, the page is loading for around 30sec and the user is navigated to an error page and the operation is terminated.

### Expected behavior
The item gets deleted from the Avalon system.

### Actual behavior
The item is not getting deleted.

1. Login to mallorn system.
2. Add a 100 section item.
3. After the item is added navigate to the preview page of the item.
4. Click on the Delete button on that page to delete the item.
5. The page loads for some time and is navigated to a error page.
delete 100 sec item issue_1
6. Open a new tab after some time and browse for the item, it is not deleted (the delete operation might have been terminated).

### Related work

phuongdh:


cjcolvar:
We should remember that this doesn't fix the underlying issue only the error or lockup being seen by the user.  We should write a new ticket to fix the really slow destroy for media objects with many sections.

joncameron:
#2047 is the ticket that Chris references as needing to be written.

PR #2030, opened 2017-06-07
Use isGovernedBy predicate to link back to media objects

### LINKED ISSUE #2007: Lease has_many media objects uses wrong predicate
Leases can't lookup the media objects they apply to because they are using isMemberOfCollection, but media objects link to leases using isGovernedBy.  We need to update the has_many relationship in whatever way to fix this.  This is not a data issue as media objects already have the correct RDF triple and leases don't have a triple for this.  The relationship is used for ActiveFedora to correctly look things up.

cjcolvar:
Fixes #2007.


cjcolvar:
Hold off on merging this as I'm going to add a test.

Issue #1954, opened 2017-05-25
New media objects should have a recordIdentifier for its Fedora4 uri


cjcolvar:
### Description
MediaObjects in Avalon 5 have a MODS recordIdentifier of their Fedora3 pid and source "Fedora".  When migrated the MODS gets an additional recordIdentifier containing the Fedora4 uri and source "Fedora4".

## Observed behavior
New media objects in Avalon 6 end up with MODS containing an empy recordIdentifier with source "Fedora".

## Expected behavior
New media objects in Avalon 6 get a recordIdentifier in MODS containing their Fedora4 uris with source "Fedora4".

### Done Looks Like
All Media Objects in an Avalon 6.1 system have a recordIdentifier element in their descMetadata MODS document containing their Fedora4 uris and with source "Fedora4"

### Tasks
- [x] Fix MODS OM template so source is "Fedora4"
- [x] Add the Fedora 4 uris of the media objects to recordIdentifier
- [x] Write a migration to fix all existing MODS documents
- [x] Update any documentation to reflect change of source to "Fedora4"

davidschober:
This looks good.

davidschober:
@cjcolvar can you throw some points on it?

cjcolvar:
I'm verifying this on spruce.

cjcolvar:
Fixed on spruce but still need to write migration and update documentation.

cjcolvar:
Added a post-upgrade step to upgrade documentation: https://wiki.dlib.indiana.edu/display/VarVideo/Upgrading+Avalon+6.0+to+Avalon+6.1

PR #2044, opened 2017-06-09
Add migration for adding Fedora4 record identifiers

bkeese:
References #1954

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -142,6 +142,14 @@ EOC
         mo.descMetadata.save
       end
     end
+
+    desc "Add Fedora4 identifier as mods record_identifier"
+    task record_identifier: :environment do
+      MediaObject.find_each({},{batch_size:5}) do |mo|
+        mo.save if mo.descMetadata.record_identifier.empty?
+      end
+    end
+
   end
 
   desc 'migrate databases for the rails app and the active annotations gem'


db/migrate/20170609132731_add_fedora4_record_identifiers.rb
```diff
@@ -0,0 +1,19 @@
+class AddFedora4RecordIdentifiers < ActiveRecord::Migration
+
+  def up
+    say_with_time("Add Fedora4 record identifers") do
+      MediaObject.find_each({},{batch_size:5}) do |mo|
+        say("MediaObject: #{mo.id}, URI: #{mo.uri}")
+        if mo.descMetadata.record_identifier.empty?
+          mo.descMetadata.ensure_identifier_exists! mo.uri
```
cjcolvar: Do you really need this?  Wouldn't a mo.save be sufficient to trigger the callback which runs ensure_identifier_exists!?

bkeese:
Yep. Fixed.

cjcolvar:
I'm wondering if we should just make this a rake task instead of a DB migration.  I feel like that is the approach taken by sufia and what we did last time (https://github.com/avalonmediasystem/avalon/blob/develop/lib/tasks/avalon.rake#L131-L144).  That way this doesn't bog down everyone who doesn't need to run it and it fixes the problem with the build as a bonus!

bkeese:
Changed to rake task.

PR #2051, opened 2017-06-12
Allow configurable persistent flash messages

bkeese:
Add a configuration like this to avalon.yml:
```
  flash_message:
    type: 'error'
    message: '<p>This is a test warning message. テスト</p>'
```
type must be from list: success, notice, error, alert


bkeese:
Added example.

PR #2071, opened 2017-06-16
Migration tweak

cjcolvar:
This still needs to be tested by IU on sodium.


cjcolvar:
@bkeese could you rereview please?  
The changes I made in the last commit make it actually work (correctly it appears) in a migration run with the following options:
`bundle exec rake avalon:migrate:repo parallel_processes=1 pidfile=delta.test.pidfile CONFIRM=yes overwrite=yes`

PR #2073, opened 2017-06-16
Bump versions of ruby tested

cjcolvar:
This bumps the versions of ruby that we test and support.  IU production is moving to ruby 2.3 and 2.4 has been out for 6 months.  So I'm suggesting we drop testing of ruby 2.2 and pick up 2.4 in Avalon 6.2.  Don't merge this until after Avalon 6.1 has been released.


cjcolvar:
@mbklein @carrickr @phuongdh Looks like ruby 2.2 is in security maintenance mode now until all support stops 3/2018.  Are you okay with dropping 2.2 testing (and thus support)?

PR #2089, opened 2017-06-21
Properly handle embed authentication popup

### LINKED ISSUE #2034: Embed Logins Don't Redirect Correctly
### Description
Embedded players with content that is restricted don't gracefully handle the user clicking the 'sign in' button and passing that back to the player.

As a user, after authenticating I expect to go back to the location where I expect my content to be.

### Current Behavior Example
jun-09-2017 09-34-04

### Done Looks Like
- After the user authenticates in the pop-up browser window, the window closes and the content the user was trying to view appears

mbklein:
Resolves #2034

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -62,8 +62,11 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     end
 
-    if request['target_id']
-      redirect_to objects_path(request['target_id'])
+    if params['login_popup'].present?
+      flash[:success] = nil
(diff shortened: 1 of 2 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -62,8 +62,11 @@ def find_user(auth_type)
 
     end
 
-    if request['target_id']
-      redirect_to objects_path(request['target_id'])
+    if params['login_popup'].present?
+      flash[:success] = nil
+      render inline: '<html><head><script>window.close();</script></head><body></body><html>'.html_safe
+    elsif request['target_id']
+      redirect_to object_path(request['target_id'])
```
cjcolvar: This needs to be changed to the plural objects_path.

PR #2098, opened 2017-06-26
Form attributes

### LINKED ISSUE #2094: Import Button Doesn't Trigger Event in Microsoft Browsers
### Description
Clicking the "Import" button in the pop-up that occurs when performing bibimport on records with existing metadata doesn't trigger perform any action. See example gif below.

### Platforms
Microsoft IE 11 and Edge on media.dlib.indiana.edu, Avalon 6.1rcx

### Initial user report
I’m trying to import the finished bib record for vad8467 (IU Philharmonic Orchestra /  Strauss’s Alpine Symphony), and I get no response when I click "import."

### Secondary User Report
When you are editing the resource description for an item, if you click to import, in the bibliographic id area, nothing happens.  In the gui, the button appears to be pressed but it does not appear that anything is trying to load.
we have tried on a few different computer.  we have tried with a few different id types, we have tired different users and different audio only items.

### Example
jun-22-2017 14-38-55

cjcolvar:
This PR takes the form attributes polyfill and makes it available throughout the app with the options of passing a jquery selector to act on.  The default selector parameter has been relaxed to all elements with type submit to include buttons instead of only inputs.

Fixes #2094.

### DIFF
--- a/app/assets/javascripts/import_button.js.coffee
+++ b/app/assets/javascripts/import_button.js.coffee
@@ -15,7 +15,7 @@
 
 $ ->
   form = $('div.import-button').closest('form').prop('id')
-  import_button_html = '<div class="input-group-btn"><button id="media_object_bibliographic_id_btn" type="submit" name="media_object[import_bib_record]" class="btn btn-success" value="yes" form="'+form+'">Import</button></div>'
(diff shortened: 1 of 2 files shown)


cjcolvar:
@bkeese changes since you approved are removing the form attribute from the large grey Import button and fixing how it is added to the red confirmation Import button.  I also pushed the fix function call down one line to be after the popover is added to the page.

PR #2107, opened 2017-06-27
No invalid objs

### LINKED ISSUE #2045: Media Objects Can Be Saved Without Required Fields (Title & Date)
### Descriptive summary
Browser: Firefox 53
Operating System: macOS 10.11.6

### Issue
Media objects can be saved without a title or date, which are required fields—objects should not save if these are absent.

### Expected behavior
Save operations are not allowed if there is no value in the Title or Date fields of the Resource Description form.

### Actual behavior
jun-09-2017 10-36-47
The record saves and the Fedora 4 ID becomes the Title.

1. Login to spruce.
2. Go to the resource description for a record, delete the Title or Date fields, and click Save
3. Warnings appear to the user (bark but no bite)
4. Go to the Preview/Object View and see that the Title and Date values are set

### Related work
Originally reported in #2039

cjcolvar:
Please thoroughly review this before merging.
Fixes #2045.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem 'devise'
 #gem 'devise-guests', '~> 0.3'
 
 # Avalon-specific
-gem 'avalon-workflow', git: "https://github.com/avalonmediasystem/avalon-workflow.git", tag: 'avalon-r6.1'
+gem 'avalon-workflow', git: "https://github.com/avalonmediasystem/avalon-workflow.git", branch: 'no_invalid_objs'
 gem 'active_encode', '~> 0.1.1'
 gem 'hooks'
 gem 'iconv'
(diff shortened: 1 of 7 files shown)


cjcolvar:
We could wait on merging this a little bit since it has the chance to break ingest badly.

carrickr:
Tagged it WIP just to ensure it doesn't get merged until we want it, probably after Chris gets back

cjcolvar:
After some testing I found a few UI oddities that need to be addressed:

1. If you create a valid media object by adding title/date issued and saving then try to save without a title you get a flash error message as expected BUT it hangs around for one request if you click away to a different step of the workflow even though the save never persisted.

1. If you advance to the resource description step then go back to manage files before saving the required fields, you can upload a file and it looks okay but the master file never gets associated with the media object and it doesn't appear on the manage files page.

cjcolvar:
I've resolved the two issues I identified and this is ready to be reviewed.  We'll still want to do more regression testing after merging but I think it is ready to go.

PR #2117, opened 2017-06-30
Update MARC21slim2MODS3-5-avalon.xsl

jlhardes:
Changing language code variable so empty or invalid values ('|||','###','N/A') are removed to make variable empty.


cjcolvar:
The error in the failing tests are:
runtime error: element variable
        Failed to evaluate the expression of variable 'controlField008-35-37'.

bkeese:
Nokogiri apparently only supports XSLT V1.0. This change uses 'replace', which is V2.0 only.

jlhardes:
We're thinking we have a problem with XSL 1.0 vs 2.0. I tried putting in a replace() function but that only works in 2.0. The XSL says it's in 2.0 but @bkeese is saying that the Nokogiri gem avalon uses for XSL only supports 1.0.

jlhardes:
@bkeese that makes sense, the other 2 places that do this same check use the $controlField008-35-37 variable as well so I just made that change where you suggested.

jlhardes:
Looks like using the variable there means we end up possibly making invalid MODS?

PR #2127, opened 2017-07-05
Release 6.1

cjcolvar:


cjcolvar:
Can we turn off codeclimate on this branch?  Seems like we should only have it on for develop and feature branches if we're going to be having it on at all.

bkeese:
:+1: to removing code climate

PR #2150, opened 2017-07-10
Add xray to dev mode

lealeelu:
I think adding [xray-rails](https://github.com/brentd/xray-rails) to the dev environment would be handy to have. When you press crtl+shft+x, it shows how the page html.erb files it's using and where. 
image
****

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,6 +109,9 @@ group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   gem 'web-console', '~> 2.0'
 
+  # ctrl+shift+x to see erb files rendered on page
+  gem 'xray-rails'
+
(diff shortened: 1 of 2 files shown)


carrickr:
I don't see any reason not to, but I'll flag @adamjarling as a reviewer since he is the UI guru and may have thoughts on if this is the best dev tool, etc.

Gemfile
```diff
@@ -103,6 +103,7 @@ group :development, :test do
   gem 'fcrepo_wrapper'
   gem 'rspec-rails'
   gem 'dotenv-rails'
+  gem 'xray-rails'
```
carrickr: This should move down to under the `group: development` on 109.  This adds to dev and test, which means extra overhead on Travis tests.

lealeelu: woops good catch

PR #2152, opened 2017-07-11
Make upload button clickable not just text

lealeelu:
If you click the upload button and not the upload text, nothing will happen. I made it so that the whole button works.

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -1,12 +1,12 @@
-/* 
+/*
  * Copyright 2011-2017, The Trustees of Indiana University and Northwestern
  *   University.  Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
- * 
+ *
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software distributed
  *   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
  *   CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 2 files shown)


lealeelu:
@bkeese It doesn't but I know what you're talking about. I can fix that up too.

carrickr:
@lealeelu Are you planning to update this PR to fix that?

lealeelu:
Yup, I'll update this PR @carrickr

lealeelu:
@bkeese @carrickr ok, Added white text fix. I think I'm done with this one unless you find something else.

bkeese:
As long as you thoroughly tested the upload button after making this change. There was some reason we put that z-index in there.  I think there were cases where the button wasn't clickable. Maybe certain browsers? I can't remember for sure.

lealeelu:
I found this commit you did three years ago that added it.
https://github.com/avalonmediasystem/avalon/commit/82b1e100d954533234aa40a4c7022b5280ab0a50

Do you know why we would want it at the same level as the modal?

bkeese:
I think I chose that because it was an available variable that would put an element over the page. There were no modals on that page at the time. Then browse everything and later that "loading" dialog came along and screwed things up. Anyway, the point was to keep the upload button at the top and thus clickable.



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

cjcolvar:
Is this ready to go?  Is there anything else that needs to be done?

bkeese:
I am afraid that completely removing that z-index will re-break the upload button.

lealeelu:
I want to get more testing done on something other than chrome.

cjcolvar:
@lealeelu browserstack could be helpful for this.

joncameron:
@lealeelu Definitely, you should feel free to use our main BrowserStack account: https://wiki.dlib.indiana.edu/display/VarVideo/BrowserStack

lealeelu:
Checked it in safari, android, edge and chrome. I wasn't able to break it so I think this pull request is good to go.

PR #2156, opened 2017-07-14
Copy Playlist

### LINKED ISSUE #1536: As a User I Want to Make Copies of Playlists
### Descriptive summary
As a user, I want to make a copy of a playlist so that I can easily create multiple versions of a playlist.

### Use Case
The ability to copy playlists means that a master playlist can be created, copied, and modified to meet different audiences and needs. This is especially the case for things like listening tests: you might circulate a longer playlist to students for studying, which you use to create the playlist for the exam in class.

### Done Looks Like
Users can create a copy of a playlist from the playlist index.
* A "Copy" button is available in the _Actions_ column of the playlist index, with the same styling as the Edit and Delete icons and with the clone icon from Font Awesome (http://fontawesome.io/icon/clone/).
* Clicking the copy button presents a modal to the user with elements:
  * Editable text field containing the string "#{source_playlist_name} Copy"
  * OK button
  * Cancel button
* After clicking the OK button in the modal
  * A copy of the playlist is created with the exact same items as the source playlist and with the name from the modal.
  * The new playlist appears in the playlist index

### Design Examples
screen shot 2017-07-06 at 4 24 43 pm

screen shot 2017-07-06 at 4 24 33 pm

### Tasks

lealeelu:
Fixes #1536 

This is still in progress, so don't merge it just yet.

### DIFF
--- a/app/models/playlist_item.rb
+++ b/app/models/playlist_item.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2017, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
@@ -25,5 +25,20 @@ class PlaylistItem < ActiveRecord::Base
   delegate :title, :comment, :start_time, :end_time, :title=, :comment=, :start_time=, :end_time=, :master_file, to: :clip
   before_save do
     clip.save
+  end
+
+  def replicate
+    new_clip = clip.dup
+    new_clip.save
+
+    new_playlist_item = PlaylistItem.create(playlist: @playlist, clip: new_clip)
+
+    marker.each do |old_marker|
+      new_marker = old_marker.dup
+      new_marker.save
+      new_playlist_item.marker << new_marker
+    end
+    new_playlist_item.save
+    new_playlist_item
   end
 end
--- a/app/models/playlist_item.rb
+++ b/app/models/playlist_item.rb
@@ -27,18 +27,18 @@ class PlaylistItem < ActiveRecord::Base
     clip.save
   end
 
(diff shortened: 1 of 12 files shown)


app/models/playlist_item.rb
```diff
@@ -26,4 +26,19 @@ class PlaylistItem < ActiveRecord::Base
   before_save do
     clip.save
   end
+
+  def replicate
```
cjcolvar: How about making this replicate! and changing all of the .save calls to .save! so exceptions are raised and bubble up instead of being silently skipped over.

spec/controllers/playlists_controller_spec.rb
```diff
@@ -174,6 +174,53 @@
...
+    let(:playlist) { FactoryGirl.create(:playlist, new_attributes) }
+
+    context 'blank playlist' do
+      it 'replicate a blank playlist' do
+        post :replicate, format: 'json', old_playlist_id: playlist.id,
+          playlist: { 'title' => playlist.title, 'comment' => playlist.comment, 'visibility' => playlist.visibility }
+        expect(response.body).not_to be_empty
+        parsed_response = JSON.parse(response.body)
+
+        new_playlist = parsed_response['playlist']
```
cjcolvar: How about looking the playlist back up from the DB to ensure that it got persisted like in the test below?

spec/models/playlist_item_spec.rb
```diff
@@ -87,4 +87,18 @@
...
     end
   end
+
+  describe '#replicate!' do
+    let(:master_file) { FactoryGirl.create(:master_file, :with_media_object) }
+    let(:playlist_item) { FactoryGirl.create(:playlist_item, playlist: playlist, clip: avalon_clip) }
+    let(:avalon_clip) { FactoryGirl.create(:avalon_clip, master_file: master_file) }
+    let(:playlist) { FactoryGirl.create(:playlist, visibility: Playlist::PUBLIC) }
+
+    it 'replicating an item' do
```
cjcolvar: How about changing this to "it 'replicates an item'"?

spec/models/playlist_item_spec.rb
```diff
@@ -87,4 +87,18 @@
...
+      new_item = playlist_item.replicate!
+      expect(new_item.id).not_to eq playlist_item.id
+      expect(new_item.playlist_id).to eq playlist_item.playlist_id
+      expect(new_item.clip_id).not_to eq playlist_item.clip_id
+    end
```
cjcolvar: I think it would be good to add a test showing new_item.persisted? is true to help explain what the method does.

PR #2164, opened 2017-07-19
Allow override of pid namespace on the migration rake task

cjcolvar:
@mbklein I noticed this same thing while writing the documentation today.  Too late for 6.1 but it could go into 6.1.1/6.2.


cjcolvar:
Will need to add this to the documentation once this is merged: https://wiki.dlib.indiana.edu/display/VarVideo/Migration+Tool

PR #2183, opened 2017-08-01
[WIP] Always resize player and controls in case they got missed during init

### LINKED ISSUE #2167: Clicking Playlist Marker Icon Moves Playhead to Marker Offset
### Description
As a user, I want the action of clicking a playlist marker icon to move the player position to the time offset property of that marker.

Current mouseover behavior for a marker icon:
- Cursor changes to hand icon
- The same events happen as though the cursor were over the rail (tooltip showing timepoint on the rail under the cursor,

### Done Looks Like
- Mousing over a marker icon changes the cursor to indicate clickability (the current behavior)
- If the mouse is over the marker icon, a tooltip (in the same style as appears above the rail to indicate time) displays the marker name and offset value
- Clicking the marker moves the playhead to that time offset

### Other Enhancements
This issue is meant to be a first pass—eventually working on enhancements as in the design work in #1534.

### Tasks
 - [ ] Have a discussion with Matt and PO's about the design.
 - [ ] Implement Design

cjcolvar:
This is a workaround that needs review and possible reworking.  The player has some extra flashing during initialization due to this workaround so it would be good if it could be scoped to only the cases where it is needed.

Fixes #2167.

### DIFF
--- a/app/assets/javascripts/avalon_player.js.coffee
+++ b/app/assets/javascripts/avalon_player.js.coffee
@@ -120,6 +120,12 @@ class AvalonPlayer
           ).bind 'mouseleave', (e) ->
             $('.mejs-time-float-marker[data-marker="'+this.dataset.marker+'"]').hide()
         @player.setCurrentTime initialTime
+        #Workaround for Firefox rendering issue
+        #TODO figure out how to scope this to Firefox and/or misrendered players
+        #Might also want to try moving it to another place in the initialization code to minimize impact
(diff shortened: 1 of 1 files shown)


cjcolvar:
Some more testing shows that this doesn't fix things all the way.  If you resize the browser window the player won't resize.

cjcolvar:
After further review this looks like a Red5/RTMP only issue.  Since we're dropping both soon, I'm going to close this PR and comment the same on the issue.

Issue #2174, opened 2017-07-21
Create a query param to switch between ME4 and ME2


davidschober:
In order to make it easier to test ME we should have a swappable query parameter. That will allow us to merge ME 4 into dev and test on spruce

Done:

- create a query string to swap player
- merge into dev and deploy to spruce

davidschober:
work to represent on #2134

adamjarling:
@bkeese @cjcolvar For some reason on this branch, the AJAX request for a new section media item is not returning the same data as my previous MEJS4 branch.  So clicking on section links and changing a player in the same browser window w/o a refresh is not working since it's missing source file data.   I might need some eyes tomorrow to help me diagnose if possible.

PR #2187, opened 2017-08-02
Mediaelement.js 4 player preview

cjcolvar:
This PR is @adamjarling 's starting work on the mediaelement.js 4 player and the ability to toggle over to it using a query param.  This allows us to keep the ME.js 2 player default while still developing the ME.js 4 player.  This is important if we want to do another 6.x release before the ME.js 4 player is ready to ship, but if we don't think that will be the case then we could skip some of the complexity here and make it only the ME.js 4 player.  The other alternative is keeping this work on a branch or maintaining multiple active branches for release both of which make testing hard.

@avalonmediasystem/product-owners I'd like to hear your opinions about release cycle before merging this.


davidschober:
My optimistic side wants to say, let's just push hard for ME4. But if we've already made the swapping work, and considering we have a few more playlist improvements that we need to ship. I say we target ME4 as a 6.3 and ship the final playlist improvements as 6.2 quickly to keep end users happy. 

@joncameron does that work for you?

mcwhitaker:
I added the milestone because the work happened during sprint 132.

joncameron:
@davidschober That works for me. Let's move forward then with ME4 as 6.3 and playlist improvements as 6.2.  @cjcolvar

cjcolvar:
Fixes #2174.

PR #2190, opened 2017-08-04
Asset manifests for ME.js 2 and 4 players

cjcolvar:
This PR restores the JS and CSS for the MediaElement 2 playlist and embed player while fixing a bug in the media object player where jquery was included twice.  A subsequent PR will be needed for exposing the MediaElement 4 player in playlist and embed contexts.


cjcolvar:
Related to #2174.

Issue #2185, opened 2017-08-01
Move / re-factor media object 'Sections' JS to a .js file instead of in-lining


adamjarling:
### Description
Move and re-factor the custom javascript for controlling opening and closing of section panels, to a separate javascript file, instead of having it inline.

### Done Looks Like
- Inline JS removed
- Functionality works as normal

### Tasks
- [ ] Remove inline JS from `app/views/media_objects/_sections.html.erb`
- [ ] Move this code to a "Sections" custom Avalon Mediaelement plugin created in #2161.

adamjarling:
Relates to #2161

adamjarling:
This is pretty old, so I'll take a second look and update the issue after Playlist Items.

adamjarling:
I think we combine with with the "Sections" MEJS plugin creation work.  In the future as it may take a few days to get it right and tested.

davidschober:
This should be part of the initial work we do.

cjcolvar:
@davidschober @joncameron I'm confused.  Is this a 6.x or 7-0.1 issue?  To me it reads as a 6.x issue.

adamjarling:
@cjcolvar This should transfer over into a 7.x issue.  Your call on whether it's worth implementing in 6.3.x?

adamjarling:
@cjcolvar  I think we close this as stale, and re-evaluate how it'll work in 7.

PR #2195, opened 2017-08-05
Refactor player partials to prepare for mejs4 and move towards DRYing…

cjcolvar:
… up the players

I created a modules/player directory to start collecting partials that are common to all instances of the player then updated all of the players to use the now common section partial.  With that work done, I standardized all of the players to be aware of the mejs4 query param with partials to handle the JS for mejs2 or mejs4.  There are commented out render partial tags for the mejs4 JS for the playlist and embed players ready to be filled in.  Overall this is a step towards further DRYing up of the players and there are still improvements that could be made, but I think this is a good first step.


cjcolvar:
@adamjarling  Maybe this fixes #2185 ?

app/views/master_files/_player.html.erb
```diff
@@ -13,72 +13,22 @@ Unless required by applicable law or agreed to in writing, software distributed
...
 ---  END LICENSE_HEADER BLOCK  ---
 %>
-<% content_for :page_styles do %>
-  <%= stylesheet_link_tag 'mejs2_player' %>
-<% end %>
-<% content_for :page_scripts do %>
-  <%= javascript_include_tag "mejs2_player" %>
-<% end %>
+<% # Set MediaElement 4 query param variable
+is_mejs4 = params['mejs4'] === 'true' %>
```
adamjarling: @cjcolvar I think I put this in here at some point, but was wondering your opinion...does it make sense to set this variable higher up in the application just once, so it can be accessed from any view/controller?

cjcolvar: I'm torn between DRY and scoping it to only where it is necessary.  Maybe it would be better as a helper since those are global anyways?  I'll try that and push it up.

adamjarling: Sounds good.  Give a shout if you need me to push the Merge button.

cjcolvar: @adamjarling I pushed the change and flipped the logic so mejs2 is the special case.  Probably doesn't make a big difference but I thought it would be easier when deconstructing mejs2 support.

PR #2196, opened 2017-08-07
Playlists can be shared

### LINKED ISSUE #2110: add "share" links to playlists
## Description

Users would like to share playlists with end users. In order to facilitate this we should have a "share" link similar to individual items . Access control should follow the item (this is only a share link). 

## Done looks like

- A share icon is placed at the playlist list _and_ view page
- An end user can share a playlist by clicking on the share button and copying the link
- An end user can share a playlist that has items which require LTI auth by clicking share >>LTI then adding to something like a canvas module. This would authenticate using the LTI method as setup by the institution. 

## Tasks

cjcolvar:
This PR copies much from the MediaObject's share views and functionality.  This PR resolves #2110 and sets the stage for #2179.

There is some styling that I put inline that should probably be pulled out to an included stylesheet.  I'd welcome ideas for best place to put it.

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -15,10 +15,26 @@
 require 'avalon/variations_playlist_importer'
 
 class PlaylistsController < ApplicationController
+  include ConditionalPartials
+
   before_action :authenticate_user!, except: [:show, :refresh_info]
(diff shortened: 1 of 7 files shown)


cjcolvar:
Closed in favor of #2197.

PR #2217, opened 2017-08-17
Feature/#1943 add mediaobject to playlist

### LINKED ISSUE #1943: As a user, I want to be able to add all tracks of a CD to a playlist at once
### So that 
I don't have to add track by track.

### Items for Discussion
- Should this happen via some action on the Media Object page, like a helper button?
- Should this be per-section or per-media object?
- Should this also be available in _bulk actions_?

### Done looks like
-  A user has the ability to add all segments of a section to a playlist in one step. 
- The structural metadata is used to inform the clip sizes and names of playlist items.
- Per discussion: 07/27 The scope of this is **ONLY** for a top-level "add all" link. 
- The "add to" interface should be similar functionally to the current add to playlist.

bkeese:
Fixes #1943 

Adds new ability to easily add one or all sections of a media object to a playlist. Entire sections may be added as single playlist items, or if structural metadata is available, sections may be added as multiple playlist items based on that structure.

### DIFF
--- a/spec/controllers/media_objects_controller_spec.rb
+++ b/spec/controllers/media_objects_controller_spec.rb
@@ -1042,29 +1042,27 @@ describe MediaObjectsController, type: :controller do
       login_as 'administrator'
     end
     it "should render add_to_playlist_form with correct masterfile_id" do
-      post :add_to_playlist_form, id: media_object.id, scope: 'master_file', masterfile_id: media_object.master_file_ids[0]
+      get :add_to_playlist_form, id: media_object.id, scope: 'master_file', masterfile_id: media_object.master_file_ids[0]
       expect(response).to render_template(:_add_to_playlist_form)
       expect(response.body).to include(media_object.master_file_ids[0])
     end
     it "should render the correct label for scope=master_file" do
-      post :add_to_playlist_form, id: media_object.id, scope: 'master_file', masterfile_id: media_object.master_file_ids[0]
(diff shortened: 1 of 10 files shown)


cjcolvar:
Related to #1943

spec/controllers/media_objects_controller_spec.rb
```diff
@@ -1034,4 +1034,67 @@
...
+  end
+
+  describe "#add_to_playlist" do
+    let(:media_object) { FactoryGirl.create(:media_object, :with_master_file, title: 'Test Item') }
+    let(:master_file_with_structure) { FactoryGirl.create(:master_file, :with_structure) }
+    let(:playlist) { FactoryGirl.create(:playlist) }
+
+    before do
+      login_as 'administrator'
+      media_object.ordered_master_files += [master_file_with_structure]
```
cjcolvar: You can achieve this in the create without the need for this extra handling:
`let(:media_object) { FactoryGirl.create(:media_object, ordered_master_files: [FactoryGirl.create(:master_file), master_file_with_structure], title: 'Test Item') }`

bkeese:
I went back and forth on which controller to use and landed on MediaObject because that's what we're really operating on here. I don't think this will be reusable for the other naming story, as that work will happen client-side.

cjcolvar:
Looks good. We'll need to test with large items and see if we need to create a background job.

PR #2218, opened 2017-08-17
Fix bug with playlist comment field

lealeelu:
Comments with newlines would break the JavaScript. We get out of that by using string literals.


bkeese:
escape_javascript  http://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html

Issue #1537, opened 2017-01-18
As a User I Want to Copy Playlist Items


joncameron:
### Descriptive summary

Users would like to ability to copy items from one playlist to another, rather than only having the option to move the playlist items from one playlist to another. Both the playlists and the items to be copied are _owned_ by the user. 

### Done Looks Like
Playlist items may be copied to another playlist by the same means as users can currently move items—by selecting on the playlist edit page, and using the drop down menu with actions. Items can be copied to the **top** of the receiving list. 

- Follows "move to" playlist design
- A button with "Copy to..." appears and shows the same list of playlists when clicked as "Move to..."
- when items are copied to, the item exists in both the original and the receiving.

### Tasks
- [ ] Add Copy items drop down to edit playlist page
- [ ] Add copy handling to PlaylistsController#update_multiple using PlaylistItem#duplicate
- [ ] Add Controller tests for new behavior

lealeelu:
@joncameron, what does bullet point 2 mean? The "move to" is a dropdown where you select the playlist you want to send the item to. Do you want something different than that?

joncameron:
@lealeelu That was written very confusingly! What I had meant was that the design for the button should be the same. Here's what I'm imagining:
![jul-25-2017 11-25-48](https://user-images.githubusercontent.com/1406618/28580014-4bc92fc6-712c-11e7-92e3-23898296673c.gif)

lealeelu:
Ok, that makes sense now. Thanks for the demo.

PR #2219, opened 2017-08-17
Limit by user playlist then by access

lealeelu:
Limiting what playlists an administrator sees in the copy and move item dropdowns by creating a method get_all_other_playlists and checking that the current user is also an owner.


cjcolvar:
Related to #2110 and #1537

PR #2231, opened 2017-08-18
Copying playlist fixes

### LINKED ISSUE #2222: Copy playlist items appends instead of prepends
When copying playlist items, they need to be appended onto the target playlist, not pre-pended like it is now.

lealeelu:
fixes #2222 and #2230  
This change checks to make sure a playlist item to be copied has a masterfile. If it doesn't it just skips it.
It also appends the copied items onto the end of the items list.

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -138,8 +138,11 @@ class PlaylistsController < ApplicationController
 
       #copy items
       old_playlist.items.each do |item|
+        next if item.nil?
         copy_item = item.duplicate!
-        @playlist.items << copy_item
+        copy_item.playlist_id  = @playlist.id
+        copy_item.save!
+        copy_item.move_to_bottom
       end
 
       respond_to do |format|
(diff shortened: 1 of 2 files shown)


app/models/playlist_item.rb
```diff
@@ -28,17 +28,19 @@ class PlaylistItem < ActiveRecord::Base
   end
 
   def duplicate!
-    new_clip = clip.dup
-    new_clip.save!
+    unless (clip.master_file.nil?)
```
cjcolvar: Instead of wrapping this all in an unless, how about:
`return nil if clip.master_file.nil?`

lealeelu: We'll definitely need tests, I'll make sure to add them later
I'll make that change. It looks better.

app/controllers/playlists_controller.rb
```diff
@@ -184,17 +184,25 @@ def update
   def update_multiple
     if request.request_method=='DELETE'
       PlaylistItem.where(id: params[:clip_ids]).to_a.map(&:destroy)
-    elsif params[:new_playlist_id].present? and params[:clip_ids] #PATCH
+    elsif params[:new_playlist_id].present? and params[:clip_ids]
       @new_playlist = Playlist.find(params[:new_playlist_id])
       playlist_items = PlaylistItem.where(id: params[:clip_ids])
+      playlist_items = playlist_items.sort_by { |item| item.position }
```
cjcolvar: I don't think we need this line now.

app/controllers/playlists_controller.rb
```diff
@@ -184,17 +184,25 @@ def update
...
       if (params[:action_type] == 'move_to_playlist')
-        @new_playlist.items += playlist_items
-        @playlist.items -= playlist_items
+        playlist_items.each do |item|
+          if (!item.clip.master_file.nil?)
```
cjcolvar: How about doing? `next if item.clip.master_file.nil?` like below.

app/controllers/playlists_controller.rb
```diff
@@ -184,17 +184,25 @@ def update
...
+            item.playlist_id = @new_playlist.id
+            item.move_to_bottom
+          end
+        end
         @playlist.save!
       else
         playlist_items.each do |item|
-          new_item = item.duplicate!
-          @new_playlist.items << new_item
+          if (!item.clip.master_file.nil?)
```
cjcolvar: Same as move.

app/controllers/playlists_controller.rb
```diff
@@ -139,7 +139,7 @@ def duplicate
       #copy items
       old_playlist.items.each do |item|
         copy_item = item.duplicate!
-        @playlist.items << copy_item
+        @playlist.items << copy_item unless copy_item.nil?
```
cjcolvar: Does this need the same pattern as copy/move below?

cjcolvar:
Fixes #2230

PR #2243, opened 2017-08-24
Playlist share by obfuscated url

lealeelu:


app/models/playlist.rb
```diff
@@ -45,6 +47,10 @@ def generate_access_token
     end
   end
 
+  def access_token_url
+    url_helpers.playlist_path(self, only_path: false) + '?token=' + self.access_token
```
cjcolvar: Instead of bringing the url_helpers into the model, how about moving this method into the playlist_helpers.rb?  I think that makes the most sense.  This is how I'm thinking of it: the access_token is data so it should live on the model while the url is a rendering of that data so it belongs outside the model.

app/controllers/playlists_controller.rb
```diff
@@ -206,6 +207,13 @@ def update_multiple
     redirect_to edit_playlist_path(@playlist), notice: 'Playlist was successfully updated.'
   end
 
+  # PATCH/PUT /playlists/1
```
cjcolvar: It would be good to have this comment (and the one above) match the routes: 
# PATCH/PUT /playlists/1/regenerate_access_token

Issue #982, opened 2016-07-20
Titles for Newly Created Playlist Items Only Include Media Object Title


joncameron:
### Current Behavior
- Click the "Add Playlist Item" icon
- Look at the Title field. The default title is "MediaObject_Title -"

### Possible Changes to Behavior
Examples would be:
- "Parent label - track label" (idea provided by the IU Music Library)
- "track label"
- or letting this be user-defined somewhere

### Desired Behavior for Population of Title Field For New Playlist Items
- If there is only a single section and no structural metadata XML exists, use MediaObject's Title
- If no structural metadata XML exists or it exists but no structural segment is active, use "MediaObject_Title - SectionLabel"
- If structural metadata XML exists and a segment is currently highlighted, use "#{Label value of parent node if one exists} - #{Label of span which is currently active}"

### Done Looks Like
- The default _Title_ value for a playlist item is populated based on the desired behavior above.

### Real world example
The title of this playlist item defaults to the name of the media object, when it could be "Schuber, Sonata ... Track 3. Allegreto"
![screen shot 2017-01-18 at 2 25 25 pm](https://cloud.githubusercontent.com/assets/1406618/22079750/e926dd08-dd8a-11e6-81c5-e1e291246081.png)

### Tasks
- [ ] Add controller function to masterfile_controller for generating the playlist label.
- [ ] Add tests for said function
- [ ] Use function in appropriate views

davidschober:
per @bkeese currently this is the label for the section or the media object label. Adding structural label is a new feature.

carrickr:
Need decisions made on which prepopulation option(s) should be implemented.  Can task after that,

davidschober:
Per discussion john and David, the default should the section title OR the label from the  currently selected structural metadata

PR #2245, opened 2017-08-24
Update mediaelement-add-to-playlist

bkeese:


app/models/concerns/master_file_behavior.rb
```diff
@@ -68,7 +68,7 @@ def stream_details(token,host=nil)
   end
 
   def embed_title
-    "#{ self.media_object.title } - #{ self.title || self.file_location.split( "/" ).last }"
+    "#{ self.media_object.title } - #{ self.title.present? ? self.title : self.file_location.split( "/" ).last }"
```
cjcolvar: This is another place where a nil file_location could bite us, but since it is only if title is also blank then I think it is  even more of a corner case.

cjcolvar:
Fixes https://github.com/avalonmediasystem/avalon/issues/982.

PR #2253, opened 2017-08-25
Add playlist share options to modal

lealeelu:


app/controllers/playlists_controller.rb
```diff
@@ -78,7 +78,7 @@ def paged_index
         [
           view_context.link_to(playlist.title, playlist_path(playlist), title: playlist.comment),
           "#{playlist.items.size} items",
-          playlist.visibility =='private'? '<span class="fa fa-lock fa-lg"></span> Only me' : '<span class="fa fa-globe fa-lg"></span> Public',
```
lealeelu: @cjcolvar this line is why the tests are failing. I removed the "only me" because I wanted it to match everywhere else. If that's ok with you, I'm going to change the test to reflect that.

PR #2282, opened 2017-08-31
[WIP] Failing Capybara tests

### LINKED ISSUE #2281: Travis test failing randomly
Travis builds have been failing because of a bug in capybara tests. So far it looks like the have_content test is causing the problem and might need to be replaced with assert_text. 

Here's an example of a failed build:
https://travis-ci.org/avalonmediasystem/avalon/jobs/269302128

lealeelu:
fixes #2281 
fix capybara tests that keep on failing on us


app/controllers/bookmarks_controller.rb
```diff
@@ -63,7 +63,6 @@ def user_can? action
   def verify_permissions
     @response, @documents = action_documents
     @valid_user_actions = [:delete, :unpublish, :publish, :move, :update_access_control, :add_to_playlist]
-    mos = @documents.collect { |doc| MediaObject.find( doc.id ) }
```
lealeelu: mos was never being used.

config/avalon.yml.example
```diff
@@ -75,6 +75,3 @@ test:
     protocol: sru
     url: http://zgate.example.edu:9000/exampledb
     query: rec.id='%s'
-  flash_message:
```
bkeese: I think this needs to stay as an example for non-docker deploys.

lealeelu:
Going in a different direction for this.

PR #2286, opened 2017-09-01
Make sure vendored assets and stylesheets behave w/r/t asset pipeline

mbklein:
Our testing should probably include at least one system that turns on-the-fly asset serving off.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -1,21 +1,21 @@
 # Be sure to restart your server when you modify this file.
 
-# Version of your assets, change this if you want to expire all your assets.
-Rails.application.config.assets.version = '1.0'
+Rails.application.config.assets.tap do |assets|
+  # Version of your assets, change this if you want to expire all your assets.
+  assets.version = '1.0'
 
-# Add additional assets to the asset load path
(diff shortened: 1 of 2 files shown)


config/initializers/assets.rb
```diff
@@ -1,21 +1,21 @@
...
-# MediaElement 4 files
-Rails.application.config.assets.precompile += %w( mejs4_player.js )
-Rails.application.config.assets.precompile += %w( mejs4_player.scss )
-
-# MediaElement 2 files
-Rails.application.config.assets.precompile += %w( mejs2_player.js )
-Rails.application.config.assets.precompile += %w( mejs2_player.scss )
+  # MediaElement 4 files
+  assets.precompile += %w( mejs4_player.js mejs4_player.scss )
+  assets.precompile += Dir[Rails.root + 'vendor/assets/stylesheets/mediaelement/mejs-controls.*']
```
cjcolvar: Does Rails.root always have a ending slash or would File.join be better?

mbklein: Rails.root is a Pathname, so addition acts like `File.join`. But I've found another issue with this PR so I'm going to close it and resubmit when I've got it worked out.

PR #2289, opened 2017-09-01
Fix auth for add_to_playlist

bkeese:


app/controllers/media_objects_controller.rb
```diff
@@ -67,15 +67,22 @@ def new
...
+          render template: '500.html', status: 500
+        end
       end
     end
   end
 
   # POST /media_objects/avalon:1/add_to_playlist
   def add_to_playlist
+    @media_object = MediaObject.find(params[:id])
+    authorize! :read, @media_object
```
cjcolvar: The current_user also needs to be able to :update playlist, right?

app/controllers/media_objects_controller.rb
```diff
@@ -67,15 +67,22 @@ def new
...
   # POST /media_objects/avalon:1/add_to_playlist_form
   def add_to_playlist_form
+    @media_object = MediaObject.find(params[:id])
     respond_to do |format|
       format.html do
-        render partial: 'add_to_playlist_form', locals: { scope: params[:scope], masterfile_id: params[:masterfile_id] }
+        if can? :read, @media_object
+          render partial: 'add_to_playlist_form', locals: { scope: params[:scope], masterfile_id: params[:masterfile_id] }
+        else
+          render template: '500.html', status: 500
```
cjcolvar: Is this the right thing to render?  Should we render a custom error instead even though it is an edge case?

PR #2302, opened 2017-09-08
Turn off feature tests in travis until randomly failing tests are fixed

cjcolvar:
The randomly failing feature tests are really slowing down PRs.  There is work ongoing on #2281 but until that work is finished I would like to turn off feature tests when rspec is run by travis.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,6 +23,11 @@ task default: [:ci]
 
 Rails.application.load_tasks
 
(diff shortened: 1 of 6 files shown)


bkeese:
Looks like feature tests are still running (and randomly failing)?

cjcolvar:
Yup, :-1:

PR #2307, opened 2017-09-11
Add markers to playlist custom mejs4 plugin - #1885

### LINKED ISSUE #1885: Media element 4 - Add Marker
### Description
Add markers to videos in 'playlist' view.

### Done Looks Like
- Investigate current implementation of adding markers to a video (and audio?) and understand the ux/flow.
- Re-create add marker functionality from 6.x Avalon player, to Media  Element 4 player, building on the markers Mediaelement 4 plugin.

### Tasks
- [ ] Investigate current method and/or plugin for adding/editing markers
- [x] Pull in markers plugin
- [ ] Wire up add marker functionality to proper endpoints

adamjarling:
Fixes #1885 

This PR makes use of MediaElement 4's 'markers' plugin in our application, and also builds upon it's foundation with our own custom 'Add markers to playlist' plugin.  The custom plugin does the following:
- Handles adding a new marker to a playlist
- Handles all interactions between the 'Markers' table in the UI, and MEJS4 player

Also in this PR I created a JSON endpoint to return all markers for a playlist item.  This endpoint is being hit a few times in the custom plugin, and also in the main MEJS4 js wrapper file.

One note for QA, is that I'm using ES6 Javascript promises (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) to handle AJAX request execution timing.  It's not clear to me in the documentation for `gem 'sprockets-es6'` whether the gem contains the polyfill for promises.  I checked in all major browsers via Browserstack, and the promises are not throwing any errors, and behavior looks as expected, but QA should give this a thorough testing in Windows environments and older browsers (whatever Avalon supports down to).

There are some remaining issues in the Playlists page which need to be worked out.  These are more with the MEJS4 setup in general, then specific to markers issues.  I'll create sub-issues for those and link them here as well.

The following work needs to be built on top of this:
- #2308 - Playlists page - playlist items list wired up to MEJS4


adamjarling:
Looks like one travis build succeeded and one failed.  I think this error failed on me last time too in travis.  Is it real?  It's passing locally.

```
Failures:
  1) Playlist is able to change public playlist to private
     Failure/Error: page.should have_content('Private')
       expected to find text "Private" in "Skip to main content moore.era Sign out Browse Manage Content Manage Groups Manage Selected Items (0) Playlists Search × Playlist was successfully updated. × Test flash message configured in [secure].yml. Type can be success, notice, error, or alert. Editing playlist View Playlist Delete Playlist Playlist Details Namepublic_playlist DescriptionThis is test Visibility Public - This playlist can be viewed by anyone on the web. There are currently no items in this playlist. Avalon Media System Project Website Contact Us Avalon Media System Release 6.1.0". (However, . it was found 1 time including non-visible text.)
     # ./spec/features/capybara_playlist_spec.rb:171:in `block (2 levels) in <top (required)>'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:112:in `block in run'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:101:in `loop'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:101:in `run'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # ./spec/spec_helper.rb:29:in `block (2 levels) in <top (required)>'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:112:in `block in run'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:101:in `loop'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:101:in `run'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /home/travis/.rvm/gems/ruby-2.4.0/gems/rspec-retry-0.5.4/lib/rspec/retry.rb:30:in `block (2 levels) in setup'
```

app/controllers/playlist_items_controller.rb
```diff
@@ -62,6 +62,27 @@ def update
     render json: { message: "Item was not updated: #{error.message}" }, status: 500 and return
   end
 
+  def show
+    playlistItem = PlaylistItem.find(params[:id])
```
cjcolvar: I think we need authorization here to ensure that the requestor `can? :read, playlistItem`.  This would probably be easiest to do by removing the find on line 66 and adding at the top of the class:
`load_and_authorize_resource only: [:show]`
This will find the playlist item and should put it in `@playlist_item` after performing the `can? :read` check.  If the check fails we should handle it returning json instead of the standard redirect with flash message.  For example look at the create method for an example or doing that although I think we'd want two rescues: 
 `CanCan::AccessDenied` (403) and `StandardError` (500).

adamjarling: @cjcolvar Ok, sounds good.  Whatever you recommend for `playlist_items_controller.rb`, let's do it.  I can give it a shot, but might need some assistance.

bkeese: Maybe since this is a show action for playlist_items and not just markers, it would be wise to return {markers: [marker, marker, ...]}, instead of just [marker, marker, ...], and parse it as necessary on the other end.

adamjarling: Any chance either of you guys can just adjust `playlist_items_controller.rb` however you think is best, then I'll make updates on the JS side of things?   I'm not really sure how to move on Chris's original comment...   Or if you want to pair on it, I can share a screen and you could  instruct me where to place stuff?

bkeese: I'll work on the controller changes this afternoon.

adamjarling: Thanks Brian

bkeese:
You can see what is returned for those errors in the controller (line 67 and 87, I think).

cjcolvar:
These are showing on the scrubber now but are hard to see.  @avalonmediasystem/product-owners what are your thoughts on styling?  Do they need to have the mouse over popup like in the mejs2 player?

joncameron:
Yes, I'd like to have the same styling we had previously. Right now the color of the progress on the rail obscures the marker once it moves past the white vertical line, and navigation takes a hit compared to the clickable markers that had the timecode pop-up in the mejs2 version. The previous marker icons weren't elegant but they were functional and should be retained.

adamjarling:
@joncameron @cjcolvar  FYI currently markers for MEJS4 are using in the Mediaelement core "Markers" plugin for visual display, which handles the styling, but most importantly it handles the calculation of marker placement in the time rail.

If we're going to change back to the old way markers looked, we're going to have to hack the Markers plugin, and this will take time.  I could adjust the CSS to make the display easier to see if you'd want to see how that looks like.  ie. Different color, pixel width, possibly move outside the time rail element...?    We could set up a meeting where I could make the adjustments in real time and see if anything works?

bkeese:
It probably wouldn't be too difficult to add the mouse-over functionality on top of the default markers too.

bkeese:
Here's some css I've come up with for altering the look of the markers. Tweak or toss at will:
```
.mejs__time-marker {
    top: 0;
    background: none !important;
}
.mejs__time-marker:before {
    content: "\2b27";
    width: 1em;
    transform: translate(-50%, 0);
    color: white;
}
```

adamjarling:
@bkeese I like these.  Putting in the CSS now.

cjcolvar:
@bkeese How about mouse over fills in the diamond with a yellow or something along with displaying the tooltip?

bkeese:
Mouse event handlers are next level work overriding the standard mejs marker plugin. I could only get them to work for mejs2 by putting the markers on their own rail. The standard plugin leaves them on the normal scrubber time rail.

davidschober:
I’d like to stick to default me4 as much as possible.

cjcolvar:
Sure, but I think the issue here is that default markers are not very visible and thus not as usable.  I think the amount of code required for customizing the markers is not going to be very much in comparison to the amount we already have.

bkeese:
Actually, making the markers more visible is easy (and already done). The issue I was referring to was adding mouse interactions. That's more significant.

cjcolvar:
@bkeese what would you size a ticket to add back the mouse over popup on markers similar to the MEJS2 player?

bkeese:
When I did it for MEJS2, it was a 5. Maybe it would be a 3 now that I understand the problem better.

davidschober:
@bkeese did you add an issue for it?  Can you reference.

bkeese:
I did not.

PR #2321, opened 2017-09-15
Exclude feature tests from travis

cjcolvar:
The randomly failing feature tests are really slowing down PRs. There is work ongoing on #2281 but until that work is finished I would like to turn off feature tests when rspec is run by travis.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,9 +23,14 @@ task default: [:ci]
 
 Rails.application.load_tasks
 
+RSpec::Core::RakeTask.new('rspec') do |task|
+  # Disable feature tests in travis until the intermittent failures are fixed
+  task.rspec_opts = "--tag ~type:feature" if ENV['TRAVIS']
+end
+
 task :ci do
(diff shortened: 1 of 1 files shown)


bkeese:
Feature tests are valuable and I fear merging this opens to the door to forgetting about this issue. I think instead of turning off the feature tests, we should just let PRs be merged with failing tests.

cjcolvar:
I don't really want to have to dig into the failing travis builds (up to 4 of them) to inspect the logs to find the tests that failed to ensure they are the feature tests we expect to fail.  Allowing PRs with failing tests introduces the risk of accidentally merging PRs that have failing tests that aren't these feature tests.

An alternative to all feature tests would be to flag just those few tests (2-3 I think) which have been causing the false positives.

Fixing the failing tests is of course ideal, but I'm open to other approaches.  Let's keep up the conversation.

cjcolvar:
Abandoned in favor of skipping tests via "pending".

PR #2325, opened 2017-09-15
Ensure currentPlaylistItem is set on the audio/video element on first…

### LINKED ISSUE #2322: Cannot add markers for the playlist items
### Descriptive summary

Server: mallorn.dlib.indiana.edu
Testing: Regression
Browsers: Chrome
Operating System: Windows10
Issue: Cannot add markers for the playlist items

1. Add items to any playlist.
2. Go to the playlist and play an item.
3. Try to add marker for that item.
4. It is returning and error message stating that the user is unauthorised.
5. The page reloads but you can see that the items are still present in first playlist and the second playlist is still empty.
image

cjcolvar:
… load of playlist

Fixes #2322.

I'm setting this on the playlist player initialization success callback because it is specific to playlists.
This regression was introduced when I refactored the player partials and is not present in Avalon 6.1.

### DIFF
--- a/app/views/playlists/_mejs2_player_js.html.erb
+++ b/app/views/playlists/_mejs2_player_js.html.erb
@@ -36,6 +36,7 @@
         displayTrackScrubber: true,
         autostart: <%= params[:autostart] == 'true' ? 'true' : 'false' %>,
         success: function(mediaElement, domObject, player) {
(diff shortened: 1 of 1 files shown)


cjcolvar:
In doing this work I also noticed that the mediaelement-add-to-playlist gem isn't locked down in the Gemfile.  We'll need to tag the gem and pin to that in the Gemfile.  I can add that to this PR if others want.

PR #2331, opened 2017-09-18
Mark intermittently failing tests as pending so they run but don't affect build results

cjcolvar:
This is a fallback if we're not able to solve this issue during the timebox of #2281.


cjcolvar:
It looks like `pending` will fail tests that pass instead of letting them pass.  The alternative to `pending` is `skip` which won't run the tests.  I wish there was an `optional` directive which would pass if the test passes and pass with the pending message if the test fails.

PR #2336, opened 2017-09-19
Move authorization of playlists that are private with token into the ability.

### LINKED ISSUE #2332: Markers Do Not Display for Viewers In Playlists with "Share by link" Permissions
### Description
When a user views a playlist with "Share by link" permissions, the markers set on playlist items do not appear.

### Done Looks Like
When viewing a playlist item within a playlist context that contains markers, if the user has permissions to access the media object which is the source of playlist item the markers from that playlist should appear.

cjcolvar:
Fixes #2332 and other bug mentioned this morning that doesn't have a ticket yet: playlists that are private with token don't show the copy playlist button on the view page.

The fix to exposing the copy playlist button reveals a different problem that causes copying from playlists that are private with token to fail.  @lealeelu is working on it.

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -18,8 +18,9 @@ class PlaylistsController < ApplicationController
   include ConditionalPartials
 
   before_action :authenticate_user!, except: [:show, :refresh_info]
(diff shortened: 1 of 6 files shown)


cjcolvar:
Closing this in favor of two PRs with the original targeting 6.2 and the follow on work targetting 6.3.

PR #2341, opened 2017-09-21
Bugfix/markers share with link

### LINKED ISSUE #2333: Copy Button Not Working or Displaying for Viewers of Playlists with "Share by link" Permissions
### Description
When a playlist is set to "Share by link" for its permissions, the Copy button that usually appears doesn't show up for anyone except the playlist owner.

### Done Looks Like
For a viewer of a playlist that is "Share by link", the copy button should appear as it does when viewing a public playlist.

lealeelu:
Playlist tokens are passed into abilities
fixes #2333 and other errors that popped up while fixing this.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -164,14 +164,10 @@ class Ability
     if @user.id.present?
       can :manage, Playlist, user: @user
       # can :create, Playlist
-      can :duplicate, Playlist, visibility: Playlist::PUBLIC
-      can :duplicate, Playlist do |playlist|
+      can [:read, :duplicate], Playlist, visibility: Playlist::PUBLIC
+      can [:read, :duplicate], Playlist do |playlist|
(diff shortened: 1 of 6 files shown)


app/models/ability.rb
```diff
@@ -164,11 +164,10 @@ def playlist_permissions
     if @user.id.present?
       can :manage, Playlist, user: @user
       # can :create, Playlist
-      can :duplicate, Playlist, visibility: Playlist::PUBLIC
-    end
-    can :read, Playlist, visibility: Playlist::PUBLIC
-    can :read, Playlist do |playlist|
-      playlist.valid_token?(@options[:playlist_token])
+      can [:read, :duplicate], Playlist, visibility: Playlist::PUBLIC
```
bkeese: Did we decide only logged-in users can view public/shared playlists?

PR #2363, opened 2017-10-02
Playlist Tags

### LINKED ISSUE #1538: As a User I Want A System for Categorizing and Organizing Playlists
## Description
As a user, I want to categorize and organize my playlists for each of searching and sorting. 

### Use Case
For users with tens or hundreds of playlists, navigation can quickly become cumbersome. Besides the name of the playlist, there is no way to order or arrange lists of playlists. For a faculty member who wants to share a subset of playlists with students of a particular course, for example, there would have to be a naming convention used in the name or some external means of organization for them to be in one place.

#### Tags
In an initial planning meeting tags were decided upon as the means to categorize playlists for the purposes of searching and sorting.

Tags would be user-defined and allow a single playlist to “belong” to multiple courses, topics, etc.
Allowing arbitrary tags to be placed on a playlist will aid in searching and sorting. For example, a faculty member may tag a number of playlists with the tag 'MUS 152', indicating that playlists with that tag correspond to a particular course.

Users can add one or more _tags_—descriptive strings— to playlists in the playlists view. These tags appear in their own column. A search bar is available to search only for specific tag

### Done Look Like
Initial design work is done and circulated to key users for feedback before implementation.

### Tasks
 - [ ] Add tags to the Playlist Model and corresponding unit tests
 - [ ] Add tag list to the playlist edit page
 - [ ] Create tag editing field to be re-used whenever someone wants to edit tags
 - [ ] Add tag list and edit button to the playlist index page

### LINKED ISSUE #2333: Copy Button Not Working or Displaying for Viewers of Playlists with "Share by link" Permissions

### LINKED ISSUE #2359: Playlist Tags: Add filter to playlist index page
Refers to #1538

lealeelu:
Fixes #1538 
Fixes #2359 
Fixes #2333 

Adding the ability to add tags to playlists and filter by those tags.
(and fixes playlist#duplicate action)

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -1014,7 +1014,7 @@ div.structure_edit {
 }
 
 /* Playlist index page */
-#playlists_top {
+.playlists_top {
   text-align: right;
 }
 
@@ -1030,22 +1030,27 @@ div.structure_edit {
   float: right;
 }
 
-#playlist_filter_container {
+.playlist_filter_container {
   display: inline-block;
   margin-left: 1em;
 }
 
-#tag_filter_container {
+.tag_filter_container {
   display: inline-block;
 }
 
-#tag_filter {
+.tag_filter {
   display: inline;
   width: auto;
 }
+
+.tag-button-wrapper {
+  margin-bottom:.4em;
+}
+
 /* Playlist show page */
 .queue {
-  opacity:.80;
+  opacity:.8;
   &:hover {
     opacity:1;
   }
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -1013,6 +1013,36 @@ div.structure_edit {
   visibility: visible !important;
 }
 
+/* Playlist index page */
+#playlists_top {
+  text-align: right;
+}
+
+#Playlists_info {
+  float: left;
+}
+
+#Playlists_length {
+  display: inline-block;
+}
+
+#Playlists_paginate {
+  float: right;
+}
+
+#playlist_filter_container {
+  display: inline-block;
+  margin-left: 1em;
+}
+
+#tag_filter_container {
+  display: inline-block;
+}
+
+#tag_filter {
+  display: inline;
+  width: auto;
+}
 /* Playlist show page */
 .queue {
   opacity:.80;
@@ -1136,6 +1166,11 @@ h5.panel-title {
     padding-left:20px;
     float:right;
   }
+}
+
+.info-text-gray {
+  font-style:italic;
+  color: gray;
 }
 
 /* Mediaelement styles */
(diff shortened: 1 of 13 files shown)


joncameron:
ao7okxyawt9sw

app/assets/stylesheets/avalon.scss
```diff
@@ -1013,6 +1013,36 @@ div.structure_edit {
...
+}
+
+#tag_filter_container {
+  display: inline-block;
+}
+
+#tag_filter {
+  display: inline;
+  width: auto;
+}
```
adamjarling: Just a note, it's generally best practice to target `class`, instead of `id` in CSS.  A lot of code quality tools (like Hound for example), will complain.

app/assets/stylesheets/avalon.scss
```diff
@@ -1013,6 +1013,36 @@ div.structure_edit {
...
 /* Playlist show page */
 .queue {
   opacity:.80;
```
adamjarling: I'd change this to `.8` instead for decimals.  Another general best practice pattern in CSS.

app/views/playlists/_item_list.html.erb
```diff
@@ -45,6 +45,13 @@ Unless required by applicable law or agreed to in writing, software distributed
     <%= @playlist.comment %>
   </div>
 <% end %>
+
+<div style="margin-bottom:.4em;">
```
adamjarling: Again, not a huge deal but moving forward we should avoid in-lining styles in the current structure of this application, and create classes  ie. `<div class="my-awesome-buttons-wrapper">` or something hopefully re-usable and/or consistent with margin/padding definitions.

app/views/playlists/index.html.erb
```diff
@@ -27,6 +27,15 @@ Unless required by applicable law or agreed to in writing, software distributed
     <div class="col-lg-10">
       <h1 class=>Playlists <small>(<%= playlists.size %> total)</small></h1>
       <% unless playlists.empty? %>
+        <div id="tag_filter_container">
+          <div>Filter: <select id="tag_filter" name="tag_filter" class="form-control input-sm" style="width:auto">
```
adamjarling: Is this wrapping `<div>` around the `<select>` element necessary?

lealeelu:
@adamjarling made the requested changes. Cleaned up some of the inline CSS.

adamjarling:
@lealeelu Looks good.  Just as a follow up, if you need to the `id` attribute on any of those elements (for example, you're using the `id` as a jquery selector, which is performant and fast), it's ok to sometimes have `<div id="my-element" class="my-element"></div>`.  As long as you're not losing any selector action, the changes look great.

lealeelu:
Cool! Thanks @adamjarling. That's good to know.

PR #2370, opened 2017-10-06
Remove old config examples

phuongdh:
settings.yml become settings.yml.example for easier cap deploy


cjcolvar:
Looks like you need to copy `settings.yml.example` to `settings.yml` in `.travis.yml` (and remove the existing `cp` commands)?

PR #2382, opened 2017-10-13
Make tag filtering use exact search

### LINKED ISSUE #2379: Filter option in playlist page is doing a wildcard search rather than the exact string for tags.
### Descriptive summary
Testing: Testing the ticket #1538 
Browser: Chrome(61.0.3163.100 (64-bit))
Operating System: Windows10
Issue: Filter option in playlist page is doing a wildcard search rather than the exact string for tags.

### Steps to reproduce
1. Create a set of playlists, tag the playlists and make sure that some tags of a playlist matches the sub string of some tags in other playlist.
2. Go to the playlist page, click on Filter drop-down and select the tag that is also a sub string of other tags.
3. The expected result would be the exact match of the tag, but it is also returning the tags that contains the selected tag as sub string.
image

bkeese:
Fixes #2379 

Makes tag filter for playlists use exact search.

This is complicated by the fact that tags are a 'serialized' activerecord field and the search is over the serialized data. This *might* be dependent on dbms implemented.

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -55,12 +55,12 @@ class PlaylistsController < ApplicationController
     playlistsFiltered = playlists.where("title LIKE ?", "%#{params['search']['value']}%")
     # Apply tag filter if requested
     tag_filter = params['columns']['5']['search']['value']
-    playlistsFiltered = playlistsFiltered.where("tags LIKE ?", "%#{tag_filter}%") if tag_filter.present?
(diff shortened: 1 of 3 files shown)


bkeese:
I made that comment before understanding that activerecord serializes these fields as yaml. I now have high confidence that this will work regardless of dbms.

cjcolvar:
Could you add a test for this?

bkeese:
Do these scopes accumulate? The final list of playlists should apply the user, title, and tags scopes simultaneously.

cjcolvar:
My first commit wasn't chaining them so it was bad.  I redid it and now that github has finally attached my second commit to this PR, I think it should be good.  So please look again.

PR #2401, opened 2017-10-17
remove the batch file after

### LINKED ISSUE #2390: Batch ingest going into a loop and ingesting the same items multiple times
### Descriptive summary
Testing: Regression 
Browser: Chrome(61.0.3163.100 (64-bit))
Operating System: Windows10
Issue: Batch ingest going into a loop and ingesting the same items multiple times

### Steps to reproduce
1. Create a manifest file and start the batch ingest using it.
2. Keep checking the e-mail notifications and the ingest process in avalon.
3. Multiple mails will be coming saying that batch process is successful, and can observe the same item ingesting multiple times in the collection.

carrickr:
Fixes #2390

### DIFF
--- a/spec/lib/avalon/batch_ingest_spec.rb
+++ b/spec/lib/avalon/batch_ingest_spec.rb
@@ -73,7 +73,7 @@ describe Avalon::Batch::Ingest do
 
     it 'deletes the manifest after registering' do
        batch_ingest.scan_for_packages
-       expect(File.exist?(@batch.manifest.file)).to be_falsey
+       expect(FileLocator.new(@batch.manifest.file).exists?).to be_falsey
     end
 
     it 'does not persist anything to fedora' do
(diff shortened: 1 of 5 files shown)


cjcolvar:
This works for local filesystem, but I think we'll need to change this into a method on `Avalon::Batch::FileManifest` and `Avalon::Batch::S3Manifest` (maybe also `Avalon::Batch::Manifest`?).

carrickr:
@cjcolvar I think we need to tag in @mbklein and maybe have a larger talk on this.  We have the issue of needing the AWS gem for Avalon on AWS builds, but what about people running on IBM's cloud, etc (and we have people doing that).  We need to talk pattern.  Is it up to individual institutions to just find and replace FileUtil with the proper RubyGem (AWS/IBM/etc) or what?

spec/lib/avalon/batch_ingest_spec.rb
```diff
@@ -114,14 +119,16 @@
     end
 
     it 'should ingest batch with spaces in name' do
-      space_batch_path = File.join('spec/fixtures/dropbox/example batch ingest', 'batch manifest with spaces.xlsx')
+      FileUtils.cp_r 'spec/fixtures/dropbox/example batch ingest', @dropbox_dir
```
cjcolvar: Do you need this change (and in the test below)?  I'd rather not create files during the tests since they don't always get cleaned up.

PR #2403, opened 2017-10-17
Install bootstrap-toggle-rails from git instead of rubygems

mbklein:
The released version of `bootstrap-toggle-rails` was built with restrictive (`0640`) permissions. Installing on AWS (or any other platform where the user running the app has a different uid/gid from the user that ran `bundle install`) results in an `Errno::EACCES` error during asset compilation. Installing from git instead of Rubygems fixes the issue.


cjcolvar:
Do you want to lock this to specific branch or tag?

mbklein:
There are only three commits in the repo and the most recent is more than a year old. There's only one release tag, and it's pointing at the HEAD commit. I'd be OK pinning it later, but for now, I feel like it's pretty safe to do it this way.

PR #2404, opened 2017-10-17
rake tasks and schedule for other batch jobs

### LINKED ISSUE #2398: Write Rake Task for Batch Complete Email
Currently nothing checks to see if the batch is complete and alert anyone

### LINKED ISSUE #2399: Write Rake Task to Alert on Stalled Batch Jobs
Currently not checked by anything

carrickr:
Fixes #2399 
Fixes #2398

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -223,6 +223,16 @@ EOC
         end
       end
     end
+
+    desc "Starts Status Checking and Email Notification of Existing Batches"
+    task :ingest_status_check => :environment do
+      IngestFinished.perform_later
+    end
+
+    desc "Status Checking and Email Notification for Stalled Batches"
+    task :ingest_stalled_check => :environment do
+      StalledJob.perform_later
+    end
   end
   namespace :user do
     desc "Create user (assumes identity authentication)"
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -6,7 +6,7 @@ every 1.minute do
   locking_rake "avalon:batch:ingest", :lock_name => "batch_ingest", :environment => ENV['RAILS_ENV'] || 'production'
 end
 
-every 1.minute do
+every 15.minutes do
   locking_rake "avalon:batch:ingest_status_check ", :lock_name => "batch_ingest", :environment => ENV['RAILS_ENV'] || 'production'
 end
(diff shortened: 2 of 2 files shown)


lib/tasks/avalon.rake
```diff
@@ -223,6 +223,16 @@ EOC
         end
       end
     end
+
+    desc "Starts Status Checking and Email Notification of Existing Batches"
+    task :ingest_status_check => :environment do
+      IngestFinished.perform_later
```
cjcolvar: If this is running every minute, I think we need to protect it from sending duplicate emails.  Maybe we could do this by wrapping it in a `WithLocking` like `avalon:batch:ingest` and changing it to `perform_now`?

lib/tasks/avalon.rake
```diff
@@ -223,6 +223,16 @@ EOC
...
+    end
+
+    desc "Status Checking and Email Notification for Stalled Batches"
+    task :ingest_stalled_check => :environment do
+      StalledJob.perform_later
```
cjcolvar: If this is run once a day then I don't think we need to worry about multiple of the jobs running at the same time or we'd have some other pretty major issues.  But maybe if they were given a low priority it might be possible to happen?

lib/tasks/avalon.rake
carrickr: I just have it at one minute for testing, my intent was to back it off to every 15 minutes or so in production, no reason to hit our DB that hard, I'll back it off

lib/tasks/avalon.rake
carrickr: I'm not quite clear on the proposed change/fix here

cjcolvar: I think this is okay for now unless we wanted to be consistent across the three tasks.  I think there is still a risk to double running things albeit very very low.

lib/tasks/avalon.rake
cjcolvar: I think the main cost here is loading the rails context every minute.  It makes sense to be as responsive as possible when the batch is getting loaded but every 15 minutes on reporting completed is probably fine for now.  In that case, if we push it to 15 minutes then we could skip the locking and see if the issue ever comes up.

PR #2416, opened 2017-10-23
Handle section clicked transitions

adamjarling:
This PR code does quite a bit of re-work to how section links interact with the player, and new instances of the player loading in the page.

The main takeaways are probably:
- Added `id` attributes to some of the section links to better target elements (slight optimization?)
- Section highlighting now works with clicking on section links

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -25,11 +25,18 @@ class MEJSPlayer {
     this.addSectionsClickListener()
     // audio or video file?
     this.mediaType = (this.currentStreamInfo.is_video === true) ? 'video' : 'audio'
-    // Flag whether to play a specified range of media clip (ie. 00:15 - 1:23)
-    // Default 'false' indicates play from beginning (00:00)
-    this.playRangeFlag = false
-    // Temporary holder for clip range data
-    this.playRangeData = {}
+
+    // Helper object when loading a new MEJS player instance (ie. a different media object source section link clicked)
+    this.switchPlayerHelper = {
+      active: false,
+      data: {},
+      paused: false
+    }
+
+    // Array of all current segments for media object
+    this.segmentsMap = this.createSegmentsMap()
+    // Holder for currently active segment DOM element 'id' attribute
+    this.activeSegmentId = ''
 
     // Initialize the player
     this.initializePlayer()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -119,6 +145,63 @@ class MEJSPlayer {
...
+    // Use the active segment fragment,
+    try {
+      t = [parseFloat(segment.fragmentbegin), parseFloat(segment.fragmentend)]
+    } catch(e) {
+      t = this.currentStreamInfo.t.slice(0)
+    }
+
+    // Ensure t range array has valid values
+    t[0] = (isNaN(parseFloat(t[0]))) ? 0 : t[0]
+    t[1] = (t.length < 2 || isNaN(parseFloat(t[1]))) ? 100 : t[1]
```
bkeese: I think the fallback value here should be duration, not 100.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -264,66 +342,98 @@ class MEJSPlayer {
...
+    const currentTime = this.player.getCurrentTime()
+    let activeId = ''
+
+    // Get the segment for player's current time value (if one exists)
+    for (let segmentId in segmentsMap) {
+      let begin = parseFloat(segmentsMap[segmentId].fragmentbegin)
+      let end = parseFloat(segmentsMap[segmentId].fragmentend)
+
+      if ( (segmentsMap.hasOwnProperty(segmentId)) && (currentTime >= begin && currentTime < end) ) {
+        activeId = segmentId
```
bkeese: Maybe break the for-loop here? No need to continue through the structure. Side effect is that for items with structure that overlaps, it will pick the first match instead of the last.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
adamjarling: Good call. Think I got this mixed up with "100%".

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
adamjarling: Totally, another good catch.

adamjarling: Actually, this is a tricky one.  With the `break`, it picks the first, however in an overlapped test item, it wrongly picks the first if a later, overlapping item is clicked upon.

adamjarling: Here's the structured metadata I'm testing against.  I could be missing something, but for now this will only work with the break in there.  Feel free to pull down and let me know if we're envisioning different test data?

```
<?xml version="1.0"?>
<Item label="Bunny Video">
    <Div label="Funny Parts">
        <Span label="Funny Part 1" begin="0" end="00:10"/>
        <Span label="Funny Part 2" begin="00:12" end="00:17"/>
    </Div>
    <Div label="Scary Parts">
        <Span label="Scary Part 1" begin="00:25" end="00:35"/>
        <Span label="Scary Part Overlap" begin="00:30" end="00:40"/>
        <Span label="Scary Part Overlap" begin="00:35" end="00:46"/>
        <Span label="Scary Part 2" begin="00:38" end="00:49"/>
    </Div>
</Item>
```

bkeese: I think we need to find out if overlapping is even legal. I suspect it shouldn't be, because expected behavior is not well-defined.

adamjarling:
@bkeese I pushed up the latest updates to this branch.  The `break` line you suggested in the potentially overlapping sections is same as it was originally.  I walked through the code, and seems this works and breaking out early selects the wrong item.  For what I'm testing at least.

PR #2430, opened 2017-10-27
Add mejs4 icon svg backgrounds

lealeelu:


lealeelu:
I rebased on develop, I don't understand why that extra code is there.

lealeelu:
Ok, redid the commit without those extra changes. Problem now is that I'm getting this error when trying to watch a video from the video's page. Playlist playing works fine for some reason. 
image
It seems unrelated to what I was doing.

bkeese:
@adamjarling: I think this is a bug in the MEJS4 segment PR. Happens when there is a mediaobject with only one masterfile and no structure, right? In that case, there is no accordian for the new js to reference.

adamjarling:
@bkeese Ah, I'll pull it down and look at that scenario.  Sorry about that.

PR #2441, opened 2017-11-10
Display error when cannot set cookie in embed context

### LINKED ISSUE #2319: LTI doesn't login user in certain cases
Reported bug:
A user who has never visited the Avalon instance tries to access Avalon via LTI within Canvas (iframe not new window) and ends up at the normal facet query but without any results because the user wasn't logged in.

This was reported on Chrome and reproduced on Safari on Mac (@jwd's machine and browserstack).

Steps to reproduce:
1) Clear cookies and cache
2) Go to canvas
3) Click on an embedded course

The underlying issue here is that Safari (and some versions of other browsers) will not accept cookies from iframes for sites that have never been visited before.  (One explanation of this here: http://labs.fundbox.com/third-party-cookies-with-ie-at-2am/)

Workaround right now it to visit Avalon before trying to load embed (including Canvas).

To solve this we'll either need to:
1) do some sorcery to get the user to view Avalon directly and then reload the embed context
2) make it so Avalon embeds don't need cookies

@pdinh has been working on approach #1.  I'm not sure if approach #2 is even feasible but if it is it might be larger/riskier than #1.

phuongdh:
Fixes #2319

### DIFF
--- a/app/views/modules/_flash_messages.html.erb
+++ b/app/views/modules/_flash_messages.html.erb
@@ -40,3 +40,18 @@ Unless required by applicable law or agreed to in writing, software distributed
   <% end %>
   </div>
 <% end %>
+
+<div class="alert alert-danger hidden" id="cookieless">
+  <p>Could not set cookie, this may happen if you've never visited this website before. If you see this message and cannot log in or see content, please open <%= link_to root_url, root_url, target: :_blank %> in a new window, then come back and refresh this page.</p>
+</div>
+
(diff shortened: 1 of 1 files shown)


adamjarling:
Looks good to me.

cjcolvar:
@phuongdh I think this is good, but there are failing feature tests in `spec/features/capybara_homepage_spec.rb`.  Could you take a look at fixing those?

Issue #1879, opened 2017-05-04
Media element 4 - Track Scrubbing


adamjarling:
### Description
Re-create track scrubbing (lil' scrub) functionality from Avalon 6.x player into MEJS4 player.  After a discussion with @joncameron we're going to limit the scope of this ticket to just the mediaobject view, and not the playlist page.

### Done Looks Like
lil' scrub is implemented in  Mediaelement 4, along with the highlighting on the main scrub bar (#2184).

### Tasks
- [x] Add scrubber icon (magnifying glass) to player
- [ ] Make scrubber icon toggle lil' scrub.
- [ ] Add track highlighted on main scrubber (#2184)
- [ ] Verify track scrubbing functionality works as expected

davidschober:
@adamjarling for this one you're looking at the little zoom scrubber, correct?

adamjarling:
@davidschober Yea I was going to try using the scrubber plugin, see how to wire that into the Hyrax app.

adamjarling:
Pointing this currently at a 5, as some of the plugin documentation to bring into a Rails application is somewhat limited.   First approach is to re-use previously created mediaelement plugin.  If that doesn't work with current version of mediaelement, will need to create from scratch which will be significantly more work.

davidschober:
@adamjarling is this issue still relevant?

adamjarling:
@davidschober Yes, it's still relevant as an overarching work item, but the description, Done Looks Like, and Tasks, may need to be revised somewhat.  I do that before starting on this story.

adamjarling:
I think it makes sense to complete #2184 first, before working on this ticket.

PR #2444, opened 2017-11-13
MEJS4 track scrubber in media object page #1879

adamjarling:
There's an issue in Safari on the Playlists page I noticed for both MEJS2 and MEJS4 (different issues), where it doesn't correctly produce the track scrubber current time visual.  I created a separate story for this.  

Also on inspection in mobile, it was really hard to tell if it's working as expected or not, testing locally against Browserstack.   Maybe we can get it up to Spruce (the dev server?) first then I can Browserstack it from there for hopefully better streaming performance.

For the most part I took all @bkeese coffeescript code for the track scrubber plugin and converted it to ES6 js, and made the tweaks to get it to work with MEJS4, and our new JS player wrapper script.  Some things I didn't want to get too deep into now (ie. the `initializeTrackScrubber` method), but at some point when there's more time I think we could optimize this a touch and break it down.

I updated the color scheme of the scrubber visual itself (to Avalon lime green), temporarily just to highlight the fact it's MEJS4 instead of 2.  And made it larger.

Any questions... let me know.


adamjarling:
@bkeese Yea, just saw those and I'll try to eliminate them.

adamjarling:
@bkeese The 2 existing codeclimate errors are due to function complexity.  I'd prefer to address this at a later time, when I look at the Safari issue, as some re-factoring should be done to clean up some tech debt.  That work for you?   Ok to merge?

PR #2446, opened 2017-11-14
Use a count solr query instead of loading every MediaObject

cjcolvar:
This reduces the time needed to render a collection as json.  In support of IU's MDPI work.


bkeese:
Care to fix the indentation issues? Otherwise +1

cjcolvar:
Hmm...the indentation matches the previous methods that are under private and are 2 in from the private which is what I think is typical but maybe our rubocop profile doesn't do.  I say forget about it and merge.

PR #2455, opened 2017-11-16
Restore find_for_identity to ensure users have usernames

### LINKED ISSUE #2449: Username shows up as a number instead of email
After the recent AWS branch merge, new user shows up as the row id, e.g. "41", instead of the email, e.g. "archivist1@example.com". 

I believe this is a user_key setting in Devise and it should default to email..

cjcolvar:
Without this users get created with numbers for usernames causing lots of issues.

Fixes #2449.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -59,7 +59,13 @@ class User < ActiveRecord::Base
   def self.find_for_generic(access_token, signed_in_resource=nil)
     username = access_token.uid
     email = access_token.info.email
-    User.find_by_username(username) || User.find_by_email(email) || User.create(username: username, email: email)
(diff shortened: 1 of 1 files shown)


mbklein:
Instead of writing different methods for different providers, I'd much rather abstract the `username`/`email` lookup (the way Devise does with `user_key`), so `find_for_generic` can do the heavy lifting while we just pull the lookup keys out of `Settings`. 

I found I got partway there in one of my AWS commits, but not to the point of the fix described above.

I'd be willing to accept this PR as a temporary solution but tag it for a refactor.

cjcolvar:
@mbklein If you're okay with this, can you please merge it?

PR #2457, opened 2017-11-16
Skip completed objects when rerunning the migration

cjcolvar:
In support of IU's migration of our large Avalon instance.

@pdinh will you please test and provide feedback?


cjcolvar:
Squashed.

cjcolvar:
@phuongdh I thought code climate was turned off for the migration code but 4 issues got flagged.

PR #2458, opened 2017-11-16
Use Bixby for Samvera-style style checking

cjcolvar:
(This will also deal with the rubocop version issue by updating to 0.50.0.  Although the warning will persist until we release to master.)


cjcolvar:
@phuongdh Can you review this?

PR #2476, opened 2017-11-28
Update MEJS4 add to playlist not logged in bug

adamjarling:
This fixes a bug where when not logged in, the Add to Playlist control button on MEJS4 player still displayed.

### DIFF
--- a/app/views/media_objects/_mejs4_player_js.html.erb
+++ b/app/views/media_objects/_mejs4_player_js.html.erb
@@ -4,12 +4,16 @@
 <% content_for :page_scripts do %>
 <%= javascript_include_tag "mejs4_player" %>
 <% end %>
-
 <script>
+var mejs4ConfigFeatures = ['playpause', 'current', 'progress', 'duration', 'trackScrubber', 'volume', 'quality', 'createThumbnail', 'addToPlaylist', 'fullscreen'];
+
+if (<%= cannot? :create, Playlist %>) {
+  mejs4ConfigFeatures.splice(mejs4ConfigFeatures.indexOf('addToPlaylist'), 1);
+}
 // Define MediaElement 4 configuration here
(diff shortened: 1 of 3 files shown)


cjcolvar:
I wonder if we can do this another way than passing the current user into the JS.

adamjarling:
@cjcolvar Sure, I'm all ears.  Is there some kind of flag available?  I need some lessons I think in deciphering and reading what `byebug` object state output values are in Rails, at some point.  It's hard to comb through the multi-line stuff.  Bet there's a trick.

adamjarling:
@cjcolvar Should I instead set a boolean value whether `current_user` exists, and pass that to the JS instead of the user object?  I'll test that.

app/views/media_objects/_mejs4_player_js.html.erb
```diff
@@ -9,6 +9,7 @@
 // Define MediaElement 4 configuration here
 var mejs4AvalonPlayer = new MEJSPlayer({
   currentStreamInfo: <%= section_info.to_json.html_safe %>,
+  currentUserExists: <%= current_user_exists %>,
   features: ['playpause', 'current', 'progress', 'duration', 'trackScrubber', 'volume', 'quality', 'createThumbnail', 'addToPlaylist', 'fullscreen'],
```
cjcolvar: I was thinking that we could just skip the 'addToPlaylist' feature here if the user `cannot? :create, Playlist`.  If that works out then we wouldn't need to pass anything into the JS.

adamjarling: Ok, I'll try implementing this.

adamjarling: Pushing up this approach now.  I added an extra global JS var to store the features array which I'm not super thrilled about, but seems better than repeating the entire array text in a conditional block.

PR #2477, opened 2017-11-29
Issue#2426 refactor breakup js files

adamjarling:
This PR mainly re-organizes the MEJS4 JS files, breaks out smaller classes for better structure and easier to navigate.

### DIFF
--- a/app/assets/stylesheets/mejs4/mejs4_add_marker_to_playlist.scss
+++ b/app/assets/stylesheets/mejs4/mejs4_add_marker_to_playlist.scss
@@ -8,6 +8,10 @@
   width: 20px;
 }
 
+.mejs__time-marker {
+  top: 0;
+}
+
 #markers {
   table {
     tr {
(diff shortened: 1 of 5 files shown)


cjcolvar:
Now that I merged #2476, looks like you need to update this PR to reflect the changes there.

adamjarling:
Ah, got it.  Just fixed the conflict and re-pushed.  Thanks for the heads up.

app/assets/stylesheets/mejs4/mejs4_add_marker_to_playlist.scss
```diff
@@ -8,6 +8,10 @@
   width: 20px;
 }
 
+.mejs__time-marker {
```
cjcolvar: Why is this needed in the refactor?

adamjarling: Consider this an enhancement to the refactor.   It was a small style bug most likely introduced when I committed the track scrubber code, but which I never caught.  And only happens when I call a method on the MEJS4 core markers plugin directly, to "re-paint" the markers after an update.

adamjarling:
Thanks @cjcolvar .  An added benefit of re-structuring this way will make unit testing (when we introduce) much more manageable.

Issue #2279, opened 2017-08-31
MediaElement 4 transition code cleanup


adamjarling:
### Description

Once MediaElement4 is fully implemented, tested and ready for release, we will need to clean up and flip all development conditional code which currently allows swapping MediaElement 2 from MediaElement 4 players via the query string utility currently in place. ME4 should be the default and ME2 should be the query string flippable. 

### Done Looks Like
- MediaElement4 is the default library used in codebase.
- ME2 and ME4 switch is flipped so that a person can test ME2 if needed. 

### Tasks

- [ ] Refactor any conditional code pertaining to MediaElement2 / MediaElement4 switching via query param tool used in development.

@davidschober @joncameron

cjcolvar:
@adamjarling Is this story done or done enough given that MEJS4 is default now?

adamjarling:
Hmmm, your call.  If we're still supporting MEJS2 in this release, then I'd say we could close this story and just leave all MEJS2 code in place.  If we're abandoning MEJS2, then it might be good to do some housekeeping and see what looks like CoffeeScript and/or partials and other stuff lingering around.

cjcolvar:
@avalonmediasystem/product-owners what do you think?  I'd vote for leave it in place for now for easier debugging and because it's easier and less risky than removing it now.

davidschober:
I’m ok with leaving it for comparison.

davidschober:
That said, why is less risky? I’m not sure I like leaving old code around.

cjcolvar:
I think my thought here was that we might accidentally remove something that is needed and break things when we've been testing so far with the code still in place.  I'm all for removing but maybe just keeping that for 6.4.

davidschober:
OK. I'll call this 6.x

davidschober:
Scope of this is to look over for health of the code.

joncameron:
This was mostly completed a while ago, and finished in the Rails 5.2 push.

PR #2480, opened 2017-11-30
Use MEJS4 unless told to do otherwise

cjcolvar:
Make the switch since we're close and don't really need to test with MEJS2 anymore.  This will make testing easier for 6.3 release.


cjcolvar:
Relates to #2279

Issue #2481, opened 2017-11-30
Ensure Player API works


cjcolvar:
## Description

We need to verify the player API continues to work. 

## Done looks like 

- [x] verification
- [x] issues created and prioritized

cjcolvar:
https://wiki.dlib.indiana.edu/display/VarVideo/Player+API

cjcolvar:
This could be tested by modifying this wiki page (https://wiki.dlib.indiana.edu/display/VarVideo/Test+Embed) and making sure everything described in the player api wiki page works.

bkeese:
@yingfeng-iu did you get this email?

yingfeng-iu:
Hi Adam, after adding currentPlayer I'm now getting new TypeError: currentPlayer is null, which means mejs4AvalonPlayer.player is not initiated, or the type is wrong. I see that in mejs2, currentPlayer =  AvalonPlayer.player, which has type MediaElementPlayer, in mejs4, I was setting it to MEJSPlayer.player, which also is MediaElementPlayer instance. it's assigned in MEJSPlayer.initializePlayer. I'm going to debug it a bit, if you have suggestions please let me know. thanks!

adamjarling:
@yingfeng-iu Could this be a timing issue (maybe when the currentPlayer is set the player is not yet initialized?)  Or is the problem the iFrame window context?   I need to split for the day, but if you want to screen share tomorrow morning and show me what's going on maybe that'd help.

yingfeng-iu:
Adam/Brian, thanks for helping!

Populating the currentPlayer inside success event didn't work, that method didn't get called during loading of the page. I ended up adding it inside the mejs4 player initialize method, but also have to declare it as global var in the script files (otherwise it's not visible outside the class).
Tested locally, all functions work except that toggle loop doesn't seem to have any effect. Maybe we need to turn on some mejs4 setting for that?

yingfeng-iu:
after deploy, tested it again pointing to spruce and the buttons all work. closing the ticket. didn't find issues so no new ticket created

bkeese:
I also tested this and it looks good!

PR #2491, opened 2017-12-04
#2481: add definition of currentPlayer

yingfeng-iu:
Issue #2481: add definition of currentPlayer


cjcolvar:
@adamjarling we are still using it in a few places the most notable being the Player API:
 https://github.com/avalonmediasystem/avalon/blob/develop/app/views/master_files/_player.html.erb#L31-L38
Is there a better way to reference the currently loaded media element player?  Should this code be pushed into the player instead of on the page?

yingfeng-iu:
Chris, can I or someone squash the commits when merging the PR? 

Adam, thanks for the comment. currentPlayer is referenced in window.addEventListener in _player.html.erb, that's where the console error came from. In the current code, both mejs2 and mejs4 are provided as options. I guess if we don't need mejs2 player anymore, we can just use mejs4AvalonPlayer.player directly in place of currentPlayer. Would you prefer me to do that? Although I just took a quick look, looks like refresh_info.js.erb also reference currentPlayer, so we'll need to change there too. Let me know what's the preference. Having currentPlayer might have a small advantage for the future if we want to keep flexibility of having multiple options for players.

phuongdh:
@yingfeng-iu I notice you're using 4-space tab for indentation while our code use 2-space indent.

app/views/master_files/_mejs4_player_js.html.erb
```diff
@@ -12,4 +12,5 @@ var mejs4AvalonPlayer = new MEJSPlayer({
   features: ['playpause', 'current', 'progress', 'duration', 'volume', 'quality', 'fullscreen'],
   highlightRail: true
 });
+	currentPlayer = mejs4AvalonPlayer.player;
```
phuongdh: I don't think there should be an indentation here?

app/views/media_objects/_mejs4_player_js.html.erb
```diff
@@ -16,4 +16,5 @@ var mejs4AvalonPlayer = new MEJSPlayer({
   features: mejs4ConfigFeatures,
   highlightRail: true
 });
+	currentPlayer = mejs4AvalonPlayer.player;
```
phuongdh: Same indentation issue in these similar places

adamjarling:
Ah, ok.  @cjcolvar In general, it'd be better to handle player events on the player instance object rather than `window`, but there might be a reason for doing so... I'm not sure how/why it was implemented.

@yingfeng-iu Yea, referencing the player via `mejs4AvalonPlayer.player` may be a way to reference the current player instance as well.  Once I get this playlist items page worked out (and a bunch of more structural changes I have to make to the main wrapper file to get it working), we should have a 10,000 foot look at our JS approach with the player, plugins and talk through how the whole thing works.  Assuming it'll work:)

cjcolvar:
@adamjarling For the player api it is really the window that needs to be listening for the event: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
This code is for the player embedded within an iframe to be controlled by the parent page.

adamjarling:
@cjcolvar Got it. I keep forgetting the embed scenario.

app/views/master_files/_mejs4_player_js.html.erb
yingfeng-iu: removed indentations

app/views/media_objects/_mejs4_player_js.html.erb
yingfeng-iu: removed indentations

yingfeng-iu: removed indentations

PR #2494, opened 2017-12-04
Bugfix/#2024 close fullscreen

### LINKED ISSUE #2024: Actions performed using player controls (thumbnail/scrubber/add-to-playlist) are not visible in full-screen mode
### Observations
We should:
 1.) Hide the player controllers that bring up modal pages (playlist, thumbnail) when in full screen mode; or 
2.) Exit full screen mode when user clicks on them (the operations initiated by these controls hidden from the user in full screen mode). 
3.) A third option, which seems harder, would be to unhide the modal pages even when in full screen mode.

### Done Looks Like
- When a user clicks on a player controller in full screen mode, full screen mode exits.

### Tasks
- [ ] In the new MediaElement4 player, set up a click event handler for specified controls when in full screen mode. 
- [ ] Handle the event either by simply exiting full screen mode, or continuing to execute the clicked control.

bkeese:
Fixes #2024 
When in fullscreen mode and a control is clicked that requires a modal, exit fullscreen mode before showing modal.

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_add_to_playlist.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_add_to_playlist.es6
@@ -228,7 +228,9 @@ Object.assign(MediaElementPlayer.prototype, {
       handleControlClick: function (e) {
         const t = this;
         let addToPlayListObj = t.addToPlayListObj;
-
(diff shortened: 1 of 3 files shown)


cjcolvar:
@bkeese Can you remake this PR with just the second commit since the first is already in #2492?

PR #2500, opened 2017-12-05
Update docker-compose for AWS changes

bkeese:
Update docker config for AWS streaming config changes


bkeese:
Docker encoding and streaming no longer worked because of changes introduced in the AWS PR. This PR makes it work again.

phuongdh:
Beside these, I suspect there might be more broken configs that we haven't found yet.

cjcolvar:
So it sounds like we need documentation changes to account for the key changes of these configs.

bkeese:
I created an upgrade instructions page and put it there. https://wiki.dlib.indiana.edu/display/VarVideo/Upgrading+Avalon+6.2+to+Avalon+6.3

PR #2513, opened 2017-12-06
issue#2481: add declaration of global var currentPlayer and initialize

yingfeng-iu:
#2481: fixed the issue with previous PR, tested locally


yingfeng-iu:
Please squash the commit when merging

Issue #2504, opened 2017-12-06
ME 4js Punchlist Master Issue


davidschober:
# Playlist
- [x] Adding new marker refreshes page losing context like current playlist item and play time #2505
- [x] Markers not showing on scrubber #2307 
- [x] Switching between playlist items doesn't change video source #2507
- [x] Autoplay not working in playlists #2499
# Item player
- [ ] Playling canvas is wider than player background and hangs off a little on the right (covered by https://github.com/avalonmediasystem/avalon/issues/2266 ?) Update: fix later.
- [x] Quality control missing tooltip text #2508
- [x] Playthrough on item view page not working #2534
- [x] Can't navigate into player via keyboard (from either before or after in the DOM) (https://github.com/avalonmediasystem/avalon/issues/2130)
~- [ ] Escape key doesn't close thumbnail grabber modal (https://github.com/avalonmediasystem/avalon/issues/2130)~
~- [ ] Thumbnail grabber submit doesn't refresh the player page (won't fix?)~
- [x] Little scrub much bigger than in MEJS2 
- [ ] Poster image is cropped width-wise (related to #2266 ?)
# Embed player
- [x] Embedded player size needs to be the same as the MEJS2 player #2510
- [x] Embed missing title plugin (in progress in https://github.com/avalonmediasystem/avalon/issues/2096)
- [x] Embed missing info icon link to repository plugin (in progress as part of https://github.com/avalonmediasystem/avalon/issues/2096)

cjcolvar:
@davidschober @joncameron can we close this in favor of the remaining individual tickets?

joncameron:
@davidschober @cjcolvar I would say let's do it. I think this issue has accomplished what it needed to.

davidschober:
Done

PR #2515, opened 2017-12-07
Make player fill window in embed view #2510

### LINKED ISSUE #2510: ensure embed size of me4js player is the same as me2
## Description

Currently the embedded me4js player is not the same size as it's ME2 counterpart. Ensure that it's the same size.  

audio:

width: 600
Height 50

Width: 600
Height: 337

phuongdh:
Fixes #2510.

Part of #2504.


adamjarling:
Yea, looks good.

Issue #2521, opened 2017-12-11
MEJS4 player missing captions button


cjcolvar:
The MEJS4 player (including the embedded player) isn't showing the captions button.

Karthikeya212:
Tested on Spruce

PR #2522, opened 2017-12-11
Issue#2521: Add caption button to mejs4 player

yingfeng-iu:
issue#2521: add caption button to mejs4 player


cjcolvar:
Resolves #2521.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -294,7 +294,8 @@ class MEJSPlayer {
       success: this.handleSuccess.bind(this),
       embed_title: currentStreamInfo.embed_title,
       link_back_url: currentStreamInfo.link_back_url,
-      qualityText: 'Stream Quality'
+      qualityText: 'Stream Quality',
+      toggleCaptionsButtonWhenOnlyOne: 'true'
```
adamjarling: Can we make `toggleCaptionsButtonWhenOnlyOne`'s value `boolean` rather than type `string`?   ie: `toggleCaptionsButtonWhenOnlyOne: true` instead of ``toggleCaptionsButtonWhenOnlyOne: 'true'`

cjcolvar: I can try that and see if it works.

cjcolvar: That worked.  I've updated this PR.

PR #2526, opened 2017-12-12
Issue#2308 playlists page items mejs4

adamjarling:
It may be helpful to keep these commits in case we need to go back and inspect anything in particular.  A lot has moved around.  I'm fine either way though...


adamjarling:
The following items still need to get worked out:
- Track scrubber for Playlists.  
  - If time, abstract out the scrubber code exclusively to it's plugin, or to a helper class.
- Autoplay vs. Autoadvance.  I think I have this backwards.

app/controllers/playlist_items_controller.rb
```diff
@@ -62,6 +62,41 @@ def update
...
+    end
+  end
+
+  # GET /playlists/1/items/2/related_items
+  def related_items
+    @playlist_item = PlaylistItem.find(params['playlist_item_id'])
+    @related_clips = @playlist.related_clips(@playlist_item)
+    respond_to do |format|
+      format.html do
+        unless @related_clips.blank?
```
adamjarling: @bkeese Codeclimate is complaining about this duplicate declaration.  Let me know which one you want, or how to adjust and I can add it into my last commit.

adamjarling:
@cjcolvar Just a heads up I added another JavaScript code quality config file to the latest commit here: `.prettierrc`.  Prettier.  https://prettier.io   It's pretty popular these days and will help keep our JS formatting consistent, and also will help ahead of Codeclimate errors.   Some people use it instead of eslint these days, but think it's good to run both. 

I followed the same pattern as you set up ESLint.  A dev can choose to install Prettier globally via yarn or npm, then run CLI commands from within `/avalon` and it will pick up on the basic `.prettierrc` configuration.  

An example command would be:
`prettier --write "app/assets/javascripts/media_player_wrapper/*.es6"`

cjcolvar:
@adamjarling That looks cool and there appears to be good atom integration: https://github.com/prettier/prettier-atom

app/controllers/playlist_items_controller.rb
```diff
@@ -62,6 +62,41 @@ def update
     render json: { message: "Item was not updated: #{error.message}" }, status: 500 and return
   end
 
+  # GET /playlists/1/items/2/markers
```
cjcolvar: It looks like we need to add authorization checks to these three actions and corresponding tests to ensure they are working as expected (security and functionality).

Issue #2536, opened 2017-12-12
Playlist Items Page: represent playlist position in URL


adamjarling:
### Description
(from Pull Request notes)
Do we want to represent playlist item sections in the url (ie. http://localhost:3000/playlists/1?position=2)? It looks like copying the url for each item and pasting in the browser resolves to a route which displays the appropriate playlist item. (Low priority)

To be done post 6.3.

cjcolvar:
This should apply to links in the playlist item list on the right side as well as the related item links.

bkeese:
Clicking related items actually just finds and clicks the appropriate item on the right side, so handling it just on the right side should cover both.

joncameron:
No longer applicable.

PR #2545, opened 2017-12-13
Make related items clickable

### LINKED ISSUE #2529: Playlist Items Page - Related panel content list wire up
### Description
In the playlist items page the "Related" accordion panel body has a list of related (if they exist) playlist items.  We need to wire-up clicking on each item.

### Done Looks Like
- Clicking on a playlist item title in this list will act the same as clicking a playlist item in the right sidebar.

### Tasks
- [x] Add click listeners to the playlist items in the Related panel content.
- [x] Handle the click listeners in this file: `app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6`
- [x] Hook into existing functions to achieve expected behavior.

@davidschober @joncameron

bkeese:
Fixes #2529 

Clicking related item link locates and click associated playlist item. 
Also decouples playlist item data from the DOM structure a bit.


app/views/playlists/_related_items.html.erb
```diff
@@ -22,9 +22,14 @@ Unless required by applicable law or agreed to in writing, software distributed
         <tr>
           <% position = clip.playlist_position(@playlist.id) %>
           <td class='clip_title'>
-            <%= link_to playlist_path(@playlist, position: position) do %>
-              <%= clip.title %>
-            <% end %>
+            <a
```
cjcolvar: Will this link be able to be copy+pasted by a user?

bkeese: @adamjarling is working on that for a different story.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -213,11 +233,7 @@ Object.assign(MediaElementPlayer.prototype, {
     },
 
     isSamePlaylistItem(el) {
-      return (
-        $(el)
-          .parent('li')
-          .data('playlistItemId') === this.$nowPlayingLi.data('playlistItemId')
-      );
+      return $(el).data('playlistItemId') === this.$nowPlayingLi.data('playlistItemId');
```
cjcolvar: You use `dataset.playlistItemId` elsewhere in this PR.  Can you use it here?

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -79,6 +80,27 @@ Object.assign(MediaElementPlayer.prototype, {
...
+    /**
+     * Add click listeners forplaylist item links
+     * @function addRelatedItemListeners
+     * @return {void}
+     */
+    addRelatedItemListeners() {
+      // Handle click on entire RelatedItems area
+      // Filter only <a> element clicks; disregard all others
+      if (this.$relatedItems) {
+        this.$relatedItems.on('click', (e) => {
```
cjcolvar: Why do you need the click handler on the entire related items instead of attaching to the `<a>` tags themselves?

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -79,6 +80,27 @@ Object.assign(MediaElementPlayer.prototype, {
...
+          if (e.target.nodeName === 'A') {
+            // find correct playlist item in sidebar and click it.
+            let playlist_id  = e.target.dataset.playlistId
+            let playlist_item_id = e.target.dataset.playlistItemId
+            let related = this.$sidePlaylist.find('a[data-playlist-id='+playlist_id+'][data-playlist-item-id='+playlist_item_id+']')
```
cjcolvar: Shouldn't `a[data-playlist-item-id]` be sufficient to uniquely identify the matching `<a>` given that the playlist-id should be shared by everything in the playlist?

app/views/playlists/_related_items.html.erb
cjcolvar: Ok, I didn't realize that was in scope of that story, but that makes sense.  I'll add a note to it to make sure it is explicit.

app/views/playlists/_related_items.html.erb
```diff
@@ -22,9 +22,14 @@ Unless required by applicable law or agreed to in writing, software distributed
         <tr>
```
cjcolvar: Could we `render 'playlist_items/_related_items'` here instead of duplicating it?

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -79,6 +80,27 @@ Object.assign(MediaElementPlayer.prototype, {
...
+     * @return {void}
+     */
+    addRelatedItemListeners() {
+      // Handle click on entire RelatedItems area
+      // Filter only <a> element clicks; disregard all others
+      if (this.$relatedItems) {
+        this.$relatedItems.on('click', (e) => {
+          if (e.target.nodeName === 'A') {
+            // find correct playlist item in sidebar and click it.
+            let playlist_id  = e.target.dataset.playlistId
```
adamjarling: Not to get all nitpicky, but in general for our JavaScript can we follow camelCase convention for variables?

cjcolvar: Would eslint or prettier pick up on this?

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -79,6 +80,27 @@ Object.assign(MediaElementPlayer.prototype, {
...
   playlistItemsObj: {
+    /**
+     * Add click listeners forplaylist item links
+     * @function addRelatedItemListeners
+     * @return {void}
+     */
+    addRelatedItemListeners() {
+      // Handle click on entire RelatedItems area
+      // Filter only <a> element clicks; disregard all others
+      if (this.$relatedItems) {
```
adamjarling: If `$relatedItems` is a jQuery variable reference (preceding a variable title with `$` is a common, good pattern to follow), then you might need to check `if (this.$relatedItems.length > 0)` as an empty jQuery variable reference will return `true`.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
adamjarling: @cjcolvar Sometimes this approach is more performant in event delegation, attaching one click handler to a parent element, rather than a bunch of click handlers to all child items.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
bkeese: javascript vs jquery. different types of objects require different methods of access

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
bkeese: because the tags themselves aren't always visible and so can't be found to attach handlers.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
bkeese: It's a matter of efficiency. Reducing the search scope.

app/views/playlists/_related_items.html.erb
bkeese: I've got a fix in place to make these links copyable. The URL in the address bar will still need to be updated when transitioning to different playlist item ala #2536.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
adamjarling: @cjcolvar I don't think either will change variable titles.  No big deal for now, just made the comment as a 'in the future' type thing.   Let's get this sucker working for now.

PR #2550, opened 2017-12-13
Fix refresh_token.js console error by validating token variable

### LINKED ISSUE #2549: Refresh token console error
### Description
I keep seeing this console error, developing locally.  Anyone else?

refresh-token-error2

refresh-token-error

adamjarling:
Fixes #2549 

This is a pretty quick fix to get rid of the console error.  @cjcolvar Does the fact `token` doesn't exist mean there's another issue?


cjcolvar:
I was wondering that too.  I'm not sure.  If there isn't a streaming token being kept fresh then at worst the player will stop working at some point (depending on how it is configured) and require a page refresh.  This change should be safe though so I see no reason not to merge it.

PR #2555, opened 2017-12-13
Bugfix/2505

### LINKED ISSUE #2505: Adding new marker refreshes page losing context
## Description

when adding a new marker, the page refreshes and the user looses context such as  current playlist item and play time. 

### Tasks
- [ ] Verify this is an issue when there are no markers, and user creates their first marker.
- [ ] Dig into why a page refresh is necessary (refresh my memory)
- [ ] See if we can update the HTML markup for the accordion panels with the new REST end points.

bkeese:
Fixes #2505 

Handles playlist items that have no markers.

Hides marker sections if no markers exist on a playlist item.
Handles showing/hiding marker sections when transitioning between playlist items.
Shows marker sections when first marker is added to a playlist item.
De-duplicates the playlist/playlist_item marker partials.
Adds close button to alerts.
Also handles these situations for related_items and source_details sections.


app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6
```diff
@@ -301,6 +301,13 @@ class MEJSMarkersHelper {
         $('#markers').replaceWith(response);
         // Add event listeners to newly created row
         t.addMarkersTableListeners();
+        // hide marker sections if no markers (first row is header)
+        if ($('#markers').find('tr').length == 1) {
+          $('#markers_section').collapse('hide')
```
adamjarling: Is `.collapse()` jQuery?  Or a plugin maybe?  Guessing it works, just new to me.

bkeese: It's bootstrap. For the accordion stuff.

app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6
```diff
@@ -301,6 +301,13 @@ class MEJSMarkersHelper {
         $('#markers').replaceWith(response);
         // Add event listeners to newly created row
         t.addMarkersTableListeners();
+        // hide marker sections if no markers (first row is header)
+        if ($('#markers').find('tr').length == 1) {
```
adamjarling: In JS, should always (99% of the time) use `===` instead of `==`.  Just noticed this on CodeClimate.

PR #2556, opened 2017-12-14
Auto advance through sections on media view page

adamjarling:


app/helpers/media_objects_helper.rb
```diff
@@ -131,10 +131,11 @@ def hide_sections? sections
      def structure_html section, index, show_progress
        current = is_current_section? section
        progress_div = show_progress ? '<div class="status-detail alert" style="display: none"></div>' : ''
-       playlist_btn = current_ability.can?(:create, Playlist) ? '<button type="button" title="Add section to playlist" aria-label="Add section to playlist" class="structure_add_to_playlist outline_on btn btn-primary" data-scope="master_file" data-masterfile-id="#{section.id}"></button>' : ''
+
+       playlist_btn = current_ability.can?(:create, Playlist) ? "<button type=\"button\" title=\"Add section to playlist\" aria-label=\"Add section to playlist\" class=\"structure_add_to_playlist outline_on btn btn-primary\" data-scope=\"master_file\" data-masterfile-id=\"#{section.id}\"></button>" : ''
 
        headeropen = <<EOF
-    <div class="panel-heading" role="tab" id="heading#{index}">
+    <div class="panel-heading" role="tab" id="heading#{index}" data-media-object-id="#{section.isPartOf}" data-section-id="#{section.id}">
```
cjcolvar: I think`isPartOf` works here, but we usually use `media_object_id` instead.  Do you mind switching it?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -284,11 +328,20 @@ class MEJSPlayer {
...
+  hasSections() {
+    return Object.keys(this.segmentsMap).length > 0;
+  }
+
   /**
    * Update section links to reflect active section playing
    * @function highlightSectionLink
    * @param  {string} segmentId - HTML node of section link clicked on <a>
-   * @return {void}
+   * @return {void} Boolean value whether the UI has navigatable "Sections" content
```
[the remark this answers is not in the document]
adamjarling: Good catch!  You're right.

app/helpers/media_objects_helper.rb
adamjarling: Yea, totally.  I didn't know that was available.  I just did a `<% byebug %> to see what I could grab.  One of these days I need a Rails tutorial on how to view full object state or something.  Thanks for looking at this...

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -140,6 +140,47 @@ class MEJSPlayer {
...
+    // No sections content on this page, go no further
+    if (!t.hasSections()) {
+      return;
+    }
+
+    const $sections = $('#accordion').find('.panel-heading[data-section-id]');
+    const sectionsIdArray = $sections.map((index, item) =>
+      $(item).data('sectionId')
+    );
+    const currentIdIndex = [...sectionsIdArray].indexOf(t.currentStreamInfo.id);
```
bkeese: I'm curious about this. Why not just do sectionsIdArray.indexOf?

adamjarling: Ah, good catch it's not necessary.  Originally I was calling `indexOf` on an array of javascript elements, or jQuery objects or something, so was using the es6 spread operator `[...myArray]` to coerce the returned "JS node-y" array to a standard array you can call `Array` methods on.   I just didn't take the spread operator out.  I'll update that in a future commit.

adamjarling:
@cjcolvar Looks like Codeclimate is not liking this file:  https://codeclimate.com/github/avalonmediasystem/avalon/app/helpers/media_objects_helper.rb/source

I tried the remove the empty line in the line of code I adjusted, but that didn't seem to help.  Merge anyway?   Or any ideas?

PR #2572, opened 2017-12-14
Fix buggy player loading over itself twice in Playlist Items page pla…

adamjarling:
…yer instances

This is something we'll need to be aware of, and something @cjcolvar I believe pointed out earlier, that Mediaelement player events can fire with a slight time delay.  

For example, we may `pause()` the player, but the player still may emit a few more `timeupdate` events after it's been paused.   Just something to keep in mind if you're scratching your head on odd behavior.


bkeese:
The same red flag went up for me. This is a very tricky thing to get right, and I'm not confident I can tell if this is.

adamjarling:
Yea, maybe there's something else to grab onto.  I'll try another angle.

Issue #2561, opened 2017-12-14
Don't Show Actions Column When User Cannot Take Action on Markers


joncameron:
### Description
Users who are viewing a playlist are shown the column for actions to be taken in the markers accordion, but no button shows up as they do not have permissions to take action on the set of markers for the playlist. 

### Done Looks Like
- The action column does not appear to users who do not have permissions to edit the markers on the playlist being viewed.

cjcolvar:
Since #2561 got merged marker actions don't show for any user including the playlist owner.

yingfeng-iu:
was working locally when i tested, will check again

joncameron:
I tested this as well—looks good!

PR #2581, opened 2017-12-15
Disable button for Add Marker To Playlist Item when not authorized

### LINKED ISSUE #2570: Add Marker button showing when not authorized to add markers
### Description
If you go to a playlist that you don't own, the player still has the add marker button even though it doesn't do anything when you click it.

### Done Looks Like
- Add marker button doesn't show when you're not authorized to add markers to a playlist item.

bkeese:
Fixes #2570 

Hides the 'Add Marker to Playlist Item' button when a user doesn't have permission to edit the item.

### DIFF
--- a/app/views/playlist_items/_markers.html.erb
+++ b/app/views/playlist_items/_markers.html.erb
@@ -13,7 +13,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
-<div id="markers" class="container">
+<div id="markers" class="container" data-can-edit-playlist-item="<%= can? :edit, @playlist_item %>">
   <%# Alert error container %>
   <div id="marker_item_edit_alert" class="mejs-form-alert alert alert-danger" style="display: none;">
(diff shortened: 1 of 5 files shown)


app/views/playlist_items/_markers.html.erb
```diff
@@ -13,7 +13,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
-<div id="markers" class="container">
+<div id="markers" class="container" data-can-edit-playlist-item="<%= can? :edit, @playlist_item %>">
```
cjcolvar: I'm wondering if we should do this more universally and get rid of all marker abilities since they are tied to the playlist item.  This would probably make #2561 simpler.

bkeese: Makes sense. Maybe we could put this attribute on each item in the right-hand panel?

cjcolvar: I don't think we need that yet since this seems to work for now.

bkeese: Oh. Then I guess I don't know what you meant.

PR #2585, opened 2017-12-15
issue#2561: fix permission checking so Markers Actions column only shows

yingfeng-iu:


app/views/playlist_items/_markers.html.erb
```diff
@@ -58,7 +61,7 @@ Unless required by applicable law or agreed to in writing, software distributed
                 <% end %>
               <% end %>
               <%# Delete button %>
-              <% if can? :delete, marker %>
+              <% if showActions %>
```
cjcolvar: Now that we're relying on this boolean flag instead of the two `can?` checks, could the two `if` blocks be collapsed into one?

yingfeng-iu: sure, i was just waiting for someone to say that or anything else :-)

Issue #2578, opened 2017-12-15
Markers not working in Safari


davidschober:
_From @davidschober on December 15, 2017 16:17_

To Recreate:

- Go to a playlist <https://spruce.dlib.indiana.edu/playlists/100> using safari
- click on a marker
- see that the playhead is not moved forward to proper position

_Copied from original issue: avalonmediasystem/avalon-planning#221_

cjcolvar:
Might be fixed by #2586

cjcolvar:
I wasn't able to reproduce with the latest on spruce.

joncameron:
Couldn't reproduce this, but I filed another Safari marker bug: #2620

PR #2586, opened 2017-12-15
Dynamically load playlist info panels instead of render them directly…

### LINKED ISSUE #2505: Adding new marker refreshes page losing context

cjcolvar:
… into the page
Fixes #2505.

### DIFF
--- a/app/views/playlist_items/_current_item.html.erb
+++ b/app/views/playlist_items/_current_item.html.erb
@@ -17,7 +17,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% @current_clip = AvalonClip.find(@current_playlist_item.clip_id) %>
 <% @current_masterfile = MasterFile.find(@current_playlist_item.clip.source.split('/').last) %>
 <% @current_mediaobject = MediaObject.find(@current_masterfile.media_object_id) %>
-<section id="metadata">
+<section id="source_details">
 <div id="metadata_header">
   <h3>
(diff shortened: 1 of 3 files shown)


app/views/playlist_items/_current_item.html.erb
```diff
@@ -17,7 +17,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% @current_clip = AvalonClip.find(@current_playlist_item.clip_id) %>
 <% @current_masterfile = MasterFile.find(@current_playlist_item.clip.source.split('/').last) %>
 <% @current_mediaobject = MediaObject.find(@current_masterfile.media_object_id) %>
-<section id="metadata">
+<section id="source_details">
```
bkeese: This partial is now only used by mejs2. I wonder if changing this id will break mejs2 stuff.

cjcolvar: It looks like it is still working from a quick test locally.

cjcolvar:
Also I can't reproduce #2578 with this branch.

PR #2592, opened 2017-12-15
Update to Autoplay functionality on playlist items page

adamjarling:
Updated with custom threshold variable to handle timeupdate timing event calculations.

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
@@ -60,6 +60,7 @@ Object.assign(MediaElementPlayer.prototype, {
       playlistItemsObj.handleUserSeeking.bind(this)
     );
 
+    // Set default values for helper boolean flags
     playlistItemsObj.refreshInProgress = false;
 
     // Handle continuous MEJS time update event
@@ -150,6 +151,7 @@ Object.assign(MediaElementPlayer.prototype, {
 
     /**
      * Analyze the new playlist item and determine how to proceed...
+     * @function analyzeNewItemSource
      * @param  {HTMLElement} el <a> anchor element of the new playlist item being processed
      * @return {void}
      */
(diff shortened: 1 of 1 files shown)


adamjarling:
So I used `Math.round()` because the final `currentTime` value on `timeupdate` event for a 14 second media file was 13.98 or something.  It never hit the whole integer, so some code wasn't firing.

`Math.round()` will for example make `15.49 => 15`, and `15.51 => 16`.  Is that too loose in calculations do you think?   `Math.ceil()` would make `15.01 => 16`...

bkeese:
I remember struggling with this. Here's my problem scenario: The duration on a file is 15.55. The final time update event is 15.49. Round down and you won't hit that code either. But the other scenario: file duration is 15.95, and round up from 15.01, then you miss the last second of the file. Where do you draw the line? Not sure.

adamjarling:
Could so something with subtracting the currentTime from endTime, and if it's greater than a certain threshold value, like > .25 seconds or something, then return `true`?

adamjarling:
Ok, I'll see how tricky that is to make it actually work:)

adamjarling:
Updated this commit removing `Math.round()`, and am using a custom "threshold" variable (which we can adjust if need be).  Seems to work pretty well set at 0.25 seconds.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -270,58 +287,101 @@ Object.assign(MediaElementPlayer.prototype, {
...
+     * Note the 'threshold' helper variable, and see description below for why it's needed.
+     * @function isCurrentTimeInRange
+     * @param  {number}  currentTime Current time of Mediaelement player
+     * @return {Boolean}
+     */
+    isCurrentTimeInRange(currentTime) {
+      const currentTimeAdjusted = currentTime + this.threshold;
+      const startEndTimes = this.startEndTimes;
+      return (
+        currentTimeAdjusted >= startEndTimes.start &&
```
bkeese: This could probably be ```currentTime >= startEndTimes.start```. I don't think we need to adjust for the start times.

adamjarling: Sure.  I'll adjust both.

PR #2597, opened 2017-12-18
Alternative way to track and handle additional timeupdate events

adamjarling:


app/assets/javascripts/media_player_wrapper/mejs4_plugin_add_marker_to_playlist.es6
```diff
@@ -91,14 +91,18 @@ Object.assign(MediaElementPlayer.prototype, {
...
 
-    t.addMarkerObj.addButton.removeEventListener(
-      'click',
-      t.addMarkerObj.bindHandleAdd
-    );
-    t.addMarkerObj.cancelButton.removeEventListener(
-      'click',
-      t.addMarkerObj.bindHandleCancel
-    );
+    if (t.addMarkerObj.addButton) {
```
cjcolvar: Is there a safe chaining call in es6?  I remember seeing one in ruby (`&.`): http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

adamjarling: Not directly in JavaScript (yet at least).  You could create some type of proxy function to do the check, but this introduces additional code as well.

PR #2600, opened 2017-12-18
Handle marker title clicks past playlist item end time

adamjarling:


adamjarling:
@cjcolvar @bkeese We might need to update marker click handling after Brian's new markers commit.  To make the behavior the same when clicking on a marker past the playlist item end point...which this commit handles.

I'm rebasing this locally on develop now to see how it behaves...

adamjarling:
It works great, as expected.  Should I re-push my branch up after re-basing?

cjcolvar:
If it works fine then I don't think you need to.

adamjarling:
Yea, re: the custom event, I kind of had to do that since adding a click listener to the marker link from within the playlist items plugin file, wasn't working since that content is now being loaded in dynamically.  

What might benefit this application at some point is a state management system to handle all these interactions with the player.  Like RXJS, or Flux, or something.  We're going to implement RXJS or Redux come January, see if it makes life easier...

PR #2601, opened 2017-12-18
Hide instead of delete dismissable Bootstrap alerts

### LINKED ISSUE #2042: long-lasting flash message
## Description
When you add a marker to a playlist item, you get the success message:

image

The problem is that the message does not disappear and does not have the usual "x" that gives the user the ability to dismiss it. 

### Done looks like
User is given the option to get rid of that message.

### Tasks
- [ ]  Add the 'X' markup to the flash message

phuongdh:
Fixes #2042


cjcolvar:
Are those code climate issues syntax issues since it is coffeescript?

PR #2624, opened 2017-12-19
Fix playlist markers in Safari

adamjarling:
Note the only thing I really changed here creating a `"canplay"` event handling function in the playlist items plugin.  When that event fires, then we have `duration`, and everything works as expected in Safari.

The other diffs here are mostly alphabetizing functions in the plugin/class.  It's really helpful to keep these organized, and saves time (at least for me).

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -65,7 +65,10 @@ class MEJSPlayer {
   addSectionsClickListener() {
     const $accordionEl = $('#accordion.media-show-page');
     if ($accordionEl.length > 0) {
-      $accordionEl[0].addEventListener('click', this.handleSectionClick.bind(this));
+      $accordionEl[0].addEventListener(
+        'click',
+        this.handleSectionClick.bind(this)
+      );
     }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -93,13 +96,8 @@ class MEJSPlayer {
    * @return {void}
    */
   emitSuccessEvent() {
```
bkeese: Is this function being used?

adamjarling: Not yet, but I put it in a while ago in case we need to hook into it.  I tried hooking into it for this scenario, but it didn't do the trick.  I can remove it if you prefer.

cjcolvar:
I think this looks good, but I'm kind of afraid that the `canplay` event isn't always fired but I'm fine with merging and checking on the other side.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
bkeese: No, leave it. I like having it.

PR #2630, opened 2017-12-19
issue#2605: fix permission issue for creating thumbnail

yingfeng-iu:
tested locally


app/views/media_objects/_mejs4_player_js.html.erb
```diff
@@ -7,9 +7,16 @@
 <script>
 var mejs4ConfigFeatures = ['playpause', 'current', 'progress', 'duration', 'trackScrubber', 'volume', 'tracks', 'quality', 'createThumbnail', 'addToPlaylist', 'fullscreen'];
 
+// do not show createThumbnail button if current user can't edit mediaObject
+if (<%= cannot? :edit, MediaObject %>) {
```
cjcolvar: The master file or media object instance will need to be checked instead of the whole class.  My instincts would be to test the current master file:
`cannot? :edit, @currentStream`
but maybe media object would be just as good:
`cannot? :edit, @media_object`

PR #2636, opened 2017-12-20
Update to Mediaelement quality.js quality selector plugin - bugfix

adamjarling:
This PR:

1. Updates the quality selector default plugin
2. Fixes a bug in the event delegation.  

It's a small one, but I'm going to submit a PR upstream for this fix to the plugin repo.


cjcolvar:
It's pretty much impossible to see what changed in the github diff even ignoring whitespace.  I'm going to trust you on this one, but maybe different PRs or commits for each step (update from upstream, change, prettier) would be better in the future.

adamjarling:
@cjcolvar Got it.  I ran `prettier` on the plugin file just to make heads and tails of it, and make it legible.  

Without re-formatting the code, the fix was to handle the `blur` event replacing the`focusout` event, in only one place.

PR #2637, opened 2017-12-20
Add HD toggle button to embed audio player

### LINKED ISSUE #2604: Audio embed quality switching popover hidden
In the MEJS2 player a plugin was used that made an "HD" button in the control bar which would toggle between medium and high qualities.  In the MEJS4 player, the stock quality switcher with popover with list of available qualities is used.  The problem is that audio embed players have a fixed height that doesn't allow for the popover to show.  We either need to bring back that plugin or figure out a different way that works within the vertical space.

phuongdh:
Fixes #2604


app/views/master_files/_mejs4_player_js.html.erb
```diff
@@ -6,10 +6,11 @@
 <% end %>
 
 <script>
+console.log()
```
cjcolvar: You probably meant to remove this.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_hd_toggle.es6
```diff
@@ -0,0 +1,190 @@
...
+    // Add control button to player
+    t.addControlElement(player.hdtoggleButton, 'hdToggle');
+
+    if (player.options.hdToggleOn && player.qualities[0] != null) {
+      player.hdtoggleButton.className += " mejs__hdtoggle-on";
+      player.switchStream(player.qualities[0]);
+    } else if (player.qualities[1] != null) {
+      player.switchStream(player.qualities[1]);
+    } else {
+      alert("Did not find " + hdToggleBetween[0] + " and " + hdToggleBetween[1] + " streams");
```
cjcolvar: Is this how we want to handle this exception case?  I know we have alerts in at least one other place but is there a better way to handle this like maybe through the flash message area.

adamjarling: I agree.  If there's a pattern for handling messaging to the user, let's try to go that route.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_hd_toggle.es6
```diff
@@ -0,0 +1,190 @@
...
+  buildhdToggle(player, controls, layers, media) {
+    const t = this;
+    const hdToggleText = mejs.Utils.isString(t.options.hdToggleText) ? t.options.hdToggleText : mejs.i18n.t('mejs.hd-toggle')
+    player.qualities = [];
+
+    player.sources = $(player.domNode).find("source");
+    for (var i = 0; i < player.sources.length; i++) {
+      var src = player.sources[i];
+      for (var j = 0; j < player.options.hdToggleBetween.length; j++) {
+        if (src.getAttribute("data-quality") == player.options.hdToggleBetween[j]) {
```
adamjarling: Yea, in JS let's use === to avoid weird errors slipping through.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_hd_toggle.es6
```diff
@@ -0,0 +1,190 @@
...
+  cleanhdToggle(player, controls, layers, media) {
+    if (player && player.hdtoggleButton) {
+        player.hdtoggleButton.remove();
+      }
+  },
+
+  // Other optional public methods (all documented according to JSDoc specifications)
+
+  toggleQuality: function() {
+    var btn = $(this.hdtoggleButton);
```
adamjarling: I mentioned this to Brian previously... don't change this now, but in the future we should prefix jQuery variables with a "$" for clarity.  For example `var btn = $(this.hdtoggleButton)` becomes `let $btn = $(this.hdtoggleButton`).

app/assets/javascripts/media_player_wrapper/mejs4_plugin_hd_toggle.es6
phuongdh: There's no room to do that in the embedded audio player.

phuongdh: This is how it was before, but I'm ok with removing this alert altogether since it only happens when: 1. audio tags' `data-quality` attributes are misconfigured. 2. there're no audio tags.

adamjarling: That's right, I keep forgetting about the embed scenario.  Would the alert show up in non-embed as well, or can we check for that and handle each environment?

phuongdh: We're only using this for embed audio.

PR #2644, opened 2017-12-20
use name over collection id

### LINKED ISSUE #2641: Batch Ingest Success Email: Use Collection Name In Subject Line
### Description
Currently, the batch ingest success email uses the collection ID in the subject line instead of the name of the collection. This isn't very useful to collection managers who will want to quickly identify the collection that the message is referring to, and don't think in terms of Fedora IDs.

### Example
all_success
Instead of _bn999679s_, the collection name should be used.

### Done Looks Like
- Subject line for batch ingest success emails uses the name of the collection instead of the collection NOID

carrickr:
Fixes #2641


carrickr:
done

PR #2645, opened 2017-12-20
Use custom event handler which skips player resize if embedded

cjcolvar:
I tested locally for non-iOS embedding and it looks fine and the normal player looks fine to me.  I'm not able to test on iOS from my local dev since I don't have https setup.

This PR replicates the change made by @mbklein here: https://github.com/avalonmediasystem/mediaelement/commit/882db358b11aa057b43d10cb52367c3ae1daa3a9

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -274,6 +274,10 @@ class MEJSPlayer {
       this.player = this.mediaElement;
     }
 
+    // Replace resize event handler with custom handler
+    this.player.globalUnbind('resize', this.player.globalResizeCallback);
+    this.player.globalBind('resize', this.customGlobalResizeCallback);
+
     this.emitSuccessEvent();
 
     // Handle 'canplay' events fired by player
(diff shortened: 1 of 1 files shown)


bkeese:
The marker stuff relies on the resize event. This might break it.
https://github.com/avalonmediasystem/avalon/blob/develop/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6#L473

bkeese:
I guess this change is unbinding a different handler, so this should be ok. Right?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -488,6 +492,10 @@ class MEJSPlayer {
     if (!this.player.paused) {
       this.player.pause();
     }
+
+    // Unbind custom resize event handler
+    this.player.globalUnbind('resize', this.customGlobalResizeCallback)
```
bkeese: I don't think this is necessary, since remove() will remove all bound event handlers too.

bkeese: Nevermind. I see that globalbind actually binds to the document, not the player.

cjcolvar: This was to replicate what was going on here (https://github.com/mediaelement/mediaelement/blob/26ec30053d3fd59c973fec0164989bda984bee21/src/js/player.js#L2023) in mediaelement's `remove()` since we're binding something new to it now.

PR #2646, opened 2017-12-20
include links to the successful media objects

### LINKED ISSUE #2640: Batch Ingest Emails: Hyperlinks for Media Object IDs
### Description
In batch ingest emails where an object has been successfully processed, the object ID should be  a hyperlink to that object in the repository. In the image below, each successful item is listed by ID, but it would be nice if a user could click one to go directly to the media object page.

all_success

### Done Looks Like
- IDs in the _Media Object ID_ row in the batch ingest success email are hyperlinked to the media object page. For example, n009w2411 would be a hyperlink to https://spruce.dlib.indiana.edu/media_objects/n009w2411

carrickr:
Fixes #2640 

Yes I know enchancement, extra c

@cjcolvar I feel link of bad about this one in terms of the ugly code in the `.erb`.  I was wondering moving it into the mailer job, but `@be_completed` is an array of `BatchEntries`, so I can't just (easily) inject URLs into those.  I did consider making BatchEntries track a URL like how Switchyard does and I'd be okay with that, except that is messing with the model really close to code freeze.  Your thoughts are welcome.

### DIFF
--- a/app/views/batch_registries_mailer/batch_registration_finished_mailer.html.erb
+++ b/app/views/batch_registries_mailer/batch_registration_finished_mailer.html.erb
@@ -47,8 +47,11 @@ Unless required by applicable law or agreed to in writing, software distributed
     </tr>
     <% @completed_items.each do |be_completed| %>
       <tr>
+        <% media_object = MediaObject.find(be_completed.media_object_pid) %>
+        <% link_url = media_object.permalink %>
+        <% link_url = media_object_url(media_object) if link_url.empty? %>
         <td> <%= be_completed.position + 2 %> </td>
-        <td> <%= be_completed.media_object_pid %> </td>
+        <td> <% content_tag(:a, href: link_url) do %><span><%=be_completed.media_object_pid%></span><%end%></td>
       </tr>
     <% end %>
   </table>


app/views/batch_registries_mailer/batch_registration_finished_mailer.html.erb
```diff
@@ -47,8 +47,11 @@ Unless required by applicable law or agreed to in writing, software distributed
     </tr>
     <% @completed_items.each do |be_completed| %>
       <tr>
+        <% media_object = MediaObject.find(be_completed.media_object_pid) %>
+        <% link_url = media_object.permalink %>
+        <% link_url = Settings.domain.protocol + '://' +  Settings.domain.host + '/media_objects/' + media_object.id if link_url.empty? %>
         <td> <%= be_completed.position + 2 %> </td>
-        <td> <%= be_completed.media_object_pid %> </td>
+        <td> <% content_tag(:a, href: link_url) do %><span>be_completed.media_object_pid</span><%end%></td>
```
bkeese: You're missing the ```<%=  %>``` around the ```be_completed.media_object_pid```

app/views/batch_registries_mailer/batch_registration_finished_mailer.html.erb
```diff
@@ -47,8 +47,11 @@ Unless required by applicable law or agreed to in writing, software distributed
     </tr>
     <% @completed_items.each do |be_completed| %>
       <tr>
+        <% media_object = MediaObject.find(be_completed.media_object_pid) %>
+        <% link_url = media_object.permalink %>
+        <% link_url = Settings.domain.protocol + '://' +  Settings.domain.host + '/media_objects/' + media_object.id if link_url.empty? %>
```
cjcolvar: Is it possible to use the url helper here?
`media_object_url(media_object)`

PR #2648, opened 2017-12-20
Update share links when switching sections

### LINKED ISSUE #2639: Embed Code Does Not Change When Section Is Switched
### Description
For a media object, regardless of the section currently selected, the embed code shows the URL for the first masterfile in the media object and does not change.

### Environment
Spruce, Firefox 57, macOS 10.11.3

### Example
dec-20-2017 09-25-10

### Done Looks Like
- When a section is selected (either by user action or playthrough), the embed code in the Share div changes to reflect the embed URL for that masterfile

bkeese:
Fixes #2639 

Updates the share and embed links when switching sections on media_object show page.


bkeese:
It will only act on elements it finds.

PR #2650, opened 2017-12-20
Position audio player at bottom of iframe when embedding

### LINKED ISSUE #2510: ensure embed size of me4js player is the same as me2

phuongdh:
Fixes #2510 
Where time float getting clipped off by iframe.

### DIFF
--- a/app/assets/stylesheets/embed.scss
+++ b/app/assets/stylesheets/embed.scss
@@ -35,3 +35,8 @@ body {
   height: 100%;
   max-width: none;
 }
+
+.mejs__audio {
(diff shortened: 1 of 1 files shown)


adamjarling:
Yea, as long as it works and it's parent element has `position: relative`, seems like it makes sense.

bkeese:
Seems ok. Won't know until I see it in action.

PR #2655, opened 2017-12-20
Fix updateSharedLinks method error - check that var exists

### LINKED ISSUE #2654: updateShareLinks() method throwing a console error Chrome/Safari
### Description
updateShareLinks() method throwing a console error Chrome/Safari.  Looks like this line: 
 `var sectionShareLink = currentLink.data()['shareLink'];`

adamjarling:
Fixes #2654

This suppresses the error.  @bkeese want to make sure it still does the stuff its supposed to?

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -538,29 +538,20 @@ class MEJSPlayer {
   }
 
   /**
-   * Update share links when switching sections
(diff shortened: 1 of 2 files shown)


bkeese:
This feels better to me. Adding everything needed into stream_info.

cjcolvar:
The only small thing I notice is that the page loads with a section share link using a different route than the ones grabbed via ajax.  They are effectively the same because they resolve to the same location but it might be nice to be consistent.

PR #2656, opened 2017-12-21
Replace MEJS4 resize handler to stop embed player from infinitely exp…

phuongdh:
…anding on mobile Safari

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -39,5 +39,6 @@
 //= stub media_player_wrapper/mejs4_link_back
 //= stub media_player_wrapper/mejs4_plugin_playlist_items
 //= stub media_player_wrapper/mejs4_plugin_hd_toggle
+//= stub media_player_wrapper/mejs4_plugin_no_resize
 
 //= require_tree .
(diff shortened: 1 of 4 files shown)


bkeese:
Interesting approach!

cjcolvar:
Neat.  I like how it is scoped to the embed player!
@phuongdh do you think it is guaranteed that this plugin will be built (and therefore override the event handler) before the resize globalBind happens?

phuongdh:
Hmm yeah I thought it wouldn't matter but on second thought I'll deploy this branch to Spruce and test.

cjcolvar:
I think it is safe because the plugins are built synchronously and that happens before the resize globalBind, but now I'm wondering if the MEJS code will override the plugin since it runs later.
buildfeatures: https://github.com/mediaelement/mediaelement/blob/master/src/js/player.js#L683
resize handling: https://github.com/mediaelement/mediaelement/blob/master/src/js/player.js#L923-L935

phuongdh:
That's what I'm fiddling with now. Changing quality makes it grow again!

phuongdh:
Giving up for now. Feel free to mess with the branch and redeploy to Spruce

PR #2657, opened 2017-12-21
Emit custom event like table marker click to handle out of clip markers the same way

### LINKED ISSUE #2651: Refinement to Playlist MEJS4 Behavior
### Description
- On load, a user cannot click past the end of the playlist item range to move the playhead to that point, but:
  - if a users clicks a marker link in the markers accordion, the playhead moves to that point and the user can click freely at any point after the playlist item range
  - if a user double (sometimes triple) clicks the rail past the end of the playlist item range, the playhead moves to that point and the user can click to move the playhead freely past that point
Users should be able to click around on the rail at a point past the playlist item without having the playhead snap back to the beginning of the playlist item, in the same way that a user can click to move the playhead in the time range before the beginning of the playlist item. 

### Environment
Spruce, Firefox 57, macOS 10.11.3
Spruce, Safari 11, macOS 10.11.3

### Done Looks Like
- Users can single-click a point past the end of a playlist item to move the playhead to that point, without the playhead snapping back to the beginning of the playlist item

### Examples
dec-20-2017 15-58-38

dec-20-2017 16-00-16

cjcolvar:
Fixes #2651.

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6
@@ -11,7 +11,6 @@ class MEJSMarkersHelper {
   /**
    * Add event listeners for elements in Markers table rows
    * @function addMarkersTableListeners
-   * @param player Instance of player
    * @return {void}
    */
   addMarkersTableListeners() {
@@ -27,14 +26,14 @@ class MEJSMarkersHelper {
       const offset = $(e.target)
         .parents('tr')
         .data('offset');
-      const event = new CustomEvent('markerTitleClicked', {
+      const event = new CustomEvent('markerClicked', {
         detail: {
           offset: +offset
         }
       });
 
-      player.setCurrentTime(+offset);
       document.dispatchEvent(event);
+      player.setCurrentTime(+offset);
     });
 
     // Edit button click
@@ -292,14 +291,14 @@ class MEJSMarkersHelper {
         '"></span>'
     )
       .bind('click', e => {
-        const event = new CustomEvent('markerTitleClicked', {
+        const event = new CustomEvent('markerClicked', {
           detail: {
             offset: +offset
           }
         });
 
-        mejs4AvalonPlayer.player.setCurrentTime(offset);
         document.dispatchEvent(event);
+        mejs4AvalonPlayer.player.setCurrentTime(offset);
       })
       .bind('mouseenter', e => {
         $(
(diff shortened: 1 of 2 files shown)


bkeese:
Can you give more details about what this PR is about?

cjcolvar:
Now, if you click a marker (icon rendered below the scrubber) that is past the end of the clip it will skip to the next playlist item if autoplay is on and will reset to the beginning of the clip if autoplay is off.  This PR makes the click behave the same as if you clicked on the marker link in the markers info panel.

cjcolvar:
This is basically a copy of how the marker links in the info panel are being handled: https://github.com/avalonmediasystem/avalon/blob/develop/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6#L25-L38
This event is being handled here: https://github.com/avalonmediasystem/avalon/blob/develop/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6#L119-L125

bkeese:
👍 I just recommend changing the name of the custom event to describe both marker title clicks and scrubber marker clicks, maybe 'markerClicked'.

cjcolvar:
Looking at this again I'm wondering if we need to emit this custom event before we `setCurrentTime` to avoid a race condition where the code that handles the set time might trigger autoplay before the custom event gets handled.
@adamjarling does it make sense to flip these two lines?
https://github.com/avalonmediasystem/avalon/blob/develop/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6#L36-L37

adamjarling:
@bkeese Can you update the comment here to make it more generic now that we're handling this event from multiple firing spots?
https://github.com/avalonmediasystem/avalon/blob/30820c527329d5038ecf2ad256496d93a5cf2b24/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6#L117

adamjarling:
@cjcolvar Yea, agreed that probably makes sense to handle the event before setting the time.

cjcolvar:
I can take care of the updates since it is my PR.

adamjarling:
And can you remove this `@param`?  Looks like it's not used anymore.
https://github.com/avalonmediasystem/avalon/blob/30820c527329d5038ecf2ad256496d93a5cf2b24/app/assets/javascripts/media_player_wrapper/mejs4_helper_markers.es6#L14

cjcolvar:
@adamjarling @bkeese Updated.  Thanks for all of the feedback!

PR #2658, opened 2017-12-21
Directly patch MEJS: avoid player resize if embedded

### LINKED ISSUE #2632: Embedded Player Zooms Infinitely on iPad
### Description
Viewing an embedded player on mobile Safari results in a slow, constant expansion of the media player's size. Must be seen to be believed.

### Environment
iPad 4 running latest mobile Safari on Browserstack

### Example
bbb2

cjcolvar:
Fixes #2632 

Tested on spruce and physical ipad 4.


bkeese:
Instead of changing the callback, can we just remove the listener?

cjcolvar:
My thoughts when making this PR were "we're out of time and this is known to work so why don't we put this in and make a ticket for the follow up work of making it better".

PR #2662, opened 2017-12-21
Use consistent route for section links

bkeese:
Use consistent route for section links in media_object page and stream_details ajax endpoint.


cjcolvar:
@bkeese Looks like the two failing tests need to be moved from the model to the controller.

cjcolvar:
This is ready to go as the failing tests are the 25 that have been failing since yesterday.

bkeese:
:+1:

PR #2665, opened 2017-12-21
Fix for an extra var causing console errors in iOS

adamjarling:


bkeese:
Crazy. Will you push this upstream too?

adamjarling:
This was my fault I think...left it in when I was debugging this plugin.

PR #2670, opened 2017-12-21
Fix for iOS iPad; removes quality plugin loading; loads playlist item…

adamjarling:
… panel content on page load.

The issue causing this ticket is that in certain iOS environments, MEJS events like 'canplay', 'loadedmetadata', cannot be counted on and are erratic between iPhone/iPad then versions with each.   Some fire, some don't.

So, this `poller()` pattern is something else we could maybe think about as a more reliable way to handle application timing, in certain scenarios?   Just something to think about.

Note I also included code here which filters out the quality selector plugin for mobile, which was causing errors for some reason in iOS.  Even after my other PR fix for the quality selector plugin...

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_helper_utility.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_helper_utility.es6
@@ -114,6 +114,15 @@ class MEJSUtility {
   }
 
   /**
+   * Helper function - is current environment mobile?
+   * @return {Boolean}
+   */
+  isMobile() {
+    const features = mejs.Features;
+    return features.isAndroid || features.isiOS;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


cjcolvar:
@joncameron and I both think that we want to keep the quality selector in mobile views.  My testing on a physical ipad 4 had the quality switching working well enough.  I don't believe it was working at all on the android device I tested on.  Maybe something has changed since I last tested?  Either way, could you please revert that part of this PR leaving the `handleCanPlayMobile`?

adamjarling:
Sure thing.  Updating the PR now.

cjcolvar:
Thanks!  So I'm thinking we merge this and try it out and if it causes any regressions we revert and mark as 6.x.  Sound okay to you?

PR #2684, opened 2017-12-22
Change version to 6.3. Pin bootstrap-toggle-rails to tag

phuongdh:


cjcolvar:
I think we need to revert to pulling from the github tag based upon @mbklein's comments here: https://github.com/avalonmediasystem/avalon/pull/2403

PR #2687, opened 2017-12-22
Fix dropbox settings for docker-compose

phuongdh:


bkeese:
Could you add these changes to the wiki upgrade doc?

phuongdh:
@bkeese do you have a link?

PR #2691, opened 2017-12-22
Fix related items not showing #2682

### LINKED ISSUE #2682: Issues with Related Items section in playlists
### Descriptive summary
Server: https://spruce.dlib.indiana.edu
Testing: Regression Testing
Browser: Chrome 62.0.3202.94, EDGE 40
Operating System: Windows10
Issue: Issues with Related Items section in playlists

1. If there are other items already present in the playlist that are unrelated, now if new items that are related the playlist is not creating the "Related Items" section for these items.
image

2. If I add the related items to an empty playlist, then the "Related Items" section is created properly in the playlist, but then if I add any unrelated items to the playlist the Related Items section doesn't load any content for the previous items.
image

Note: In case-2, if I refresh the page and only play the related items, then the Related section is showing content properly, but once I play the unrelated item and go back to the related items the Related section is no longer showing content.

phuongdh:
Fixes #2682

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
@@ -514,7 +514,7 @@ Object.assign(MediaElementPlayer.prototype, {
         .ajaxPlaylistItemsHTML(playlistId, playlistItemId, panel)
         .then(response => {
           // Insert the fresh HTML content
-          $('#' + panel).replaceWith(response);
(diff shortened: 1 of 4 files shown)


bkeese:
It'd be kinda cool if you made the same change for the markers rebuilding.

cjcolvar:
@bkeese I agree it would be nice to be consistent but better not to rock the boat at this point.

phuongdh:
@bkeese Sounds like an enhancement story waiting to be created!

PR #2702, opened 2018-01-02
Redis store update

carrickr:
Updated to resque 1.2.7 and the newest redis-rails, which kicks redis-store up to 1.4.1.

Unless there was a reason not to update one of these, this should fix the redis-store being less than 1.4.0 issue.


phuongdh:
@carrickr the redis-store commit looks good but 2.5 is breaking on Travis, can you put the Travis commit on another PR?

PR #2730, opened 2018-01-08
Monkey patch resque task until PR is merged

cjcolvar:
This should fix the issue with resque failing to start on spruce.

This can be reverted once https://github.com/resque/resque/pull/1597 is merged.


cjcolvar:
The one codeclimate issue is from the original code that I copied from resque so it should be ignored because it can't be fixed without breaking the resque tasks.

PR #2742, opened 2018-01-11
Fix hls.js cookie and mejs4 svg for AWS

phuongdh:


phuongdh:
Since this requires some specific headers from the streaming server:
`Access-Control-Allow-Origin` cannot be '*'
`Access-Control-Allow-Credentials` must be `'true'`

It should remain for AWS only.

PR #2775, opened 2018-01-25
Update the batch_ingest_job to work with the new 6.3 batch ingest and…

carrickr:
… refactor ingest.rb slightly

Code climate issues predate this code and are just there because I'm touching it

### DIFF
--- a/app/jobs/batch_ingest_job.rb
+++ b/app/jobs/batch_ingest_job.rb
@@ -12,8 +12,25 @@
 #   specific language governing permissions and limitations under the License.
 # ---  END LICENSE_HEADER BLOCK  ---
 
+# This job is primarily intended for use within an AWS or similar environment
+# When a new spreadsheet is uploaded S3 will trigger an AWS Lambda, which in
+# turn queues the job below.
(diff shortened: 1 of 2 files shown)


app/jobs/batch_ingest_job.rb
```diff
@@ -25,13 +42,13 @@ def perform(filename)
     ingest = Avalon::Batch::Ingest.new(collection)
     begin
       package = Avalon::Batch::Package.new(filename, collection)
-      ingest.ingest_package(package)
+      ingest.process_valid_package(package: package)
```
cjcolvar: Is there a reason why you are skipping the package validation here?  Wouldn't you still want to do that or is the lambda taking care of that part?

cjcolvar: I'm okay with the lambda if it gets included in the cloud formation in https://github.com/avalonmediasystem/avalon-aws

It would be nice to maybe make a note to that effect above.  Maybe in line 17 "an AWS Lambda, which runs validations then queues the job below."?

PR #2776, opened 2018-01-25
Don't force-load all sections just to get a count

mbklein:
`MediaObject#ordered_master_files` is expensive. `MediaObject#master_file_ids` is cheap.


joncameron:
giphy

PR #2780, opened 2018-01-27
Restore the FontAwesome ➡️ arrow in the margin before the active section

mbklein:
Restores the `fa-arrow-circle-right` glyph to the section list in front of the currently playing section.

Before PR:
no arrow

After PR:
arrow

Unlike the Avalon pre-6.3 version of this effect, it's entirely CSS-driven with no DOM manipulation.

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -122,6 +122,7 @@ class MEJSPlayer {
    * @return {void}
    */
   getNewStreamAjax(id, url, playlistItemsT) {
+    $('.media-show-page').removeClass('ready-to-play')
     $.ajax({
       url: url + '/stream',
       dataType: 'json',
@@ -149,6 +150,7 @@ class MEJSPlayer {
   handleCanPlay() {
     this.mediaElement.removeEventListener('canplay');
     // Do we play a specified range of the media file?
+    $('.media-show-page').addClass('ready-to-play')
     if (this.switchPlayerHelper.active) {
       this.playRange();
     }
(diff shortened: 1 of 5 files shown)


mbklein:
This change addresses nulib/avalon#335

cjcolvar:
After talking with @joncameron, we think it would be good to have the arrow on both the section and current structure node.  @mbklein can you add that and then we can test it out on spruce after merge?

adamjarling:
Is this going into core Avalon, or our NU Avalon?  I assumed  Is this not working as expected in core?

mbklein:
It's an issue in both core and in nulib/avalon.

@cjcolvar @joncameron Are you saying you want an arrow on the current section _and_ on the current structure node within that section, or _only_ on the most granular level available? i.e., should there only ever be one arrow showing?

bkeese:
It used to only show at the most granular level. But, I think they are saying it's ok if it shows at both. One issue, if going with the old behavior, is what happens when the structure section for the currently playing section is collapsed?

cjcolvar:
@mbklein Yes, @joncameron and I thought it best to have both arrows display at the same time partially based upon the argument @bkeese mentioned.

mbklein:
@bkeese gives me an idea. I'm going to play with this for a little bit.

mbklein:
## UPDATE

* Arrows now show on the currently playing section and structured subsection (if any).
* Arrows are hidden during load and transitions between sections.
* Everything is implemented through CSS classes; no elements are dynamically created or deleted.
* Playlist styling has been updated to match (e.g., arrow outside line item numbers; consistent indenting).
yasss - avalon media system 2018-01-30 11-08-52

bkeese:
My memories of crazy RTMP event inconsistencies makes me a bit uncomfortable relying on the canPlay event. Otherwise 👍

PR #2781, opened 2018-01-29
Avoid reifying the object for show view

cjcolvar:
Makes a 88 section item load in ~1.5 sec which is the same as a 1 section item.


cjcolvar:
I fixed the two rubocop issues that were easy.  I suggest we forget about the indentation ones since that really requires fixing the entire file and opens a can of worms.

PR #2789, opened 2018-01-31
Develop

cjcolvar:


cjcolvar:
Not actually needed right now so closing until we're closer to release testing (or policy decided on to push sooner than that).

PR #2797, opened 2018-02-01
Add bulk action for intercom push

bkeese:
Adds bulk action for intercom push.
Refactors intercom push forms and javascript so they can be used also by bulk action.

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2018, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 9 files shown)


app/controllers/bookmarks_controller.rb
```diff
@@ -37,6 +37,8 @@ class BookmarksController < CatalogController
 
   self.add_show_tools_partial( :add_to_playlist, callback: :add_to_playlist_action )
 
+  self.add_show_tools_partial( :intercom_push, callback: :intercom_push_action, if: Proc.new { |context, config, options| context.user_can? :intercom_push } )
```
cjcolvar: Should this whole line have a conditional to avoid it loading unless Settings.intercom.present?

app/controllers/bookmarks_controller.rb
```diff
@@ -196,4 +200,30 @@ def move_action documents
...
+        media_object = MediaObject.find(id)
+        if cannot? :intercom_push, media_object
+          errors += ["#{media_object.title} (#{id}) #{t('blacklight.messages.permission_denied')}."]
+        else
+          success_ids << id
+        end
+      end
+      if success_ids.present?
+        session[:intercom_default_collection] = params[:collection_id]
+        BulkActionJobs::IntercomPush.perform_later success_ids, current_user.user_key, params
```
cjcolvar: This might be a bit paranoid, but I wonder if we should do `params.slice(:collection_id, :include_structure)` to avoid potential serialization/deserialization hacks since `params` would include anything the user submitted.

cjcolvar: @bkeese and I discussed this and decided that a follow on ticket can be created to apply the slice to each of the bulk actions in bookmarks_controller if it is determined this is a real concern.

cjcolvar:
Looks like my suggestion was a bad one.  The two tests are failing with that extra guard conditional added in the last commit.

app/controllers/bookmarks_controller.rb
bkeese: I think because of timing of evaluation of this line, putting the conditional here didn't work. Reverting.

PR #2821, opened 2018-02-09
Persist quality setting over section switches and handle missing deri…

### LINKED ISSUE #2820: MEJS Quality Plugin needs work
The MEJS quality plug-in will not display the quality choice modal if a derivative with the default quality does not exist. That is, unless the default is 'auto', in which case it will look for a derivative with the quality 'auto' or find the first available derivative if 'auto' isn't available.

We need to change the plug-in so that it will find the first available derivative when the default it not there, no matter what the default is.

This change should probably be pushed upstream.

bkeese:
…vatives

Fixes #2820

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/mejs4_helper_quality.es6
+++ b/app/assets/javascripts/media_player_wrapper/mejs4_helper_quality.es6
@@ -17,7 +17,10 @@
  */
 class MEJSQualityHelper {
   constructor() {
-    $(document).on('mejs4handleSuccess', this.addQualitySelectorListeners);
+    $(document).on(
+      'mejs4handleSuccess',
+      this.addQualitySelectorListeners.bind(this)
+    );
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/media_player_wrapper/mejs4_helper_quality.es6
```diff
@@ -33,18 +36,21 @@ class MEJSQualityHelper {
...
+  }
+
+  sendQualitySelection(quality) {
     $.ajax({
       type: 'POST',
       url: '/media_objects/set_session_quality',
-      data: { quality: e.target.value },
+      data: { quality: quality },
       dataType: 'json'
     });
```
adamjarling: Does this need any type of callback where the UI needs to respond or do something?  Or is the POST request good enough to live on it's own?

vendor/assets/javascripts/mediaelement/plugins/quality-avalon.js
```diff
@@ -92,16 +92,12 @@
...
-                if (value === 'auto') {
-                  var keyExist = t.keyExist(qualityMap, value);
-                  if (keyExist) {
-                    label = value;
-                  } else {
-                    var keyValue = t.getMapIndex(qualityMap, 0);
-                    label = keyValue.key;
-                  }
-                } else {
+                var keyExist = t.keyExist(qualityMap, value);
```
adamjarling: Are we changing the source code of this quality plugin?  If so, we should make a very explicit comment someplace, or override the file on our end so if we do a plugin update, it doesn't mysteriously break.

app/assets/javascripts/media_player_wrapper/mejs4_helper_quality.es6
bkeese: No, the post is good enough.

vendor/assets/javascripts/mediaelement/plugins/quality-avalon.js
bkeese: From the name, I thought this was our override. Isn't it?

adamjarling: Ah, yea sorry.  I just saw the `vendor/...` at the beginning.  I think I originally modified this for some weird focus behavior fix.

PR #2841, opened 2018-02-20
Warn user about large upload sets and prevent re-tries with modal

### LINKED ISSUE #2811: Investigation of Issues with Upload of Large Item Sets from GUI
### Description
Investigation needs to be done regarding bug #2801 before issues can be created to deal with the work needed to resolve.

### Done Looks Like
- Timeboxed investigation is performed to seek out probably causes and work needed to for  #2801

### Reference Issue
research for #2801

bkeese:
Fixes #2811 

Adds warning to dropbox modal about uploading too many files.
Displays "Uploading" modal to prevent further concurrent uploads.


bkeese:
Regarding the codeclimate check: I don't know what the line limit is, but I think it is too short.

PR #2860, opened 2018-03-08
Add comment field to media_object and master_file models

bkeese:


app/models/media_object.rb
```diff
@@ -189,6 +192,14 @@ def update_dependent_properties!
     self.set_resource_types!
   end
 
+  def all_comments
+    comment.sort + ordered_master_files.to_a.collect do |mf|
+      mf.comment.collect do |c|
+        c.present? ? "[#{mf.display_title}] #{c}" : nil
+      end.sort
+    end.flatten.uniq
```
cjcolvar: Is `uniq` is really necessary here?

app/models/media_object.rb
```diff
@@ -189,6 +192,14 @@ def update_dependent_properties!
     self.set_resource_types!
   end
 
+  def all_comments
+    comment.sort + ordered_master_files.to_a.collect do |mf|
+      mf.comment.collect do |c|
```
cjcolvar: How about rejecting `nil`s before the collect?

PR #2862, opened 2018-03-09
Update API for new comment fields

bkeese:
Adds new comment fields to api for media_objects and master_files.

Rails strong parameters weren't working correctly for scalar array fields nested within hashes (comments and other_identifiers), so refer to those fields directly.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -225,7 +225,7 @@ class MediaObjectsController < ApplicationController
       error_messages += ['Failed to create media object:']+@media_object.errors.full_messages
     elsif master_files_params.respond_to?('each')
       old_ordered_master_files = @media_object.ordered_master_files.to_a.collect(&:id)
-      master_files_params.each do |file_spec|
+      master_files_params.each_with_index do |file_spec, index|
         master_file = MasterFile.new(file_spec.except(:structure, :captions, :captions_type, :files, :other_identifier, :label))
         # master_file.media_object = @media_object
         master_file.structuralMetadata.content = file_spec[:structure] if file_spec[:structure].present?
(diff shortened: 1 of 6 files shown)


spec/factories/master_files.rb
```diff
@@ -57,5 +57,12 @@
         mf.save
       end
     end
+    trait :with_comments do
+      status_code 'COMPLETED'
+      after(:create) do |mf|
+        mf.comment += ['MF Comment 1', 'MF Comment 2']
```
cjcolvar: Do these have to be added after create?  If they can be added before then the second save could be avoided.

app/controllers/media_objects_controller.rb
```diff
@@ -237,6 +237,7 @@ def update_media_object
         master_file.title = file_spec[:label] if file_spec[:label].present?
         master_file.date_digitized = DateTime.parse(file_spec[:date_digitized]).to_time.utc.iso8601 if file_spec[:date_digitized].present?
         master_file.identifier += Array(file_spec[:other_identifier])
+        master_file.comment += Array(params[:files][index][:comment])
```
cjcolvar: Is the comment not in `file_spec` or is this just to workaround the strong params issue?

bkeese: Correct. Adding the comment to the strong parameters worked, in so far as the comment parameter wasn't rejected. However the values weren't there in the permitted parameter.

spec/factories/master_files.rb
bkeese: I did it this way because that is how we do all the other traits.

cjcolvar: The other traits in this factory deal with separate fedora objects so they require that the master file be created first to attach to.  This trait is just a property on the master file so is safe to do directly without the after_create.

PR #2885, opened 2018-03-21
2882 - thumbnail/poster

### LINKED ISSUE #2882: Unable to update thumbnail on items pushed to Spruce through intercom
### Description
When a video item is pushed to Spruce through intercom from Mallorn, thumbnail is not successfully transferred, and also user cannot update the thumbnail on this item in Spruce.

eg: https://spruce.dlib.indiana.edu/media_objects/hh63sw016

bkeese:
Fixes #2882 

Generates thumbnail and poster upon API ingest.
Fixes thumbnail player plug-in. Allow it to work when poster doesn't already exist.

Oh, and updates code version.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -585,6 +585,8 @@ class MediaObjectsController < ApplicationController
                              :other_identifier,
                              :structure,
                              :physical_description,
+                             :width,
+                             :height,
                              files: [:label,
                                      :id,
                                      :url,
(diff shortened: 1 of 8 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -242,6 +242,7 @@ def update_media_object
         if file_spec[:files].present?
           if master_file.update_derivatives(file_spec[:files], false)
             @media_object.ordered_master_files += [master_file]
+            master_file.update_stills_from_offset!
```
cjcolvar: Could you please add a test for this?  I think it could just expect that that method gets called and could probably live in this test: https://github.com/avalonmediasystem/avalon/blob/a8b38aaeac4364433c0d9988976e26fb09f0e457/spec/controllers/media_objects_controller_spec.rb#L413-L419

PR #2901, opened 2018-04-02
Move cache_store and TrafficControl config to initializers

phuongdh:
The problem is environment variables not able to override Settings config at the stage where application.rb is loaded, thus I couldn't change the Redis host to something that's not "localhost".


config/initializers/traffic_control.rb
```diff
@@ -0,0 +1,2 @@
+ActiveJob::TrafficControl.client = Redis.new(Rails.application.config.cache_store[1])
```
cjcolvar: Please add a comment that this initializer depends on the cache store initializer being run first and the cache store being redis.

PR #2918, opened 2018-04-16
Fix adaptive streaming hls playlists when using Wowza and native playback

cjcolvar:
Avalon's naive adaptive streaming hls master playlists are broken because HLS client implementations only follow one level of nesting of m3u8 playlists and Wowza only returns a master playlist with a playlist reference inside (instead of the child .ts files).  See https://www.wowza.com/docs/how-to-troubleshoot-apple-hls-playback#callflow.

This PR fetches the wowza playlists, parses them, and passes that info along to the Avalon adaptive streaming hls manifest generator.  This is a non-ideal approach but is workable at the moment.  This request takes between 0.5 and 1.5 seconds to return and appears to be run for every page load.

This PR also removes the PROGRAM-ID from the hls manifest since it isn't necessary and is listed as deprecated by the Apple hls stream validator. (Tool available with Apple Developer ID at https://developer.apple.com/go/?id=http-live-streaming-tools)

IU will test locally and if successful this PR can be merged.

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -268,8 +268,7 @@ class MasterFilesController < ApplicationController
   def hls_adaptive_manifest
     master_file = MasterFile.find(params[:id])
     authorize! :read, master_file
-    stream_info = secure_streams(master_file.stream_details)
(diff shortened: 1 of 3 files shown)


joncameron:
IU Issue for this: https://bugs.dlib.indiana.edu/browse/VOV-5706

PR #2956, opened 2018-05-08
Feature/2903

### LINKED ISSUE #2903: Files In Dropbox Are Not Automatically Removed During Ingest
### Description
When a file placed in an Avalon dropbox directory is ingested, it is never deleted from the original dropbox location. This results in files that are never moved or removed from the original dropbox location, and so these dropbox locations fill up with files that have already been ingested and processed, as well as creating confusion as to whether the files have actually been managed properly during ingest. This is also a regression from how Avalon used to deal with these files. 

### Possible Causes
@phuongdh investigated and thought that the issue may lie in the master_file model

### Done Looks Like
If configured to do so, masterfiles are deleted after ingest from their dropbox directories.

carrickr:
fixes #2903

### DIFF
--- a/app/jobs/cleanup_working_file_job.rb
+++ b/app/jobs/cleanup_working_file_job.rb
@@ -15,7 +15,7 @@
 class CleanupWorkingFileJob < ActiveJob::Base
   def perform(masterfile_id)
     masterfile = MasterFile.find(masterfile_id)
-    path = masterfile.working_file_path(File.basename(masterfile.file_location))
+    path = masterfile.working_file_path
     File.delete(path) if File.exist?(path)
   end
 end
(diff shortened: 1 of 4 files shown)


app/jobs/cleanup_working_file_job.rb
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+class CleanupWorkingFileJob < ActiveJob::Base
+  def perform(masterfile_id)
+    masterfile = MasterFile.find(masterfile_id)
+    path = masterfile.working_file_path(File.basename(masterfile.file_location))
```
phuongdh: `working_file_path` doesn't take an argument.

app/jobs/cleanup_working_file_job.rb
```diff
@@ -13,9 +13,9 @@
 # ---  END LICENSE_HEADER BLOCK  ---
 
 class CleanupWorkingFileJob < ActiveJob::Base
-  def perform(masterfile_id)
+  def performapp/jobs/cleanup_working_file_job.rb
```
bkeese: cut/paste error?

PR #2974, opened 2018-06-25
Configurable require for aws/rails/mailer

### LINKED ISSUE #2973: Should not always require aws/rails/mailer when server is EC2
This initializer requires aws mailer whenever it detects the server is an EC2 instance
https://github.com/avalonmediasystem/avalon/blob/master/config/initializers/mailer.rb

However when running Avalon on a cloud service somewhere else that might use EC2 (Dockerhub) it still triggers and fails.

phuongdh:
Fixes #2973


phuongdh:
@mbklein if this is merged you may need to add SETTINGS__EMAIL__MAILER to the the Beanstalk environment.

PR #2990, opened 2018-07-17
Upgrade to Rails 5.1

cjcolvar:
***DON'T MERGE!!! This needs to wait for PO approval before merging since it is major and might break everything.***

This PR is an upgrade to Rails 5.1 in preparation for a LTS release of avalon 6.5.  This will need very thorough testing and detailed upgrade instructions.

Mediaelement 2 code removed due to dependency on rails < 5
Gemfile.lock deleted and recreated
~Devise omniauth provider initialization disabled due to change in ActionController API.~
~Next step is ensure tests are passing and then deal with deprecation warnings~
(See comment below about not being able to upgrade to rails 5.2 yet.)


cjcolvar:
Upgrading to Rails 5.2 is currently blocked by ActiveFedora's support for Rails 5.2 not being in a release yet.  This means that `speedy_af` can't relax it's restriction of `activesupport < 5.2`.

cjcolvar:
So I think the next steps on this is local smoke testing.  Then merging and full regression testing whenever we get the okay from the POs.

PR #3004, opened 2018-07-27
Allow JSON access to public published media objects without an API token

cwant:
Also modify the way unauthorized JSON requests are handled:
return 401 instead of redirect to root path.


cjcolvar:
Closed in favor of https://github.com/avalonmediasystem/avalon/pull/3063 which incorporates this PR.

PR #3025, opened 2018-08-06
Issue#3006

yingfeng-iu:


app/views/media_objects/_metadata_display.html.erb
```diff
@@ -52,6 +52,9 @@ Unless required by applicable law or agreed to in writing, software distributed
 </dl>
 <% if can? :inspect, @media_object %>
   <%= link_to 'View Raw Metadata', inspect_media_object_path(@media_object.id, 'descMetadata'), :target => 'blank', :class => "btn btn-default btn-small" %>
+  <%# TODO 3006 move this to structure metadata page
```
cjcolvar: How about removing this since it is commented out and the work of wiring up this to the UI will be in a future story.

spec/models/master_file_spec.rb
```diff
@@ -544,6 +544,20 @@ class MyEncoder < ActiveEncode::Base
...
+    let(:master_file) { FactoryGirl.create(:master_file) }
+    it 'has a caption' do
+      expect(master_file.captions).to be_kind_of IndexedFile
+    end
+  end
+
+  describe 'waveforms' do
+    let(:master_file) { FactoryGirl.create(:master_file) }
+    it 'sets original_name to default value' do
+      expect(master_file.waveform.original_name).to eq 'waveform.json'
```
cjcolvar: How about adding an expectation here just like line 550 above checking `be_kind_of IndexedFile`?

PR #3051, opened 2018-08-24
Fix working file path

cjcolvar:
The working file path is only used when Settings.matterhorn.media_path is configured.  The code changes in prior commits did not take into account all of the ways ingest happens: web upload, web upload skip transcoding, web dropbox, web dropbox skip transcoding, batch ingest, batch ingest skip transcoding, and batch ingest pre-transcoded derivatives.  The approach to this commit was to write tests for each of these scenarios with both the Settings.matterhorn.media_path configuration set and not set then make all of the tests pass.  The underlying use case for the working file path is when Matterhorn is running on a different server than Avalon and they don't share the same filesystem (and filesystem paths) for both the Rails web upload directory and dropbox.  The working_file_path is a copy of the originally uploaded file to where Matterhorn has access and at a path shared by both servers.  For example, if Matterhorn is running on a different server and Matterhorn's ingest directory is '/srv/matterhorn/ingest' and mounted on the Avalon server at the same location and Avalon's dropbox directory is '/srv/avalon/dropbox' then Settings.matterhorn.media_path should be configured to be '/srv/matterhorn/ingest'.  Note that Matterhorn's ingset directory needs to be mounted at the same location on both the Avalon and Matterhorn servers for Avalon to be able to ingest due to the use of absolute paths and not relative paths.

Changes in this PR:
- Pass `working_file_path` to Matterhorn if present
- Change `working_file_path` to be a stored value on `MasterFile`
- Thoroughly test all cases  (AWS not tested because it shouldn't have `Settings.matterhorn.media_path` set.)

### DIFF
--- a/spec/integration/working_file_path_spec.rb
+++ b/spec/integration/working_file_path_spec.rb
@@ -1,0 +1,298 @@
+require 'rails_helper'
+
+# MasterFile#working_file_path has been a source of problems since it was introduced in 6.4.3
+# This spec file is meant to thoroughly test it in order to find any remaining bugs and to
+# ensure the intended functionality doesn't get broken accidentally in the future.
+#
+# Need to test all of the possiblilites:
+# batch ingest with single file
+# batch ingest with single file skip transcoding
+# batch ingest with pre-transcoded derivatives
+# web upload
+# web upload skip transcoding
+# BE dropbox upload
+# BE dropbox skip transcoding
+# Repeat all of these with and without media path set.
+# Also with dropbox mounted on matterhorn and not.?
+#
+# Pre-existing tests that are related (or duplicative)
+# spec/models/master_file_spec.rb:262
+# spec/models/master_file_spec.rb:500
+# spec/lib/avalon/batch/entry_spec.rb:80
+#
+describe "MasterFile#working_file_path" do
+  let(:master_file) { FactoryGirl.build(:master_file) }
+  let(:media_object) { FactoryGirl.create(:media_object) }
+  let(:workflow) { 'avalon' }
+
+  context "with Settings.matterhorn.media_path set" do
+    let(:media_path) { Dir.mktmpdir }
+
+    around(:example) do |example|
+      begin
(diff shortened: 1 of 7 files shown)


mcwhitaker:
@cjcolvar What is the size of this ticket?

cjcolvar:
@mcwhitaker I think it is an 8.

cjcolvar:
Really fixes https://github.com/avalonmediasystem/avalon/issues/2903

spec/integration/working_file_path_spec.rb
```diff
@@ -0,0 +1,298 @@
...
+# This spec file is meant to thoroughly test it in order to find any remaining bugs and to
+# ensure the intended functionality doesn't get broken accidentally in the future.
+#
+# Need to test all of the possiblilites:
+# batch ingest with single file
+# batch ingest with single file skip transcoding
+# batch ingest with pre-transcoded derivatives
+# web upload
+# web upload skip transcoding
+# BE dropbox upload
```
cwant: No clue what BE is, might be worth unrolling the acronym (back end?)

spec/integration/working_file_path_spec.rb
```diff
@@ -0,0 +1,298 @@
...
+# BE dropbox skip transcoding
+# Repeat all of these with and without media path set.
```
cwant: One thing that I haven't been able to find in the documentation: under what situations will media path be set, and when will it not be set? Might be worth a comment (somewhere, not necessarily here) about why it will or will not be set.

cjcolvar: :+1: I think we definitely need to add more documentation for guidance around setting/not setting `Settings.matterhorn.media_path`

spec/integration/working_file_path_spec.rb
cjcolvar: Sure, I'll change that now.

mcwhitaker:
Tests passed on MCO-staging.

PR #3061, opened 2018-08-31
Batch ingest post-encoding email.

cwant:
It uses an unused attribute called `processed_email_sent` in the `BatchRegistries` class
to flag whether an email was sent already about this. Added some behaviour to the
`BatchEntries` and `BatchRegistries` classes to track encoding status.

Initiated via the rake task `avalon:batch:ingest_status_check` that is run via cron.
See app/jobs/ingest_batch_status_email_jobs.rb for mail entry point.


joncameron:
seal-of-approval2

cjcolvar:
Incorported and closed by #3104.  Thanks @cwant!

PR #3074, opened 2018-09-26
Update locking_runner script to delete stale lock files

mohideen:
We had a power interruption which caused a stale lock file to stick around and prevent batch_ingest cron job from running. I have updated the locking runner script to delete lock file if there is no running process with the PID in the lock file.

### DIFF
--- a/script/locking_runner
+++ b/script/locking_runner
@@ -5,6 +5,13 @@
 
 SEMAPHORE=$1
 LOCKFILE=tmp/pids/$SEMAPHORE
+
+if ( [ -e $LOCKFILE ] && ! kill -0 $(< $LOCKFILE) ) 2> /dev/null;
+then
+  echo "$0: Deleting stale lock file: $LOCKFILE"
+  rm -f "$LOCKFILE";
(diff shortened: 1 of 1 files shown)


mohideen:
Let me know if the changes need any improvements.

script/locking_runner
```diff
@@ -5,6 +5,13 @@
 
 SEMAPHORE=$1
 LOCKFILE=tmp/pids/$SEMAPHORE
+
+if ( [ -e $LOCKFILE ] && ! kill -0 $(< $LOCKFILE) ) 2> /dev/null;
```
phuongdh: `kill -0` will fail if the user doesn't have permission to send signal to the process, which could happen in some corner cases. Did you consider `ps -p`, since it doesn't have this problem?

cjcolvar:
@mohideen Sorry about taking a long time to get back to this!  It looks good and ready to go.  Are you able to switch the target base branch to `develop` and rebase it if necessary?

mohideen:
Created new PR #3102

PR #3075, opened 2018-09-28
Display user exists message on duplicate registration

mohideen:
Currently, omniauth-identity will just log in a user if the user account
already exists. This adds a failure message and redirects the user back
to the form.

---

We had a user who was trying to register for an account forgetting that
they have already registered and they were shown an error page without
any indication of why the error happened.

Chris (@cfitz) worked on this feature to show a notice instead of throwing
an error on duplicate registration. I have tested this locally and is working
fine.


joncameron:
seal-of-approval2

cjcolvar:
Fixed with #3161.

PR #3119, opened 2018-12-10
Update hydra access_controls

bkeese:


cjcolvar:
It looks like all three builds failed on the same three tests.  Maybe if this PR is rerun conservatively (`bundle install` or `bundle update --conservative hydra-access-controls`) then it will fix the tests?

PR #3128, opened 2018-12-20
Generate iiif manifests using iiif_manifest

### LINKED ISSUE #3113: Avalon Generates IIIF Presentation 3.0 Manifests

cjcolvar:
This PR generates IIIF presentation 3.0 manifests using the `iiif_manifest` gem and code copied from `hyrax-iiif_av` into new presenter objects.  This is a first pass which should mostly be right but the manifests have not been validated (no validator currently exists) and currently don't work in the timeliner (could be any reason) and don't playback in the UV examples page (probably due to HLS).  These manifests require authorization that is tied to the media object so if you can read the media object then you can view the iiif manifest.  Also included is `rack-cors` to provide a better way to configure and handle CORS headers.

Fixes #3113.

### Future work
- Refine and reorder the descriptive metadata included in the manifest
- Implement IIIF Auth
- Add thumbnails to individual canvases (probably requires changes to `iiif_manifest`)
- Performance improvements (ensure that `speedy_af` is used and nothing requires reifying)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,8 @@ gem 'bootstrap-toggle-rails', git: 'https://github.com/rkallensee/bootstrap-togg
 gem 'bootstrap_form'
 gem 'speedy-af', '~> 0.1.1'
 gem 'recaptcha', require: 'recaptcha/rails'
+gem 'iiif_manifest'
+gem 'rack-cors', require: 'rack/cors'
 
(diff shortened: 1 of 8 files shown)


cjcolvar:
I fixed a few rubocop/codeclimate issues but I vote for ignoring the remaining 6.

PR #3136, opened 2019-01-07
Add CORS config for waveform and structure endpoints

bkeese:


cjcolvar:
Thinking about this again, what do we need CORS for?  Shouldn't these always be being accessed from the same domain?

bkeese:
I think that if we are creating API endpoints, it makes sense to make them available outside our the application's domain.

PR #3140, opened 2019-01-10
Test with ruby 2.6 and drop 2.3

cjcolvar:
Ruby 2.6 was released on 12/25/2018 and Ruby 2.3 is EOL on 3/31/2019.


bkeese:
Need to update rubocop target ruby version to something higher than 2.3 first.  And this will break all sorts of rubocop tests.

PR #3141, opened 2019-01-14
Bypass cancan authorization for structure post temporarily

cjcolvar:
We'll need a story to determine how to handle the authentication/authorization for the structure metadata editor.  If this structure endpoint is going to be an API for content owners to use then we should have a more robust Authentication mechanism than what exists for the ingest API.  If we don't plan on giving the users access to the api but instead rely on the SME living within the rails app then I don't think we need to handle authentication yet.


davidschober:
I think at this point the SME/ Content owner would live within the rails app. Programmatically updating structure is a great idea, but that sounds like something a dev would do at this point.

Issue #3019, opened 2018-08-03
Launch Structural Metadata Editor


joncameron:
### Description
Decisions need to be made on how to display the structural metadata editor within the application. The current editor launches within a modal on the page, but this might not be the right choice for the visual editor.

### Done Looks Like
- Avenues are explored (most likely via a meeting) talk through the following—
- Decision is made about how the visual and XML editors are launched within the edit page for an item
- Decision is made about how to present the visual editor within the app—within a modal, open in a complete new page for more room, etc.

### Notes
- Structure section of the edit page contains buttons for "edit" and "advanced edit"—"edit" opening the visual editor, and "advanced" the XML editor
- A large modal, much like the current one for editing structure XML, will appear on the page after the "edit" button is clicked.

### Buttons on the Structure edit page for a media object 
<img width="942" alt="screen shot 2018-12-20 at 11 20 17 am" src="https://user-images.githubusercontent.com/1406618/50296920-890d8880-0449-11e9-96d0-00fa7537efd5.png">

### Example of the editor within a modal on the page, much like the XML structure editor
![editorlaunchexample](https://user-images.githubusercontent.com/1406618/51939031-886af900-23dc-11e9-96dd-88a78633cee4.png)

child of https://github.com/samvera-labs/avalon-bundle/issues/231

joncameron:
<img width="942" alt="screen shot 2018-12-20 at 11 20 17 am" src="https://user-images.githubusercontent.com/1406618/50296920-890d8880-0449-11e9-96d0-00fa7537efd5.png">

cjcolvar:
This mockup is implemented in #3142.  The next step is packaging the SME as an npm package and bringing that into the rails app and rendering it on the modal.

adamjarling:
Putting off work on packaging as NPM for now, and focus on integrating directly into Avalon 6.

PR #3142, opened 2019-01-15
React-rails and webpacker setup for bringing in the structural metadata editor

cjcolvar:
This follows the pattern used by `hyrax-iiif_av`.

Part of #3019.


cjcolvar:
We can safely ignore the codeclimate issues since they are in the binstubs generated for webpack.

PR #3163, opened 2019-01-31
[WIP] Test HydraHead#469

cjcolvar:
This PR is to run the Avalon test suite against https://github.com/samvera/hydra-head/pull/469


cjcolvar:
Hydra-head 10.6.1 was released and we can bundle update to it when we're ready.

Issue #3173, opened 2019-02-04
Add a hook for auth service in iiif_manifest


cjcolvar:
Part of #3171.

cjcolvar:
PR: samvera-labs/iiif_manifest #30

PR #3178, opened 2019-02-07
IIIF Auth Support

### LINKED ISSUE #3171: IIIF auth work

### LINKED ISSUE #3189: IIIF Auth token service with existing query params
The IIIF Auth implementation added to Avalon 6.x  in #3171 uses a token service with a query param.  This causes problems in UV and the digirati auth testing tool because they naively append their query params with an additional `?` separator leading to errors in avalon.

We need to either adjust our Avalon implementation or ask UV and digirati to fix their implementations.

cjcolvar:
This PR adds IIIF Auth support by adding auth service definitions on all content resources.  We took the approach of making our probe service be the content resource.  Also included is a shift from writing direct streaming links with stream tokens into the manifest with an endpoint which returns the direct streaming HLS manifest with the stream token.

This work was manually tested using https://digirati-co-uk.github.io/iiif-auth-client/?sources=https://spruce.dlib.indiana.edu/iiif/index.json and the UV examples page.

Closes #3171, closes #3189 
This depends on https://github.com/samvera-labs/iiif_manifest/tree/auth_service (#3173).


cjcolvar:
I think this is ready for review.  The 4 code climate issues I propose we don't fix and the code coverage issue is due to modifying existing code that doesn't have any tests.

app/models/iiif_canvas_presenter.rb
```diff
@@ -119,4 +121,32 @@ def structure_ng_xml
       # under root or each Div node, otherwise Nokogiri::XML parser will report error, and raise exception here.
       @structure_ng_xml ||= (s = master_file.structuralMetadata.content).nil? ? Nokogiri::XML(nil) : Nokogiri::XML(s)
     end
+
+    def auth_service
```
phuongdh: Can you add test for this new code?

config/routes.rb
```diff
@@ -124,10 +124,11 @@
       post 'attach_captions'
       get :captions
       get :waveform
-      get 'adaptive', to: 'master_files#hls_adaptive_manifest'
+      match ':quality.m3u8', to: 'master_files#hls_manifest', via: [:get], as: :hls_manifest
       get 'structure', to: 'master_files#structure', constraints: { format: 'json' }
       post 'structure', to: 'master_files#set_structure', constraints: { format: 'json' }
       delete 'structure', to: 'master_files#delete_structure', constraints: { format: 'json' }
+      get 'auth_token', to: 'master_files#auth_token'
```
phuongdh: When you do move this route, can we consider a more descriptive name like `iiif_auth_token` or `iiif_token`?

cjcolvar:
@bkeese Route change looks good.

PR #3212, opened 2019-02-27
Make player volume sticky between section changes

### LINKED ISSUE #2541: As a user, I want the player to remember my settings for volume and captions
Currently player resets volume and caption settings when I switch to another section in item view or another item in playlist view. It's not convenient.

bkeese:
Closes #2541 

Stores the last volume level used in browser's localStorage. Then when section or page changes occur, resets the volume to the last level used.

Also stores and reuses the last caption setting.

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -170,13 +170,24 @@ class MEJSPlayer {
   }
 
   /**
-   * Event handler for MediaElement's 'volume' event
+   * Event handler for MediaElement's 'volumechange' event
    * Save new volume to localStorage for initializing new players with that vol
    * @function handleVolumeChange
    * @return {void}
    */
   handleVolumeChange() {
     this.localStorage.setItem('startVolume', this.mediaElement.volume)
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -451,7 +468,8 @@ class MEJSPlayer {
       link_back_url: currentStreamInfo.link_back_url,
       qualityText: 'Stream Quality',
       defaultQuality: this.defaultQuality,
-      toggleCaptionsButtonWhenOnlyOne: true
+      toggleCaptionsButtonWhenOnlyOne: true,
+      startVolume: this.localStorage.getItem('startVolume') || 0.8
```
cjcolvar: @joncameron @davidschober Do you think we can default to volume at 100% instead of the mediaelement.js default of 80%?  I always find myself trying to crank it to 100% anyways.

joncameron:
```suggestion
      startVolume: this.localStorage.getItem('startVolume') || 1.0
```

joncameron: @cjcolvar Yes, let's do that as the default.

Issue #3218, opened 2019-03-01
Rails 5.2 Support


joncameron:
### Description
Rails 6 will come out soon, at RailsConf. Versions prior to 5.2 will be deprecated and only get severe security updates.

### Done Looks Like
The version of Rails Avalon uses is upgraded to 5.2.

cjcolvar:
Gems that need releases
- [x] ActiveFedora
- [x] SpeedyAF
- [ ] active_fedora-datastreams
- [ ] blacklight-access_controls
- [ ] hydra-head

Avalon PR: https://github.com/avalonmediasystem/avalon/pull/3219

PR #3219, opened 2019-03-01
Rails 5.2 support

cjcolvar:


cjcolvar:
Fixes #3218

cjcolvar:
I think we can't pin bootstrap-sass since we need a newer version to deal with the security vulnerability.

Issue #3223, opened 2019-03-07
Bring Structural Metadata Editor into Avalon


joncameron:
### Description
Figure out the best way to bring the structural metadata editor into the Avalon 6 environment.

### Done Looks Like
- Decision made about delivering the structural metadata editor within Avalon 6
- This work undertaken and the editor can be launched in the Avalon 6 context

### To Do
- [x] Investigate whether `nbw` is still best way to turn a `create-react-app` project into a npm package (#3241).
- [x] POC work for packaging up SME application, using a given strategy/tool (#3242).
- [x] Test importing of npm package from a Github branch (so we don't have to officially publish to npm yet) (create issue).

### Note
Adam noted that ideally this would be through a single module import.

joncameron:
Adam will detail out todos and steps needed in this ticket.

adamjarling:
Some beginning documentation created below:

@joncameron @cjcolvar @Dananji 

https://github.com/avalonmediasystem/structured-metadata-editor-react/wiki/Packaging-the-module-for-export

https://github.com/avalonmediasystem/react-structural-metadata-editor
This is incomplete, but to test this, next week or what not here are the main steps:

1. Follow the instructions to package the files in the `structured-metadata-editor-react` repo.
2. In `react-structural-metadata-editor` `/src` folder, delete everything, except `index.js` file (which is the unique entrance file for the packaging we want static here).
3. Manually copy over the files in the the `/src` directory.
4. Continue with packaging instructions on the README.

joncameron:
@adamjarling Should I create a new ticket for future work on packaging? Based on the instructions and things across the two separate repos it sounds like something more elegant is in the works eventually but I don't think we have an issue about that yet.

PR #3278, opened 2019-04-09
Pull in Structural Metadata Editor

cjcolvar:
Child of #3223

### DIFF
--- a/app/javascript/components/sme/app.js
+++ b/app/javascript/components/sme/app.js
@@ -1,9 +1,10 @@
 // app/javascript/structural-metadata-editor/index.js
 import React from 'react'
 import Test from './test';
+import RSME from 'react-structural-metadata-editor';
 
 const App = props => {
-  return <Test />
+  return <RSME />
 };
 
 export default App;
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -32,6 +32,7 @@ default: &default
     - .ttf
     - .woff
     - .woff2
+    - .mp3
 
   extensions:
     - .jsx
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "prop-types": "^15.7.2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-structural-metadata-editor": "https://github.com/avalonmediasystem/react-structural-metadata-editor.git#v1.1.0",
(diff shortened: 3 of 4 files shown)


adamjarling:
Maybe this entry gets around the issue mentioned above?  

https://github.com/avalonmediasystem/avalon/blob/a1a0783b7213f7853547997b100e1e6f4ceeb9bf/package.json#L4

I'd prefer to keep `@babel/runtime` out of `react-structural-metadata-editor` if possible.

cjcolvar:
I'll make that change.

cjcolvar:
@adamjarling It looks like the dependency on `babel-runtime` is coming from the compilation: https://github.com/avalonmediasystem/react-structural-metadata-editor/blob/master/es/containers/WaveformContainer.js#L1

I was able to figure out the issue with yarn asking you to choose a version to install.  The issue is that `babel-runtime` and `@babel/runtime` are different things and that line above references `babel-runtime` which only has a 6.26.0 release and no formal 7.0.0 release.  I created a branch in the SME packaging repo with `babel-runtime` included and removed it from avalon and things render without error.

adamjarling:
@cjcolvar Ok yea lets do whatever works to get this running in Avalon.  

My thinking was that I can pull  `https://github.com/avalonmediasystem/react-structural-metadata-editor/tree/v1.1.0-without-babel-runtime` into an application which is not Avalon and it works fine.  So it seemed `babel-runtime` or `@babel-runtime` (whichever version `webpacker` uses, and yea with v `7` the packages are now prefixed with the `@`...learned that this week too)  would be an Avalon environment dependency.

Let's do whatever works though for now, and as long as we can pull `react-structural-metadata-editor` into an independent application and it still runs and doesn't seem to impact anything, then I'm fine with it.

Let me know what existing branches work for you, let's maybe get those tightened up and delete whatever becomes stale (so I don't get confused:)

adamjarling:
Just testing around here, and in Avalon if I blow out my node modules folder, then do a `yarn install` given this `package.json` file, it loads:

https://github.com/avalonmediasystem/avalon/blob/3223-pull-sme-into-avalon/package.json

Something tells me leaving that `babel-runtime` in the exportable module is going to cause some weird issue in the future.  As long as we're using `nwb` package...something could end up conflicting.

cjcolvar:
Child of #3223.

cjcolvar:
Replaced with #3316

PR #3279, opened 2019-04-09
Add LTI share ability to timelines.

bkeese:


app/views/timelines/_form.html.erb
```diff
@@ -42,6 +42,24 @@
       </div>
     </div>
   </div>
+  <div id="timeline_lti_share" class="row form-group container collapse <%= 'in' if Avalon::Authentication.lti_configured? && @timeline.visibility != 'private' %>">
```
cjcolvar: How about moving the if `lti_configured?` up a level to wrap this whole div?  Would the JS that shows/hides need to be changed?

cjcolvar:
Are there any tests for this LTI integration for playlists that could be copied for at least some coverage?  Maybe for the `lti_share_url_for` helper?

bkeese:
@cjcolvar Good ideas. Done.

PR #3290, opened 2019-04-15
Add access info to media_object json and restrict to admins

bkeese:
Enhance json view of MediaObject.
Restrict access control and file data to admins.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -334,7 +334,9 @@ class MediaObjectsController < ApplicationController
         end
       end
       format.json do
-        render json: @media_object.to_json
+        response_json = @media_object.as_json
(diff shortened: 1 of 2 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -334,7 +334,9 @@ def show
         end
       end
       format.json do
-        render json: @media_object.to_json
+        response_json = @media_object.to_json
+        response_json = JSON.parse(response_json).except('files', 'visibility', 'read_groups').to_json unless current_ability.is_administrator?
```
cjcolvar: I think we can avoid the JSON parse by using `as_json` which should return the hash that gets used to create the JSON string:
```
response_json = @media_object.as_json
response_json = response_json.except('files', 'visibility', 'read_groups') unless current_ability.is_administrator?
render json: response_json.to_json
```

Issue #3314, opened 2019-04-30
Waveform.json not created when ingesting new files


Dananji:
### Description

When ingesting new media files the waveform is not getting created for most files (tried with .mp3 and .mp4 file)

cjcolvar:
Two parts of this:
1) Determine why waveforms are being created and fix it.
2) Add a hook to the SME show controller action that kicks off the waveform generating job when it encounters the waveform missing.  Change the SME error message to tell the user that the waveform is being generated and to check back later.

PR #3323, opened 2019-05-08
Generate waveform after processing is completed instead of before it starts.

cjcolvar:
Related to #3314


cjcolvar:
The diff coverage dropped because I refactored to deal with a rubocop issue so I vote to not worry about it since the code actually being changed is tested.

PR #3354, opened 2019-05-15
Enable Devise Login

orangewolf:
This branch is party of the sponsored Avalon work. It enables pure devise login for non-sso type users.  It makes omniauth-identity completely optional and enabled sign in, sign up (with a Settings flag) and forgot password.  It also installs devise invitable, but configuration of the invitable interface will be left to a future PR.


orangewolf:
@cjcolvar not sure why the upload coverage fails, but this the actual tests are green now

cjcolvar:
I wonder if the coverage uploading is an issue with this PR being from a fork and circleci not using an environment variable.  @phuongdh do you have any ideas on solving this?

orangewolf:
@cjcolvar totally makes sense to me. Our other work will ultimately depend on this, but you don't want that in before the release cycle ends either.

cjcolvar:
@orangewolf FYI I just force-pushed to the develop branch again.  You should be able to rebase without any need to resolve conflicts.  The release was merged to master with a rebase instead of a merge commit so only the refs changed.

app/assets/javascripts/sessions_new.js.coffee
```diff
@@ -0,0 +1,6 @@
+$ ->
```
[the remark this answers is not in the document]
orangewolf: no problem

config/settings.yml
```diff
@@ -68,6 +68,7 @@ encode_throttling:
   update_jobs_throttle_threshold: 3
   update_jobs_spacing: 10
 auth:
+  registerable: true
```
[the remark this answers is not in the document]
orangewolf: I agree it is kind of a toss up.  I think long term on by default means folks spinning up a test instance can more easily get it going, but that it will need to be clearly mentioned in the release notes.

app/views/modules/_footer.html.erb
```diff
@@ -19,6 +19,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 	   	 <li><%= image_tag "AvalonMediaSystem_Logo_A.png", alt: "Avalon Media System - #{t(:release_label)} #{Avalon::VERSION}" %>
 	     <li><a href="http://www.avalonmediasystem.org">Avalon Media System Project Website</a> </li>
        <li><%= link_to "Contact Us", comments_path %></li>
+       <li><%= link_to "Admin", new_session_path('user', admin: true) %></li>
```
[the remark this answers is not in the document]
orangewolf: my notes show the decision going the other way - that we wanted it in the footer and just not up in the top nav. Sounds like we should revisit this during todays call.

orangewolf:
@bkeese based on our conversation today I believe this PR is ready to go.  Note that upload-coverage still fails, but we think that is a fork PR issue (see Chris's comments above).  There will be a second PR with the actual engine enabled once we make the interface tweaks we talked about in today's call.

config/environments/development.rb
```diff
@@ -2,7 +2,7 @@
 
 Rails.application.configure do
   # Verifies that versions and hashed value of the package contents in the project's package.json
-config.webpacker.check_yarn_integrity = true
+config.webpacker.check_yarn_integrity = false
```
bkeese: Was leaving this change in intentional?

config/environments/development.rb
```diff
@@ -66,4 +66,7 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  # Docker/Vagrant debugging
+  config.web_console.whitelisted_ips = ['172.16.0.0/12', '192.168.0.0/16', '10.0.0.0/8']
```
bkeese: Also here. Same question.

config/environments/development.rb
orangewolf: great catch - I've put it back

config/environments/development.rb
orangewolf: @bkeese I lean toward saying this is universally a good change.  I basically means that Docker and Vagrant debugging works out to the box.  Having said that it probably isn't directly related to the scope of this PR, so I've taken it out of here. If y'all want it in separately, I do recommend it though.

bkeese: @phuongdh Want to weigh in on this one? I'm in favor of keeping it now that I know what it's about.

orangewolf: @bkeese if @phuongdh wants it in I can make it part of the second PR

phuongdh: Looks good. :+1: to keeping it through 2nd PR.

cjcolvar:
@orangewolf I skimmed back through and it looks good to me.  Do you think we need a DB migration that will populate the password attribute for all existing Users?  I see the DB migration set the default to `""` and wanted to make sure we weren't opening up all existing accounts to password-less logins.

orangewolf:
@cjcolvar great question. Those users simply will not be able to log in via username/password until a password is added to their accounts (via forgot password or admin). It does not default to allowing passwordless login (it did once upon a time, but that was in like 0. something ;-). Since they will still be able to use their existing SSO approach things should be ok.

cjcolvar:
:+1: Great!

cjcolvar:
@orangewolf I'm going to go ahead and merge this even though it doesn't meet the diff-coverage threshold.  Maybe the next PR could add a little bit more testing so we don't lose too much coverage.

PR #3373, opened 2019-05-22
Add/update license header to 2019

bkeese:


app/assets/javascripts/blacklight/facet_load.js
```diff
@@ -1,3 +1,19 @@
+/*
```
cjcolvar: It looks like we pull this one in via Blacklight so the header should be removed or replaced with one for blacklight.
https://github.com/projectblacklight/blacklight/blob/v6.20.0/app/assets/javascripts/blacklight/facet_load.js

app/assets/javascripts/modernizr.js
```diff
@@ -1,3 +1,19 @@
+/*
```
cjcolvar: This one should probably be removed as well.  It already specifies it's header as MIT & BSD.

app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,5 @@
-/*
- * Copyright 2011-2018, The Trustees of Indiana University and Northwestern
+/*
```
cjcolvar: Odd that the scss files are getting a space at the end of their first lines.  Probably not worth dealing with though.

app/assets/stylesheets/datatables.css
```diff
@@ -1,3 +1,19 @@
+/*
```
cjcolvar: I'm guessing this and the datatables.js come from `jquery-datatables` so we should remove the headers on them.

app/javascript/iiif-timeliner-styles.css
```diff
@@ -1,3 +1,19 @@
+/*
```
cjcolvar: @joncameron @mcwhitaker What should the license say about this and the timeliner pack since it was work for hire?

spec/fixtures/meow.json
```diff
@@ -1 +1,17 @@
+/*
```
cjcolvar: I don't think we really need the headers in the fixtures.

spec/lib/avalon/solr_config_uploader_spec.rb
```diff
@@ -1,3 +1,17 @@
+# Copyright 2011-2019, The Trustees of Indiana University and Northwestern
```
cjcolvar: It looks like this is verbatim from hyku so the header should be removed.

lib/avalon/solr_config_uploader.rb
```diff
@@ -1,3 +1,17 @@
+# Copyright 2011-2019, The Trustees of Indiana University and Northwestern
```
cjcolvar: This header should probably be removed since it was taken from hyku.

app/javascript/iiif-timeliner-styles.css
joncameron: Good question... I think @mcwhitaker can speak to that better than I could.

app/assets/stylesheets/datatables.css
cjcolvar: :+1: but it looks like datatables.js is still needing its header removed.

spec/fixtures/meow.json
cjcolvar: :+1: It looks like three more json fixtures could have headers removed: structure.json, structure_with_html.json, and waveform.json.

app/javascript/iiif-timeliner-styles.css
cjcolvar: I went ahead and merged this PR and we can make a smaller PR to fix these two files if we need to.

mcwhitaker:
@cjcolvar - all the timeliner work is IU's copyright.

PR #3374, opened 2019-05-23
Bump up timeouts to deal with failings tests in ci builds

cjcolvar:
This PR upgrade `rsolr` to 2.x in order to get faraday for better configurability.  Since we're using AF 12+ we can remove the initializer override for timeouts but I had to add one for solr.


bkeese:
For after release, right?

cjcolvar:
I guess so since it bumps the version of rsolr which is a key dependency.

cjcolvar:
It looks like the rsolr version bump is actually causing a failing test.

PR #3375, opened 2019-05-23
SME add props to pass media file duration

Dananji:
New prop for passing media file duration to fill in missing end attributes of timespans in the structure.


cjcolvar:
Since the SME PR got merged, how about including the package.json change to update the SME in avalon in this PR?

PR #3417, opened 2019-06-10
Update MediaInfo CLI URL (fixes #3416)

### LINKED ISSUE #3416: README has stale URL for Mediainfo CLI
Under “[Setting Up an Avalon Media System Development Environment](https://github.com/avalonmediasystem/avalon#setting-up-an-avalon-media-system-development-environment)”, the README gives http://mediainfo.sourceforge.net; the current URL appears to be https://mediaarea.net/en/MediaInfo/Download/Mac_OS.

dmolesUC:


README.md
```diff
@@ -23,7 +23,7 @@ The following steps will let you run the avalon stack locally in order to
 explore the out-of-the-box functionality or do basic development.
 
 * Ensure that you're running one of the Ruby versions listed in under rvm in ".travis.yml".
-* Install [Mediainfo cli](http://mediainfo.sourceforge.net)
+* Install [Mediainfo cli](https://mediaarea.net/en/MediaInfo/Download/Mac_OS)
```
cjcolvar: I think it would be best to keep this link OS-independent and provide some context similar to the old link.  How about this link instead?
```suggestion
* Install [Mediainfo cli](https://mediaarea.net/en/MediaInfo)
```

dmolesUC: From the first sentence of that section (“For developers using OS X, you can get a full Avalon development environment, including transcoding and streaming using docker”) I assumed that the whole section was intended for OS X developers. If not, then I'd agree, but maybe that should also be clarified?

cjcolvar: That first sentence is meant to give an alternative to the rest of the section.  I think it might be best to just remove the rest of the section entirely or move it further down the page and add warnings that it is not the recommended path.

cjcolvar: I'll make a PR that should hopefully clarify this.

cjcolvar: Do you think this version (proposed as #3418) makes things clear?
https://github.com/avalonmediasystem/avalon/blob/5be75c0eb52f3638a677c686fc478f173fceb3e9/README.md

cjcolvar:
Fixed by #3418.

PR #3450, opened 2019-07-08
Implement Samvera Persona

orangewolf:
Installs a new engine - samvera-persona
Support soft delete
Modify all layout view partials to append main_app to links, allowing engines to use the layout
Add link to user management section for admins
Show a message when becoming another user and allow logging out of second account and back in to primary account

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,12 +10,15 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_04_174803) do
+ActiveRecord::Schema.define(version: 2019_06_28_003746) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
 
   create_table "annotations", force: :cascade do |t|
     t.string "uuid"
     t.string "source_uri"
-    t.integer "playlist_item_id"
+    t.bigint "playlist_item_id"
     t.text "annotation"
     t.string "type"
     t.index ["playlist_item_id"], name: "index_annotations_on_playlist_item_id"
@@ -26,23 +29,23 @@ ActiveRecord::Schema.define(version: 2019_06_04_174803) do
     t.string "token", null: false
     t.string "username", null: false
     t.string "email", null: false
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
     t.index ["token"], name: "index_api_tokens_on_token", unique: true
     t.index ["username"], name: "index_api_tokens_on_username"
   end
 
   create_table "batch_entries", force: :cascade do |t|
-    t.integer "batch_registries_id"
-    t.text "payload", limit: 4294967295
+    t.bigint "batch_registries_id"
+    t.text "payload"
     t.boolean "complete", default: false, null: false
     t.boolean "error", default: false, null: false
     t.string "current_status"
-    t.string "error_message"
+    t.text "error_message"
     t.string "media_object_pid"
     t.integer "position"
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
     t.index ["batch_registries_id"], name: "index_batch_entries_on_batch_registries_id"
     t.index ["position"], name: "index_batch_entries_on_position"
   end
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/sessions_new.js.coffee
```diff
@@ -0,0 +1,7 @@
+$ ->
```
cjcolvar: Even though we have coffeescript in our repo, we're trying to only bring in new JS as plain old JS or ES6.  Do you mind converting this?

db/schema.rb
```diff
@@ -214,10 +217,13 @@
     t.integer "invitation_limit"
     t.integer "invited_by_id"
     t.string "invited_by_type"
+    t.datetime "deleted_at"
```
cjcolvar: It looks like you missed committing the migration that goes with this schema change.

app/assets/javascripts/sessions_new.js.coffee
orangewolf: no problem. sorry you had to tell me the same thing twice - do to the various rebases we lost a few of the things we fixed from the last PR. I'm going to go back and make sure we got all of those.

db/schema.rb
orangewolf: that schema change comes from a migration in the engine itself. the reason for that is the dependency that uses it is also part of the gem, not in Avalon itself. Does that make sense or would you prefer the migration be copied over?

cjcolvar: I've generally copied them over via the install:migration rake task so the db can always regenerate from the migrations.

app/assets/javascripts/sessions_new.js.coffee
cjcolvar: No worries.  We haven't done anything to make it apparent without having to rely on memory.

db/schema.rb
orangewolf: no problemo

orangewolf: or maybe not - as long as the gem is installed these migrations will run.  if we copy them over rails gets upset because it then sees them twice.  I do think that not copying them is the better course here simply because if you have samvera-persona installed you get these migrations on any db:migrate. If you don't have samvera-persona installed (or comment it out) you never get these migrations. What do you think?

cjcolvar: I'll have to take a look at this tomorrow and give a quick ponder.

cjcolvar: I still think it would be cleaner and safer to follow the existing rails process of having the engine's migrations copied over via the install:migration rake task (https://edgeguides.rubyonrails.org/engines.html#engine-setup).  If we're concerned about the optional nature of this then I think It could be done as part of the persona install process.  I think devs would expect to see all of the migrations that led to the current schema in the db/migrations directory so keeping them in the gem is not transparent.  If persona is removed from the gemfile, then the migrations disappear from the migrations path but they still stay in the DB unless they were reversed.  We could chat about this synchronously if it would be helpful to resolve it.

orangewolf: I can understand that.  I've basically been using this pattern - https://content.pivotal.io/blog/leave-your-migrations-in-your-rails-engines since 2011 or so, but I've expanded the code to handle both cases. That makes it nice for everyone as they can either include them or not as they see fit. I'll add the migrations back in on the Avalon side now and make sure it still runs.

orangewolf:
giving it another try w/ the updated gem - if the specs pass I'll resquash

orangewolf:
@cjcolvar this has been updated so you can have your migrations and eat them too (or something like that)

orangewolf:
@cjcolvar this passed then I squashed it, so I think the pipeline just needs a re-run?

orangewolf:
> Thanks, the changes look good. Only question I still have is that I'm seeing samvera-persona 0.1.3 only in rubygems and not in github. Could you please push the changes that went into this release? I'd like to see what you did to accommodate both approaches.

No problem.  I thought I was, but it was going to the old remote, not the samvera-labs one. Basically we just check for the existence of the persona migrations in your app and do not load the migration path if it exists. That way if you've installed them locally you use them, if not you do not.

https://github.com/samvera-labs/samvera-persona/blob/master/lib/samvera-persona/engine.rb#L15

cjcolvar:
@orangewolf Heads up...we just ran into this issue when our deploy build failed because it uses ruby 2.4.
https://github.com/samvera-labs/samvera-persona/issues/1

orangewolf:
0.1.4 is out with your proposed fix

PR #3475, opened 2019-07-26
Fixes #3474 files not cleaned up after extract_frame

### LINKED ISSUE #3474: .ts and .jpeg files left behind by MasterFile extract_still
When creating thumbnail/poster from the stream

The ts files are only removed when /tmp is cleaned by the OS
The jpeg files are supposed to be garbage collected but when thousands of files are being ingested, they could clog up the disk.

Solution: *always* `File.unlink` any temporary files.

phuongdh:
Co-authored-by: Brian Keese <bkeese@indiana.edu>

Fixes #3474

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -572,7 +572,10 @@ class MasterFile < ActiveFedora::Base
         secure_url = SecurityHandler.secure_url(playlist_url, target: self.id)
         playlist = Avalon::M3U8Reader.read(secure_url)
         details = playlist.at(options[:offset])
-        target = File.join(Dir.tmpdir,File.basename(details[:location]))
+
+        # Fixes https://github.com/avalonmediasystem/avalon/issues/3474
+        target_location = File.basename(details[:location]).split('?')[0]
+        target = File.join(Dir.tmpdir, target_location)
         File.open(target,'wb') { |f| open(details[:location]) { |io| f.write(io.read) } }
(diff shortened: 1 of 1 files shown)


app/models/master_file.rb
```diff
@@ -640,7 +643,9 @@ def get_ffmpeg_frame_data frame_source, new_width, new_height
         end
         data
       ensure
-        File.unlink(file_source) unless file_source =~ %r(https?://)
+        File.unlink file_source unless file_source =~ %r(https?://)
+        File.unlink frame_source[:source] unless frame_source[:source] =~ %r(https?://)
```
cjcolvar: Should this be `unless frame_source[:master] or frame_source[:source] =~ %r(https?://)` so it doesn't try to unlink a master file that it didn't create?

app/models/master_file.rb
```diff
@@ -640,7 +643,9 @@ def get_ffmpeg_frame_data frame_source, new_width, new_height
...
+        File.unlink jpeg
```
cjcolvar: Odd that this is necessary and that it isn't automatically being cleaned up.

phuongdh: It can take many days to clean up, could fill up the drive when thousands of items are being ingested.

app/models/master_file.rb
phuongdh: Good catch, fixed.

PR #3476, opened 2019-07-29
Protect workaround from undefined errors.

cjcolvar:
This is tested as working in Chrome and Safari and android.


cjcolvar:
I'm not sure if the extra guard is really necessary but it doesn't hurt and that was the final form that we tested on staging.

cjcolvar:
I forgot to squash these when merging but I did it in develop post merge.

PR #3477, opened 2019-07-29
Fix mediaelement.js rounding off errors in playlist items and markers

Dananji:


cjcolvar:
@Dananji I was writing some tests for `#pretty_time` and had a couple questions:
1. It returns time formatted like `00:00:00.00` but since we work in milliseconds should this be `00:00:00.000` so we have the full precision?
2. If not, should we round to the nearest second?  (Currently it is truncating.)
I'm going to go ahead and push my commit since it adds some error handling and a fix.  Feel free to rework my changes.

Dananji:
@cjcolvar Yes you are correct, we should go with the `00:00:00.000` format instead of `00:00:00.00` for full precision. Rounding off to the closest second was problem before. I think this looks good :+1:

joncameron:
With Chris's latest commit to fix the decimal count this looks good! Thumbs up, thanks @bkeese for noticing this.

PR #3478, opened 2019-07-30
Fix player time code update from end of a minute to next

Dananji:
Fix for the time code in the player getting updated from `01:59 -> 01:60` instead of `01:59 -> 02:00` when playing an item.

### DIFF
--- a/vendor/assets/javascripts/mediaelement/mediaelement-and-player.js
+++ b/vendor/assets/javascripts/mediaelement/mediaelement-and-player.js
@@ -8068,12 +8068,14 @@ function secondsToTimeCode(time) {
 			seconds = (f / timeBase % 60).toFixed(secondsDecimalLength);
 		}
 	} else {
-		hours = Math.floor(time / 3600) % 24;
-		minutes = Math.floor(time / 60) % 60;
+		var sec_num = time.toFixed(secondsDecimalLength);
+		hours   = Math.floor(sec_num / 3600) % 24;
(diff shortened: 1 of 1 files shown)


vendor/assets/javascripts/mediaelement/mediaelement-and-player.js
```diff
@@ -8068,12 +8068,14 @@ function secondsToTimeCode(time) {
 			seconds = (f / timeBase % 60).toFixed(secondsDecimalLength);
 		}
 	} else {
-		hours = Math.floor(time / 3600) % 24;
-		minutes = Math.floor(time / 60) % 60;
+		var sec_num = time.toFixed(secondsDecimalLength);
+		hours   = Math.floor(sec_num / 3600);
```
cjcolvar: Was the `% 24` meant to be removed from this line?

Dananji: I missed it. Thanks!

Dananji:
@cjcolvar Can you re-run the circleci? I think it failed due a connection failure and I don't have permission to restart it.
Thanks!

phuongdh:
I tried the latest ME.js and it has fixed this issue: https://codepen.io/phuongdh/pen/ZgKgVK

We should consider upgrading our old (who-knows-which) version from 2017 to avoid adding more custom code that will inevitably make it harder to upgrade in the future.

PR #3485, opened 2019-08-02
Migration report works in Rails 5.2 and returns 404 when full migration report not found.

### LINKED ISSUE #3452: Migration report needs to be updated for rails 5.2
@jlhardes was trying to access the migration report in IU's production avalon and running into 500s dealing with strong parameters: `ActionView::Template::Error: unable to convert unpermitted parameters to hash`

https://app.honeybadger.io/projects/54117/faults/51973920

cjcolvar:
Fixes #3452

### DIFF
--- a/app/controllers/migration_status_controller.rb
+++ b/app/controllers/migration_status_controller.rb
@@ -52,11 +52,13 @@ class MigrationStatusController < ApplicationController
   end
 
   def report
-    filename = "#{params[:id].sub(/:/,'_')}.json"
-    # Sanitize filename for security
-    bar_chars = ['/', '\\']
-    filename.gsub!(bar_chars,'_')
-    send_file File.join(Rails.root, 'migration_report', filename), type: 'application/json', disposition: 'inline'
+    requested_filename = "#{params[:id].sub(/:/, '_')}.json"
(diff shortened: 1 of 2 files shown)


app/controllers/migration_status_controller.rb
```diff
@@ -52,11 +52,13 @@ def detail
   end
 
   def report
-    filename = "#{params[:id].sub(/:/,'_')}.json"
-    # Sanitize filename for security
-    bar_chars = ['/', '\\']
```
bkeese: We don't need to care about these anymore?

cjcolvar: I don't think.  I'm pretty sure forward or backslashes would never be in the `id` param.
I manually tested on mco-staging and I was only seeing links with plain old ids e.g. `avalon:12345`.

bkeese: Was the concern that someone malicious could pass a relative path to get the contents of their target?

cjcolvar: That's probably what it was, but now since I switched it to whitelisting of children of the migration directory I don't think it should be a concern.

Issue #3484, opened 2019-08-02
persona punchlist


davidschober:
@orangewolf I moved this to a new issue for easier tracking for the team. 

The Avalon team got together, and I realized there were a few parts I didn't understand about how persona was working under the hood. Hopefully this is a punchlist we can work from:

- [x] The identity provider should be listed in the "manage users" table. There's a column for it, but it is not being updated.
- [x] Change / Edit should be greyed out, not visible if the user is not a local user (devise). We discussed this as a solution to managers trying to edit passwords/reset passwords when they can't.
- [x]  Remove admin from the footer. My notes had us settling on documentation for those setting up Avalon. 
- [x] @cjcolvar was investigating whether there's an existing screen to use instead of the screen that comes up when you click "admin". At a minimum LTI needs to be removed. 
- [x] LTI Should be removed from screen as it's not a login method. 
- [x] Add some spacing around the "forgot password" and "sign up"
- [x] Failed login for email/password goes back to /users/sign_in with the provider choice buttons revealed and the login form hidden.
- [x] Add Hard Deletes!

davidschober:
relates to #3451

cjcolvar:
Things I'm seeing now on spruce:
1. Failed login for email/password goes back to /users/sign_in with the provider choice buttons revealed and the login form hidden.
2. Back to Admin button doesn't work and presents the error: "You are not authorized to manage users"
3. I'm still a bit confused by soft deletes and wonder if we should use hard deletes for now to avoid errors and confusion.  I also can't seem to delete a user even when using `#really_destroy!` in the rails console.

orangewolf:
2. sounds like a bug, I'll look at it.

3. is.a bug
 https://cl.ly/e5a73db50431/Screen%20Shot%202019-08-14%20at%208.57.15%20AM.png
`really_delete` works as it should and `destroy_without_paranoia`  works, but something is conflicting on `really_destroy!`

cjcolvar:
@orangewolf Thanks for fixing `really_destroy!`.  I still would like to talk through soft deletes with the POs since I think it is confusing from an end user perspective without more functionality that isn't currently there yet.

cjcolvar:
@orangewolf So I did a little exploration and #2 is due to the `auth` `before_action` causing the post to `stop_impersonating` throw a `AccessDenied` error.  Skipping that `before_action` in a monkey patch initializer makes things work as expected.
```
Samvera::Persona::UsersController.class_eval do
[...]

PR #3489, opened 2019-08-06
assing providers to users when they are created for SSO

orangewolf:
This allows samvera-persona to display the SSO provider and to limit editing. Ref #3484 3484

### DIFF
--- a/db/migrate/20190806193607_fill_in_providers.rb
+++ b/db/migrate/20190806193607_fill_in_providers.rb
@@ -1,0 +1,14 @@
+class FillInProviders < ActiveRecord::Migration[5.2]
+  def change
+    User.where('provider IS NULL').find_each do |user|
+      identity = Identity.find_by(email: user.email) if user.email
+      if identity
+        user.update_attribute(:provider, 'identity')
+      elsif user.uid.present?
+        user.update_attribute(:provider, Avalon::Authentication::VisibleProviders.first[:provider])
+      else
+        user.update_attribute(:provider, 'local')
+      end
+    end
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,10 +76,10 @@ class User < ActiveRecord::Base
     user
   end
 
-  def self.find_or_create_by_username_or_email(username, email)
+  def self.find_or_create_by_username_or_email(username, email, provider = 'local')
     find_and_verify_by_username(username) ||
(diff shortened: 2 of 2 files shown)


db/migrate/20190806193607_fill_in_providers.rb
```diff
@@ -0,0 +1,14 @@
+class FillInProviders < ActiveRecord::Migration[5.2]
+  def change
+    User.where('provider IS NULL').find_each do |user|
+      identity = Identity.find_by(email: user.email) if user.email
+      if identity
+        user.update_attribute(:provider, 'identity')
+      elsif user.uid.present?
```
cjcolvar: I wonder if we ever set `uid` since we never set `provider`.  I'm guessing not but having this `elsif` block is probably good to have even if it might never trigger.

cjcolvar: But if that's the case then do you think we should add `uid` just like how `provider` is being set on users now?

orangewolf: in the case of Identity, uid would be email. This _might_ protect against someone who changes their email getting disconnected from the Identity object, but I think 1) that's not supported and 2) that's pretty out of scope since fixing the dev only Identity provider isn't really the point here.

I check uid not for the Identity case, but for other providers.  SAML or Shibboleth for example would need UID set. We don't want to guess local on those. TBH almost all providers would need to set UID other than the Identity case, making this a nice safety feature for SSO having implementations.

cjcolvar: I don't think we have ever set `uid` even for SSO omniauth providers.  At least we don't here at IU with a CAS provider and if I'm reading our user model code we only set username and email (and now provider with this PR).

orangewolf: I think you're making the case that UID should not be changed then and the code is ok as is? or am I misunderstanding something?

cjcolvar: ~I guess I'm suggesting that we make `VisibleProviders.first[:provider]` be the `else` clause instead of `local`.  Since users don't have `uid`s then line 8 will never be executed and I think the configured provider is what users will expect more than `local`. Hmmm...but there could be incorrect entries then if users started with LTI...so maybe `local` makes the most sense.~  So coming full circle, I'm fine with how this is written now.

PR #3490, opened 2019-08-06
hide lti from sso screen, remove admin link from footer

orangewolf:
- Hides all hidden providers
- Removes admin link in footer

Ref #3484


orangewolf:
looks like the spec needs to be re-run - the failure is a timeout on the feature specs that I'm fairly sure is unrelated to this code.

PR #3491, opened 2019-08-06
style forgot password and session#new

orangewolf:
Ref #3484

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -22,17 +22,19 @@
       </div>
 
       <% if devise_mapping.rememberable? -%>
-        <div class="field">
+        <div class="form-group">
           <%= f.check_box :remember_me %>
         </div>
       <% end -%>
 
-      <div class="actions">
+      <div class="actions form-group">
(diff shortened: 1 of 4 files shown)


app/views/devise/sessions/new.html.erb
```diff
@@ -22,17 +22,19 @@
...
-      <div class="actions">
+      <div class="actions form-group">
         <%= f.submit "Connect", class: 'btn btn-primary' %>
       </div>
     <% end %>
-    <%= render "devise/shared/links" %>
-    <p align='center'><a href='javascript:;' data-trigger='show-email'>Cancel</a></p>
+    <p align='center' class='devise-links'>
+      <%= render "devise/shared/links" %>
+      <span class='devise-link'><a href='javascript:;' data-trigger='show-email'>Cancel</a></span>
```
cjcolvar: I see the `devise-links` class above on line 34 and `devise-link` here on line 36 but I only see `devise-links` defined in `admin.scss`.

orangewolf: good point - I meant to scope them in the CSS so that we didn't leave just any span in that partial to become targeted in the future. I've updated the css

PR #3504, opened 2019-08-14
Fix really_destroy!

orangewolf:
The destroy method override was blocking paranoia destroy from working. This moves the bookmark deletion to a before_destroy and therefore no longer overrides destroy.


orangewolf:
ref #3484

PR #3509, opened 2019-08-15
Clean up working_file_path even if MasterFile is gone

phuongdh:


app/jobs/cleanup_working_file_job.rb
```diff
@@ -13,13 +13,14 @@
...
-    masterfile.working_file_path.map do |path|
+  def perform(masterfile_id, working_file_path)
+    working_file_path.map do |path|
       parent_directory = File.dirname(path)
       File.delete(path) if File.exist?(path)
       Dir.delete(parent_directory) if Dir.exist?(parent_directory)
     end
+
+    masterfile = MasterFile.find(masterfile_id)
     masterfile.working_file_path = nil
```
bkeese: Do we care if the job still fails here if the mf is gone?

phuongdh: I think it should still raise some error in case the master file wasn't intentionally deleted.

PR #3523, opened 2019-08-19
Rough draft of collection index and landing page

cjcolvar:
This PR adds routes and controller actions for the end-user focused collection index and view pages.  Gated discovery is applied through media objects: if you can discover at least one item in a collection then you can see the collection.  Conversely though if you can't discover at least one item in a collection then it doesn't show up in the index nor are you allowed to see the landing page.  Alternatively (or in addition) we could give collections themselves visibility and special exceptions which would explicitly allow the ability to discover/read a collection.

### DIFF
--- a/.solr_wrapper
+++ b/.solr_wrapper
@@ -1,5 +1,5 @@
 # Place any default configuration for solr_wrapper here
-version: 7.1.0
+version: 7.7.2
 # port: 8983
 instance_dir: tmp/solr-development
 collection:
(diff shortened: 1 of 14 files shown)


Dananji:
@cjcolvar When I click on `Collections` tab, I get this error;

collection_index_error

Is there anything I need to do?

cjcolvar:
@Dananji Try restarting your avalon docker container and see if that fixes it.

Dananji:
Thanks @cjcolvar ! It worked.
Then this happened;

nil_error

When there is at least one collection without a description, this error occurs.

And when there are descriptions for all collections, it looks as follows;

collection_index

cjcolvar:
Good catch.  I'll take care of that now.

Issue #3444, opened 2019-07-02
LTI Link for Timelines Not Working


joncameron:
### Description
When using LTI links for Timelines in Avalon within Canvas, Avalon returns a 500.

### Environment
IU Canvas implementation, Firefox 67 on macOS 10.13.5.

### Honeybadger Report
[54117-95e81045-3497-4d0d-80f3-d3f0c4d5ea2c.jsonl.gz](https://github.com/avalonmediasystem/avalon/files/3350900/54117-95e81045-3497-4d0d-80f3-d3f0c4d5ea2c.jsonl.gz)

cjcolvar:
Since we don't have a test LMS server, testing this will have to wait until IU's canvas has been updated to whitelist IU's staging server.

PR #3531, opened 2019-08-23
Fix LTI share links

cjcolvar:
There was an ordering issue after the previous fix so lti share links were being overwritten by the general lti redirect to search results.


cjcolvar:
Prior work in #3444

PR #3540, opened 2019-08-28
Change CSS for hover event for nav links

Dananji:


adamjarling:
@Dananji @cjcolvar  Sorry I was late to the game on checking this one out, but I'm sure it's great and much needed improvement.

Just a thought, for checking stuff out like this that's more visual in nature, one thing they used to like to do in Hyrax PRs for UI-related work (when I was on the working group), was include screen shots in PRs of "before" / "after".    Could be handy for future reference, end of sprint retros, or letting another developer know what they should be seeing if they pull down the branch locally to run it.

Just a thought...

mcwhitaker:
I like that suggestion, Adam. It would also help non-developers to know better what the PR is about. Thank you for the suggestion!

PR #3544, opened 2019-08-29
Move create collection button to top

Dananji:


cjcolvar:
@Dananji Could you please add before/after pictures?

cjcolvar:
Don't worry about the failed test.  It isn't related to this PR and I don't want to restart it.

Dananji:
Sorry I forgot that.

Before:

before_create_coll

After:
after_create_coll

mcwhitaker:
@joncameron @Dananji Wouldn't it be better if the button were placed more to the left? It looks too far off to the right. People read from left to right and our attention starts on the left because of this.

cjcolvar:
I think it looks better on spruce than in the screenshots above.
https://spruce.dlib.indiana.edu/admin/collections

mcwhitaker:
It looks far off to the left to me... @adamjarling, what do you think? And @Dananji?

adamjarling:
@mcwhitaker I think consistency is key to buttons and button placement.  So whatever pattern is established for the "primary" button, or button groups in the existing application should probably be followed.

What if we looked at Matt's new designs and see how he handles form buttons, and use that as our guide moving forward?

mcwhitaker:
@adamjarling @Dananji I don't think Avalon is consistent about this. This is why Dananji had to move this particular button to begin with. 

If one of you can glean a pattern from Matt's page designs, we can go with that.

adamjarling:
This may be a good time to create a story or epic for an Avalon Style Guide eyeing Avalon 7?  It could address:

1. Bootstrap 3 or 4 migration plan.
2. Reinforce that Bootstrap UI is the default "style guide" (and any deviances from it in the app should be corrected).
3. Show common UI patterns, such as button placement, form labels, form input widths, container widths (fluid full-width, constrained), modals, etc.

mcwhitaker:
@adamjarling Are you talking about Avalon on Hyrax? Avalon 7 is now an upgrade of Avalon 6.5 based on the Avalon 6 code base. 
I am asking because I am not sure there is time for many changes to style for Avalon 7, scheduled for Thanksgiving.

Dananji:
@cjcolvar This is the current look, when you don't have any collections.

no_collections

PR #3552, opened 2019-08-30
Add class to collections nav link

Dananji:
For the CSS to show the active tab to work, `class='active'` need to be present in the `<li>` tag. This PR adds this class to the newly added `Collections` tab.

NOTE: The current path given for the `Manage Users` tab doesn't work as well. And could not figure out the controller path for that. @cjcolvar If you can please check it out when you have time, that'd be great! Thanks.


cjcolvar:
@Dananji Looks like it should be `samvera/persona/users` for `Manage Users`.  Do you mind adding that to your PR and then I'll review it?

PR #3557, opened 2019-09-03
Invite was broken by 33a932c15, this fixes it

orangewolf:
Related to #3484


cjcolvar:
@orangewolf Good catch!  Do you mind taking care of the 4 code climate styling issues?

PR #3558, opened 2019-09-03
disable soft delete for persona

orangewolf:
Ref #3484


cjcolvar:
I'd like to see how you did this in persona but I can't find the code.  Could you please push the latest persona code to samvera-labs github?

orangewolf:
@cjcolvar yes and I removed the old github remote so that I stop pushing to the wrong one

PR #3567, opened 2019-09-06
Allow configuring of bib retrievers per bib id type

cjcolvar:
For a local need, but this has come up in the past as a potential feature.


cjcolvar:
I think we'll just need to add some notes to the migration wiki page.

Issue #2874, opened 2018-03-16
[Android] Quality Selector Popup Not Disappearing Upon Selection


joncameron:
### Description
The popover doesn't disappear when clicking a quality. It only goes away when a click somewhere else happens.

davidschober:
Need to retest. This is old @SumithBaddam can you retest? If it's no longer an issue, let us know and we'll close.

bkeese:
NB: There is on open PR that addresses this.

davidschober:
Great! Do you know which one? Can we attach it to this issue and move the issue to Review?

PR #3609, opened 2019-09-20
Fix rails highlight in player

Dananji:
Before: 
ezgif com-video-to-gif (2)

After:
ezgif com-video-to-gif (3)

Along with the fix for the connected issue, this PR contains a fix for the following issue I came across while working on this.

Description: When navigating only using keyboard the quality selector in the media player does not disappear when the element is out of focus as seen in below;

ezgif com-video-to-gif (1)

After the fix:
ezgif com-video-to-gif


joncameron:
Great examples, thanks @Dananji!

bkeese:
Is it #2874 that this fixes? (Quality widget not disappearing)

PR #3638, opened 2019-09-27
Add cropper to collection poster upload.

### LINKED ISSUE #3627: Collections - Normalize Poster Image Upon Upload
### Description
Normalize the poster image so it fits into a specific size. This should happen when a user uploads a poster image for a collection.

### Done Looks Like
- Libraries to handle this are investigated
- Uploaded images are normalized 

### Notes
- Current poster images are 700 w/l max (anything below is not upscaled)
- anything below that won’t get upscaled

bkeese:
fixes #3627


app/views/admin/collections/show.html.erb
```diff
@@ -99,10 +124,112 @@ Unless required by applicable law or agreed to in writing, software distributed
 
 <%= render "form", modal_title: "Edit Collection Infomation" %>
 
+<% content_for :page_styles do %>
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/cropperjs/1.5.5/cropper.min.css" integrity="sha256-mS+0WtWHI0Mw4UuIK5xU+b+pB/vfWZiEHV4sJWx7az8=" crossorigin="anonymous" />
+<% end %>
+
 <% content_for :page_scripts do %>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/cropperjs/1.5.5/cropper.min.js" integrity="sha256-vshROrHusjRqmlrpFKS7oAAd4YV7Stzn6DlDmtIBG/4=" crossorigin="anonymous"></script>
```
cjcolvar: Is it better to use the CDN here or manage this via yarn?

adamjarling: I'd vote for `yarn` to keep the dependency management consistent when looking for stuff.

app/views/admin/collections/show.html.erb
```diff
@@ -99,10 +124,112 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+              $alert.show().addClass('alert-warning').text('Upload error');
+            },
+            complete: function () {
+              $progress.hide();
+            },
+          });
+        });
+      }
+    });
   });
```
adamjarling: Would it be a pain to add this to its own `.js` file instead of inlining?

app/views/admin/collections/show.html.erb
bkeese: OK. How would that be done?

adamjarling: Yea, guess depends how Avalon is wired up.  In the project root in a terminal window you could try `yard add cropperjs`.  Then start up the docker container?

Here's the package:
https://www.npmjs.com/package/cropperjs

bkeese: I've added cropperjs to the yarn package, but I don't know how to import it when I need in the app. What is the path that it ends up at?

adamjarling: @bkeese Did you get this figured out?

app/views/admin/collections/show.html.erb
```diff
@@ -30,21 +50,26 @@ Unless required by applicable law or agreed to in writing, software distributed
...
       <p><strong>Poster:</strong></p>
-      <p><%= @collection.poster.original_name if @collection.poster.present? %></p>
-      <% if @collection.poster.present? %>
+      <p id="poster_original_name"><%= @collection.poster.original_name if @collection.poster.present? %></p>
       <div class="well-vertical-spacer">
-        <%= image_tag(poster_admin_collection_path(@collection), class: 'img-responsive') %>
+        <% if @collection.poster.present? %>
+          <%= image_tag(poster_admin_collection_path(@collection), class: 'img-responsive', id: 'poster-image') %>
+        <% else %>
+          <img src="" class="img-responsive" id="poster-image">
```
adamjarling: Does this need to be in there with an empty `src` tag?

app/views/admin/collections/show.html.erb
```diff
@@ -30,21 +50,26 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-        <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" />
-        <% else %>
+        <%= form.file_field :poster, id: 'poster_input', class: "filedata", style: "height:0px;width:0px;" %>
         <input type="button" class="btn btn-primary btn-xs" onclick="$('.filedata').click();" value="Upload Poster" />
+        <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" />
         <% end %>
-        <% end %>
+      </div>
+      <div class="well-vertical-spacer">
+        <div id='cropper-alert' class="alert" role="alert" style="display: none; padding: 5px; margin-top: 5px"></div>
```
adamjarling: I'd give this element an extra class noting what type of alert function its providing.  `alert-info`, `alert-warning`, `alert-danger`, etc...  https://getbootstrap.com/docs/3.3/components/#alerts

app/views/admin/collections/show.html.erb
bkeese: Yeah. Thanks.

app/views/admin/collections/show.html.erb
bkeese: Yeah, the cropper js requires this and adds a src value.

app/views/admin/collections/show.html.erb
bkeese: The cropper js adds the appropriate alert type class depending on error state it generates..

adamjarling: Ah, nice.

PR #3640, opened 2019-10-01
Use active_encode master and default to using Ffmpeg adapter

### LINKED ISSUE #3574: Upgrade to Latest active_encode Version
### Description
To replace Matterhorn, the active_encode version used needs to be updated to a current version and the adapter used needs to be changed.

### Done Looks Like
- The active_encode version in Avalon is upgraded to the latest version needed
- The default adapter used is changed from the Matterhorn adapter to the local FFmpeg adapter 
- This is working outside of Docker (on Spruce)

### Notes
- The Elastic Transcoder integration will break once this happens (separate issue)
- The 'skip transcoding' feature will break once this happens (separate issue)
- This will break Docker :( (separate issue)

cjcolvar:
Configure Ffmpeg adapter to use streaming server directory.
With this Avalon no longer attempts to delete derivative files when the Derivative object is destroyed.
Future work includes making the ffmpeg profiles configurable (and the other stories in the backlog).

After this is merged, we'll need to reconfigure Spruce slightly and test there then we can close #3574.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,9 +33,6 @@ jobs:
       parallelism:
         type: integer
         default: 4
-      bundler_version:
-        type: string
-        default: "2.0.1"
 
     working_directory: /home/app/avalon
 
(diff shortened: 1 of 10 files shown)


cjcolvar:
@phuongdh I rolled back the changes in Gemfile.lock and slimed it down to just active_encode.  We can deal with the nokogiri update in a separate issue.

cjcolvar:
Well that didn't work either.  I'm guessing that it has been working up to this point with a cached bundle and now that we have some changes to the Gemfile it is rebuilding the gemset including rebuilding the nokogiri native extensions even though it is installing 1.10.3.

PR #3649, opened 2019-10-04
Default to WatchedEncode which uses ActiveEncode mixins and replaces …

cjcolvar:
…ActiveEncodeJob


cjcolvar:
Closing because #3651 includes all of these commits.

PR #3680, opened 2019-10-17
Use presenter consistently

bkeese:
And use persisted values directly.


app/presenters/encode_presenter.rb
```diff
@@ -52,22 +52,30 @@ def title
     @encode_record.title.split('/').last
   end
 
+  def display_title
+    @encode_record.display_title
+  end
+
   def master_file_id
-    @create_options['master_file_id']
+    @encode_record['master_file_id']
```
cjcolvar: Since the `master_file_id` is a field on the encode_record, do you mind updating the `after_completed` callback in `WatchedEncode` to use it instead of having to parse the `create_options`? https://github.com/avalonmediasystem/avalon/blob/003fc62c096b1d07f50a509c54cd25dcdaca7886/app/models/watched_encode.rb#L31-L32

PR #3684, opened 2019-10-25
Encode dashbord ui

Dananji:


cjcolvar:
I think the consensus yesterday was to stick with the datatables.js api (or at least something close to it) for now.  So maybe a path forward is to resurrect the old controller code and make the components work with it.

config/routes.rb
```diff
@@ -216,6 +212,9 @@
   end
   get '/jobs', to: redirect('/')
 
+  # Catchall route matching any routes coming from React, EncodeDashboard.js
+  match '*path', to: 'encode_records#index', via: :all
```
cjcolvar: Can we make this matcher more specific?

Dananji: I think React routes should work with normal route configurations. I put this just to catch all the routes that aren't matched in the routes above.

joncameron:
We're going to continue to use DataTables, so this will go unmerged but remain on a branch for possible future use.

cjcolvar:
I'm going to close this for now, but leave the branch around as Jon stated.

PR #3697, opened 2019-11-05
Include some further Sass modular organization

adamjarling:
I think we consider this a work in progress, as its nearly impossible to manually determine (for generically named classes), which styles are used where in the application.  There also appear to be no clear Rails solutions for identifying unused CSS that I found in a quick search.  

So, I'd suggest as developers interact with parts of the application and notice certain styles might be used in given part of the app, they move those styles into an appropriate module folder.   Over the course of time it will clean itself up.  

At some point it would be good to do a deep-cleaning of unused styles, but some risk comes with that and it might take a good chunk of time.

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -17,12 +17,84 @@
 @import 'branding';
 @import 'bootstrap/variables';
 @import 'avalon/mixins';
-@import 'avalon/header';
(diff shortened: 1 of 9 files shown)


adamjarling:
@cjcolvar Sorry about that...I just updated my local environment to ignore the file w/o adding to `.gitignore`.  And it should be re-pushed now.

PR #3701, opened 2019-11-06
Collections index page updates to use Cards

adamjarling:
Some general cleanup, minor component re-organization, and now looks like the following:

image

image

image


adamjarling:
Hey @Dananji Hmm, what browser are you using by chance?  Actually I'll run it through Browserstack and see if I can find it and fix it.  Good catch.

adamjarling:
@Dananji This should now be fixed.

Issue #3688, opened 2019-10-28
Transcoding Dashboard UI


joncameron:
### Description
Styling for the transcoding dashboard. The dashboard is now ported from Hyrax but has carried over the default styling. This should be refined so that it fits with our current styling for the application.

### Done Looks Like
- List of styling changes for the transcoding dashboard enumerated

joncameron:
### Notes from Discussion
####  Styling Changes
- alternating row color or gridlines for table
- changing the sort icons (I like the DataTables website triangles, instead of our current up/down arrows)
- links to the encode job page should be easier to understand as hyperlinks (color scheme currently makes it difficult to see)
- show 25 records by default (make this 20 if possible), reverse sorted by default so newest jobs show first
- dynamic progress should be shown
- change the progress bar color? red when failed, etc.

#### Notes
- "job started" string represents the searchable database value, which is why it looks odd

bkeese:
Here is the datatables event that can be used to help setup dynamic updates for progress bars:
https://datatables.net/reference/event/draw

bkeese:
NOTE: Changes to default sorting and per page options will not be immediately visible. This is because we cache previously chosen options. To see the defaults, one needs to open an incognito window or to clear their cache.

joncameron:
This looks amazing! I'm super happy with how this has turned out. My thoughts:

- Possibly change the title of the page to "Encoding Dashboard"—this might be more familiar to users and is the language used in the nav bar. 
- TBD

cjcolvar:
Things are looking better after #3757 but I'm seeing a few things on spruce that could be cleaned up:
- Failed encodes are showing as green instead of red
- In progress show blue but don't turn to green when they move to complete dynamically before a page refresh
- The stripes should probably be removed for completed encodes

PR #3713, opened 2019-11-12
Style and option updates for encode job table.

bkeese:
References #3688 

Addresses style and table default parts of #3688. 
Does not address making progress bars dynamic.

### DIFF
--- a/app/views/encode_records/index.html.erb
+++ b/app/views/encode_records/index.html.erb
@@ -28,6 +28,9 @@ $(document).ready( function () {
     stateSave: true,
     processing: true,
     serverSide: true,
+    order: [[ 6, 'desc' ]],
+    lengthMenu: [10, 20, 50, 100],
+    pageLength: 20,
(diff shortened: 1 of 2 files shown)


app/views/encode_records/index.html.erb
```diff
@@ -28,6 +28,9 @@ $(document).ready( function () {
     stateSave: true,
     processing: true,
     serverSide: true,
+    order: [[ 6, 'desc' ]],
+    lengthMenu: [ [10, 20, 50, -1], [10, 20, 50, "All"] ],
```
cjcolvar: Do we really want an "All" option?  I'm just afraid of the impact on server/DB/client.  Or maybe it won't be that bad because it will probably take a long time to get to 100,000+.

bkeese: Good point. I didn't really consider the implications; I just copied from an example.

PR #3714, opened 2019-11-12
Split Resque to their own containers

### LINKED ISSUE #3580: Docker-compose Works With FFmpeg Adapter in Latest Active_Encode
### Description
Updating our Docker setup to work with the local FFmpeg adapter in active_encode updates to Avalon.

### Done Looks Like
- A new worker container is made which runs the FFmpeg adapter
- The worker listens to the background job queue
- Multiple worker containers are possible 
- Writes to the Avalon/active_encode database 
- Deploys derivatives to streaming server store
- Streaming should work

### Notes
- This issue depends on code changes to avalon (#3578)
- General goal for splitting resources out for transcoding is two-fold: Docker setup which allows you to use multiple containers, and splitting FFmpeg across a separate machine or machines

phuongdh:
Fixes #3580 partially

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -77,7 +77,7 @@ services:
     <<: *redis
 
   avalon: &avalon
-    image: avalonmediasystem/avalon:6.x-slim-dev
+    image: avalonmediasystem/avalon:6.6-slim-dev
     build: .
     command: bash -c "/docker_init.sh && rails server -b 0.0.0.0"
     depends_on:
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -90,53 +90,36 @@ services:
...
-      - EMAIL_NOTIFICATION
-      - EMAIL_SUPPORT
-      - ENCODE_WORK_DIR=/streamfiles
-      - FEDORA_BASE_PATH
       - FEDORA_NAMESPACE=avalon
       - FEDORA_URL=http://fedoraAdmin:fedoraAdmin@fedora:8080/fedora/rest
       - SETTINGS__FFMPEG__PATH=/usr/bin/ffmpeg
-      - MASTER_FILE_PATH
-      - MASTER_FILE_STRATEGY=delete
+      - MASTER_FILE_STRATEGY=none
```
cjcolvar: Does it make sense to default this to `none` or should it get set back to `delete`?

phuongdh: In development I found it easier to debug if masterfiles aren't deleted. These are visible outside of the container and can be easily managed if necessary.

cjcolvar: But this docker-compose.yml is being used for more than just development, right?  If it is then I think we should try to mirror best practices and not what might be best for specific development tasks.  Is this a change that you can just make locally for now?

phuongdh: This docker-compose is just used for development. Actually I'll just get rid of it and use the default settings.yml.

PR #3719, opened 2019-11-13
Dynamic progress

### LINKED ISSUE #3688: Transcoding Dashboard UI

bkeese:
Fixes #3688 

Updates style, defaults and adds polling to encode dashboard.

### DIFF
--- a/app/views/encode_records/index.html.erb
+++ b/app/views/encode_records/index.html.erb
@@ -36,9 +36,10 @@ update_encode_progress = function() {
     function(progress_data) {
       let stop = true;
       $('.encode-progress').each(function(i, bar) {
-        let progress = progress_data[$(bar).data('encode-id')]
-        stop = stop && (progress == 100)
-        $(bar).val(progress);
+        let encode_data = progress_data[$(bar).data('encode-id')]
+        $(bar).closest('tr').find('.encode-status').text(encode_data['status'])
+        $(bar).val(encode_data['progress']);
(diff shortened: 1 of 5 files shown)


app/views/encode_records/index.html.erb
```diff
@@ -23,8 +26,29 @@
...
+update_encode_progress = function() {
+  let ids = $('.encode-progress').map(function(d){ return $(this).data('encode-id') }).get()
+  $.post($('#encode-records').data('progress-url'),
+    {ids},
+    function(progress_data) {
+      let stop = true;
+      $('.encode-progress').each(function(i, bar) {
+        let progress = progress_data[$(bar).data('encode-id')]
+        stop = stop && (progress == 100)
+        $(bar).css('width', progress + "%");
```
cjcolvar: Is this supposed to be setting the progress value instead of css width?

bkeese: Good point. I just copied what we're doing on the item view page, but it isn't using <progress>. Fixed.

cjcolvar:
Hmmm, I think we'll need to expand the progress endpoint to return more information since once the progress hits 100% we also need to know the status, and master file id.

PR #3731, opened 2019-11-15
Run active encode create in a background job

cjcolvar:
The FfmpegAdapter spawns the ffmpeg process as part of the encode create.  In order to run the FfmpegAdapter on a separate machine or container than avalon the create call needs to be run in a background job.

### DIFF
--- a/app/jobs/create_encode_job.rb
+++ b/app/jobs/create_encode_job.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+# Copyright 2011-2019, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


app/jobs/create_encode_job.rb
```diff
@@ -0,0 +1,22 @@
...
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+class CreateEncodeJob < ActiveJob::Base
+  def perform(input, master_file_id)
+    return unless MasterFile.exists? master_file_id
+    master_file = MasterFile.find(master_file_id)
+    return if master_file.workflow_id.present?
```
cjcolvar: Avoid rerunning the create when a workflow has already run for this master file.  But if you want to do that you'll need to nil the workflow_id before calling `master_file.process`.

PR #3736, opened 2019-11-18
Collection landing page

Dananji:


adamjarling:
Looks like this Read More button doesn't quite work as expected.  It probably should be hidden if there's not a lot of text to show/hide?

image

adamjarling:
This is more of an organizational change, but can we rename `collections/index` to `collections/list`?   `index` has a certain JavaScript connotation which might lead future devs to think it means something it doesn't.

image

app/javascript/components/collections/landing/Facets.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { makeStyles } from '@material-ui/core/styles';
```
adamjarling: We should not use this `material-ui` convention (helper function) for applying styles to Avalon components.  `material-ui` is an indirect dependency brought in by Timeliner I'm guessing, but not listed in Avalon's `package.json` file.  If Timeliner goes away, this will throw an error.  

So if we want to keep the styles inline, then should handle something like:

```
const styles = {
   root: {
      width: "100%"
   }
   ...
}

File: app/javascript/components/collections/landing/Facets.js [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+    flex: '0 1 30%'
+  },
+  facetValue: {
+    [theme.breakpoints.down('sm')]: {
+      paddingLeft: '15px'
+    }
+  }
+}));
+
+function FiltersPanel(props) {
```
adamjarling: Let's break out `FiltersPanel` into it's own file and `@import` it into this component.

app/javascript/components/collections/landing/Pagination.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+
+class Pagination extends Component {
+  constructor(props) {
```
adamjarling: Don't need this `constructor` function.  And looks like this component could be a standard functional component, as it doesn't handle any state directly.  Not a big deal, just a little cleaner.

Dananji:
> Looks like this Read More button doesn't quite work as expected. It probably should be hidden if there's not a lot of text to show/hide?
> 
> image

Good catch! :100:

app/javascript/components/collections/landing/Pagination.js
```diff
@@ -0,0 +1,45 @@
...
+    super(props);
+  }
+
+  handleClick = (page, event) => {
+    event.preventDefault();
+    this.props.search.setState({ currentPage: page });
```
adamjarling: I'm not liking calling `setState()` on a parent component from a `prop`....hmmm.  This could probably be better structured in a more conventional React way.   Like calling a callback function passed in as a `prop`, then in the parent component let it call it's own `setState()`.  Otherwise this could be confusing on where a component's state is being updated.

app/javascript/components/collections/landing/Pagination.js
```diff
@@ -0,0 +1,45 @@
...
+                              {this.props.pages.next_page != null ? (
+                                <a href="#" onClick={event => this.handleClick(this.props.pages.next_page, event) }>Next</a>
+                              ) : ( <span>Next</span> )}
+                            </div>;
+    if (this.props.pages.total_count) {
+      return paginationBlock;
+    } else if (this.props.pages.total_count === 0) {
+      return ( <p className="no-search-results">No results matched your search.</p> );
+    } else {
+      return <div></div>;
```
adamjarling: Let's return `null` instead of an empty `div`.

app/javascript/components/Search.js
```diff
@@ -1,107 +1,114 @@
...
+      this.state.query +
+      '&page=' +
+      this.state.currentPage +
+      facetFilters;
+    console.log('Performing search: ' + url);
+    Axios({ url: url }).then(function(response) {
+      console.log(response);
+      component.setState({ searchResult: response.data.response });
+    });
+  }
```
adamjarling: Seems all the string concatenation could be more readable using String template literals?   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

app/javascript/components/collection_view.js
```diff
@@ -3,8 +3,8 @@ import { render } from 'react-dom';
 import Search from './Search';
 
 const props = {
-  baseUrl: "https://spruce.dlib.indiana.edu",
-  collection: "Chris Test"
+  baseUrl: 'https://spruce.dlib.indiana.edu',
+  collection: 'Chris Test'
```
adamjarling: Maybe this shouldn't be hard-coded any more?   "Chris Test"?

app/javascript/components/collections/landing/FacetBadges.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react';
+
+class FacetBadges extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  handleClick = (index, event) => {
+    event.preventDefault();
+    console.log(index);
```
adamjarling: Let's strip these `console.log` lines out.

app/javascript/components/collections/landing/FacetBadges.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react';
+
+class FacetBadges extends Component {
+  constructor(props) {
```
adamjarling: Don't need this

app/javascript/components/collections/landing/FacetBadges.js
```diff
@@ -0,0 +1,41 @@
...
+    this.props.search.setState({ appliedFacets: newAppliedFacets });
+  };
+
+  render() {
+    if (this.props.facets != []) {
+      return (
+        <div className="mb-3 facet-badges">
+          {this.props.facets.map((facet, index) => {
+            if (facet.length === 0) {
+              return <div></div>;
```
adamjarling: Let's refactor this `return` block to get rid of embedded `returns`.  Line 24 and 26.   And with that cleanup, let's lose the empty `div` return at the bottom of the block as well.

app/javascript/components/collections/landing/SearchResults.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import '../Collection.scss';
+import SearchResultsCard from './SearchResultsCard';
+
+const SearchResults = props => {
+  return (
+    <ul className="search-within-search-results">
+      {props.documents.map((doc, index) => (
+        <SearchResultsCard
+          key={index}
```
adamjarling: Same here, wherever possible we should not use `index` as a value for `key`.  If there's some other unique identifier to grab onto...

app/javascript/components/collections/landing/SearchResultsCard.js
```diff
@@ -0,0 +1,78 @@
...
+  let hours = Math.floor(sec_num / 3600);
+  let minutes = Math.floor((sec_num % 3600) / 60);
+  let seconds = sec_num - minutes * 60 - hours * 3600;
+
+  let hourStr = hours < 10 ? `0${hours}` : `${hours}`;
+  let minStr = minutes < 10 ? `0${minutes}` : `${minutes}`;
+  let secStr = seconds.toFixed(0);
+  secStr = seconds < 10 ? `0${secStr}` : `${secStr}`;
+
+  return `${hourStr}:${minStr}:${secStr}`;
```
adamjarling: Maybe this is a good opportunity to pull in the `moment` package to  handle date / time parsing?

Dananji:
@adamjarling What's the purpose of `javascript/components/collection_view.js`, `javascript/components/collection_index.js`, `javascript/components/collection_index.html`, and `javascript/components/collection_view.html` files? 
And these files are referenced in `config/webpack/collection_view.js` and `config/webpack/collection_index.js` files, which are then used in `package.json` to build the components? I don't really understand this part :|
Do we need them to run the collections related component?

app/javascript/components/collections/landing/FacetBadges.js
Dananji: Since this does not handle state, can we convert this to a functional component as well?

adamjarling:
@Dananji That's a good question.  I'm not really sure if that's test code, or actually being used.  @cjcolvar @phuongdh ?

app/javascript/components/collections/landing/FacetBadges.js
adamjarling: Sure thing.  Thinking about keeping this code relevant for the years to come, we should not really be using `class` components, unless absolutely necessary for some reason.   If a component needs to handle state, the `{ useState }` hook is pretty straightforward to wire in.

app/javascript/components/collections/landing/SearchResultsCard.js
Dananji: Since this function works properly I think this is better than pulling in an external package for only this. If we have any other places where we can use it, then I think it's a good idea to pull it in. What do you think?
Maybe we can refactor this to a smaller code block?

cjcolvar:
Those were for easy testing since we no longer had the code split out into it's own repositories.  We could remove them but I don't think they get in the way other than being confusing.  Maybe renaming them to make their purpose explicit would be best?

app/javascript/components/collections/landing/SearchResultsCard.js
adamjarling: @Dananji Agreed.  Let's keep it easy for now and maybe just keep in mind for the future.  Good call.

adamjarling:
@Dananji Everything looks good, but can you check the searching within a Collection functionality?  It seems like its not working half the time.  When the search bar text is cleared, the Collection items also either don't return, or return after a few seconds.  At least for me.  Can you verify this seems smooth to you?

Dananji:
> @Dananji Everything looks good, but can you check the searching within a Collection functionality? It seems like its not working half the time. When the search bar text is cleared, the Collection items also either don't return, or return after a few seconds. At least for me. Can you verify this seems smooth to you?

Yes, this seems to work fine on my end.

Dananji:
> Looks good. I'd say let's get this into `develop`, then make any final adjustments we come across in next few days?

Sure :+1:

PR #3738, opened 2019-11-18
Add search bar to the hamburger menu items

Dananji:


app/views/modules/_avalon_search_form.html.erb
```diff
@@ -20,14 +20,26 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-        <button id="global-search-submit" type="submit" class="btn btn-primary">
+        <button id="global-search-submit" type="submit" class="btn">
           <i class="fa fa-search icon-white"></i>
         </button>
       </span>
     </div>
   </div>
 </form>
+
+<% content_for :page_scripts do %>
```
adamjarling: What happens if a user changes the screen size?  This might not get picked up.  Is there any way to handle this exclusively through CSS?

adamjarling: Also, looks like this code gets doubled up?  If so, means there are 2 DOM elements with the same `id`, which will cause issues and technically is invalid HTML.

image

Dananji: Good catch!! Yes, only the button changes when screen size changes not the input class. Let me see whether there's a way to do this with CSS.
This HTML component is rendered in 2 different places, for mobile and desktop views. And only one is visible at a given time.

cjcolvar: Ah, this is probably the source of the failing capybara tests too:
`rspec ./spec/features/capybara_navigation_spec.rb:73`

cjcolvar:
This looks good to me.  
@adamjarling could you review again please?

Issue #3672, opened 2019-10-17
Waveform  generation causes a zombie process


joncameron:
### Description
The cancel method isn't fully killing the ffmpeg process that has been started, and/or leaves a zombie process. This is (naturally!) not ideal.

### Done Looks Like
- FfmpegAdapter#cancel fully kills the ffmpeg process without leaving any lingering processes

phuongdh:
Couldn't reproduce on Spruce: 
- Ingested a file
- Wait till processing is half way through
- Find running ffmpeg process: there is one
- Delete master file from item
- Find running ffmpeg process: there is none

cjcolvar:
I tested again locally under docker and couldn't replicate this behavior either.

cjcolvar:
Now that we moved the encode job create into a background job we have noticed that this has started happening again.

cjcolvar:
Looks like the zombie is present on spruce. :(

cjcolvar:
If we can't fix this right away then we can provide a workaround of restarting sidekiq often which would clean up the zombies.  Really really not good, but an option.

PR #3739, opened 2019-11-18
Add Minio and Sidekiq support

phuongdh:


cjcolvar:
When upgrading will need to clear out redis: https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#the-threads-are-not-processing-jobs

cjcolvar:
Fixes #3672

spec/models/watched_encode_spec.rb
```diff
@@ -45,12 +50,28 @@
...
+
+    context 'using Minio and ffmpeg adapter' do
+      before do
+        Settings.minio = double("minio", endpoint: "http://minio:9000", public_host: "http://domain:9000")
+        allow(Settings).to receive(:encoding).and_return(double(engine_adapter: "ffmpeg", derivative_bucket: "derivs"))
+        allow(File).to receive(:delete)
+      end
+
+      it 'uploads to Minio' do
+        encode.create!
```
cjcolvar: Are there any expectations for this test?  If it is just that it doesn't raise errors then I think it could be:
```
expect { encode.create! }.not_to raise_error
```

spec/models/watched_encode_spec.rb
```diff
@@ -45,12 +50,28 @@
       # Return a completed encode so the polling job doesn't run forever.
       allow(described_class.engine_adapter).to receive(:create).and_return(running_encode)
       allow(described_class.engine_adapter).to receive(:find).and_return(completed_encode)
+      # allow(master_file).to receive(:update_progress_on_success!)
```
cjcolvar: This can be removed.

phuongdh:
I still have a few more things to do with active_encode, let's update it in a new PR.

PR #3741, opened 2019-11-18
API logins get same abilities as UI logins for all users.

cjcolvar:
In support of local work where a power user is requesting api access to push structural metadata.


joncameron:
We will need to add notes to the upgrade documentation about adding pre-existing api tokens that were being used with the content ingest api to the administrators group so they continue to have full privileges.

PR #3750, opened 2019-11-20
Make 'Manage Content' in main nav-bar, keyboard accessible

Dananji:
I tested this on Chrome, Firefox, and Safari in local development environment.

This is when using `Tab` to navigate and pressing `Enter` key to go to `Manage Groups` page:

ezgif com-video-to-gif (5)

When using `Shift` + `Tab` to go in reverse this works properly in Chrome and Safari. But in Firefox, the dropdown menu closes and opens again before going to the previous item like as seen below (but works as intended):

ezgif com-video-to-gif (6)

This happens unlike in Chrome and Safari, Firefox handles the 2 events that get fired at that particular moment with a slight time gap. 

Following is a screen grab of reverse tabbing working in Chrome;

ezgif com-video-to-gif (7)

### DIFF
--- a/app/assets/javascripts/avalon.js
+++ b/app/assets/javascripts/avalon.js
@@ -107,31 +107,19 @@ $(document).ready(function() {
 
   /* Handle dropdown list for `Manage` in nav-bar */
   // Manage user access with mouse-point
-  $('#manageDropdown').hover(
+  $('#manage-dropdown').hover(
     function() {
-      $('#manageDropdown').addClass('open');
+      $(this).addClass('open');
       $(this).attr('aria-expanded', 'true');
     },
     function() {
-      $('#manageDropdown').removeClass('open');
+      $(this).removeClass('open');
       $(this).attr('aria-expanded', 'false');
     }
   );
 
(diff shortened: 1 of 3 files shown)


Dananji:
> Is all of this extra JS keyboard event handling necessary? I wonder if we changed our markup a little if it might handle most of the cases. I'm looking at this example which makes me think most of it should be possible with HTML changes: http://jsfiddle.net/damphat/WJS9p/.
> Maybe I'm just being naive about it but I would hope that most of these events should be handled by bootstrap and we shouldn't have to add them.

@cjcolvar Yes, I agree. With this we can get rid of the `focus` event, but we still might need the others. With this solution the issue with the Firefox is not visible as much. So let me give this a try.

Dananji:
The second commit is implemented using buttons similar to the example in Chris's comment, this can be further improved (style-wise).

adamjarling:
@cjcolvar Style-wise, I think this looks pretty good.  And follows Bootstrap 3.3's CSS, so not really super excited about altering it.  I'll try to fix the conflict and re-push up this branch.

PR #3757, opened 2019-11-21
Fix progress bar colors and format timecode

Dananji:
I couldn't test the `cancelled` and `running` states locally, because for some reason the new files I ingest doesn't show up in the encoding dashboard.


app/controllers/encode_records_controller.rb
```diff
@@ -75,11 +75,11 @@ def paged_index
         [
           "<span data-encode-id=\"#{encode.id}\" class=\"encode-status\">#{encode_presenter.status}</span>",
           view_context.link_to(encode_presenter.id, Rails.application.routes.url_helpers.encode_record_path(encode)),
-          "<progress value=\"#{encode_presenter.progress}\" max=\"100\" data-encode-id=\"#{encode.id}\" class=\"encode-progress\"></progress>",
+          "<progress value=\"#{encode_presenter.progress}\" max=\"100\" data-encode-id=\"#{encode.id}\" class=\"encode-progress #{encode_presenter.status.downcase}\"></progress>",
           encode_presenter.display_title,
           view_context.link_to(encode_presenter.master_file_id, encode_presenter.master_file_url),
           view_context.link_to(encode_presenter.media_object_id, encode_presenter.media_object_url),
-          encode_presenter.created_at
+          encode_presenter.created_at.utc.strftime('%D %r')
```
bkeese: I have a concern about changing the date format because date searches might not match search results as they appear here.

app/controllers/encode_records_controller.rb
```diff
@@ -95,7 +95,13 @@ def progress
     progress_data = {}
     ::ActiveEncode::EncodeRecord.where(id: params[:ids]).each do |encode|
       presenter = EncodePresenter.new(encode)
-      progress_data[encode.id] = { progress: presenter.progress, status: presenter.status }
+      # Set progress = 200 to show the progress bar in UI when failed
+      progress =  if presenter.status.casecmp("failed")
+                    200
```
bkeese: 200 seems like a weird choice here.  Also, whatever value we set here should  also be set for failed items on line 78 above.

bkeese:
Can we get rid of the file app/assets/stylesheets/active_encode.scss? I don't think it's getting loaded anyway.

app/controllers/encode_records_controller.rb
Dananji: The search works for dates, but only in the `yyyy-mm-dd` format, which is the original format. Shall we change the format to that and keep the date formatting code here?

bkeese: The storage format is determined by something other than us, afaik (ActiveRecord or ActiveFedora, I think). I don't think there is a way for us to change that.

app/controllers/encode_records_controller.rb
Dananji: Yeah.. What I wanted to do was to get rid of progress = 1 when failed, and give a value to progress bar >= 100. Because Firefox does not have 2 different ways to style the `progress-bar` and the filled potion (value). In order to show the `failed` status as a red color bar the `<progress>` element should have a value >= 100. Let me see whether there's a better way to this.

bkeese: I think sending 100 is reasonable. Maybe we should consider setting the actual progress value in the encode record to 100 when status goes to failed? @pdinh @cjcolvar?

app/controllers/encode_records_controller.rb
Dananji: Sorry I didn't mean that we need to change the stored format, but only the format shown in the UI. If we change this line to,
```
encode_presenter.created_at.strftime('%Y-%m-%d  %H:%M:%S')
```
the dates are shown as the same format as storage format. That way when a user enters the date into the search they are not confused. What do you think?

bkeese: That sounds good.

app/controllers/encode_records_controller.rb
```diff
@@ -95,7 +95,8 @@ def progress
     progress_data = {}
     ::ActiveEncode::EncodeRecord.where(id: params[:ids]).each do |encode|
       presenter = EncodePresenter.new(encode)
-      progress_data[encode.id] = { progress: presenter.progress, status: presenter.status }
+      # Set progress = 0.0 when job failed
+      progress_data[encode.id] = { progress: format_progress(presenter.progress), status: presenter.status }
```
bkeese: format_progress(presenter)

app/controllers/encode_records_controller.rb
```diff
@@ -95,7 +95,8 @@ def progress
     progress_data = {}
     ::ActiveEncode::EncodeRecord.where(id: params[:ids]).each do |encode|
       presenter = EncodePresenter.new(encode)
-      progress_data[encode.id] = { progress: presenter.progress, status: presenter.status }
+      # Set progress = 0.0 when job failed
```
cjcolvar: Should this comment be updated to `100.0`?

Issue #3753, opened 2019-11-20
Collection Carousel on Homepage Populates from Configuration File


joncameron:
### Description
The list of collections to be within the collections carousel displays in the collection carousel on the homepage.

### Done Looks Like
- Call to collections.json from the collection carousel React component is updated to to include the carousel parameter

Issue #3752, opened 2019-11-20
Featured Collection Display on Homepage


joncameron:
### Description
On the new Avalon homepage, a featured collection is listed. The featured collection displayed is pulled from a list specified in a configuration file, and a random collection from that list is selected upon load to be displayed.

### Done Looks Like
- On page load, the featured collection is randomly selected from the list within the configuration

Issue #3751, opened 2019-11-20
Configuration File for Avalon Homepage and Endpoint Update


joncameron:
### Description
Configuration will be added to a config file in Avalon w that will control the collections that appear on the new Avalon homepage. This configuration  will specify:
- a list of collection to be under "featured collections"
- a list of collections to include under the collection carousel

The collections.json endpoint will also be updated to return information for just the list of collections under the "collections carousel" listing in the configuration file, via parameter passed.

### Done Looks Like
- Spec for configuration file determined (where does it live?)
- Name and and other factors determined; initial blank info added
- collection.json updated to handle an extra parameter that returns only those collections listed in the configuration file

Issue #3636, opened 2019-09-27
Implement new homepage design


adamjarling:
### Description
We have a new homepage design intended for Avalon on Hyrax, however we could try to implement the design now.  Updating the homepage might give a nice bow on top of recent UI work.

![avalon-homepage-draft-1](https://user-images.githubusercontent.com/3020266/65781857-0de0ad80-e112-11e9-8033-c25d686779cd.png)

### Done Looks Like
- New homepage implemented

### Tasks
- [x] Work on the HTML markup in the views
- [x] Wire up the data to the views (or at least make sure the the view has all the data it needs to make the new design work).
- [x] Consider using any React components?   Would it be easier, or any benefit?

davidschober:
Cool. Let's attack this after SC.

adamjarling:
@davidschober @joncameron Any interest in getting this into the 7 release?

mcwhitaker:
@adamjarling In theory, we are supposed to be feature-complete by the end of next week. Would changing this require time from anyone at IU? If so, I don't think we can manage it.

adamjarling:
@mcwhitaker Ah, understood.  Well, the dynamic content would be the featured collection part.  If the code is not in place to support that on the back-end, then this would require some IU dev time I'm guessing?  @cjcolvar ?   It would be a nice bow on the 7 release if possible to squeeze it in...but don't want to rush anything.

cjcolvar:
I think it might already be done since the collection listing page consumes a json endpoint which should have the same data as would be displayed on the homepage.
https://spruce.dlib.indiana.edu/collections.json

cjcolvar:
I guess the only new part would be "featured"-ness.  How does a collection become "featured" and returning that information in the json endpoint.  Probably would be fairly easy but would a little UX thought.

joncameron:
I'm definitely interested but my concern would be time. Seems like a good backlog buddies discussion.

adamjarling:
Here's where the current UI dev stands visually.  Branch pushed: `3636-new-homepage-design`

![image](https://user-images.githubusercontent.com/3020266/69259843-81d48a80-0b84-11ea-8c95-82daceeaebd9.png)

PR #3767, opened 2019-11-25
3636 new homepage design

cjcolvar:
Completes #3636, #3751, #3752, and #3753.


adamjarling:
Up to you @cjcolvar when you'd like to merge, but with this large a PR might be better sooner than later to thoroughly test it against staging.

PR #3768, opened 2019-11-25
Mobile cleanup related mostly to the Collection pages

adamjarling:
Mostly smaller cleanup, but had to update the Collection Landing page to 1.) make it just cleaner overall, and 2.) to make mobile way easier to update (basically nothing else needed with general page layout cleanup work)

New Collection Landing page:

image

image

image

### DIFF
--- a/app/views/collections/show.html.erb
+++ b/app/views/collections/show.html.erb
@@ -27,6 +27,11 @@ Unless required by applicable law or agreed to in writing, software distributed
   <div class="row">
     <div class="document-thumbnail col-sm-4 col-sm-push-8">
       <%= image_tag @doc_presenter.poster_url %>
+      <div class="filter-collection-button-wrapper">
+        <a href="<%= root_url + '?f[collection_ssim][]=' + @doc_presenter.name %>" class="btn btn-default btn-block">
+          Filter this collection
+        </a>
+      </div>
     </div>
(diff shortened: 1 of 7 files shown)


app/views/collections/show.html.erb
```diff
@@ -27,6 +27,11 @@ Unless required by applicable law or agreed to in writing, software distributed
   <div class="row">
     <div class="document-thumbnail col-sm-4 col-sm-push-8">
       <%= image_tag @doc_presenter.poster_url %>
+      <div class="filter-collection-button-wrapper">
+        <a href="<%= root_url + '?f[collection_ssim][]=' + @doc_presenter.name %>" class="btn btn-default btn-block">
```
cjcolvar: This is going to the root path which is now the new home page.  I think it should be changed to the following which goes to `/catalog`:
```suggestion
        <a href="<%= search_catalog_path('f[collection_ssim][]' => @doc_presenter.name) %>" class="btn btn-default btn-block">
```

adamjarling:
I'm having a hard time figuring out where the "Filter this collection" button makes sense.  In its previous location I kept wanting to press it to do a search, as it was in line with the search Collections text input, but extended out too far.  Which felt like it failed 2 separate UX experiences, for me at least.   Wherever you want it is fine with me.

#### Original
image

#### Alternative placement 1
image

#### Alternative placement 2
image

Issue #3103, opened 2018-11-27
Slow performance loading media objects with comments


cjcolvar:
I didn't experience this but saw it in MCO's log.  This will need a small change to the indexing or handling of this field since it isn't getting found automatically by `speedy_af`.

>Reifying MasterFile because comment called from /srv/avalon/avalon_r6/releases/20181127165933/app/helpers/media_objects_helper.rb:261:in `block in gather_all_comments'

cjcolvar:
Smoke testing found these places where objects are being reified:
```
Reifying MasterFile because operation called from /srv/avalon/avalon_r6/releases/20191127182958/app/controllers/media_objects_controller.rb:357:in `block in show_progress'
Reifying MasterFile because status_code called from /srv/avalon/avalon_r6/releases/20191127182958/app/controllers/media_objects_controller.rb:354:in `block in show_progress'
Reifying MasterFile because comment called from /srv/avalon/avalon_r6/releases/20191127182958/app/helpers/media_objects_helper.rb:251:in `block in gather_all_comments'
Reifying StructuralMetadata because section_title called from /srv/avalon/avalon_r6/releases/20191127182958/app/models/concerns/master_file_behavior.rb:104:in `display_title'
Reifying MasterFile because structuralMetadata called from /srv/avalon/avalon_r6/releases/20191127182958/app/models/concerns/master_file_behavior.rb:104:in `display_title'
Reifying MasterFile because status_code called from /srv/avalon/avalon_r6/releases/20191127182958/app/views/media_objects/_sections.html.erb:16:in `block in _app_views_media_objects__sections_html_erb__2909720380917588155_70303131231540'
```

PR #3773, opened 2019-12-02
Use SpeedyAF proxy and refactor to avoid using fedora for encode progress

cjcolvar:
Part of #3103.


cjcolvar:
Nah, the two frozen string ones might be bogus since I put the magic comment in those files albeit under the license header.  The other two make things more confusing in my opinion.

PR #3783, opened 2019-12-04
Fix playlist layout issues.

bkeese:
Fixes layout of playlist share tabs.

### DIFF
--- a/app/views/playlists/_share.html.erb
+++ b/app/views/playlists/_share.html.erb
@@ -13,7 +13,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
-<div id="share-button">
+<div id="share-button" style="margin: -1.6em 1em 1em 0;">
   <a data-toggle="collapse" href="#share-list" aria-expanded="false" aria-controls="collapseExample">
     <i class="fa fa-share-alt"></i>
     Share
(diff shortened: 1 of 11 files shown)


app/views/playlists/_share.html.erb
```diff
@@ -13,7 +13,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
-<div id="share-button">
+<div id="share-button" style="margin: -1.6em 1em 1em 0;">
```
Dananji: @bkeese I think it would be better if we could move this inline styling to a CSS file.

bkeese: Agreed. Will you fix please?

app/views/playlists/show.html.erb
```diff
@@ -23,9 +23,13 @@ Unless required by applicable law or agreed to in writing, software distributed
...
 <div class="playlist-view-wrapper row">
   <div class="col-sm-8" id="left_column">
+    <div class="page-title-wrapper playlist-title">
+      <%= icon_only_visibility @playlist.visibility %>
+      <h1 class="page-title"><%= @playlist.title %></h1>
+    </div>
+
     <% if first_playable_item.nil? %>
     <%= image_tag "media-placeholder.png", :class => "img-responsive" %>
-    <%= render 'playlist_title' %>
```
cjcolvar: Can we delete `app/views/playlists/_playlist_title.html.erb` then?

Dananji: Yes :+1:

PR #3818, opened 2019-12-09
Animate progress bar in encode dashboard

Dananji:


Dananji:
I don't understand why these tests are failing. I think one of them might be fixed if the build is restarted. If someone can do it, that'd be great! Thanks.

PR #3832, opened 2019-12-10
Fix embedded player wiggle in Safari

Dananji:


Dananji:
> This works?! Wow! Good catch!

This might be working!

In Browserstack before the fix, Safari 11.1's embedded player wiggled but not in Safari's latest version. With this fix, the behavior between the older version and the latest version interchanged. Now in Safari's latest version the player wiggles but not in Safari 11.1 :|

This needs to be tested on a MacOS device to be sure it's working.
@cjcolvar @joncameron When you have time, can you please test this? Thanks!

PR #3840, opened 2019-12-11
Bump dependencies

cjcolvar:
Based on results of snyk and github security notices.


cjcolvar:
I'm rolling back the JS dependency upgrades since I can't figure out how to fix the issue with webpacker compilation failing.

PR #3847, opened 2019-12-12
Only allow administrators to see sidekiq web UI

cjcolvar:


cjcolvar:
This doesn't solve the problem in practice because the sidekiq web UI logs you out of the session when it loads so it is unusable.

PR #3854, opened 2019-12-13
Log an error when waveform generation fails instead of raising an error and failing the job

cjcolvar:
Since waveforms aren't critical, I decided to let the job finish without failure and log the error instead of having the job rerun.  I think that if a user comes across a file without a waveform in the SME then they can raise the issue and someone can troubleshoot it.  Anyways this is the likely result even if the job were to fail and rerun.

The original issue of ffmpeg not returning any waveform data is concerning and still not resolved.


cjcolvar:
The original issue is that the video file doesn't have an audio track.  Brian and I talked about this and decided that it might make more sense to have the SME handle this case gracefully (like the Timeliner prompts but allows you to continue without audio) instead of generating a flat waveform.  Generating a flat waveform might mask cases where there was a bug or ffmpeg was misconfigured which led to no wave audio to generate the waveform from.

PR #3855, opened 2019-12-13
Fix encode dashboard filtering

bkeese:
Fix dashboard filtering for MySQL without breaking Postgres.


bkeese:
The Codeclimate sql injection warning is a false alarm.

PR #3881, opened 2019-12-18
Use omniauth controller's logic for determining post-login redirect path

cjcolvar:


app/controllers/application_controller.rb
```diff
@@ -56,16 +56,35 @@ def rewrite_v4_ids
...
+    end
   end
 
-  def after_sign_in_path_for(resource)
+  def after_sign_in_path_for(_resource)
     if request.env['QUERY_STRING']['login_popup'].present?
       root_path + "self_closing.html"
     else
-      request.env['omniauth.origin'] || stored_location_for(resource) || session[:previous_url] || root_path
+      request.env['omniauth.origin'] || find_redirect_url(nil)
```
phuongdh: Without stored_location_for, logging in from embedded player may not auto-close popup window. I think we could merge now and test this use case on Spruce.

PR #3903, opened 2019-12-23
Improve misleading deletion notice

phuongdh:


cjcolvar:
I think this change needs to be reconsidered.  With it in place if you go to the url for an item that has been deleted (potentially years ago) it reads "Item(s) being deleted." which is confusing.  What is the scenario this change was meant to cover?  I wonder if we can find a different solution for that case which doesn't affect the one above.

PR #3904, opened 2019-12-26
Make selected items page span the entire container

Dananji:
Before:

Screenshot from 2019-12-26 11-07-36
After:

Screenshot from 2019-12-26 11-06-55

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -395,8 +395,12 @@ a[data-trigger='submit'] {
   }
 }
 
-#content {
-  max-width: 870px;
+@media (max-width: $screen-xs-max) {
+  #content {
+    .btn {
+      margin-top: 2px;
+    }
+  }
 }
 
 #resource_description_content {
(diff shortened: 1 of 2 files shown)


bkeese:
This seems like a very broad selector. Are you sure we want this change everywhere it will apply?

Dananji:
In mobile screens the buttons/items will always left-align making the list of buttons/items span to multiple lines. When `margin-left: 2px`, it looks like this;
Screenshot from 2019-12-26 11-20-30

So having the margin on right instead of left will make it look aligned. And in larger screen this makes no difference, because everything is on the same line.

I think we want to make this change everywhere it applies. WDYT?

bkeese:
It's the elimination of max-width on all #co tent that I'm concerned about. I don't remember why that was there specifically, but I'll bet there was a reason, maybe on some other page.

Dananji:
Ohh.. okay. Yes it is used in other places (e.g. Timeliner page, Search results, Media player section) When I tested those scenarios, and found that `max-width: 870px;` is dominated either by other CSS classes or inline styling. Removing `max-width: 870px;`, I couldn't see it affecting the UI to change.

PR #3910, opened 2020-01-06
Test with Ruby 2.7

cjcolvar:
Don't merge this!  This is just for testing purposes.


cjcolvar:
Looks like it works fine.

PR #3917, opened 2020-01-07
Fix time float position in MEJS player

Dananji:
Pop-over time label for the embedded player from mallorn still gets cut as follows;
Screenshot from 2020-01-07 14-47-08

Because the style application in JS for `mejs__time-float` (`bottom` CSS property) from [#3839](https://github.com/avalonmediasystem/avalon/pull/3839) is overridden by the styles in in CSS files. This does not happen in local development environment.

** Need to test this once it is on Spruce.


Dananji:
It works now;
Screenshot from 2020-01-07 16-57-34

PR #3946, opened 2020-01-17
Fix ActionMailer not using SES

### LINKED ISSUE #3944: Demo Box - Password Reset Throws 500
### Description
Password resets are returning the Rails 500 page for existing user accounts (doesn't happen for email addresses not tied to an account).

### To Reproduce
At https://spruce.dlib.indiana.edu/users/password, enter archivist1@example.com as the user and click the "send me password reset instructions" button.

phuongdh:
Tested working on demobox

Fixes #3944

### DIFF
--- a/config/initializers/mailer.rb
+++ b/config/initializers/mailer.rb
@@ -1,4 +1,4 @@
-if Settings.email.mailer.present? && Settings.email.mailer == :aws_sdk
+if Settings&.email&.mailer&.to_sym == :aws_sdk
   require 'aws-sdk-rails'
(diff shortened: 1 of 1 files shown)


cjcolvar:
Is there any documentation changes that need to happen to explain the what the environment variable should look like?

phuongdh:
> Is there any documentation changes that need to happen to explain the what the environment variable should look like?

Good point, I took the opportunity to update our config wiki https://wiki.dlib.indiana.edu/display/VarVideo/Configuration+Files

PR #3954, opened 2020-01-23
Fix and standardize styling for all tables

phuongdh:
# Before
Screenshot from 2020-01-23 16-14-35

# After
Screenshot from 2020-01-23 16-14-49

Screenshot from 2020-01-23 16-16-44

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -882,7 +882,7 @@ h5.panel-title {
     text-align: left;
     position: relative;
     height: 13px;
-    margin: 8px 8px 8px 8px;
+    margin: 4px 0px;
     .completed {
       background-color: $success;
     }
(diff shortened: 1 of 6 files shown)


mcwhitaker:
Looks great!!!!

config/settings.yml
```diff
@@ -62,6 +62,8 @@ bib_retriever:
     retriever_class_require: 'avalon/bib_retriever/sru'
 controlled_vocabulary:
   path: config/controlled_vocabulary.yml
+timeliner:
+  timeliner_url: http://localhost:3000/timeliner
```
Dananji: @phuongdh Is it okay to push settings changes?

phuongdh: Hmm now that I think about it, it should probably live in `config/settings/development.yml` instead

phuongdh: Split out to a different PR

PR #3975, opened 2020-02-13
Use if/elsif instead of ||= with guard clause

cjcolvar:
The previous version needlessly required that all guard clauses run even when a
prior line had matched and initialized mf_title.  Switching to explicit
if/elsif allows for early breaking.  This switch helps speed up the view
page of an item with hundreds of sections by avoiding the third case of
file_location which turns out to be expensive (~0.25s).  

Making this change resulted in `gather_all_comments` for this item running under 2s as opposed to 360s before the change.

### DIFF
--- a/app/models/concerns/master_file_behavior.rb
+++ b/app/models/concerns/master_file_behavior.rb
@@ -101,9 +101,14 @@ module MasterFileBehavior
   end
 
   def display_title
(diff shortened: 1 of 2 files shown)


phuongdh:
https://github.com/avalonmediasystem/avalon/issues/3978

PR #3986, opened 2020-02-14
Don't sort user playlists before using them in paged_index since it does all of the sorting

cjcolvar:
This appears to have been a bug in the bugfix done in https://github.com/avalonmediasystem/avalon/pull/3783/commits/d96f7ba5a43cbca53d8606d9e04ce3d1dd5e3743
The change from that commit to `get_user_playlists` was unnecessary and the change to `get_all_other_playlists` was sufficient.

This fixes VOV-6132 by removing the `order(:title)` to keep it from competing and complicating the order set in `paged_index`.

I tested this on spruce and mco-staging and both worked as expected after this fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ jobs:
        environment:
          - POSTGRES_USER=postgres
          - POSTGRES_DB=avalon
+         - POSTGRES_PASSWORD=password
      - image: ualbertalib/docker-fcrepo4:4.7
        environment:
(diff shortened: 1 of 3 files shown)


cjcolvar:
I just pushed another commit which handles https://app.honeybadger.io/projects/54117/faults/60860157 by using `blank?` instead of `empty?` because playlist comments can be nil.

cjcolvar:
I had to set the POSTGRES_PASSWORD env variable in the circleci config for the postgres container to startup.  There must have been a new version of the postgres docker image that added this requirement.

PR #3993, opened 2020-02-21
New timeliner with time code fix

Dananji:
Changes made in PR [#40](https://github.com/IUBLibTech/timeliner/pull/40) in `IUBLibTech/timeliner` repo

### DIFF
--- a/app/views/timelines/show.html.erb
+++ b/app/views/timelines/show.html.erb
@@ -21,7 +21,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   <script>
     $(document).ready(function () {
       // Push footer to the bottom of the page content
-      $('.footer-container').css('top', $(this).height());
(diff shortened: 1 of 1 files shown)


Dananji:
Due to the huge number of lines being changed in this PR, and the fact that there should be only a couple of lines (these include line changes due to webpacker using absolute paths that includes user names, and some `date-fns` functions included due to the changes made to the import. Still should have less than 1000 line changes) both Chris and I agree we need to do regression testing on spruce for this once it is merged.

mcwhitaker:
@Dananji  Do you mean regression testing on the timeliner or in the whole app?

Dananji:
> @Dananji Do you mean regression testing on the timeliner or in the whole app?

The timeliner. I don't think it might have affected any other parts of the application.

PR #3996, opened 2020-02-24
Vendor in patched version of hls.js

cjcolvar:
Patched version of master hls.js (0.13.1) added to improve hls segment
detection by increasing the number of TS packet sync bytes checked.  The
upstream check of 3 TS packets was falsely matching an AAC hls segment
in production.  This was not a problem under the old version of hls.js
we used (0.8 I believe) because it tested for AAC before TS.  I will
submit a PR upstream from the fork I built this version: https://github.com/avalonmediasystem/hls.js/tree/stricter_ts_probing


phuongdh:
Let's try using yarn to take advantage of dependencies management https://github.com/avalonmediasystem/avalon/pull/4006

cjcolvar:
Closed in favor of #4006

Issue #3937, opened 2020-01-14
Collection poster upload failure messages not rendered on the page


cjcolvar:
We came across this when ImageMagick wasn't installed on the server.  This issue is multi-faceted:
1. The poster upload is an XHR request and the collections controller uselessly does a redirect to the collection edit page instead of returning a useful code and message.
2. The collection edit page shows the cropped poster image during upload.  This could be confusing because it appears on the page where a saved poster appears even when the upload fails.
3. The collection edit page doesn't display messages (success or failure) from the upload POST request.  These used to be shown below the poster buttons where the progress bar is but those were removed since they seemed to be in the wrong place on the page.

### Done Looks Like
UX for poster upload and display of messages from the server designed and implemented.

### Related
#3938 - ImageMagick unnecessary for collection poster upload?
#3965 - Collection Poster can be set (but not saved) by depositor

Issue #3965, opened 2020-02-03
Collection Poster Can Be Set (Not Saved) By Depositor


joncameron:
Depositors cannot set the poster for a collection, but they can use the poster upload tool and it looks to the user like they are (even though it won't save on the server side)

cjcolvar:
Solution is probably to hide or disable the buttons when the user doesn't have the ability to upload or delete a poster.

PR #4013, opened 2020-02-28
Collection poster

Dananji:
Fixes: #3937 

Show the poster added success message on top of the page, which was not shown before;
Screenshot from 2020-02-28 14-00-14

Shows the cropped new poster after the upload is successful.

Before:
ezgif com-video-to-gif (11)
After:
ezgif com-video-to-gif (12)

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -16,7 +16,7 @@ class Admin::CollectionsController < ApplicationController
   include Rails::Pagination
 
   before_action :authenticate_user!
-  load_and_authorize_resource except: [:index, :remove, :attach_poster, :poster]
+  load_and_authorize_resource except: [:index, :remove, :attach_poster, :remove_poster, :poster]
   before_action :load_and_authorize_collections, only: [:index]
   respond_to :html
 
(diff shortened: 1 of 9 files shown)


app/controllers/admin/collections_controller.rb
```diff
@@ -252,24 +252,21 @@ def attach_poster
...
+        flash[:success] = "Poster file successfully added."
       else
         flash[:error] = "Uploaded file is not a recognized poster image file"
       end
     else
       @collection.poster.content = ''
       @collection.poster.original_name = ''
-      flash[:success] = "Poster file succesfully removed."
+      flash[:success] = "Poster file successfully removed."
+      redirect_to admin_collection_path(@collection)
```
cjcolvar: I'm a little concerned about the chances of a double `redirect_to` when a delete fails to save.  Maybe it would be better to refactor the delete out to a separate action that is the same route but which is a DELETE instead of a POST?

app/controllers/admin/collections_controller.rb
```diff
@@ -252,24 +252,21 @@ def attach_poster
...
-      flash[:error] = "There was a problem storing the poster image."
-    end
-
-    respond_to do |format|
-      format.html { redirect_to admin_collection_path(@collection) }
-    end
+    return if @collection.save
+    flash[:success] = nil
+    flash[:error] = "There was a problem storing the poster image."
+    redirect_to admin_collection_path(@collection)
```
cjcolvar: It's not exactly clear to me which cases should be redirecting and which shouldn't.  Shouldn't they all?  Or maybe the flash message rendering is dependent on this not being a redirect?

Dananji: It was my understanding that a redirect is necessary for the flash messages to render on the page. Maybe we can put this at the end of the method like it was before?

app/controllers/admin/collections_controller.rb
Dananji: Yes, sounds good :+1:

Dananji:
@cjcolvar Once this work is finished and merged, I will make a PR for the other ticket (#3965) after rebasing.

config/routes.rb
```diff
@@ -75,7 +75,8 @@
         get 'remove'
         get 'items'
         get 'poster'
-        post 'attach_poster'
+        post 'poster', action: :attach_poster, as: 'attach_poster'
```
cjcolvar: I thought it was better to make this RESTful instead of having a different url.

cjcolvar:
@Dananji How does this look to you?

Dananji:
@cjcolvar This looks great :+1:

PR #4014, opened 2020-02-28
Authorize managers to delete collections

### LINKED ISSUE #3995: Managers Cannot Delete Collections
### Description
Manager users cannot delete collections, but they should be able to: https://wiki.dlib.indiana.edu/display/VarVideo/Collection+Member+Roles

### Environment
Spruce, IU production

cjcolvar:
Fixes #3995

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -116,6 +116,10 @@ class Ability
           is_editor_of?(collection)
         end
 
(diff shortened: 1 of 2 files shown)


cjcolvar:
The one code climate issue is pre-existing and outside of the scope of this PR.

PR #4019, opened 2020-03-02
Disable collection poster buttons when user is a depositor

Dananji:
Changes in this PR;

1. Disable buttons when the user is a depositor;
Screenshot from 2020-03-02 14-14-01

2. Disable `Remove Poster` button when the collection does not have a poster;
Screenshot from 2020-03-02 14-15-00

3. Changed `Upload Poster` value to `Change Poster` when the collection has a poster;
Screenshot from 2020-03-02 14-15-24

### DIFF
--- a/app/views/admin/collections/show.html.erb
+++ b/app/views/admin/collections/show.html.erb
@@ -61,7 +61,7 @@ Unless required by applicable law or agreed to in writing, software distributed
         <%= form_for @collection, url: attach_poster_admin_collection_path(@collection.id), html: {method: "post"} do |form| %>
           <%= form.file_field :poster, id: 'poster_input', class: "filedata", style: "height:0px;width:0px;" %>
           <input type="button" class="btn btn-primary btn-xs" onclick="$('.filedata').click();" value="<% if @collection.poster.present? %>Change Poster<%else%>Upload Poster<%end%>"
-            <% if is_depositor %>disabled<%end%> />
+            <% if cannot?(:edit, @collection) %>disabled<%end%> />
         <% end %>
         <%= form_for @collection, url: remove_poster_admin_collection_path(@collection.id), html: {method: "delete"} do |form| %>
           <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" <% if is_depositor || @collection.poster.blank? %>disabled<%end%> />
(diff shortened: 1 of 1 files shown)


app/views/admin/collections/show.html.erb
```diff
@@ -57,12 +57,14 @@ Unless required by applicable law or agreed to in writing, software distributed
         <% end %>
       </div>
       <div class="collection_poster">
+        <% is_depositor=can?(:read, Admin::Collection) && !can?(:manage, Admin::Collection) %>
         <%= form_for @collection, url: attach_poster_admin_collection_path(@collection.id), html: {method: "post"} do |form| %>
           <%= form.file_field :poster, id: 'poster_input', class: "filedata", style: "height:0px;width:0px;" %>
-          <input type="button" class="btn btn-primary btn-xs" onclick="$('.filedata').click();" value="Upload Poster" />
+          <input type="button" class="btn btn-primary btn-xs" onclick="$('.filedata').click();" value="<% if @collection.poster.present? %>Change Poster<%else%>Upload Poster<%end%>"
+            <% if is_depositor %>disabled<%end%> />
```
cjcolvar: I think these authorization checks would be better if they were scoped to the specific collection and action being taken:
```suggestion
            <% if cannot?(:edit, @collection) %>disabled<%end%> />
```

app/views/admin/collections/show.html.erb
```diff
@@ -57,12 +57,14 @@ Unless required by applicable law or agreed to in writing, software distributed
...
         <% end %>
         <%= form_for @collection, url: remove_poster_admin_collection_path(@collection.id), html: {method: "delete"} do |form| %>
-          <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" />
+          <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" <% if is_depositor || @collection.poster.blank? %>disabled<%end%> />
```
cjcolvar: Same here:
```suggestion
          <input type="submit" class="btn btn-danger btn-xs" value="Remove Poster" <% if @collection.poster.blank? || cannot?(:edit, @collection) %>disabled<%end%> />
```

app/views/admin/collections/show.html.erb
```diff
@@ -57,12 +57,14 @@ Unless required by applicable law or agreed to in writing, software distributed
         <% end %>
       </div>
       <div class="collection_poster">
+        <% is_depositor=can?(:read, Admin::Collection) && !can?(:manage, Admin::Collection) %>
```
cjcolvar: I don't have a strong opinion between inlining the cancancan checks below or centralizing them in this line.  Either way works for me.

Dananji: I'm okay with the inline checks, since they are not long.
```suggestion
```

PR #4020, opened 2020-03-03
Replace ImageMagick with FastImage

Dananji:
Replace ImageMagick with FastImage gem, which checks the uploaded poster file is an image with the correct dimensions. This uses the URI of the uploaded file.


app/controllers/admin/collections_controller.rb
```diff
@@ -298,12 +298,10 @@ def collection_params
...
-    image.thumbnail '700x700>'
-    image.format 'png'
-    image.tempfile.open.read
-  rescue MiniMagick::Invalid
-    return nil
+  def check_image_compliance(poster_path)
+    type = FastImage.type(poster_path) == :png
+    # Size derived from width and aspect ratio from JS code, assets/javascript/crop_upload.js:60-63
+    size = FastImage.size(poster_path) == [700, 560] # [width, height]
+    type && size
```
cjcolvar: FastImage.type() and FastImage.size() both open and read the file so I think it would be better to instantiate a FastImage object to avoid the second open and read.
```suggestion
    fastimage = FastImage.new(poster_path)
    # Size derived from width and aspect ratio from JS code, assets/javascript/crop_upload.js:60-63
    fastimage.type == :png && fastimage.size == [700, 560] # [width, height]
```

app/controllers/admin/collections_controller.rb
```diff
@@ -246,9 +246,9 @@ def attach_poster
     authorize! :edit, @collection, message: "You do not have sufficient privileges to add a poster image."
 
     poster_file = params[:admin_collection][:poster]
-    resized_poster = resize_uploaded_poster(poster_file&.path)
-    if resized_poster
-      @collection.poster.content = resized_poster
+    is_image = check_image_compliance(poster_file&.path)
+    if is_image
+      @collection.poster.content = poster_file.tempfile.open.read
```
cjcolvar: I think this could be simplified:
```suggestion
      @collection.poster.content = poster_file.read
```
https://api.rubyonrails.org/v5.2.3/classes/ActionDispatch/Http/UploadedFile.html#method-i-read

cjcolvar:
@Dananji Do you want to work together on increasing the code coverage this afternoon during team work time?

PR #4022, opened 2020-03-04
Bug fix: Media player disappearing on reload in Edge 16

Dananji:
```
if (t.domNode.textTracks) {	
			for (var i = t.domNode.textTracks.length - 1; i >= 0; i--) {	
				t.domNode.textTracks[i].mode = 'hidden';	
			}	
		}
```
This code block hides the tracks if browser supports native captions in order to load MEJS captions. But this code block fails when it runs in Edge 16. 
For one of the items this code failed in mallorn (https://mallorn.dlib.indiana.edu/media_objects/td96k2606), on the first page load there was only one `textTracks` item while on the reload  there were 2 `textTracks` items. But this does not happen in Google Chrome.
And it seems the code uses VTTCue interface to handle the captions, which is not supported by Edge 16 (https://caniuse.com/#feat=mdn-api_vttcue).
Even though this fixes the issue, I'm not sure what the exact reason for this to fail on the reload.


Dananji:
@cjcolvar I tested on Safari on Browserstack, and seems to be working fine. Yes I agree. I'll add a browser check for it instead removing the entire code block :+1:

PR #4025, opened 2020-03-06
Fix ActionMailer not using smtp configs

phuongdh:


doug-hahn:
Tested it on our Docker setup and it worked.  I got Mail!

PR #4058, opened 2020-03-20
Disable Create Timeline button until derivative loads

Dananji:


Dananji:
> Currently there is some JS which does a similar check and then reloads the page when a player can be displayed. Does it make sense to do it that way instead of adding another JS check? I'm thinking of adding the check:
> 
> ```
> <% if @currentStream.present? and @currentStream.derivatives.present? %>
> ```
> 
> around the partial render here:
> 
> https://github.com/avalonmediasystem/avalon/blob/8ccab608c9a079f8dc96a5bcd9812839eb70a022/app/views/media_objects/_item_view.html.erb#L32

@cjcolvar I tried using this if-block within the button, but I was not able to make it work. Let me try this way and see if it works :+1:

cjcolvar:
@Dananji Oh, that's a good point.  We probably just want to disable instead of not render so the UI is consistent.

Dananji:
I used `@currentStream.present? and @currentStream.derivatives.present?` condition on `timeline` partial render. It renders the `Create Timeline` button on page reload, it did not stop it from rendering as I expected. 
So I used it on the button like this;
```
<button type="button" class="btn btn-default" id="timeline-btn" data-toggle="modal" data-target="#timelineModal" style="margin-top: 10px;" <% if !@currentStream.present? and !@currentStream.derivatives.present? %>disabled<% end %> >
    Create Timeline
</button>
```
And the button is not disabled when the player shows `00:00` as the duration of the media (which means the derivative is not loaded)
It seems the condition `@currentStream.present? and @currentStream.derivatives.present?` evaluates to `true` from the start.

cjcolvar:
Ok, sorry, I think I understand now what this PR was initially trying to solve.  I was thinking this was for the corner case where the derivatives were still being generated, but maybe this is more about just disabling the button until the player finishes loading the stream (to avoid issues with initializing the custom timeline start/end times?).  If this is the case then maybe we want to do the same for the Add to Playlist button?  And maybe the button should be disabled on page render but enabled by JS?

Dananji:
@cjcolvar Yeah, I thought the same initially.
Yes, this is to avoid initializing the custom timeline start/end times in the modal form.
`Add to Playlist` button => the ones in the accordion?

cjcolvar:
I think just the icon on the player since it has custom start/end times but the buttons on the accordion don't.

PR #4064, opened 2020-03-24
After the derivative object is destroyed queue a job…

cjcolvar:
…which deletes the actual file from the streaming server.  This is based on the masterfile management delete job.

I'll add tests if this looks reasonable.


phuongdh:
> Thanks @phuongdh for adding the tests! Code-wise this looks good to me now. Did you test it out in docker?

Using minio, `absolute_location` protocol is http instead of s3 so it's not working.
```
Attempting to delete derivative http://minio:9000/derivatives/be3c7ad7-a615-4386-981d-a66784a67d86/outputs/sample-medium.mp4
Derivative http://minio:9000/derivatives/be3c7ad7-a615-4386-981d-a66784a67d86/outputs/sample-medium.mp4 does not exist
```

cjcolvar:
If that change works then :+1:.

phuongdh:
Confirmed working in development stack

PR #4065, opened 2020-03-24
Update rails to 5.2.4.2 for actionview XSS fix

phuongdh:


Dockerfile
```diff
@@ -68,6 +68,10 @@ COPY        --from=download /usr/bin/ff* /usr/bin/
 
 # Build devevelopment image
 FROM        base as dev
+RUN         apt-get install -y --no-install-recommends --allow-unauthenticated \
```
cjcolvar: Was this just missed before or needed as part of this upgrade?

phuongdh: This is needed now because I couldn't update Rails without installing nokogiri, which requires build dependencies.

PR #4073, opened 2020-03-31
Simple model for supplemental files

cjcolvar:
Supplemental files model implemented as a json string stored in a single RDF triple to avoid overhead of new object while still retaining order.  New controller created to provide RESTful JS and html with flash message responses.

This PR also creates new avalon error classes to simplify the controller.

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -1,0 +1,119 @@
+# Copyright 2011-2020, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+class SupplementalFilesController < ApplicationController
+  before_action :set_master_file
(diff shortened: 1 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -11,6 +11,7 @@
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
 #   specific language governing permissions and limitations under the License.
 # ---  END LICENSE_HEADER BLOCK  ---
+require 'avalon/errors'
```
Dananji, line 14: @cjcolvar This line gives me an error, `cannot load such file -- avalon/errors`.

cjcolvar: :+1: It worked for me at first but then failed as well.  I'll fix that.

cjcolvar: It would probably help if I committed that file! 😄 
I pushed it and now the circleci tests are running.

app/controllers/supplemental_files_controller.rb
```diff
@@ -0,0 +1,122 @@
...
+    end
+
+    def supplemental_file_params
+      # TODO: Add parameters for minio and s3
+      params.fetch(:supplemental_file, {}).permit(:label, :file)
+    end
+
+    def next_id
+      last_id = @master_file.supplemental_files.max_by { |f| f.id.to_i }
+      last_id.present? ? last_id + 1 : 1
```
Dananji, line 111: @cjcolvar I get an error here `undefined method '+' for #<MasterFile::SupplementalFile:0x00007effbaf2a128>` when adding a new file. Here `label_id` is the Supplemental_file object not the id as expected.
So I changed that line to the following to make it work;
```
last_id.present? ? last_id.id.to_i + 1 : 1
```

cjcolvar: Good catch.  I was trying to simplify that line and missed that change in return value.  I'll probably just revert that change, but it looks like I need more tests if this is failing and the tests didn't catch it.

Issue #3936, opened 2020-01-14
Deleting Masterfile Should Remove Derivatives


joncameron:
### Description
When a masterfile is deleted, Avalon should also remove derivatives from the filesystem where there are placed for streaming. This is not currently the case in current ActiveEncode. The scope of ActiveEncode is just for kicking off jobs and not managing files elsewhere, so we need to think of what piece is in charge of this: probably Avalon itself. Permissions 

### Done Looks Like
- When masterfiles are deleted in Avalon, the associated derivatives and attending directories (if applicable) are deleted as well.

phuongdh:
We have to be careful going forward because deleting an item in MCO-staging can delete its derivatives in MCO. More access isolation is needed.

cjcolvar:
That's a really good point.  I can't recall how staging is configured.  Does the sidekiq worker on staging have access to the streaming mount?  As a last resort, we could disable the job by making it a no-op on staging, but it would still be risky not to have it isolated.

joncameron:
Needs quick testing on Spruce before being moved to done.

PR #4074, opened 2020-04-01
Fix local derivative file deletion

phuongdh:
Part of #3936


phuongdh:
Tested working on Spruce.

PR #4075, opened 2020-04-08
Preserve fullscreen of player when advancing to next section

Dananji:
Co-authored-by: Sumith Reddi Baddam <sumith.reddy2@gmail.com>

Fixes included in this PR:
1. Preserve fullscreen when advancing to next section in item page
2. Preserve fullscreen in playlists, when autoplay is on and in fullscreen mode

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -286,7 +286,7 @@ class MEJSPlayer {
 
     this.reInitializeCaptions();
 
-    // this.player.load();
+    this.player.load();
     this.player.play();
   }
 
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -129,25 +129,30 @@ class MEJSPlayer {
   }
 
   /**
-   * Make AJAX request for clicked item's stream data
+   * Make AJAX request for clicked/next available items's stream data
    * @function getNewStreamAjax
-   * @param  {string} id - id of master file id
    * @param {string} url Url to get stream data ie. /media_objects/xg94hp52v/section/bc386j20b
+   * @param {boolean} startPlaying once advancing to next item, continue playing after the player is re-initialized
    * @param {Array} playlistItemT Array which contains playlist item clip start and end times.  This is sent in from playlist items plugin, when creating a new instance of the player.
    * @return {void}
    */
-  getNewStreamAjax(url, isEnded, playlistItemsT) {
+  getNewStreamAjax(url, startPlaying, playlistItemsT) {
     $('.media-show-page').removeClass('ready-to-play');
     $.ajax({
       url: url + '/stream',
       dataType: 'json'
     })
       .done(response => {
-        this.setContextVars(response, playlistItemsT);
-        if(isEnded) {
-          this.switchPlayer(isEnded, playlistItemsT);
+        // Use the existing player iff existing player and new stream are both of type video and
+        // previous item has come to its end
+        if(startPlaying && response.is_video && this.player.isVideo) {
+          // Set currentStreamInfo before switching the player: switchPlayer uses new stream info
+          this.setContextVars(response, playlistItemsT);
+          this.switchPlayer(playlistItemsT);
         } else {
+          // Set currentStreamInfo after removing the player: removePlayer uses existing stream info
           this.removePlayer();
+          this.setContextVars(response, playlistItemsT);
           this.createNewPlayer();
         }
         this.updateShareLinks();
@@ -238,11 +243,10 @@ class MEJSPlayer {
    * Swap the media source and track in the current player instance with values from
    * the new stream info
    * @function switchPlayer
-   * @param {Boolean} isEnded
    * @param {Object} playlistItemsT
    * @returns {void}
    */
-  switchPlayer(isEnded, playlistItemsT) {
+  switchPlayer(playlistItemsT) {
     let markup = '';
 
     this.node.innerHTML = '';
@@ -260,14 +264,11 @@ class MEJSPlayer {
 
     this.node.innerHTML = markup;
 
-    // Handle 'canplay' events fired by player
+    // Bind to canplay event to switch off loading when media is ready
     this.mediaElement.addEventListener(
       'canplay',
       this.handleCanPlay.bind(this)
     );
-
-    this.player.buildquality(this.player, null, null, this.mediaElement);
-    this.player.buildtracks(this.player, null, this.player.layers, this.mediaElement);
 
     // Build playlists button from the new stream when not in playlists
     if(!playlistItemsT) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -223,11 +223,70 @@ class MEJSPlayer {
...
+   * @param instance {Object} - player instance
+   */
+  switchMedia(media, node, instance) {
+    let markup = '';
+
+    node.innerHTML = '';
+    this.currentStreamInfo.stream_hls.map(source => {
+      markup += `<source src="${
+        source.url
+      }" type="application/x-mpegURL" data-quality="${source.quality}"/>`;
```
cjcolvar, line 261: I think it is okay to hardcode the type here but ideally I think we would want to get it from the stream info returned from the server (if it does return that information).

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -223,11 +223,70 @@ class MEJSPlayer {
         paused: false
       };
 
-      // Go to next section
-      this.getNewStreamAjax(
```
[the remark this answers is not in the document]
Dananji: Yes that is the only place that calls this. I will look into refactoring it :+1:

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
Dananji: I will look through `currentStreamInfo` and see whether it has the type coming in with the stream.

Dananji: @cjcolvar `currentStreamInfo` has a `mimetype` in its data for each source. But they have `null` as the value.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -134,19 +136,20 @@ class MEJSPlayer {
    * @param {Array} playlistItemT Array which contains playlist item clip start and end times.  This is sent in from playlist items plugin, when creating a new instance of the player.
    * @return {void}
    */
-  getNewStreamAjax(id, url, playlistItemsT) {
+  getNewStreamAjax(url, isEnded, playlistItemsT) {
```
cjcolvar, line 139: After looking though this commit, I wonder if `startPlaying` might be a better parameter name than `isEnded` or maybe that is conflating the actual `ended` case with starting playing once the player has be reinitialized?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -134,19 +136,20 @@ class MEJSPlayer {
    * @param {Array} playlistItemT Array which contains playlist item clip start and end times.  This is sent in from playlist items plugin, when creating a new instance of the player.
```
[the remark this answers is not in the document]
Dananji: Yes :+1:

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -217,70 +220,92 @@ class MEJSPlayer {
...
+      this.player.buildaddToPlaylist(this.player, null, null, null);
+    }
 
+    // Turn on captions
     this.toggleCaptions();
 
-    instance.load();
-    instance.play();
+    this.player.load();
+    if(isEnded) {
```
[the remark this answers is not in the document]
Dananji: Now that I think of it, we don't need this conditional check at all because it is already checked in `getNewStreamAjax` function before calling `switchPlayer`. So `isEnded` is always true.

`this.switchPlayerHelper.paused` is initialized to `false` and changed with the status of the player within `avalon_player_new.es6`. So when using `this.switchPlayerHelper.paused` when invoking this function from an external file, the value of it is always `false`.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_add_to_playlist.es6
```diff
@@ -136,6 +138,11 @@ Object.assign(MediaElementPlayer.prototype, {
         addToPlayListObj.bindHandleCancel
       );
     }
+    if (player) {
+      if (player.addPlaylistButton) {
```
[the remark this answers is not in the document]
Dananji: I don't think prettier will have an issue with it been on the same condition.

app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -228,16 +226,17 @@ Object.assign(MediaElementPlayer.prototype, {
           { id: playlistItemId, playlist_id: playlistId, position: null }
         );
 
+        // Set now playing item in playlist items
+        this.setCurrentItemInternally();
```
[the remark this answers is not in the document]
Dananji: It is called within `this.setupNextItem()`.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
Dananji: Yes, I agree :+1:

Dananji:
@cjcolvar While I was working on these fixes, I realized with the new way of re-initializing the player it doesn't highlight time rail for the new segment in the playlists. Should I work on it and then commit the changes?

cjcolvar:
@Dananji Yes, I think it would be best to restore the playlist behavior in this PR.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -260,46 +264,56 @@ class MEJSPlayer {
...
+   * in playlists and item page
+   * @function reInitializeCaptions
+   * @param playlistItemsT 
+   */
+  reInitializeCaptions(playlistItemsT) {
+    if (this.currentStreamInfo.captions_path) {
+      // Re-arrange player features when tracks are available
+      // in both playlists and item page
+      this.player.featurePosition.tracks = 6;
+      this.player.featurePosition.fullscreen = 9;
```
cjcolvar, line 303: This looks brittle.  I wonder if there is another way we could do this since I'm afraid this will break if we add other controls.  What do you think?

Dananji: This wouldn't break the code, but if we add new controls they would be out of order. And that is an issue.
This is the way I could think of to do this. But there should be a better way.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -284,32 +284,24 @@ class MEJSPlayer {
...
    */
-  reInitializeCaptions(playlistItemsT) {
+  reInitializeCaptions() {
     if (this.currentStreamInfo.captions_path) {
-      // Re-arrange player features when tracks are available
-      // in both playlists and item page
-      this.player.featurePosition.tracks = 6;
-      this.player.featurePosition.fullscreen = 9;
+      // Place tracks button after volume button when tracks are available
+      this.player.featurePosition.tracks = this.player.featurePosition.volume + 1;
```
Dananji, line 300: @cjcolvar What do think of this solution? I feel this is better than the previous one.

cjcolvar: :+1: This looks good to me.  Does it automatically push other features one position higher when you set it like this?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -284,32 +284,24 @@ class MEJSPlayer {
     // Build quality
     this.player.buildquality(this.player, null, null, this.mediaElement);
 
-    this.reInitializeCaptions(playlistItemsT);
+    this.reInitializeCaptions();
 
-    this.player.load();
+    // this.player.load();
```
cjcolvar, line 289: Is this load not necessary?

Dananji: I've commented it out by accident. I will fix it.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
Dananji: Yes it does.

Dananji:
@cjcolvar This PR has the supplemental file changes. I rebased this branch once on develop. How do I remove those commits?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
cjcolvar: :+1: Neat!

cjcolvar:
@Dananji We can work together on cleaning the git history after standup today if that works for you.

Dananji:
@cjcolvar Yes :100: . Thank you!

Dananji:
@cjcolvar You were right, the merge commit was the one that pulled in all the supplemental file commits. Steps I followed to remove those commits were;
1. Reverted to the commit before the merge commit
2. Made my last change (uncommenting the `this.player.load()` line) and made a commit
3. Force pushed
It was not as scary as I thought (I will remove the backup branch I create). Thanks for the help!

PR #4080, opened 2020-04-09
Supplemental file add/remove UI

Dananji:
Screenshot from 2020-04-13 14-38-54

### DIFF
--- a/app/assets/javascripts/supplemental_files.js
+++ b/app/assets/javascripts/supplemental_files.js
@@ -7,7 +7,7 @@ $('button[name="edit_label"]').on('click', e => {
   inputField.focus();
 });
 
-$('button[name="label_edit_cancel"]').on('click', e => {
+$('button[name="cancel_edit_label"]').on('click', e => {
   const { $row, fileId, masterFileId } = getHTMLInfo(e);
   $row.find('input[name="label_' + masterFileId + '_' + fileId + '"]');
   $row.removeClass('is-editing');
--- a/app/assets/javascripts/supplemental_files.js
+++ b/app/assets/javascripts/supplemental_files.js
@@ -1,37 +1,40 @@
 
-  $('button[name="edit_label"]').on('click', e => {
-    const $row = $(e.target).parents('.row');
-    const fileId = $row.data('file-id');
-    console.log(fileId);
-
-    $(e.target)
-      .parents('.row')
-      .addClass('is-editing');
-  });
+$('button[name="edit_label"]').on('click', e => {
+  const { $row, fileId, masterFileId } = getHTMLInfo(e);
+  const inputField = $row.find('input[name="label_' + masterFileId + '_' + fileId + '"]');
+  inputField.val($row.data('file-label'));
+  $row.addClass('is-editing');
+  inputField.focus();
+});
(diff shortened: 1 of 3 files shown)


app/views/media_objects/_structure.html.erb
```diff
@@ -211,6 +255,13 @@ Unless required by applicable law or agreed to in writing, software distributed
     $('#' + target).toggleClass('hidden');
   });
 
+    $('.files_toggle').on('click', function (e) {
```
cjcolvar, line 258: Should this one line be unindented?  It looks odd as it is right now.

app/assets/stylesheets/avalon.scss
```diff
@@ -576,6 +601,10 @@ div.status-detail {
     span.tool_label {
       font-weight: bold;
     }
+
+    div.transcript_file_view {
+      display: flex;
+    }
```
cjcolvar, line 607: It doesn't look like this is used.  Can it be removed?

app/assets/javascripts/supplemental_files.js
```diff
@@ -0,0 +1,40 @@
+
+$('button[name="edit_label"]').on('click', e => {
+  const { $row, fileId, masterFileId } = getHTMLInfo(e);
+  const inputField = $row.find('input[name="label_' + masterFileId + '_' + fileId + '"]');
+  inputField.val($row.data('file-label'));
+  $row.addClass('is-editing');
+  inputField.focus();
+});
+
+$('button[name="label_edit_cancel"]').on('click', e => {
```
cjcolvar, line 10: For consistency would it make sense to rename this button `cancel_edit_label`?

app/assets/javascripts/supplemental_files.js
```diff
@@ -0,0 +1,40 @@
...
+  formData.append('supplemental_file[label]', newLabel);
+  formData.append('authenticity_token', $('input[name=authenticity_token]').val());
+
+  fetch('/master_files/' + masterFileId + '/supplemental_files/' + fileId, {
+    method: "PUT",
+    body: formData
+  }).then(() => {
+    $row.removeClass('is-editing');
+    // Page reload to show the flash message
+    location.reload();
```
cjcolvar, line 31: I wonder if it is really necessary to reload the page for a label update.  Maybe an inline indication of success would be better?  But I'd wait on feedback from PO before changing this since it looks like it already works.

app/assets/stylesheets/avalon.scss
Dananji: Yes, good catch :+1:

app/assets/javascripts/supplemental_files.js
Dananji: Yeah, that sounds good to me. Now that I think about it, some inline edit forms do not have feedback flash messages (e.g. editing markers in playlists, add to playlist in item page. these are the inline forms I can think of right now).

Dananji:
> I think this is ready now but I think it makes sense for it to wait to be merged until the model/controller PR has been merged (which will hopefully be soon). Is that okay with you?

Yes I agree :100:

PR #4085, opened 2020-04-14
Supplemental file storage

cjcolvar:
Use ActiveStorage to allow for avalon to be configured to store files on disk or in s3 (or other cloud storage).


cjcolvar:
After the upgrade of aws-sdk there was one failing test.  The issue was that aws-s3-sdk was subverting FakeFS when doing download_file so the file didn’t exist in the FakeFS filesystem but it did in the real filesystem.  Switching to Tempfile felt better in the real code and allowed for getting rid of FakeFS in the test.

PR #4088, opened 2020-04-21
Attempt to fetch Derivative.absolute_location in WaveformJob

phuongdh:
To avoid pulling from the streaming server if possible.

### DIFF
--- a/app/jobs/waveform_job.rb
+++ b/app/jobs/waveform_job.rb
@@ -24,12 +24,9 @@ class WaveformJob < ActiveJob::Base
     return if master_file.waveform.content.present? && !regenerate
 
     service = WaveformService.new(8, SAMPLES_PER_FRAME)
-    uri = file_uri(master_file) || playlist_url(master_file)
+    uri = file_uri(master_file) || derivative_file_uri(master_file) || playlist_url(master_file)
     json = service.get_waveform_json(uri)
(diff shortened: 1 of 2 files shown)


app/jobs/waveform_job.rb
```diff
@@ -51,6 +48,22 @@ def file_uri(master_file)
       end
     end
 
+    def derivative_file_uri(master_file)
+      derivatives = master_file.derivatives
+      uri = nil
+
+      # Find the lowest quality stream
+      ['high', 'medium', 'low'].each do |quality|
+        d = derivatives.select { |d| d.quality == quality }.first
```
cjcolvar, line 57: You should be able to simplify this with `find`:
```suggestion
        d = derivatives.find { |derivative| derivative.quality == quality }
```

phuongdh: Thanks!

phuongdh: Nvm, that didn't work, I remember starting out with this now. 
```
Failure/Error: d = derivatives.find { |derivative| derivative.quality == quality }
     ArgumentError:
       Couldn't find Derivative without an ID
```

cjcolvar: Ah, probably needs to be `derivatives.to_a.find` due to AF.

PR #4091, opened 2020-04-23
Collection contact information

### LINKED ISSUE #3604: Contact Information on Collections Page
### Description
As a collection manager, I want the  ability to provide contact information for a collection on the collections page so that users can find contact information or other

### Done Looks Like
New fields added to a Collection: 
  - Contact Email address
  - Website
These fields are optional (like description, not required). These are fields able to be edited in the same modal where name, description etc. is changed on the collection's admin page.

This information is rendered on the collection's landing page. We'll move the "Show more" text to the end of the cut-off text and put the email/url info where that is now.

#### Collection Edit Details Form
Screen Shot 2020-04-21 at 10 31 10 AM

cjcolvar:
Resolves #3604


app/javascript/components/CollectionDetails.js
```diff
@@ -67,7 +67,7 @@ const CollectionDetails = ({ content = '', email = '', website = '' }) => {
         { website &&
           <>
             <dt style={descriptionStyle}>Website:</dt>
-            <dd><a href={website}>{website}</a></dd>
+            <dd dangerouslySetInnerHTML={{ __html: website }}></dd>
```
cjcolvar, line 70: @phuongdh @Dananji This is equivalent to `innerHTML` and while it is called `dangerously` because of the possibility of XSS I think it should be safe here since we're in control of the html.  I thought it made more sense for the presenter to return a full a tag instead of the label and url separately.  What do you two think?

cjcolvar:
@Dananji I know we talked about keeping the UI consistent with having the field labels present even when there isn't a value but I went back on that when seeing it in the collection landing page.  Was that a bad idea?  Maybe we can ask @joncameron tomorrow.

app/javascript/components/CollectionDetails.js
Dananji: Yeah, sounds good to me :+1:

joncameron:
When this gets merged and is on Spruce I'll take a look and we can tweak it. In this case I don't think it makes sense to have labels with no value.

spec/models/collection_spec.rb
```diff
@@ -147,6 +155,10 @@
     end
     it {is_expected.to validate_presence_of(:unit)}
     it {is_expected.to validate_inclusion_of(:unit).in_array(Admin::Collection.units)}
+    it { is_expected.to allow_value('collection@example.com').for(:contact_email) }
+    it { is_expected.not_to allow_value('collection@').for(:contact_email) }
+    it { is_expected.to allow_value('https://collection.example.com').for(:website_url) }
+    it { is_expected.not_to allow_value('collection.example.com').for(:website_url) }
```
phuongdh, line 161: This may be a little a little too strict in practice as people aren't used to typing out the "https://" part, but I guess it makes for a safer validator.

app/models/admin/collection.rb
```diff
@@ -38,6 +40,15 @@ class Admin::Collection < ActiveFedora::Base
...
   end
+  property :contact_email, predicate: Avalon::RDFVocab::Collection.contact_email, multiple: false do |index|
+    index.as :stored_sortable
+  end
+  property :website_label, predicate: Avalon::RDFVocab::Collection.website_label, multiple: false do |index|
+    index.as :stored_sortable
+  end
+  property :website_url, predicate: Avalon::RDFVocab::Collection.website_url, multiple: false do |index|
+    index.as :stored_sortable
+  end
```
cjcolvar, line 51: @jlhardes Are there standard RDF predicates that we should be using for these?

PR #4101, opened 2020-04-27
Rebuild list of current user's virtual groups stored in user_session

### LINKED ISSUE #4039: [Persona] Virtual groups (including LDAP groups) not replaced when impersonating another user
### Description
When a user logs in the virtual (external) groups they belong to (including LDAP groups and LTI group) are calculated and put in the session.  Then when that user becomes another user via persona these virtual groups are not being recalculated.  This causes permissions to be different and gives an inaccurate view of what the impersonated user would see.

This is scoped to LDAP group membership. Becoming for other sorts of virtual groups (LTI) is out of scope.

### Done Looks Like
- When Becoming another user, LDAP group membership for that user applies to the permissions on view/edit/etc.

cjcolvar:
when starting/stopping impersonation.

Fixes #4039.

### DIFF
--- a/app/controllers/samvera/persona/users_controller.rb
+++ b/app/controllers/samvera/persona/users_controller.rb
@@ -14,6 +14,7 @@
 
 # Copied here in full from samvera-persona 0.1.7
 # Added prepend_view_path
+# Added LDAP group recalculation in impersonate and stop_impersonating
 # TODO: Determine why this is necessary and remove this override
 module Samvera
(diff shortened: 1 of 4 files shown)


cjcolvar:
I copied the full controller tests from persona into avalon since we're completely overriding the controller.

cjcolvar:
I'm a bit at a loss here.  CircleCI is reporting three failing tests but when I run those tests in a console on the running CircleCI container they pass. 🤷‍♂️

PR #4104, opened 2020-04-28
Display supplemental files in item view page

phuongdh:
Co-authored-by: Chris Colvard <cjcolvar@indiana.edu>
Co-authored-by: Dananji Withana <dwithana@iu.edu>

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -341,9 +341,13 @@ a[data-trigger='submit'] {
   }
 }
 
-#creation_metadata dd {
-  margin-left: 10px;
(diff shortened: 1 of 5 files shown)


phuongdh:
@joncameron do you have a preference as to where we put the "View Raw" link? It's in a tab right now but should probably move somewhere else.

image

joncameron:
@phuongdh Let's move it to the bottom of "Details", below the descriptive metadata, like so:
Screen Shot 2020-05-05 at 9 41 49 AM

cjcolvar:
:+1: Let's use "Raw Metadata" and "View MODS XML" for the label and link.

PR #4108, opened 2020-05-01
Fix batch replay

### LINKED ISSUE #4109: Batch ingest replay broken
Reported in slack.

cjcolvar:
Replay complete testing and reset to pending both used a different means of making the payload than registering a new entry.

Fixes #4109 .


cjcolvar:
The test that is failing is not the one I added.  I'm not getting the failure locally and I haven't been able to deduce why it is failing in circleci yet.  I'll try to debug this more through ssh to running containers but what I'm seeing right now is instead of an admin collection being created a media object is which breaks everything.

lib/avalon/batch/ingest.rb
```diff
@@ -104,15 +104,15 @@ def replay_entry(entry, position)
...
+        return nil if previous_entry.payload == entry.to_json && previous_entry.complete
 
         # Case 1, if there is a payload change and the item never completed, reset to pending
         # Also reset to pending if there is an error
         reset_to_pending(previous_entry, entry) if !previous_entry.complete || previous_entry.error
 
         # Case 2, if there is a published media object we cannot replay
         unless previous_entry.media_object_pid.nil?
-          mo = MediaObject.find(previous_entry.media_object_pid)
+          mo = MediaObject.find(previous_entry.media_object_pid) if MediaObject.exists?(previous_entry.media_object_pid)
```
cjcolvar, line 115: Maybe this should be handled in a separate PR since it might include more code changes and need a test?  
I would think that if the batch entry has a media object id but the media object can't be found then the batch entry should probably be reset_to_pending so that it will be recreated.  
@joncameron Do you have an opinion on this?

cjcolvar: I went ahead and removed it so this PR can go ahead and be merged, but I think it would be good to create a follow-up issue or PR so it doesn't get lost.

PR #4110, opened 2020-05-04
Zero width

### LINKED ISSUE #4061: Avalon Doesn't Remove Zero Width Space Character
### Description
When adding a value to a field within Avalon, a frequent user action is to copy and paste this value (corresponding to the name of an external group, in this instance) into the text field and save. A problem occurs when this copy action picks up non-printing characters that are retained but not stripped by Avalon. This results in situations where to the user the value seems fine—exactly the characters they submitted—but since the strings aren't equal due to these hidden characters, permissions don't apply correctly. I encountered with with the zero width space character (https://unicode-table.com/en/200B/) but it may also apply to other unicode characters.

### To Reproduce
Add a value to "External Groups" under the Special Access of an item's access control with a zero width space character in it. The group will be unable to access this item because the group name, "MCO LTI Test" for example, is "MCO LTI Testu+200B". 

Copying and pasting from the Canvas course ID name contained in an email into a certain editors show this extra information. In Vim for example the unicode character is displayed as:
MCO LTI Test​<200b>

### Example
Item 1: https://media.dlib.indiana.edu/media_objects/rb68xg57v/edit?step=access-control
Item 2: https://media.dlib.indiana.edu/media_objects/d791sk993/edit?step=access-control

Item 1 included the u/200b, and does not appear in the "MCO LTI Test" course. Item 2 doesn't contain this character and appears normally.

### Done Looks Like
Avalon forms are tested to find all places where whitespace trimming is not occurring.  Controllers updated to fix all of these instances.

SumithBaddam:
Fixes #4061 Remove Zero-width characters from user input for collections, groups and users. Changes made to add a module names "remove_zero_width_chars" to String class. Striping the input by calling this function before performing operations.


lib/string_additions.rb
```diff
@@ -0,0 +1,8 @@
+module StringAdditions
+  def remove_zero_width_chars
+    self.gsub(/^[\u200B-\u200D\uFEFF]/, '').gsub(/[\u200B-\u200D\uFEFF]$/, '') #Begins with and Ends with zero-width character
```
cjcolvar, line 3: This is looking good!  Do you think we need to handle `\u2060` as well? (https://en.wikipedia.org/wiki/Word_joiner)

spec/lib/string_additions_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+require 'rails_helper'
+
+describe 'StringAdditions' do
```
cjcolvar, line 17: Please add tests for \u200c, \uFEFF, and \u2060 (if you decide to add it).

app/controllers/admin/collections_controller.rb
```diff
@@ -143,7 +143,7 @@ def update
       ["group", "class", "user", "ipaddress"].each do |title|
         if params["submit_add_#{title}"].present?
           if params["add_#{title}"].present?
-            val = params["add_#{title}"].strip
+            val = params["add_#{title}"].strip.remove_zero_width_chars
```
cjcolvar, line 146: Please remove this call to `remove_zero_width_chars`

app/models/access_control_step.rb
```diff
@@ -39,7 +41,7 @@ def execute context
         end_time = context["add_#{title}_end"].blank? ? nil : context["add_#{title}_end"]
         create_lease = begin_time.present? || end_time.present?
         if context["add_#{title}"].present?
-          val = context["add_#{title}"].strip
+          val = context["add_#{title}"].strip.remove_zero_width_chars
```
cjcolvar, line 44: Please remove this call to remove_zero_width_chars

app/controllers/admin/collections_controller.rb
```diff
@@ -143,7 +143,7 @@ def update
       ["group", "class", "user", "ipaddress"].each do |title|
         if params["submit_add_#{title}"].present?
           if params["add_#{title}"].present?
-            val = params["add_#{title}"].strip.remove_zero_width_chars
+            val = params["add_#{title}"]
```
cjcolvar, line 146: I think we may still need the `.strip` unless if we want to move that into the ApplicationController before_action as well, but I think that could wait for another PR.

PR #4112, opened 2020-05-05
Use standard rdf predicates

cjcolvar:
Also make website label optional and use plain text fields because built in validations of email and url cause odd behaviors with the modal.

### DIFF
--- a/app/views/admin/collections/show.html.erb
+++ b/app/views/admin/collections/show.html.erb
@@ -51,8 +51,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
       <% if @collection.website_url.present? %>
         <p><strong>Website:</strong></p>
-        <% website_label = @collection.website_label.present? ? @collection.website_label : @collection.website_url %>
-        <p><%= link_to(website_label, @collection.website_url)%></p>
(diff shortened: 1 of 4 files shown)


app/views/admin/collections/show.html.erb
```diff
@@ -51,7 +51,8 @@ Unless required by applicable law or agreed to in writing, software distributed
       <% end %>
       <% if @collection.website_url.present? %>
         <p><strong>Website:</strong></p>
-        <p><%= link_to(@collection.website_label, @collection.website_url)%></p>
+        <% website_label = @collection.website_label.present? ? @collection.website_label : @collection.website_url %>
```
phuongdh, line 54: Would this be cleaner? `website_label = @collection.website_label || @collection.website_url`
It can be merged with the next line too.

cjcolvar: We need the `present?` check because it could be empty string.  Or we could try to handle that in the model or controller somewhere by turning `""` into `nil`.

phuongdh: I see. Looks like `presence` will handle that https://apidock.com/rails/Object/presence
i.e. `@collection.website_label.presence || @collection.website_url`

cjcolvar: Neat!  I've never seen that before.

PR #4121, opened 2020-05-08
Switch yarn to use PnP instead of node_modules

cjcolvar:
It's just this easy. Thanks for pointing this out @phuongdh!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,7 @@ yarn-debug.log*
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+*.pnp*
 
 # Cypress test output
 /cypress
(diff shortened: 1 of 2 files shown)


cjcolvar:
This is breaking tests with failures like this:
```
ActionView::Template::Error:
couldn't find file 'cropperjs/dist/cropper.min' with type 'application/javascript'
```

merceyz:
If you want to use PnP you should be using Yarn V2 as it has a better implementation https://yarnpkg.com/getting-started/install#per-project-install

cjcolvar:
Switched to shakapacker and it doesn't support PnP so closing this.  We can revisit in the furture when we switch off of webpacker/shakapacker to a different JS bundling solution.

Issue #4093, opened 2020-04-24
Check SpeedyAF Support for Supplemental Files


joncameron:
### Description
Need to make sure that loading from Fedora isn't causing slowdown when loading media objects that contain or have masterfiles that contain supplemental files.

cjcolvar:
The change needed so far was so small that I just included it in the open PR: #4104

cjcolvar:
Looks like reifying is still happening:
`Reifying MasterFile because supplemental_files_json called from /srv/avalon/avalon_r6/releases/20200507195919/app/presenters/speedy_af/proxy/master_file.rb:42:in `supplemental_files'`

PR #4125, opened 2020-05-11
Handle errors when attaching files

### LINKED ISSUE #4113: Supplemental File records still created when ActiveStorage misconfigured
### Steps to Reproduce
Change the ActiveStorage configuration to be something that isn't configured correctly then try uploading a file to a section on the structure step.

### Expected Behavior
If the attaching of the file to the supplemental file fails then the supplemental file record should be deleted.

### Actual Behavior
500 page displayed but going back to the structure step the supplemental file is displayed as if the upload succeeded.

cjcolvar:
Fixes #4113, finishes #4093, and fixes issues with failing CI builds due to tests added in #4104.

### DIFF
--- a/app/presenters/speedy_af/proxy/master_file.rb
+++ b/app/presenters/speedy_af/proxy/master_file.rb
@@ -22,7 +22,9 @@ class SpeedyAF::Proxy::MasterFile < SpeedyAF::Base
   end
 
   def find_encoder_class(klass_name)
-    ActiveEncode::Base.descendants.find { |c| c.name == klass_name }
+    return nil if klass_name.blank? || !Object.const_defined?(klass_name)
+    klass = Object.const_get(klass_name)
+    return klass if klass.ancestors.include?(ActiveEncode::Base)
   end
 
   def display_title
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -515,6 +515,10 @@ class MediaObjectsController < ApplicationController
   protected
 
   def master_file_presenters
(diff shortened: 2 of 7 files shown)


app/presenters/speedy_af/proxy/master_file.rb
```diff
@@ -22,7 +22,9 @@ def encoder_class
   end
 
   def find_encoder_class(klass_name)
-    ActiveEncode::Base.descendants.find { |c| c.name == klass_name }
+    return nil if klass_name.blank? || !Object.const_defined?(klass_name)
+    klass = Object.const_get(klass_name)
+    return klass if klass.ancestors.include?(ActiveEncode::Base)
   end
```
phuongdh, line 28: There's a lot of duplicated code (not just this method) between this proxy and the model, is there a way to dry it out using mixin or something?

cjcolvar: Sure, this is probably a good time to clean it up.

cjcolvar: I looked but couldn't come up with any good alternative to this duplication right now.  I think the proper solution is the separation of the UI from the model by true presenters powered only by solr documents instead of speedy_af proxies powered by both solr and fedora, but that is considerable more work to go through the UI and make sure that we are using the presenter only.  It is a larger refactor that probably shouldn't hold up this PR.  Maybe there could be changes made to speedy_af that could move us in that direction without having to do quite as large a refactor?  But this effort might not really be worth it given the eventual move to avalon-bundle.

PR #4127, opened 2020-05-14
Increase default DB pool setting for prod

phuongdh:
So sidekiq can run multiple jobs in parallel.


cjcolvar:
I'm merging despite test failures since they are being addressed in #4125.

Issue #4081, opened 2020-04-10
Test Universal Viewer Support for Avalon Manifests


joncameron:
### Description
We should explore and list issues of Universal Viewer playing Avalon-generated manifests. There are still some issues with UV playing our media, and it would be good for us to know what these are and what is causing them. 

### Done Looks Like
- Timeboxed investigation into issues with using Avalon manifests in UV
- List of issues compiled and possibly shared with UV development team

Dananji:
Observed:

1. Avalon manifests are missing the property `format` for items (https://iiif.io/api/presentation/3.0/#format), and UV expects it to be present. Because of this UV can't load the manifest.

2. I updated a manifest from Spruce in JSONBin.io (link: https://api.jsonbin.io/b/5ea9a95e66e603359fe0cd7d/5) and tried loading it in UV, and it loads the player. When it is being played there's no sound even though the time is updated in the player. In the console in DevTools it keeps logging `loading media` indefinitely.

![Screenshot from 2020-04-29 13-54-59.png](https://images.zenhubusercontent.com/5cc1c66b3244e00362de97ea/eeb2fc95-3fae-419c-a730-04da2058e670)

It seems the `loadedmetadata` event in ME.js is not getting fired. And I can't figure out what's the reason for this.

3. IIIF 3.0 spec presentation manifests in maple loads and plays as expected.

Dananji:
Tested on https://uv.netlify.app/;

1. Manifest from maple, (https://dlib.indiana.edu/iiif_av/lunchroom_manners/lunchroom_manners_v6_jwd.json) works well in this app
2. Manifests from both spruce and mco-staging, does not work in this. It fails trying to fetch additional data from the stream URL. 
For sample manifests with images, this action happens for fetching additional information for the images from the an image service. 
For Avalon manifests, the error occurs when trying to go through the `service` section in the items for authentication. According to 3.0 spec [here](https://iiif.io/api/presentation/3.0/#service), the `services` block has some changes from previous version. I read through it and couldn't find why this would fail from the manifest side.

cjcolvar:
[...]

PR #4129, opened 2020-05-18
Add probe service for UV support

cjcolvar:
As noted as needed in https://github.com/avalonmediasystem/avalon/issues/4081#issuecomment-625469328

This needs to be tested in UV and timeliner before including in a release.


cjcolvar:
I think this implementation isn't quite right since it sets the probe service to the auto quality stream every time instead of using the quality of the stream that the probe service is being specified for.  I'll create a follow on PR.

PR #4136, opened 2020-05-19
Catch top-most Exception since LoadError can be thrown

cjcolvar:
With spruce configured to use the minio storage configuration without minio setup, a LoadError is thrown when trying to upload a supplemental file.  `LoadError` is not a subclass of `StandardError` so we need to catch `Exception` instead.

Read and Delete also have errors but I'm not sure if we really need to handle them.


cjcolvar:
@phuongdh I tried to come up with something for destroy but failed.  It looks like default behavior is already to `purge_later` the attached file when the parent record is deleted.  It looks like the attempt to `purge_later` is causing the parent record's destroy to fail and rollback.  So maybe this is okay for now and we can handle it better when we have a real failing case?

phuongdh:
We last talked about catching both StandardError and LoadError

PR #4151, opened 2020-06-01
ME.js player keyboard support

Dananji:
This PR includes the following;
- Add support for jump forward and backwards in ME.js player using keyboard
- Add buttons in the player for jump forward and backwards
Screenshot from 2020-06-01 13-48-04
- Focus relevant keyboard actions on player, when it is on page. For example: when on item page, pressing spacbar will play/pause the player, pressing `f` key will make the player enter/exit fullscreen, etc.


app/views/master_files/_mejs4_player_js.html.erb
```diff
@@ -26,7 +26,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 <!-- Missing title and logo features -->
 var mejs4AvalonPlayer = new MEJSPlayer({
   currentStreamInfo: <%= section_info.to_json.html_safe %>,
-  features: ['playpause', 'current', 'progress', 'duration', 'volume', 'tracks', <%= quality_feature.html_safe %>'linkBack', 'fullscreen'],
+  features: ['playpause', 'skipback', 'jumpforward', 'current', 'progress', 'duration', 'volume', 'tracks', <%= quality_feature.html_safe %>'linkBack', 'fullscreen'],
```
cjcolvar, line 29: I wonder if we really want this in the embed player since we have limited space within the iframe, but it would be helpful in fullscreen mode and it might have to be in the normal size in order to be in fullscreen.  What do other people think?
@joncameron

joncameron: I feel like this might be too much and cramp the rail even further. But I'm interested in at least testing it out on Spruce and feel it out. Maybe these could also sit outside of the MEjs features row.

PR #4152, opened 2020-06-01
Make active-storage configurable

### LINKED ISSUE #4114: Allow choosing activestorage configuration from settings.yml
Currently the activestorage adapter is hard coded in the matching environment file (e.g. config/environments/development.yml).  This leads to problems when deploying to spruce since it isn't setup to use minio which is the default so developers can use docker.

### Done Looks Like
- Config file is decided
- Application updated to use config

phuongdh:
Fixes #4114

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,5 +47,7 @@ module Avalon
         resource '/timelines/*/manifest.json', headers: :any, methods: [:get, :post]
       end
     end
+
+    config.active_storage.service = (Settings&.active_storage&.service.presence || "local").to_sym
   end
 end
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -47,5 +47,7 @@ class Application < Rails::Application
         resource '/timelines/*/manifest.json', headers: :any, methods: [:get, :post]
       end
     end
+
+    config.active_storage.service = (Settings&.active_storage&.service.presence || "local").to_sym
```
cjcolvar, line 51: I'm curious what does `.presence` do here?

phuongdh: `presence` returns nil when it's an empty string, since `("" || "local" = "").to_sym = :""`

phuongdh:
> Looks fine to me. Is there any difference between application and environment here? Maybe load order?

I pulled it out of development.rb because all environments (test is debatable) need to be configurable to fit different setup.

PR #4154, opened 2020-06-02
Cache s3 object locally for faster waveform job

phuongdh:


app/services/file_locator.rb
```diff
@@ -30,6 +30,14 @@ def initialize(uri)
     def object
       @object ||= Aws::S3::Object.new(bucket_name: bucket, key: key)
     end
+
+    def localize(dir = "/tmp")
+      new_dir = File.join dir, SecureRandom.uuid
+      new_path = File.join new_dir, File.basename(key)
+      FileUtils.mkdir_p new_dir
+      object.download_file new_path
```
cjcolvar, line 38: I'm curious.  Is this a blocking operation or does it download in the background?  If it is blocking, should we add a comment saying that it could be a long running method?

cjcolvar: There should also be a comment about cleaning up `/tmp` in case it uses up available space.

app/services/file_locator.rb
```diff
@@ -30,6 +30,14 @@ def initialize(uri)
     def object
       @object ||= Aws::S3::Object.new(bucket_name: bucket, key: key)
     end
+
+    def localize(dir = "/tmp")
+      new_dir = File.join dir, SecureRandom.uuid
+      new_path = File.join new_dir, File.basename(key)
+      FileUtils.mkdir_p new_dir
```
cjcolvar, line 37: What about using Tempfile instead of manually handling it in `/tmp`?  Maybe also add a method like `destroy_local_copy`?
https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close

phuongdh: Auto delete would be nice but we don't know when ffmpeg would be done with the file. How would you implement that?

app/services/file_locator.rb
phuongdh: It's blocking. I can add a comment.

spec/jobs/waveform_job_spec.rb
```diff
@@ -21,7 +21,7 @@
   let(:service) { instance_double("WaveformService") }
   let(:derivative_path) { URI.parse(master_file.derivatives.first.absolute_location).path }
 
-  describe "perform" do
+  describe "perform" do
```
cjcolvar, line 24:
```suggestion
  describe "perform" do
```

PR #4160, opened 2020-06-03
Fix for breaking autoplay in playlists

Dananji:


app/assets/javascripts/media_player_wrapper/mejs4_plugin_playlist_items.es6
```diff
@@ -286,7 +286,12 @@ Object.assign(MediaElementPlayer.prototype, {
      * @return {Object} jQuery object
      */
     getNextItem() {
-      return this.$nowPlayingLi.next('li');
+      let nextItem = this.$nowPlayingLi.next('li');
+      // When next item is not a valid playlist item (e.g. from a deleted item) get the item after that
+      if(nextItem[0].className != 'queue') {
+        nextItem =  nextItem.next('li');
+      }
```
cjcolvar, line 293: Code Climate suggests this change.  What do you think?
```suggestion
      if(nextItem[0].className !== 'queue') {
        nextItem = nextItem.next('li');
      }
```

Dananji: Yeah sounds good :+1: I will add this.

PR #4169, opened 2020-06-08
Show error messages when creating/updating collections

Dananji:
How errors are shown when the modal is closed;

When there is only one error:
Screenshot from 2020-06-09 14-23-50

When there are multiple errors:
Screenshot from 2020-06-09 14-23-19

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -38,6 +38,11 @@ body {
   }
 }
 
+/* Override Bootstrap CSS */
+.alert {
+  margin-bottom: 10px;
+}
+
 main {
   padding-bottom: 20px;
 }
--- a/app/views/admin/collections/_form.html.erb
+++ b/app/views/admin/collections/_form.html.erb
@@ -46,7 +46,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% content_for :page_scripts do %>
 <script>
   function inject_alert(message) {
-    $('#header-navbar').closest('.container-fluid').after(
+    $('.page-title').closest('.page-title-wrapper').before(
       '<div class="container-fluid" id="alerts"><div class="alert alert-danger"><button type="button" class="close" data-dismiss="alert">&times;</button>' +
       message + '</div></div>');
   }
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/avalon.scss
```diff
@@ -38,6 +38,11 @@ body {
   }
 }
 
+/* Override Bootstrap CSS */
+.alert {
+  margin-bottom: 10px;
+}
+
```
cjcolvar, line 45: Does this negatively impact any other pages with alert messages?

app/views/admin/collections/_form.html.erb
```diff
@@ -46,15 +46,15 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+    $('.page-title').closest('.page-title-wrapper').before(
       '<div class="container-fluid" id="alerts"><div class="alert alert-danger"><button type="button" class="close" data-dismiss="alert">&times;</button>' +
       message + '</div></div>');
   }
 
   $('form[data-remote][modal]').on('ajax:complete', function (jq_event, xhr) {
     if (xhr.responseJSON.hasOwnProperty('errors')) {
       $(this).closest('.modal').modal('hide')
-      inject_alert(xhr.responseJSON['errors'].join('<br/>'));
+      inject_alert(xhr.responseJSON['errors'].join(' '));
```
cjcolvar, line 57: Does a space end up rendering better than a line break when there are multiple errors in the response?

app/assets/stylesheets/avalon.scss
Dananji: I checked this in other pages and seems to not have any negative effect.

app/views/admin/collections/_form.html.erb
Dananji: Oh it didn't cross my mind. But for the collections at the moment name is the only attribute that user can miss when submitting this form right?

cjcolvar: I think there can be multiple errors.  Most likely scenario where this might happen is if the collection name is missing or not unique and the contact email/url is improperly formatted.  I don't think it is possible to get an error on the unit or managers with the current UI.

https://github.com/avalonmediasystem/avalon/blob/develop/app/models/admin/collection.rb#L28-L32

Dananji: Yes :+1: I will fix this.

PR #4179, opened 2020-06-10
Remove 'Remove' button for structure and captions when there is no structure/captions

Dananji:
For structure:

Screenshot from 2020-06-10 15-33-31

For captions:

Screenshot from 2020-06-10 15-34-06


cjcolvar:
I like how the label changes from Upload to Replace for captions.  Would you please do the same for structure?

PR #4216, opened 2020-07-07
Add avalon:user:admin rake task

mcritchlow:
When using an omniauth provider such as google, it is helpful to be able
to assign initial administrators to an Avalon application. This commit
adds a rake task to prompt the user for an email address of a user to
assign the administrator role.

Usage:
```
[avalon@lib-avalon-dev current]$ bundle exec rake avalon:user:admin
Assign user as an administrator
Email address for user: mcritchlow@ucsd.edu
Successfully assigned mcritchlow as an administrator
```

If a user does not exist in the system:
```
[avalon@lib-avalon-dev current]$ bundle exec rake avalon:user:admin
Assign user as an administrator
Email address for user: notauser@example.com
User with email address notauser@example.com not found
```

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -290,6 +290,26 @@ EOC
 
       puts "Updated password for user #{username}"
     end
+
+    desc "Assign user an an administrator"
+    task admin: :environment do
+      puts "Assign user as an administrator"
+      print "Email address for user: "
+      email_address = $stdin.gets.chomp
+      begin
+        new_administrator = User.find_by_email(email_address).user_key
+      rescue NoMethodError
+        abort "User with email address #{email_address} not found"
+      end
+      admin_group = Admin::Group.find('administrator')
+      if admin_group.users.any? new_administrator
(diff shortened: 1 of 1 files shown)


mcritchlow:
@phuongdh - Ok, hopefully this `develop` based PR will do the trick. Thanks again for the info. I'll know better for next time.

mcritchlow:
@phuongdh - The codeclimate error about the ruby hash syntax, would you like me to fix that? I followed the same convention used in the existing rake file, but whatever you think makes sense.

phuongdh:
> @phuongdh - The codeclimate error about the ruby hash syntax, would you like me to fix that? I followed the same convention used in the existing rake file, but whatever you think makes sense.

The newer tasks in there seem to not use the hash rocket so it'd be better to use the new style. I'll come around and fix the rest later. Thanks

mcritchlow:
> The newer tasks in there seem to not use the hash rocket so it'd be better to use the new style. I'll come around and fix the rest later. Thanks

Ah, yeah I've seen that before. Ok, codeclimate is green.

PR #4223, opened 2020-07-09
Delete dropbox directory when deleting a collection

Dananji:
Deletes collection's dropbox on S3 bucket/file system based on the settings when the collection is deleted. And shows the user the following warning when deleting the collection.

Screenshot from 2020-07-29 15-05-26

I'm not sure whether we should add something to the reassign note to make sure the user is aware that the dropbox files are not re-assigned but deleted permanently. @joncameron what do you think?

### DIFF
--- a/app/jobs/delete_dropbox_job.rb
+++ b/app/jobs/delete_dropbox_job.rb
@@ -17,13 +17,8 @@ class DeleteDropboxJob < ActiveJob::Base
   queue_as :default
   def perform(path)
     Rails.logger.debug "Attempting to delete dropbox directory #{path}"
-    locator = FileLocator.new(path)
     begin
-      if Settings.dropbox.path.match? %r{^s3://}
-        locator.destroy_s3_dropbox_directory(path)
-      else
-        locator.destroy_fs_dropbox_directory(path)
-      end
+      FileLocator.remove_dir(path)
     rescue StandardError => err
       Rails.logger.warn "Error deleting dropbox directory #{path}: #{err.message}"
     end
--- a/app/jobs/delete_dropbox_job.rb
+++ b/app/jobs/delete_dropbox_job.rb
@@ -17,10 +17,6 @@ class DeleteDropboxJob < ActiveJob::Base
   queue_as :default
   def perform(path)
     Rails.logger.debug "Attempting to delete dropbox directory #{path}"
-    begin
-      FileLocator.remove_dir(path)
-    rescue StandardError => err
-      Rails.logger.warn "Error deleting dropbox directory #{path}: #{err.message}"
-    end
+    FileLocator.remove_dir(path)
   end
 end
--- a/app/jobs/delete_dropbox_job.rb
+++ b/app/jobs/delete_dropbox_job.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+# Copyright 2011-2020, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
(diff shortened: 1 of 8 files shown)


app/models/admin/collection.rb
```diff
@@ -298,6 +316,11 @@ def create_s3_dropbox_directory!
       self.dropbox_directory_name = name
     end
 
+    def destroy_s3_dropbox_directory!
+      objects = Aws::S3::Bucket.new(name: Settings.encoding.masterfile_bucket).objects(prefix: "/dropbox/#{dropbox_directory_name}")
```
phuongdh, line 320: Same as above regarding `dropbox_absolute_path`

app/models/admin/collection.rb
```diff
@@ -298,6 +316,11 @@ def create_s3_dropbox_directory!
...
+      objects.batch_delete!
```
phuongdh, line 321: This could take a long time if the collection is large. IMO it should be in a background job to prevent blocking user.

app/models/admin/collection.rb
```diff
@@ -271,6 +281,14 @@ def create_dropbox_directory!
       end
     end
 
+    def destroy_dropbox_directory!
+      if Settings.dropbox.path =~ %r(^s3://)
+        destroy_s3_dropbox_directory!
+      else
+        destroy_fs_dropbox_directory!
+      end
```
phuongdh, line 289: IMO these (and the other) directory handling logic would be better off in a service class (FileLocator or something like that)

Dananji:
TODO:: Delete S3 dropbox when there are no files.

app/models/admin/collection.rb
```diff
@@ -208,6 +210,17 @@ def dropbox_absolute_path( name = nil )
     File.join(Settings.dropbox.path, name || dropbox_directory_name)
   end
 
+  def dropbox_object_count
+    if Settings.dropbox.path =~ %r(^s3://)
+      # Build the prefix with dropbox directory name and collection name
+      obj_prefix = dropbox_absolute_path.split('/').last(2).join('/')
```
phuongdh, line 216: `URI.parse(path).host` or `.path`

app/jobs/delete_dropbox_job.rb
```diff
@@ -0,0 +1,31 @@
...
+  queue_as :default
+  def perform(path)
+    Rails.logger.debug "Attempting to delete dropbox directory #{path}"
+    locator = FileLocator.new(path)
+    begin
+      if Settings.dropbox.path.match? %r{^s3://}
+        locator.destroy_s3_dropbox_directory(path)
+      else
+        locator.destroy_fs_dropbox_directory(path)
+      end
```
phuongdh, line 26: Extract to service class

app/services/file_locator.rb
```diff
@@ -124,4 +124,24 @@ def attachment
       location
     end
   end
+
+  def destroy_s3_dropbox_directory(dropbox_path)
```
phuongdh, line 128: Let's use a more generic name

Dananji:
@phuongdh The tests fail with the message `Too long with no output (exceeded 10m0s): context deadline exceeded`. Do you know how to fix this?

Dananji:
@phuongdh I skipped the tests I added in this PR and ran the test suite in my machine and the same thing happened. And new tests pass when I run those 2 test files separately.

phuongdh:
> @phuongdh I skipped the tests I added in this PR and ran the test suite in my machine and the same thing happened. And new tests pass when I run those 2 test files separately.

Some other tests could be messing up the configurations needed for your tests. How did you run the tests locally? This seems like a CircleCI specific thing https://support.circleci.com/hc/en-us/articles/360045268074-Build-Fails-with-Too-long-with-no-output-exceeded-10m0s-context-deadline-exceeded-

Dananji:
@phuongdh To run the whole test suite I ran `docker-compose exec test bash -c "bundle exec rspec`

spec/services/file_locator_spec.rb
```diff
@@ -121,4 +121,66 @@
...
+        allow(File).to receive(:exist?).with(file_path) { true }
+        expect(locator.exist?).to be_truthy
+        FileLocator.remove_dir("file://tmp/dropbox/test")
+        allow(File).to receive(:exist?).with(file_path) { false }
+        expect(locator.exist?).to be_falsey
+      end
+    end
+  
+    describe '#remove_s3_dir' do
+      let(:old_bucket) { Settings.encoding.masterfile_bucket }
```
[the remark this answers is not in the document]
Dananji: Oh okay :+1: Thanks! I'll keep that in mind the next time :100:
Since this PR has been going on for a couple of weeks I'll merge this for now.

PR #4227, opened 2020-07-13
Collection pounding names

### LINKED ISSUE #4172: Collection Names with Pound Sign
### Description
Collection names that include pound signs (#) are problematic. As an example, in the case that two collections are created sequentially with these names:
- Test 1
- Test #1
Avalon will encounter an error because it will attempt to use the same name for the dropbox folder for each. The error here is most likely happening in the logic used for creating and naming directories in the Avalon dropbox directory.

This is probably the case for more characters than just #, but this was the circumstance that exposed the bad behavior.

### Done Looks Like
- Review sanitizer to make sure collisions don't occur
- Make changes as needed

### Text of other issue created for the same behavior:
Environment: S3 (Aws, Minio)

Steps:

    Create Collection "Test 1"
    Create Collection "Test #1"

Expected:

    2 directories created in dropbox folder

Observed:

    Only 1 directory created.

Cause: s3 strip special chars from collection name, thus colliding with existing directory.

SumithBaddam:
Fixes #4172 

Added a section of code to replace the S3 avoidable special characters for object creation to be replaced with '_'

### DIFF
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -296,7 +296,6 @@ class Admin::Collection < ActiveFedora::Base
       avoid_special_chars.each { |spcl_char|
         name = name.sub(spcl_char, "_")
       }
-      #name = name.sub("#", "_")
       absolute_path = base_uri.join(name).to_s + '/'
       obj = FileLocator::S3File.new(absolute_path).object
       Aws::S3::Client.new.put_object(bucket: obj.bucket_name, key: obj.key)
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -292,6 +292,11 @@ class Admin::Collection < ActiveFedora::Base
         obj = FileLocator::S3File.new(base_uri.join(n).to_s + '/').object
         obj.exists?
       end
+      avoid_special_chars = ['{', '}', '^', '#', '%', '`', '[', ']', '!', "'", '"', '>', '<', '~', '|']
+      avoid_special_chars.each { |spcl_char|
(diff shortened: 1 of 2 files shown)


app/models/admin/collection.rb
```diff
@@ -292,6 +292,10 @@ def create_s3_dropbox_directory!
         obj = FileLocator::S3File.new(base_uri.join(n).to_s + '/').object
         obj.exists?
       end
+      avoid_special_chars = ['{', '}', '^', '#', '%', '`', '[', ']', '!', "'", '"', '>', '<', '~', '|']
```
Dananji, line 295: @SumithBaddam Do you think using a regex like in this [line](https://github.com/avalonmediasystem/avalon/blob/6c54fe5736ac7aeafe20d1cd7a14a92829bc2a0e/app/models/admin/collection.rb#L267) instead of this array would look more clean?

Dananji:
This looks good! If you can you please add a test that'd be good I think. And can you remove `db/schema.rb` file from the PR?

app/models/admin/collection.rb
Dananji: Something like this;
``name.gsub(/[{}^#`%\[\]!'"><~|]/,  '_')``

phuongdh:
@SumithBaddam can you remove the db/schema.rb changes from this PR? It may break MySQL.

SumithBaddam:
This is an old PR I guess. I actually created a new one with the test scenarios. It is here - https://github.com/avalonmediasystem/avalon/pull/4230

PR #4237, opened 2020-07-21
Require Git SCM plugin for Capistrano

mcritchlow:
To prevent deprecation warnings such as:

```
[Deprecation Notice] Future versions of Capistrano will not load the Git SCM
plugin by default. To silence this deprecation warning, add the following to
your Capfile after `require "capistrano/deploy"`:
    require "capistrano/scm/git"
    install_plugin Capistrano::SCM::Git
```

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -3,6 +3,8 @@ require "capistrano/setup"
 
 # Include default deployment tasks
 require "capistrano/deploy"
+require "capistrano/scm/git"
+install_plugin Capistrano::SCM::Git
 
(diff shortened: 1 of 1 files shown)


phuongdh:
@mcritchlow thank you for the PR. If you can fix the one CodeClimate complaint we'll be good to go.

mcritchlow:
@phuongdh - Derp. Yeah, sorry about that. Pushed a fixed commit w/o the extra line.

PR #4239, opened 2020-07-23
Add an intermediate page when accessing restricted content

Dananji:
Co-authored-by: Sumith Reddi Baddam <sumith.reddy2@gmail.com>

When user is not logged in:
Screenshot from 2020-07-23 15-12-24

After logging in and the user is still not authorized to access the content:
Screenshot from 2020-07-23 15-11-46

@joncameron When the user is not logged in the page has 2 `Sign in` buttons, do you think maybe we should make the `Sign in` in the text just a link?

### DIFF
--- a/app/views/errors/restricted_pid.html.erb
+++ b/app/views/errors/restricted_pid.html.erb
@@ -1,0 +1,26 @@
+<%#
+Copyright 2011-2020, The Trustees of Indiana University and Northwestern
+  University.  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed
+  under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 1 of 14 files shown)


app/views/errors/restricted_pid.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  CONDITIONS OF ANY KIND, either express or implied. See the License for the
+  specific language governing permissions and limitations under the License.
+---  END LICENSE_HEADER BLOCK  ---
+%>
+<div id="main_content_container" class="container-fluid">
+  <div class="content" class="col-md-8">
+    <h2>Restricted Content</h2>
+    <% if current_user.nil? %>
+      <p>You're not signed in. You may be able to view this item after signing in.</p>
+      <%= link_to 'Sign in', main_app.new_user_session_path, class: "btn btn-info" %>
```
phuongdh, line 21: Previously, after signing in the user would get redirected to the item, is that still the case here?

Dananji: If the user doesn't have permission to see the content after signing in; then he/she will be redirected to the restricted content page (this page) with the message `You are not authorized to access this content`. 
If not; the user will be redirected to the item page.

PR #4244, opened 2020-07-27
New SME build with bug fix for adding new timespans at root level of structure

Dananji:


Dananji:
This PR fixes the bug reported by users when adding timespans at the root level in SME.

PR #4250, opened 2020-07-31
Use okcomputer gem for health check

Dananji:
Co-authored-by: Phil Dinh <phuongdh@gmail.com>

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,0 +1,58 @@
+require "net/https"
+require "uri"
+
+class AvalonCheck < OkComputer::Check
+  def time_limit
+    5
+  end
+
+  def status
+    Timeout.timeout(time_limit) do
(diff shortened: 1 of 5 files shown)


config/initializers/okcomputer.rb
```diff
@@ -1,58 +1,61 @@
-require "net/https"
-require "uri"
-
-class AvalonCheck < OkComputer::Check
-  def time_limit
-    5
-  end
+if defined? OkComputer
```
Dananji, line 1: @phuongdh I understood the other changes, but this one. I have 2 questions;
1. What does this do?
2. What happens if this fails? Does it fall back to `about-page` implementation?

phuongdh: This is needed because we only use okcomputer in production, so it's not available in development & test and would break the app without this check.

PR #4261, opened 2020-08-26
Generate and use an empty waveform file in SME when masterfile is missing the waveform.json

Dananji:
Screenshot from 2020-08-26 13-26-36

When the masterfile is missing the waveform.json data, SME makes a request through Peaks.js to Avalon to get an empty set of waveform data like in the attached screenshot. And in addition to that SME will show an alert with the message `Requested waveform data is not available`.
NOTE: In the attached screenshot the derivative in the masterfile is missing. Therefore adding new timespans is disabled since it requires player in Peaks instance, which is created upon successful fetching of the derivative.

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -43,12 +43,7 @@ class MasterFilesController < ApplicationController
     @master_file = MasterFile.find(params[:id])
     authorize! :read, @master_file
 
-    ds = @master_file.waveform
-
-    # Generate empty waveform data
-    if params[:empty]
-      ds = WaveformService.new(8, samples_per_frame).empty_waveform(@master_file)
-    end
+    ds = params[:empty] ? WaveformService.new(8, samples_per_frame).empty_waveform(@master_file) : @master_file.waveform
 
     if ds.nil? || ds.empty?
       render plain: 'Not Found', status: :not_found
@@ -433,6 +428,6 @@ protected
   end
 
   def samples_per_frame
-    Settings.waveform.sample_rate * Settings.waveform.finest_zoom_in_sec / Settings.waveform.player_width_in_px
+    Settings.waveform.sample_rate * Settings.waveform.finest_zoom / Settings.waveform.player_width
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/master_files_controller.rb
```diff
@@ -42,7 +42,14 @@ def captions
   def waveform
     @master_file = MasterFile.find(params[:id])
     authorize! :read, @master_file
+
     ds = @master_file.waveform
```
phuongdh, line 46: I don't think this is needed when generating an empty waveform (params[:empty] exists)

app/jobs/waveform_job.rb
```diff
@@ -15,9 +15,9 @@
 class WaveformJob < ActiveJob::Base
   queue_as :waveform
 
-  PLAYER_WIDTH_IN_PX = 1200
-  FINEST_ZOOM_IN_SEC = 5
-  SAMPLES_PER_FRAME = (44_100 * FINEST_ZOOM_IN_SEC) / PLAYER_WIDTH_IN_PX
+  PLAYER_WIDTH_IN_PX = Settings.waveform.player_width_in_px
+  FINEST_ZOOM_IN_SEC = Settings.waveform.finest_zoom_in_sec
+  SAMPLES_PER_FRAME = (Settings.waveform.sample_rate * FINEST_ZOOM_IN_SEC) / PLAYER_WIDTH_IN_PX
```
phuongdh, line 20: Cool, good idea to make it configurable.

config/settings.yml
```diff
@@ -85,3 +85,7 @@ auth:
 # google_analytics_tracking_id: "someid"
 supplemental_files:
   proxy: false
+waveform:
+  player_width_in_px: 1200
+  finest_zoom_in_sec: 5
```
phuongdh, line 90: Minor point, but since no other settings have unit in their names, should we shorten the names and add a comment, ie 
```
player_width: 1200 #pixels
finest_zoom: 5 #seconds
```

phuongdh:
Looks like you need to rebase develop into this branch.

PR #4271, opened 2020-09-14
Display access restrictions on item view page

phuongdh:


Dananji (approved):
This looks great :+1:

PR #4293, opened 2020-09-22
Run cron jobs on Sidekiq

phuongdh:
Simplify deployment & management


Dananji:
@phuongdh Question: why do we have an empty file `config/initializers/worker_cron.rb` ?

Dananji (approved):
Looks good :+1:

PR #4296, opened 2020-09-24
Fix bug in timeline creation when using a custom scope

Dananji:
When custom scope is given to create a new timeline and initialize it with existing structure, it checks for the existing timespans in the structure. With these timespans it compiles an array called `structures` which goes into the timeline's manifest.

The bug here was, when the custom scope does not overlaps existing timespans/does not wrap existing timespans the `structures` array returned is empty. This empty array is then used to create the final structure for the timeline and it assumes the `structures` array is not empty.


joncameron (dismissed):
Looks good!

phuongdh (approved):
:+1:

PR #4298, opened 2020-09-30
Move timeline configuration from abilities to view

Dananji:
When timeline configuration is not in place, the user can still type in the address bar `/timelines` and use that route. Because when we move configuration check from abilities the route is not restricted to the other (not admin) users anymore, and this doesn't take them to the restricted content page. So I put in a piece of code that will show something like below;

Screenshot from 2020-09-30 15-08-07

This is what I came up with, but there must be better ways to do this :)
@joncameron @phuongdh What do you think?

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -18,7 +18,7 @@ class Ability
   include Hydra::MultiplePolicyAwareAbility
 
   self.ability_logic += [ :playlist_permissions, :playlist_item_permissions, :marker_permissions, :encode_dashboard_permissions ]
-  self.ability_logic += [ :timeline_permissions ] if Settings['timeliner'].present?
+  self.ability_logic += [ :timeline_permissions ]
 
   def encode_dashboard_permissions
     can :read, :encode_dashboard if is_administrator?
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -27,7 +27,7 @@ Unless required by applicable law or agreed to in writing, software distributed
     <%= link_to 'Playlists', main_app.playlists_path, id:'playlists_nav' %>
   </li>
   <% end %>
-  <% if current_ability.can? :create, Timeline %>
+  <% if (current_ability.can? :create, Timeline) && Settings['timeliner'].present? %>
   <li class=<%= active_for_controller('timelines') %>>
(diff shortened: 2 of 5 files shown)


phuongdh:
That's better than not having it. Although I dont know why we need the `timeliner_url` config. Isn't it always mounted at `/timeliner`? https://github.com/avalonmediasystem/avalon/blob/develop/config/routes.rb#L207

joncameron:
👍

phuongdh:
From @joncameron  "I’m pretty sure this is an artifact of the timeliner originally being a standalone application hosted outside Avalon. I’m not sure if that’s possible at this point, but my guess would be that really isn’t possible or needed anymore since we pulled the Timeliner inside Avalon, and as you noted that route doesn’t change."

I guess the only use for this setting now is allowing Avalon to use a standalone timeliner instead of the built-in one, which seems like an unlikely scenario. 

Can we remove this setting? Or if it's needed as a toggle then it should be named accordingly. Of course that leads to the question of why does it need to be toggle-able.

Dananji:
Yes it seems timeliner is always mounted at `/timeliner` regardless of the setting being present or not, therefore the route exists all the time. And [here](https://github.com/avalonmediasystem/avalon/blob/develop/app/controllers/timelines_controller.rb#L110) Ruby code reads the URL from settings for the iframe where the React app is rendered.

We can do the following;
1. Rename the setting to reflect that it is being used to turn on/off timelines feature in Avalon
2. Build the `/timeliner` route conditionally based on that setting (like we do with [variations](https://github.com/avalonmediasystem/avalon/blob/develop/config/routes.rb#L181-L183)). That way `/timeliner` route is not mounted at all times
3. Use `timeliner_url` [in the code](https://github.com/avalonmediasystem/avalon/blob/develop/app/controllers/timelines_controller.rb#L110) instead of reading it from settings
4. For the case where the timelines is not turned on, we can show the above message

This makes timelines feature toggle-able, and it will be useful when someone is looking to use Avalon without timelines (I'm not sure if there are such use cases) and we are not exposing `/timeliner` route all the time.   
@phuongdh What do you think?

phuongdh:
@Dananji Jon agreed to remove this setting altogether, and as a result always display Timelines https://github.com/avalonmediasystem/avalon/issues/4300

Dananji:
Closing in favor of #4300

PR #4310, opened 2020-10-15
Fixes #3566

### LINKED ISSUE #3566: Links to Non-existent Fedora 3 Identifiers Return 500
### Description
Requests for media objects with Fedora 4 identifiers returns a not found message if not found, but Fedora 3 identifiers, if there is no map to a Fedora 4 ID in Avalon, will return a 500. Fedora 3 identifiers are in the format "avalon:#value"

### Examples
https://media.dlib.indiana.edu/media_objects/avalon:293820
https://spruce.dlib.indiana.edu/media_objects/avalon:2938

### Done Looks Like
- Requests with Fedora 3-style IDs for the media object route (as in example URLs) above redirect to the Avalon homepage with an "Item not found" message to the user

SumithBaddam:
Fixes #3566
Validating if there is no map to a Fedora 4 ID in Avalon and if not it will throw an item not found exception.


phuongdh:
@SumithBaddam you should make the PR to `develop` instead of `master`.

SumithBaddam:
> @SumithBaddam you should make the PR to `develop` instead of `master`.
> 
> Edit: looks like you committed directly to `develop` first and made a PR from develop to develop?

I worked on the develop branch by mistake. So when I committed and pushed it might have got committed. I was thinking I wouldn't have direct access to develop branch and this would be ignored. So discarded it and created a new PR.

phuongdh:
Closing this PR to make a new one

PR #4314, opened 2020-10-19
Links to Non-existent Fedora 3 Identifiers Return 500

### LINKED ISSUE #3566: Links to Non-existent Fedora 3 Identifiers Return 500

SumithBaddam:
Fixes #3566 #4310 
Added a condition to throw ObjectNotFound exception when NIL is returned by the query.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -51,11 +51,11 @@ class ApplicationController < ActionController::Base
     return if params[:controller] =~ /migration/
 
     params.permit!
-    new_id = ActiveFedora::SolrService.query(%{identifier_ssim:"#{params[:id]}"}, rows: 1, fl: 'id')
-    if new_id.size == 0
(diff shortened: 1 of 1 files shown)


phuongdh:
@SumithBaddam  this PR was based on the bad develop branch, I made another PR with your commit on top of the clean develop here https://github.com/avalonmediasystem/avalon/pull/4315

PR #4316, opened 2020-10-26
Remove timeliner_url setting in config

Dananji:


phuongdh (approved):
Looks good. Is this part of 7.2?

Dananji:
No I don't think this was intended to be included in 7.2

PR #4323, opened 2020-11-03
Handle request for deleted m3u8

### LINKED ISSUE #3395: HLS manifest requests for deleted files returns 500
This should return a 410 Gone instead. Looks like we need to add a clause to the if/else in ApplicationController for m3u8 formats to avoid this error.  Probably just have it respond with `return head(:gone)`
https://app.honeybadger.io/projects/54117/faults/37819665

phuongdh:
Fixes #3395


Dananji (approved):
Looks good :+1:

PR #4336, opened 2020-12-09
Stop player when it reaches last playlist item's end time

Dananji:
This PR contains a fix for updating time rail highlight (displaying structure in the player) when media sources switch.


phuongdh (approved):
:+1:

PR #4356, opened 2020-12-30
Show a flash message when trying to publish an item w/o title/issued date

Dananji:
When the item doesn't have a title/date issued the following flash message is shown;

Screenshot from 2021-01-14 12-58-51

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -143,17 +143,19 @@ class BookmarksController < CatalogController
   end
 
   def status_action documents
-    errors = []
-    success_ids = []
-    status = params['action']
+    errors, success_ids = [], [], []
     Array(documents.map(&:id)).each do |id|
       media_object = MediaObject.find(id)
       if cannot? :update, media_object
         errors += ["#{media_object.title} (#{id}) #{t('blacklight.messages.permission_denied')}."]
       else
-        case status
+        case params['action']
         when 'publish'
-          success_ids << id
+          if media_object.title.nil? || media_object.date_issued.nil?
+            errors += ["#{id}, Unable to Publish Item. Missing required fields."]
+          else
+            success_ids << id
+          end
         when 'unpublish'
           if can? :unpublish, media_object
             success_ids << id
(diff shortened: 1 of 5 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -587,6 +581,32 @@ def set_active_file
     return @currentStream
   end
 
+  def update_publish(media_object)
+    errors = []
+    success_count = 0
+    if media_object.title.nil? || media_object.date_issued.nil?
+      errors += ["#{media_object.id}, missing required fields"]
+    else
```
phuongdh, line 589: I don't think this is necessary. You should already have the error messages in an `ActiveFedora::RecordInvalid` exception when calling `media_object.published!`. We can probably leave the previous code intact and add some error handling to `update_status`.

app/controllers/media_objects_controller.rb
```diff
@@ -410,31 +410,25 @@ def destroy
   # it will just toggle the state.
   def update_status
     status = params[:status]
-    errors = []
-    success_count = 0
+    errors, update_errors = [], [], []
```
phuongdh, line 413: Also one too many `[]`

phuongdh (commented):
Please see comments below.

app/controllers/bookmarks_controller.rb
```diff
@@ -143,17 +143,19 @@ def add_to_playlist_action documents
   end
 
   def status_action documents
-    errors = []
-    success_ids = []
-    status = params['action']
+    errors, success_ids = [], [], []
```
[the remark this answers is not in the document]
Dananji: When I put the same number of `[]`s codeclimate complained about it

app/controllers/media_objects_controller.rb
Dananji: Ohh... I'll check it out :+1: Thanks!

app/controllers/media_objects_controller.rb
Dananji: Same here codeclimate complained when there are equal number of `[]`s

app/controllers/media_objects_controller.rb
```diff
@@ -430,11 +431,14 @@ def update_status
             else
               errors += ["#{media_object.title} (#{id}) (permission denied)."]
             end
+          end
+        rescue ActiveFedora::RecordInvalid
+          errors += ["#{media_object.id}, validation failed."]
```
phuongdh, line 436: Can you use or include the error message from the exception itself? It lists all the fields that are missing so the user knows what to fix, otherwise it's unclear what needs to be done.

phuongdh (approved):

PR #4366, opened 2021-01-12
Transcripts UI

Dananji:
Transcript UI is using the same template used for the supplemental files UI.
And for this work transcripts UI is using supplemental files endpoints and storage. Once the transcript storage and endpoints are implemented the UI can be changed to use these endpoints (since transcripts have the same behavior as supplemental files).

Screenshot from 2021-01-12 15-59-21


app/controllers/master_files_controller.rb
```diff
@@ -206,6 +202,24 @@ def attach_captions
     end
   end
 
+  def delete_captions
+    if params[:id].blank? || (not MasterFile.exists?(params[:id]))
+      flash[:notice] = "MasterFile #{params[:id]} does not exist"
+    end
+    @master_file = MasterFile.find(params[:id])
```
phuongdh, line 209: I know you didn't start this pattern but many methods in this controller use this same code block. I think we should consolidate them into a private/protected method and use a `before_action` to hook it up.

phuongdh (commented):
Looks good overall. I have 1 small comment.

app/controllers/master_files_controller.rb
Dananji: Yeah sounds good :100:

phuongdh (approved):

PR #4399, opened 2021-02-16
Fix section edit reload

Dananji:
Editing the supplemental file labels the flash messages are shown inline:

When successfully saved,
Screencast from 02-16-2021 10_55_55 AM

When saving fails,
Screencast from 02-16-2021 10_59_33 AM

Uploading new supplemental/caption files reloads the page. The browser remembers the previously opened section accordions and collapse them after page reload:
Screencast from 02-16-2021 10_57_15 AM

### DIFF
--- a/app/assets/javascripts/supplemental_files.js
+++ b/app/assets/javascripts/supplemental_files.js
@@ -19,9 +19,8 @@
 /* Show form to edit label */
 $('button[name="edit_label"]').on('click', e => {
   const $row = getHTMLInfo(e, '.row');
-  const fileId = $row.data('file-id');
-  const masterfileId = $row.data('masterfile-id');
-  $row.find('small[name="flash-message-' + masterfileId + '-' + fileId + '"]').html('');
+  const { masterfileId, fileId } = $row[0].dataset;
+
   $row.addClass('is-editing');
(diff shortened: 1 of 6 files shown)


phuongdh (commented):
Very nice. I noticed that at certain browser size the notice pushes the buttons down, maybe the message can be shorter?

image

For consistency, can we get similar success notice when saving the section title? 

image

Btw, doing it this way if you open the sections and go somewhere else then come back to editing the item, the sections still remain open. This may even be desirable but I figure we should check with the PO.

Dananji:
> Very nice. I noticed that at certain browser size the notice pushes the buttons down, maybe the message can be shorter?
> 
> image
> 
> For consistency, can we get similar success notice when saving the section title?

Oh right! I'll make the message shorter and change the styles to make it look like the section title edit flash message :+1: 

> 
> image
> 
> Btw, doing it this way if you open the sections and go somewhere else then come back to editing the item, the sections still remain open. This may even be desirable but I figure we should check with the PO.

Hmm good point.
@joncameron Your thoughts on this is appreciated. Thanks!

joncameron:
@Dananji This looks great! I think it's OK that the sections are still open. A worthy trade-off for not having everything permanently collapse on save.

phuongdh (approved):

PR #4407, opened 2021-02-25
Add Amazon S3 for ActiveStorage

phuongdh:


Dananji:
@phuongdh Is this related to the issue with supplemental files giving 500, the one you fixed in MCO?

Dananji (approved):
:+1:

phuongdh:
> @phuongdh Is this related to the issue with supplemental files giving 500, the one you fixed in MCO?

It is somewhat related but more for cloud deployments.

PR #4415, opened 2021-03-11
Add local storage config for Docker deploy

phuongdh:


apertome (approved):
I tested enabling local storage, and it worked. I also tried configuring the path, which also worked. No issues found.

This looks good.

PR #4418, opened 2021-03-14
Allow different DB engines in Dockerfile

phuongdh:


phuongdh:
Closing this one as it's probably more suitable for the IU github.

PR #4443, opened 2021-04-23
Run rails with bundle exec to fix an error in some environments

apertome:
Some users (me) were seeing an error in the avalon container "rails not found." This fixes that issue.


Dananji (approved):
:+1:

PR #4488, opened 2021-06-24
Staging fix mediainfo

apertome:
Making mediainfo installation less fragile (intsall via apt rather than pulling down directly from website)


apertome:
Ready to merge for futher testing/deployment

mcwhitaker (approved):

PR #4535, opened 2021-09-10
Cast uri to a URI to ensure it responds to `#scheme` in WaveformService.get_waveform_json

cjcolvar:
This should fix the issue reported in VOV-6325.  I manually tested this in a rails console with an MDPI item and it successfully get wave normalized peaks data.


apertome (approved):
Looks good

PR #4553, opened 2021-10-18
Change Docker base images to Debian Bullseye/Ruby 2.7

apertome:
This PR changes the base OS for the Docker images to Bullseye, and the Ruby version to 2.7. So far it seems to work in my testing, and most or all of the rspec tests pass.

One caveat I've seen is that on both hosts I checked, the Streaming server checks on the `about` page fail and show up in red. I am not sure whether this worked prior to these changes.

From 
[http://localhost:3000/about](http://localhost:3000/about)
streaming_error


Gemfile
```diff
@@ -175,3 +175,5 @@ end
 
 extra_gems = File.expand_path("../Gemfile.local", __FILE__)
 eval File.read(extra_gems) if File.exist?(extra_gems)
+
+gem "omniauth-saml", "~> 1.10"
```
cjcolvar, line 179: Is including `omniauth-saml` intentional?  It feels outside the scope of this PR to me.

config/application.rb
```diff
@@ -8,7 +8,7 @@
 Bundler.require(*Rails.groups)
 
 module Avalon
-  VERSION = '7.2.1'
+  VERSION = '7.2.2'
```
cjcolvar, line 11: Do you think this should be `7.3`?  I can't remember if there are any new features being added or if this is all bug fixes and dependency upgrades.

Gemfile
apertome: I didn't change anything relating to circleCI. I did rename the image to 7.2.2 something; I guess we either need to update the image name circleci is using, or  build a new image using the same name.

Do we really want the version number hard coded in the circleci image name?

app/services/file_locator.rb
```diff
@@ -23,8 +23,8 @@ class S3File
 
     def initialize(uri)
       uri = Addressable::URI.parse(uri)
-      @bucket = URI.decode(uri.host)
-      @key = URI.decode(uri.path).sub(%r(^/*(.+)/*$),'\1')
+      @bucket = Addressable::URI.escape(uri.host)
+      @key = Addressable::URI.escape(uri.path).sub(%r(^/*(.+)/*$),'\1')
```
cjcolvar, line 27:
```suggestion
      @bucket = Addressable::URI.unencode(uri.host)
      @key = Addressable::URI.unencode(uri.path).sub(%r(^/*(.+)/*$),'\1')
```

Gemfile
```diff
@@ -21,47 +21,50 @@ gem 'uglifier', '>= 1.3.0'
...
 # Samvera version pins
 gem 'blacklight', '< 7.0'
-gem 'rdf', '~> 2.2'
+gem 'rdf', '~> 3.1'
 gem 'rsolr', '~> 1.0'
 
 # Rails & Samvera Plugins
 gem 'about_page', git: 'https://github.com/avalonmediasystem/about_page.git', tag: 'avalon-r6.5'
-gem 'active_annotations', '~> 0.2.2'
+gem 'active_annotations', git: 'https://github.com/avalonmediasystem/active_annotations.git', branch: 'rails5.2'
```
cjcolvar, line 39:
```suggestion
gem 'active_annotations', '~> 0.3.0'
```

cjcolvar (approved):
:tada:

PR #4562, opened 2021-11-02
Fix sorting of the last access column on manage users page

Dananji:
The `Last Access` column now sorts in chronological order instead of alpabetical order;

desc
asc


apertome (approved):
Looks good

PR #4565, opened 2021-11-04
Change playlist item source link to masterfile route from mediaobject route

Dananji:


cjcolvar:
@Dananji @joncameron What about retaining the original behavior of linking to the media object but add an additional link to the section?  E.g. "[9-bard-smedsrod.mp3](https://demo.avalonmediasystem.org/media_objects/9z903010x/section/6w924c02q) from [Bla bla bla](https://demo.avalonmediasystem.org/media_objects/9z903010x)"

Dananji:
That sounds good to me :+1:

joncameron:
@Dananji will make the change suggested by Chris

cjcolvar (approved):

PR #4569, opened 2021-11-29
IIIF transcript integration

Dananji:
transcript-avalon

### DIFF
--- a/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
+++ b/app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
@@ -31,6 +31,7 @@ class MEJSPlayer {
     this.mejsMarkersHelper = new MEJSMarkersHelper();
     this.mejsQualityHelper = new MEJSQualityHelper();
     this.localStorage = window.localStorage;
+    this.canvasIndex = 0;
 
     // Unpack player configuration object for the new player.
     // This allows for variable params to be sent in.
@@ -109,7 +110,8 @@ class MEJSPlayer {
     let itemScope = document.querySelector('[itemscope="itemscope"]');
     let node = this.mejsUtility.createHTML5MediaNode(
       this.mediaType,
-      this.currentStreamInfo
+      this.currentStreamInfo,
+      this.canvasIndex
     );
 
     // Mount new <audio> or <video> element to the DOM and initialize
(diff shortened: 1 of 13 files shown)


config/routes.rb
```diff
@@ -152,6 +152,7 @@
       post 'structure', to: 'master_files#set_structure', constraints: { format: 'json' }
       delete 'structure', to: 'master_files#delete_structure', constraints: { format: 'json' }
       post 'move'
+      get 'transcript/:t_id/:filename', to: 'master_files#transcript'
```
cjcolvar, line 155: It doesn't look like you're using the `:filename` route component in the controller action.  Could it be dropped?
I think that would make the url be more standard which might make the testing a little more straight forward as well.

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -671,7 +680,7 @@ class MEJSPlayer {
         let fullConfiguration = Object.assign({}, defaults, markerConfig);
         // Create a MediaElement instance
         this.player = new MediaElementPlayer(
-          `mejs-avalon-${this.mediaType}`,
+          'mejs-avalon-player',
```
cjcolvar, line 683: This is interesting.  Was the distinction between audio and video never being checked or used in the player code?

app/assets/javascripts/media_player_wrapper/mejs4_helper_utility.es6
```diff
@@ -41,37 +41,32 @@ class MEJSUtility {
       node.classList.add('invisible');
 
       // Add <source>s
-      currentStreamInfo.stream_hls.map(source => {
-        markup += `<source src="${
-          source.url
-        }" type="application/x-mpegURL" data-quality="${source.quality}"/>`;
+      currentStreamInfo.stream_hls.map((source) => {
+        markup += `<source src="${source.url}" type="application/x-mpegURL" data-quality="${source.quality}"/>`;
```
cjcolvar, line 45: Collapsing the markup like this one make the lines a lot more readable!

cjcolvar (commented):
This is looking really good and I think it will be ready for upgrading past the beta for the player once the routes question is resolved.

config/routes.rb
Dananji: Yes, `:filename` is not used in the controller action, but this url is being used in the player side to identify the file type with respect to the file extension and it is used to decide which parsing action to take. That's why I included this in there. Maybe I can use `Content-type` in the response in the player side instead of filename (which is more stable), so that the url looks much cleaner?

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
Dananji: We use the distinction when building the player, but it was not reflected in the player id

cjcolvar (approved):
Looks great!

PR #4573, opened 2021-12-13
Move files not already in the dropbox into the dropbox prior to processing

cjcolvar:
This attempts to resolve an issue where files that are web uploaded aren't available to be moved in post processing if background jobs are being run on a different server.  This also avoids issues with the globalness and temporalness of /tmp.


apertome (approved):
Looks good - one code climate issue, but it's just a style thing.

PR #4377, opened 2021-01-25
Upgrade omniauth to 2.0.0 for CSRF fix

phuongdh:
Replaces https://github.com/avalonmediasystem/avalon/pull/4367

PR #4574, opened 2021-12-15
Upgrade omniauth to 2.0

cjcolvar:
Replaces #4377


cjcolvar:
#4553 should go in first then this can be rebased on top of that.

cjcolvar:
If omniauth-saml is included in the Gemfile in #4553 then this PR will probably need to upgrade it to 2.0 in order to allow omniauth 2.0.  Other omniauth plugin gems may need to be upgraded as well.

cjcolvar:
Looks like we're still running into this issue: https://github.com/avalonmediasystem/avalon/pull/4377#issuecomment-776294168

Dananji (approved):
Looks good :+1:

PR #4580, opened 2022-01-03
Switch to ActiveFedora13

cjcolvar:
This PR upgrades to ActiveFedora13 as a means of eventually getting a fix for the ruby 2.7 deprecation warning when it is released.  This is an alternative route than putting the fix in the 12.x stable branch and cutting a new 12 release of ActiveFedora.

This brings us up to the latest version of ActiveFedora.  I had to make changes to the `active_fedora-datastreams` gem to allow for the newer version of AF ~so this PR currently pins to a branch until a formal release of the gem happens~.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'webpacker'
 
 # Core Samvera
 gem 'active-fedora', '~> 13.2'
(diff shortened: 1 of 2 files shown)


Dananji (approved):
Looks good! :100:

PR #4587, opened 2022-01-12
Update to latest ActiveFedora and remove monkey-patch

cjcolvar:


apertome (approved):
Thanks for getting the new AF gem released. This looks good!

PR #4590, opened 2022-01-19
Remove  from devise_for call since it adds a format=false query param…

cjcolvar:
…eter instead of avoiding non-HTML formats like it did previously


apertome (approved):
Looks good

PR #4607, opened 2022-02-16
Show intermediate page when user attempting to edit an item but isn't logged in yet

cjcolvar:
This mirrors the changes made in #4239.  A number of controller tests had to be changed to match this new behavior.


Dananji (approved):
Looks good :100:

PR #4610, opened 2022-02-18
Rails 6.0 / Blacklight 8 upgrade

cjcolvar:
This PR is a quick upgrade to Rails 6 and Blacklight 8 (including Bootstrap 4).  The tests pass but there is probably still many many things broken.  Manual testing and CSS fixing still need to happen.


apertome (approved):
Probably still broken stuff, but the idea here is to get this into `develop` to work on it there.

PR #4614, opened 2022-02-21
Support Solr 8

cjcolvar:
Upgrade to solr 8 by using newest solr schema from Samvera community.  This was put in `solr/conf` along with a copy of the existing `solr/config` directory which is for solr 6.  This will require a reindex for existing installs so this probably will need to be a major release of Avalon.

- Upgrade CircleCI to use solr 8
- Upgrade docker-compose to use solr 8
- Configure solr_wrapper to use solr 8
- Add copy of docker-compose.yml modified slightly to work with podman-compose.yml (dev/test only)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  samvera: samvera/circleci-orb@1.0.0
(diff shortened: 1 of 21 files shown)


masaball (approved):
Looks good.

PR #4641, opened 2022-03-11
Render component only when at least one section has transcript file(s)

Dananji:
With this fix, for the transcript component to be rendered on the page at least one section (masterfile, i.e. canvas in the IIIF manifest) has to have a transcript file.


cjcolvar (approved):
Looks good!  I would have thought to do this check on the ruby side but doing it in the JS makes a lot of sense.

Dananji:
> Looks good! I would have thought to do this check on the ruby side but doing it in the JS makes a lot of sense.

Yes, that's what I tried at first. But then realized deconstructing the supplemental files list for each section is a lot easier (and cleaner) on the JS side than on the ruby side.

PR #4648, opened 2022-03-21
Bootstrap4 changes

Dananji:
In this PR, bootstrap changes were made for all the major views except for item view for which I got the following error. I tried this view today, so couldn't report this earlier for Chris to take a look at it :( 

item-view-error


cjcolvar:
This got included in #4610 which got merged.

PR #4649, opened 2022-03-23
Add configured domain host to rails hosts config

cjcolvar:
avalon-dev was throwing an error after deploy and I believe this fixes it.
See https://blog.saeloun.com/2019/10/31/rails-6-adds-guard-against-dns-rebinding-attacks.html

### DIFF
--- a/config/initializers/default_host.rb
+++ b/config/initializers/default_host.rb
@@ -16,4 +16,8 @@ if server_options
   Rails.application.routes.default_url_options.merge!( server_options )
   ActionMailer::Base.default_url_options.merge!( server_options )
   ApplicationController.default_url_options = server_options
+
+  # Required for rails 6+
+  # See https://blog.saeloun.com/2019/10/31/rails-6-adds-guard-against-dns-rebinding-attacks.html
+  Rails.application.config.hosts << server_options[:host]
 end


config/initializers/default_host.rb
```diff
@@ -16,4 +16,8 @@
   Rails.application.routes.default_url_options.merge!( server_options )
   ActionMailer::Base.default_url_options.merge!( server_options )
   ApplicationController.default_url_options = server_options
+
+  # Required for rails 6+
+  # See https://blog.saeloun.com/2019/10/31/rails-6-adds-guard-against-dns-rebinding-attacks.html
+  Rails.application.config.hosts << server_options[:host]
```
cjcolvar, line 22: This is the substantive change.  The rest of the PR is fixing the tests to ensure we're using the host configured here.

spec/rails_helper.rb
```diff
@@ -70,6 +70,10 @@
 ActiveJob::Base.queue_adapter = :test
 
 Capybara.server = :webrick
+Capybara.server_host = Rails.application.routes.url_helpers.root_url
+Capybara.app_host = Rails.application.routes.url_helpers.root_url
+Capybara.asset_host = Rails.application.routes.url_helpers.root_url
+Capybara.server_port = Rails.application.default_url_options[:port]
```
cjcolvar, line 76: I'm not sure all of these are necessary but this ensures we're using the configured host and not the default of `www.example.com`

apertome: Looks good. I'd rather over-test than under

spec/rails_helper.rb
```diff
@@ -145,6 +149,10 @@
     ActiveFedora::Cleaner.clean!
   end
 
+  config.before(:each, type: :request) do
+    host! Rails.application.default_url_options[:host]
+  end
```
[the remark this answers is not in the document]
apertome: Looks good. I'd rather over-test than under

apertome (approved):

PR #4651, opened 2022-03-24
Version handling develop

apertome:


lib/avalon/build_utils.rb
```diff
@@ -0,0 +1,109 @@
...
+      version
+    end
+
+    def get_tags(version, options = {})
+
+      split = options[:split] || false
+      branch = options[:branch] || ""
+      top_level = options[:top_level] || false
+      additional_tags = options[:additional_tags].split(",") if options[:additional_tags]
+      extra_tags = additional_tags || []
```
cjcolvar, line 51: Instead of a new variable you can use `||=` to conditionally initialize `additional_tags` if it hasn't already.
```suggestion
      additional_tags ||= []
```
A different option would be to cast it to an array when you use it `Array(additional_tags)`, but I personally like the `||=` pattern.

lib/avalon/build_utils.rb
```diff
@@ -0,0 +1,109 @@
...
+      tags = []
+      1.step do |i|
+        tag = parts.slice(0, i).join('.')
+        tags.push(tag)
+        break if i >= len
+      end
+      tags
+    end
+
+    def get_commands(tags, source, dest)
```
cjcolvar, line 100: Maybe this method should be removed if it isn't used anywhere?

script/avalon_image_tags.rb
```diff
@@ -0,0 +1,54 @@
...
+ exit 1
+end
+
+utils = Avalon::BuildUtils.new
+version = utils.detect_version if version.empty?
+tags = utils.get_tags(version, options)
+
+puts tags.join(",")
+
+tags
```
cjcolvar, line 54: Is this return value getting used or can it be removed?

script/avalon_image_tags.rb
```diff
@@ -0,0 +1,54 @@
...
+    options[:additional_tags] = tags_cli
+  end
+  opts.on("-s", "--split", "Split version number, i.e. 1.2.3 becomes 1.2.3,1.2,1") do |dh|
+    options[:split] = true
+  end
+end.parse!
+
+if !options[:top_level] && (options[:branch].nil? || options[:branch].empty? )
+ warn "Error: must supply --branch and/or --top-level\n"
+ warn help_text
```
cjcolvar, line 44: If you save the parser object in a variable (e.g. `parser`) then I think you should be able to use `parser.help()` instead.

script/avalon_push_images
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+for tag in "${TAGS[@]}"; do
+  echo "tag '$tag'"
+  #exit 1 if $tag.empty?
+  command="${COMMAND_BASE} ${SOURCE} ${DESTINATION}:${tag}"
+  echo "pushing image ... command:"
+  echo $command
+  #`$command`
```
cjcolvar, line 26: Do you plan on enabling this or will this script only return the commands as guidance to be copy/paste'd?

spec/lib/avalon/build_utils_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+require 'avalon/build_utils'
+
+describe Avalon::BuildUtils do
+    before :all  do
+      Utils = Avalon::BuildUtils.new
+    end
```
cjcolvar, line 20: I think this would do well as a `subject`:
```suggestion
  subject(:utils) { described_class.new }
```
Then the tests would be like:
```
  expect(subject.detect_version(contents_arr)).to eq '1.2.3'
  # OR
  expect(utils.detect_version(contents_arr)).to eq '1.2.3'
```

spec/lib/avalon/build_utils_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+        branch = "production"
+        options = {"split": false, "branch": "production", "top_level": true}
+        expect(Utils.get_tags(version, options)).to eq(["3.2.1-production", "3.2.1", "production"])
+      end
+
+      it 'SPLITS and outputs correct tags for a 3-part version and a branch tag' do
+        version = "3.2.1"
+        options = {"split": true, "branch": "staging"}
+        expect(Utils.get_tags(version, options)).to eq(["3-staging", "3.2-staging", "3.2.1-staging", "staging"])
+      end
```
cjcolvar, line 134: Is this a duplicate of the test on lines 117-121 given that `top_level` defaults to `false`?

spec/lib/avalon/build_utils_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+        expect(Utils.get_tags(version, options)).to eq(["1-develop", "1.2-develop", "1.2.3-develop", "1.2.3.4-develop", "develop", "latest", "testing_something"])
+      end
+
+      it 'Outputs correct tags for a 4-part version and several tags' do
+        version = "1.2.3.4"
+        options = {"split": false, "branch": "develop", "additional_tags": "latest,testing_something"}
+        expect(Utils.get_tags(version, options)).to eq(["1.2.3.4-develop", "develop", "latest", "testing_something"])
+      end
+
+      it 'SPLITS and Outputs correct CSV tags for a 4-part version and several tags' do
```
cjcolvar, line 148:
```suggestion
      it 'SPLITS and Outputs correct CSV tags for a 3-part version and several tags' do
```

spec/lib/avalon/build_utils_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+        version = "11.12.13"
+        options = {"split": true, "branch": "develop", "additional_tags": "latest,testing_something"}
+        expect(Utils.get_tags(version, options)).to eq(["11-develop", "11.12-develop", "11.12.13-develop", "develop", "latest", "testing_something"])
+      end
+
+      it 'Outputs correct CSV tags for a 4-part version and several tags' do
```
cjcolvar, line 154:
```suggestion
      it 'Outputs correct CSV tags for a 3-part version and several tags' do
```

spec/lib/avalon/build_utils_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      it 'Outputs correct tags for a 3-part version and a branch tag' do
+        version = "3.2.1"
+        options = {"split": false, "branch": "staging"}
+        expect(Utils.get_tags(version, options)).to eq(["3.2.1-staging", "staging"])
+      end
+
+      it 'SPLITS and Outputs correct tags for a 3-part version and a branch tag TOP LEVEL production branch' do
+        version = "3.2.1"
+        options = {"split": true, "branch": "production", "top_level": true}
+        expect(Utils.get_tags(version, options)).to eq(["3-production", "3.2-production", "3.2.1-production", "production"])
```
cjcolvar, line 114: Given that `top_level` returns the undecorated tags when there isn't a branch, I would expect this to return the union of the branch non-top level and non-branched top level tags.  Does that make sense?  Or does the `branch` option override the `top_level` option and only one of them can be used at a time?

```suggestion
        expect(Utils.get_tags(version, options)).to eq(["3", "3.2", "3.2.1", "3-production", "3.2-production", "3.2.1-production", "production"])
```

cjcolvar (commented):
Not loading the whole rails/bundle context should make this a lot faster to run but probably made it a bit trickier!  I left some initial comments, questions, and suggestions.  I think addressing the style issues flagged by rubocop in codeclimate would help readability.

apertome:
@cjcolvar Thanks for the feedback! I had a feeling some of those bits could be cleaner/more rubyish, but wasn't sure about some of them.

Avoiding the bundle pieces wasn't for performance reasons, it was so we could run this in environments where the bundle hasn't been installed (particularly the build environment). 

Yes, I've cleaned up some codeclimate suggestions, but the ones you suggested look good; I'll review them in more depth. Thanks!

lib/avalon/build_utils.rb
apertome: Good call, removing

script/avalon_image_tags.rb
apertome: I tried that, but I couldn't get it to work. I'll try again.

script/avalon_push_images
apertome: Yes, I do want to enable it. maybe we need a --dry-run option or something to see what it's going to do without actually executing the command.

spec/lib/avalon/build_utils_spec.rb
apertome: Good eye, fixing this in a new commit

spec/lib/avalon/build_utils_spec.rb
apertome: Yes, I also changed the branch name in the prior instance to use `staging` since toplefel really should not be used with anything other than `production`

script/avalon_image_tags.rb
apertome: It's being used in the unit tests only. I could refactor the tests to use csv strings instead, which would be closer to how the scripts are using it. Think I'll do that.

script/avalon_push_images
cjcolvar: :+1: I like the idea of a `--dry-run`.

lib/avalon/build_utils.rb
```diff
@@ -0,0 +1,100 @@
...
+      branch = options[:branch] || ""
+      top_level = options[:top_level] || false
+      additional_tags_str = options[:additional_tags]
+      additional_tags_str ||= ""
+      extra_tags = additional_tags_str.split(",")
+      parts = version.split('.')
+      len = parts.length
+
+      # only accept between 2 and 4 parts
+      return if len < 2 || len > 4
```
cjcolvar, line 57: From the tests and the regex in `detect_version` it looks like only 3 and 4 part versions are supported.  Are these lines wrong or were you hoping to support 2 part version strings?

apertome:
I pushed some more improvements, but I have not quite addressed all the review feedback yet.

script/avalon_image_tags.rb
apertome: I thought you were commenting on a different file. your'e right, that  return value is unnecessary. Removing.

apertome:
I simplified some functions and created some smaller, simpler ones, but CodeClimate is still complaining about complexity. It's a lot better than it was.

Assuming the circleci tests pass, this should be ready for review

cjcolvar:
Thanks for cleaning things up.  We haven't been picky about method length and complexity so I'm not worried about those, but it's nice to get the style issues taken care of.

cjcolvar (approved):
Looks good.

PR #4657, opened 2022-03-29
Clean up initializers and wrap reloadable constants in #to_prepare blocks

cjcolvar:
This handles the deprecation warnings which were coming up on avalon-dev.  See https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-when-the-application-boots

The basic idea is that any classes or modules which are in the set of reloadable constants (mostly anything in `app/`) needs to be wrapped in a `#to_prepare` block so that if the constant gets reloaded then this block will rerun to pick up the new class/module.

Since a lot of code got touched by indentation it is probably easiest to review this PR with the hide whitespace option in github enabled.


apertome:
Looks good.

apertome (approved):

PR #4663, opened 2022-04-04
Fix collection landing page search

cjcolvar:
This PR fixes the react search component on a collection landing page due to changes in Blacklight's API.

In this PR:
- Fix Blocked Host error by wrapping only the default_url_options setting in config#to_prepare
- Add SpeedyAF proxies for MediaObject and Admin::Collection to support polymorphic_url
- Change override of Blacklight's #url_for_document to return a SpeedyAF proxy
- Bump rails version to latest 6.0.x
- Add more fields to Blacklight index configuration only when JSON request
- Helper to force section_id_ssim to be an array instead of string sentence
- Rework react components to consume new JSON structure


apertome (approved):
Looks good.

PR #4673, opened 2022-04-20
Add identifier_in_use for noid-rails

### LINKED ISSUE #4672: Missing identifier_in_use for Noid::Rails
While testing Avalon 7.3.0 I came across a situation where the Noid::Rails minter was returning identifiers that were tombstoned in Fedora. Although this situation should be extremely rare I wanted to bring this to your attention.

**Problem:** Creating new media objects would fail and return `Ldp::Gone`. Caused by a new media object identifier tombstoned in Fedora.

**Solution:** Set identifier_in_use function for Noid::Rails to check for existing or gone identifiers in Fedora

Adding the following to `config/initializers/noid-rails.rb` fixes the issue

```
config.identifier_in_use = lambda do |id|
  ActiveFedora::Base.exists?(id) || ActiveFedora::Base.gone?(id)
end
```

seanlw:
Added `identifier_in_use` for noid-rails to insure new identifiers don't exist or are gone in Fedora.

Fixes #4672

### DIFF
--- a/config/initializers/noid-rails.rb
+++ b/config/initializers/noid-rails.rb
@@ -2,4 +2,8 @@ require 'noid-rails'
 
 Noid::Rails.configure do |config|
   config.minter_class = Noid::Rails::Minter::Db
+
+  config.identifier_in_use = lambda do |id|
+    ActiveFedora::Base.exists?(id) || ActiveFedora::Base.gone?(id)
+  end
 end


cjcolvar (approved):
Thanks for fixing this!

cjcolvar:
I missed that this is a PR into the default branch.  We have a kind of wonky git flow and the default branch is a static branch of released code only.  Do you mind rebasing this on top of the `develop` branch?

cjcolvar (approved):

Issue #4675, opened 2022-04-20
Format and Position of Metadata Information on Catalog Page


joncameron:
# Description
The display of the metadata text and its layout in the listing of results in /catalog has shifted due to dependency updates. These should be restored to their previous states.

The metadata appears to be printing in a string representing an array value, for example:
> Date: ["1996"]
> Main contributors: ["Mahler, Gustav, 1860-1911."]

# Done Looks Like
- [ ] Metadata values (Date, Main contributors etc.) are presented as text and not wrapped as strings or array values
- [ ] Metadata appears to the right of the title and icon

# Screenshots
## Current
<img width="944" alt="Screen Shot 2022-04-20 at 11 47 23 AM" src="https://user-images.githubusercontent.com/1406618/164271465-b31853bd-f431-4d4d-bcf8-823c8e0d8895.png">

## Expected
<img width="930" alt="Screen Shot 2022-04-20 at 11 47 36 AM" src="https://user-images.githubusercontent.com/1406618/164271492-3fe89621-aa85-4062-846e-88ae07832653.png">

joncameron:
Probably a Blacklight override that isn't quite right.

cjcolvar:
#4685

PR #4685, opened 2022-04-25
Adjust styling for search results

cjcolvar:
This involved changing some css column size classes and adding the `row` class to the document result which required a blacklight helper override.  To avoid too many overrides we decided to change the styling slightly.  @joncameron what do you think about the change to the metadata styling?

Before:
image

After:
image

### DIFF
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -45,3 +45,7 @@
     background-color: $white;
   }
 }
+
+.dl-invert dt {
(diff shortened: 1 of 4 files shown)


cjcolvar:
Fixes #4675

joncameron (dismissed):
This looks good to me. I'm OK with merging this and then seeing how the different layout looks with some examples, and tweaking as necessary.

Dananji (approved):
:100:

PR #4688, opened 2022-04-27
Test with blacklight 7.x branch with fix

### LINKED ISSUE #4678: "more »" Link Does Not Open Modal
# Description
In /catalog view a facet's "more »" button should open a modal with a full listing, but currently clicking it results in no action. 

# Done Looks Like
- [ ] Clicking the "more" hyperlink opens the modal listing as expected

cjcolvar:
This branch has guards for if the bootstrap variable is undefined which was causing a JS error blocking the more facets modal from opening.  The blacklight dependency will need to be updated to a real release once it is cut.

Fixes #4678

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'rdf-rdfxml'
 gem 'rdf-vocab', '< 3.1.5'
 
 # Samvera version pins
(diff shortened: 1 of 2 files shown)


Dananji (approved):
:+1:

PR #4692, opened 2022-04-28
Upgrade jasny-bootstrap for file upload step

Dananji:
Previous version of `jasny-bootstrap` was using bootstrap v.3.x. Even though nothing seems to be broken with this version, this upgrade was done to keep things updated.
Along with this some code in the file upload was updated to reflect the jasny-bootstrap's file upload example usage.


app/views/media_objects/_file_upload.html.erb
```diff
@@ -112,7 +112,7 @@ Unless required by applicable law or agreed to in writing, software distributed
                     <div class="col-sm-2">
                       <div class="form-group">
                         <label>Date Digitized</label>
-                        <%= form.text_field :date_digitized, class: 'date-input' %>
+                        <%= form.text_field :date_digitized, class: 'date-input', id: section.id %>
```
cjcolvar, line 115: Why was this id added to date digitized?  I'm not seeing an id as necessary on other text fields.

cjcolvar (commented):
Looks good overall.  I was confused about one small part so I left a question.

app/views/media_objects/_file_upload.html.erb
Dananji: Yes, all the other fields works fine without the `id` when there are multiple sections. 

But when I changed the `Date Digitized` field of any section other than the first one, the change happened in the first section, not in the one where it was intended. I thought it was odd, because all the other fields work well without an `id` and this used to work well without it. 

Maybe this was a fluke in my local dev enviroment? Could you please test this out locally if possible?

cjcolvar: Sure, I'll see if I can reproduce that behavior.

cjcolvar: I'm seeing that too.  I'm guessing that the JS could be improved but without setting an id like you've done then there is id duplication which is definitely a problem.  So I think it would be good to keep your change but could it be changed to be more specific?
```suggestion
                        <%= form.text_field :date_digitized, class: 'date-input', id: "master_file_date_digitized_#{section.id}" %>
```

Dananji: This looks good to me :+1:

PR #4694, opened 2022-04-29
Fix select/deselect all buttons

masaball:


masaball:
Looks like a change in Blacklight's css classes. There's an old commit about "HTML class selectors should use hyphens per the CSSLint style guide" which changed it.

PR #4698, opened 2022-05-02
Add format property to Resources - IIIF Manifest Generation

masaball:
I looked around for a bit to see where/how the hls mimetype was brought in (i.e. as part of the master_file or stream_info, etc.). I found that it was hard coded as part of the _video_element and _audio_element partials. Since it didn't appear to be created on the fly, I have just hardcoded the format for the manifest generation as well.


cjcolvar (approved):
Looks great!

PR #4699, opened 2022-05-03
More bootstrap changes

Dananji:


app/views/playlists/_item_list.html.erb
```diff
@@ -34,14 +34,23 @@ Unless required by applicable law or agreed to in writing, software distributed
...
   </div>
+  <div id="share-list" class="mt-3 collapse">
+    <ul class="nav nav-tabs share-tabs">
+      <%= render_conditional_partials :share, section:'share-tabs' %>
+    </ul>
+
+    <div class="tab-content">
+      <%= render_conditional_partials :share, section:'tab-content' %>
+    </div>
+  </div>
```
cjcolvar, line 53: I see the `mt-3` class was added to this `<div>`, but I'm curious why did you move this div out of the share partial?

cjcolvar (commented):
Wow!  This is a lot of JS/CSS/HTML fixing.  This looks really good to me.  I only had one question.

app/views/playlists/_item_list.html.erb
Dananji: Share partial is rendered within the div with the `col-*` class, which spans the share panel to the width of that space. But it should span the entire width of the row under the share and autoplay buttons.
Do you think we need to render it from a partial instead of putting it in here, because of the condition?

cjcolvar: Now that you mention it, yes!  :slightly_smiling_face: 
I was thinking that semantically they fit better in the share partial since there is JS in that partial which references them.
So how about pushing the wrapping div into the partial instead of moving the other div out of the partial?  Would that work?

Dananji: Yes that would work! Let me try that :+1:

cjcolvar (approved):
Looks great!  Thanks!

PR #4702, opened 2022-05-05
Add sanitization of post-processing filenames

masaball:
This PR adds a configuration method to sanitize filenames. Default behavior is to replace spaces with underscores.

### DIFF
--- a/lib/avalon/configuration.rb
+++ b/lib/avalon/configuration.rb
@@ -28,6 +28,16 @@ module Avalon
       @config.send(sym, *args, &block)
     end
 
+    def sanitize_filename
+      @sanitize_filename = lambda do |filename|
+        if filename.include? ' '
+          filename.gsub(/\s+/, '_')
+        else
+          filename
+        end
+      end
+    end
+
     private
     class << self
       def coerce(value, method)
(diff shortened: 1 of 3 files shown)


lib/avalon/configuration.rb
```diff
@@ -28,6 +28,16 @@ def method_missing(sym, *args, &block)
       @config.send(sym, *args, &block)
     end
 
+    def sanitize_filename
+      @sanitize_filename = lambda do |filename|
+        if filename.include? ' '
+          filename.gsub(/\s+/, '_')
+        else
+          filename
+        end
```
cjcolvar, line 37: I think it should be safe to run this `gsub` in all cases since it will only apply if the regex matches.
```suggestion
        filename.gsub(/\s+/, '_')
```

cjcolvar (commented):
This looks great!  I'm excited to see this pattern come into Avalon.

I think that we need to add a setter method in order for this to be available to configuration.  I think this should only need an `attr_writer` and the instance variable assignment be made optional using `||=`.  (See https://github.com/samvera/hyrax/blob/a9e13ca755c3c30b8b28a1cb94df429fae932d9d/lib/hyrax/configuration.rb#L834-L839 for an example)

cjcolvar (approved):
Looks great!

PR #4703, opened 2022-05-06
Use official release of Blacklight

cjcolvar:
Blacklight 7.25.1 was released with includes the fix for #4678 and makes the changes in #4688 unnecessary.


masaball (approved):
Looks good!

PR #4707, opened 2022-05-10
Restore rendering of search constraints breadcrumbs

### LINKED ISSUE #4689: Restore search constraints
The search constraints bar is missing after the blacklight upgrade.

image

A place to start with this is to add `<%= render 'constraints' %>` before `<%= render 'search_results' %>` in `app/views/catalog/index.html.erb`.  This renders something but it looks very different than before.

cjcolvar:
I first tried to override another partial to change the remove button's class from `btn-secondary-outline` to `btn-outline` to ensure the button turns red when hovered over.  But I decided to do the override in CSS instead of overriding another blacklight partial.

Fixes #4689

### DIFF
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -46,3 +46,10 @@
     }
   }
 }
+
+// Search constraint remove button
+// Needed to override btn-secondary-outline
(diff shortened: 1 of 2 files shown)


Dananji (approved):
This is great! It's good to have less BL overrides :+1:

PR #4710, opened 2022-05-11
Fix dropbox (browse everything) for bootstrap 4 and remove warning banner

### LINKED ISSUE #4400: Change Warning Text in Browse Everything
### Description
The red text when doing dropbox upload is overkill and is a holdover from the Avalon 5 era. It doesn't offer much value and is present in all Avalon systems by default.

### Done Looks Like
- Warning box for dropbox ingest modal is removed

### Current Snapshot
Screen Shot 2021-04-22 at 9 40 47 AM

cjcolvar:
This PR covers a few issues:
- Removes warning banner (fixes #4400)
- Uses bootstrap 4 with browse everything
- Remove css that takes over background causing modal not to look like a modal
- Adds missing config option for s3 browse-everything driver

### DIFF
--- a/app/assets/javascripts/file_browse.js.coffee
+++ b/app/assets/javascripts/file_browse.js.coffee
@@ -19,13 +19,6 @@ $ ->
   initialized = false
   $('#browse-btn').browseEverything()
     .show ->
-      alertMsg = $('<div>')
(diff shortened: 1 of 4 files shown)


Dananji (approved):
Works great!! :tada:

PR #4712, opened 2022-05-13
Adjust markup, css classes, and css styling for global nav rendering …

### LINKED ISSUE #4700: Fix responsive homepage nav
### Description

Responsive nav bar on the homepage is broken. 

### Done Looks Like

- [ ] Dropdown menu includes 'Sign In/Out' link
- [ ] Styling and colors are correct
- [ ] Avalon logo on top instead of text

cjcolvar:
…for non-desktop screen sizes

I think there might be some issues with breakpoints, but I think this fixes the layout when the breakpoints have been hit.

Fixes #4700

### DIFF
--- a/app/assets/stylesheets/avalon/_nav.scss
+++ b/app/assets/stylesheets/avalon/_nav.scss
@@ -39,7 +39,7 @@ $state-info-border:              darken(adjust-hue($state-info-bg, -10), 7%) !de
   }
 
   .navbar-toggler {
-    @include media-breakpoint-down(xs) {
+    @include media-breakpoint-down(sm) {
       background: $primary;
       margin: 0.5rem 1rem 0.5rem 0.5rem;
     }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/avalon/_nav.scss
```diff
@@ -92,6 +94,8 @@ $state-info-border:              darken(adjust-hue($state-info-bg, -10), 7%) !de
 
 .navbar-header {
   @include media-breakpoint-down(md) {
+    width: 100%;
```
Dananji, line 97: I think this is not necessary at this screen size, since the `.navbar-header` is a hidden div at this point. Setting width to 100% pushes the visible nav items to the right. So, it works as expected at this screen width without setting this.

Instead I think `width: 100%` could be applied to the next breaking point -> `sm`

app/assets/stylesheets/avalon/_nav.scss
```diff
@@ -35,15 +35,17 @@ $state-info-border:              darken(adjust-hue($state-info-bg, -10), 7%) !de
     background-size: contain;
     background-color: white;
     background-position: 20%;
+    margin-top: 0;
   }
 
   .navbar-toggler {
     @include media-breakpoint-down(xs) {
```
Dananji, line 42: For this I think the breakpoint could be `sm` instead of `xs`?

Dananji (changes requested):

app/assets/stylesheets/avalon/_nav.scss
cjcolvar: Moving that line down to the next block and adjust it to `sm` from `xs` seems to have fixed a number of problem screen sizes.  Thanks for the suggestion!

Dananji (approved):
Looks great :tada:

PR #4714, opened 2022-05-16
Make Timeliner iiif manifest homepage an array

masaball:
IIIF presentation V3 requires that 'homepage' be an array. This PR adds the array structure to manifests generated by Timeliner and updates the tests to work with the new structure.

### DIFF
--- a/app/models/timeline.rb
+++ b/app/models/timeline.rb
@@ -60,7 +60,7 @@ class Timeline < ActiveRecord::Base
 
     manifest_json = JSON.parse(manifest)
     manifest_json["homepage"] ||= {}
-    manifest_json["homepage"]["id"] = "#{base_url}?t=#{media_fragment}"
+    manifest_json["homepage"][0]["id"] = "#{base_url}?t=#{media_fragment}"
     self.manifest = manifest_json.to_json
   end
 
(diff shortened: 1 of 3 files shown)


app/models/timeline.rb
```diff
@@ -60,7 +60,7 @@ def standardize_homepage
 
     manifest_json = JSON.parse(manifest)
     manifest_json["homepage"] ||= {}
```
cjcolvar, line 62: Does this initialization need to change as well?
`manifest_json["homepage"] ||= []`

cjcolvar (commented):
This looks good overall.  Glad that it was really straightforward.  I had one suggestion and there is one more style issue (indentation) that would be good to resolve before this gets merged.

app/models/timeline.rb
masaball: Not sure if it's strictly necessary, since it didn't seem to interfere with the tests but probably best to make the change. Will switch it over.

cjcolvar (commented):
Looks like my suggestion was only half the answer and that the empty hash initialization still needs to happen but for the first entry in the array.
`manifest_json["homepage"][0] = {}`

cjcolvar (approved):

PR #4715, opened 2022-05-17
Adjust styling for intercom modal

cjcolvar:
This should resolve issues with the intercom modal that pops up when you use the "Push to other avalon" button on a media object show page.  Before these fixes the modal would not display at all and the modal background would lock you out of the page.

### DIFF
--- a/app/views/media_objects/_intercom_push_form.html.erb
+++ b/app/views/media_objects/_intercom_push_form.html.erb
@@ -28,7 +28,7 @@ Unless required by applicable law or agreed to in writing, software distributed
     <div class="input-group" style="flex-flow: nowrap">
       <%= select_tag :collection_id, options_from_collection_for_select(@collections, "id", "name", session[:intercom_default_collection] ), required: true, style: 'width:100%;', class:'form-control', form:'intercom_push_form' %>
       <span class="input-group-btn">
-        <button class="btn" style="lightgrey" title="Refresh list" onclick="getIntercomCollections(true, '<%= intercom_collections_media_objects_path %>');"><i class="fa fa-refresh"></i></button>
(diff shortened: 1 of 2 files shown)


app/views/media_objects/_intercom_push_form.html.erb
```diff
@@ -17,17 +17,18 @@ Unless required by applicable law or agreed to in writing, software distributed
...
 
 <div class="modal-body">
   <div class="form-group">
     <label for "collection_id">Choose Target Collection:</label>
-    <div class="input-group">
+    <div class="input-group" style="flex-flow: nowrap">
       <%= select_tag :collection_id, options_from_collection_for_select(@collections, "id", "name", session[:intercom_default_collection] ), required: true, style: 'width:100%;', class:'form-control', form:'intercom_push_form' %>
       <span class="input-group-btn">
-        <button class="btn btn-secondary" title="Refresh list" onclick="getIntercomCollections(true, '<%= intercom_collections_media_objects_path %>');"><i class="fa fa-refresh"></i></button>
+        <button class="btn" style="lightgrey" title="Refresh list" onclick="getIntercomCollections(true, '<%= intercom_collections_media_objects_path %>');"><i class="fa fa-refresh"></i></button>
```
Dananji, line 31: I beleive this should be `class="btn btn-outline"` instead of `style="lightgrey"` which creates a white button with an outline.

Dananji (changes requested):
This looks great 🎉 I just had one question about the btn styling.

app/views/media_objects/_intercom_push_form.html.erb
cjcolvar: For reference this is what it looks like in dark avalon: 
image
And this is what it looks like with this PR:
image
The `btn-secondary` didn't look good, but I think the white button works fine.

Dananji (approved):
Looks great 💯

PR #4733, opened 2022-06-02
Fix eventsource security alert

Dananji:
Previously `eventsource` library was imported as one of the dependencies of `sockjs-client` library, and this was installed as a dependency of the `webpack-dev-server` library. So in order to upgrade just the `eventsource` library it needs to be added as a resolution in the package.json file probably requiring manually editing the yarn.lock file (https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)

`webpack-dev-server`'s latest release (v4.9.1), uses `sockjs-client` library as a devDependency instead of one of its dependencies which eventually requires us, not to install `eventsource` library as one of the dependencies in the Avalon code.

So in this PR, `webpack-dev-server` is upgraded from v3.2.1 to v4.9.1 instead of upgrading `eventsource` from v1.1.0 to v1.1.1 via resolutions. This remove `eventsource` dependency in the yarn.lock file. I think this is better going forward because it make the dependency management easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "html-webpack-plugin": "^5.3.2",
     "prop-types": "^15.7.2",
     "webpack-cli": "^3.2.3",
-    "webpack-dev-server": "^3.2.1"
+    "webpack-dev-server": "^4.9.1"
   },
   "resolutions": {
     "elliptic": "^6.5.2",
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Nice!  I wonder if this will take care of most of the other JS dependency security notices as well.

PR #4737, opened 2022-06-07
Add Checkouts to navbar

Dananji:
checkouts_nav


cjcolvar:
This PR overlaps with #4741 so depending on which goes in first the other will need to be rebased.

cjcolvar (approved):

PR #4741, opened 2022-06-10
Create "Checkouts" page

masaball:


app/controllers/checkouts_controller.rb
```diff
@@ -47,13 +48,31 @@ def destroy
     end
   end
 
+  # DELETE /checkouts
+  def return_all
```
cjcolvar, line 52: I'm thinking we may want to create a different action that returns a checkout by adjusting the return time without destroying the checkout.  It could be POST update or a new POST route. Unless you think it makes more sense to have the DELETE route be the return action.  For now, I'm thinking it might be easiest to rename this to `destroy_all` and have another PR implement the soft delete/return functionality.  What do you think?

app/models/checkout.rb
```diff
@@ -4,6 +4,8 @@ class Checkout < ApplicationRecord
   validates :user, :media_object_id, :checkout_time, :return_time, presence: true
 
   scope :active_for, ->(media_object_id) { where(media_object_id: media_object_id).where("return_time > now()") }
+  scope :checkouts, ->(user_id) { where(user_id: user_id).where("return_time > now()") }
+  scope :returned_checkouts, ->(user_id) { where(user_id: user_id).where("return_time < now()") }
```
cjcolvar, line 8: What do you think about making these scope names have the param scope in them?
```suggestion
  scope :active_for_media_object, ->(media_object_id) { where(media_object_id: media_object_id).where("return_time > now()") }
  scope :active_for_user, ->(user_id) { where(user_id: user_id).where("return_time > now()") }
  scope :returned_for_user, ->(user_id) { where(user_id: user_id).where("return_time < now()") }
```

app/views/checkouts/index.html.erb
```diff
@@ -2,28 +2,59 @@
...
-    <tr>
-      <th>User</th>
-      <th>Media object</th>
-      <th>Checkout time</th>
-      <th>Return time</th>
-      <th colspan="3"></th>
-    </tr>
-  </thead>
+<div class="Checkout-index">
+  <% if Checkout.returned_checkouts(current_user.id).count > 0 or current_user.groups.include? 'administrator' %>
```
cjcolvar, line 6: I think this could be tightened up a little.  And I have a tendency to avoid `or` because it has a different precedence than `||` which can get confusing (see https://stackoverflow.com/a/2083118).
```suggestion
  <% if Checkout.returned_checkouts(current_user.id).any? || current_ability.is_administrator? %>
```

config/routes.rb
```diff
@@ -26,7 +26,11 @@
     end
   end
 
-  resources :checkouts, only: [:index, :create, :show, :update, :destroy]
+  resources :checkouts, only: [:index, :create, :show, :update, :destroy] do
+    collection do
+      delete '', as: '', to: 'checkouts#return_all'
```
cjcolvar, line 31: Is `delete 'return_all'` the same as this longer form?  I'm never sure since routing can be a bit magical.

spec/rails_helper.rb
```diff
@@ -112,7 +112,7 @@
   config.before :suite do
     WebMock.disable_net_connect!(allow: ['localhost', '127.0.0.1', 'fedora', 'fedora-test', 'solr', 'solr-test', 'matterhorn', 'https://chromedriver.storage.googleapis.com'])
     DatabaseCleaner.allow_remote_database_url = true
-    DatabaseCleaner.url_allowlist = ['postgres://postgres:password@db/avalon', 'postgresql://postgres@localhost:5432/postgres']
+    DatabaseCleaner.url_allowlist = ['postgres://postgres:password@db/avalon', 'postgresql://postgres@localhost:5432/postgres', 'postgresql://postgres:password@db-test/avalon']
```
cjcolvar, line 115: Nice catch!

spec/views/checkouts/index.html.erb_spec.rb
```diff
@@ -1,16 +1,53 @@
...
+      before { FactoryBot.create(:checkout, user: user, return_time: DateTime.current - 1.week) }
+      it 'renders the show inactive checkouts checkbox' do
+        render
+        expect(response).to render_template('checkouts/_inactive_checkout')
+      end
+    end
+  end
+  context 'as an admin user' do
+    let(:admin) { FactoryBot.create(:admin) }
+    before { allow(view).to receive(:current_user).and_return(admin) }
```
cjcolvar, line 34: You don't need the before if you override `user` instead of defining a new `admin` variable.  This kind of overriding allows the nested contexts to DRY up the spec tests, but can also sometimes make it too DRY and hard to follow. 
```suggestion
    let(:user) { FactoryBot.create(:admin) }
```

spec/controllers/checkouts_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+
+RSpec.describe CheckoutsController, type: :controller do
```
cjcolvar, line 3: Since the rails scaffold generator created a request spec and didn't create a controller spec I was putting the tests in `spec/requests/checkouts_spec.rb`.  The general trend in rspec appears to be towards request specs and away from controller specs (https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11).  I know we have a ton of controller specs but I thought we could try doing request specs since this is new functionality.  What do you think?

cjcolvar (commented):
This is looking good functionally!  The questions I have are mainly around positioning for future work on this feature.

app/controllers/checkouts_controller.rb
masaball: I had misinterpreted the single destroy action as a temporary implementation for returns and followed that lead for this action, but since end goal functionality is that return history is retained and viewable we should change this action as you suggest and implement the soft return with a future PR.

spec/controllers/checkouts_controller_spec.rb
masaball: I was not aware of that trend. If current best practices/recommendations are to use request specs, we should probably move towards that.

config/routes.rb
masaball: I am pretty sure that when I was trying to implement this `delete 'return_all'` resulted in the path being `checkouts/return_all`. Since we are going to change this to `destroy_all` and implement a different action for our returns, it may be fine to use the shorten it and separate the action from the main `/checkouts` path.

app/controllers/checkouts_controller.rb
masaball: Although, on further thought, is it necessary to have a mass hard delete?

app/controllers/checkouts_controller.rb
```diff
@@ -47,13 +48,31 @@ def destroy
...
   private
 
     # Use callbacks to share common setup or constraints between actions.
     def set_checkout
       @checkout = Checkout.find(params[:id])
     end
 
+    def user_checkouts
+      @checkouts = if current_user.groups.include? 'administrator'
+                     Checkout.all.where("return_time > now()")
```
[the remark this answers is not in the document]
masaball: I think we can probably wait on turning this into a scope.

app/controllers/checkouts_controller.rb
cjcolvar: You're right.  I don't think it makes sense to have `delete_all` without a good use case (which I can't think of one right now).

masaball: Ok. Should we just leave it for now and remove it in the followup PR properly implementing returns, or should I scrap it before we merge?

PR #4744, opened 2022-06-15
Set checkout time and return time based upon default setting at create time

cjcolvar:
Related to #4740

### DIFF
--- a/app/models/checkout.rb
+++ b/app/models/checkout.rb
@@ -3,11 +3,26 @@ class Checkout < ApplicationRecord
 
   validates :user, :media_object_id, :checkout_time, :return_time, presence: true
 
+  after_initialize :set_checkout_return_times!
+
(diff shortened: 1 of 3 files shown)


app/models/checkout.rb
```diff
@@ -3,11 +3,26 @@ class Checkout < ApplicationRecord
...
   scope :returned_for_user, ->(user_id) { where(user_id: user_id).where("return_time < now()") }
 
   def media_object
     MediaObject.find(media_object_id)
   end
+
+  private
+
+  def set_checkout_return_times
+    self.checkout_time = DateTime.now
```
masaball, line 19: Should this be `DateTime.current`?

masaball (approved):
Looks good!

app/models/checkout.rb
cjcolvar: :100: Good catch!

masaball (approved):
Looks good.

PR #4749, opened 2022-06-21
Use join instead of merge for Addressable::URI

cjcolvar:
Addressable::URI#merge raises an error in this case due to expecting a hash.  Found when running in production.


joncameron (approved):
👍

PR #4752, opened 2022-06-21
Add missing license headers

cjcolvar:
Also update all other ones to add annoying space so future runs of the tool will be easier


Dananji (approved):
:+1:

PR #4756, opened 2022-06-23
Use real releases instead of github branches

cjcolvar:
Use releases of active_annotations, active_fedora-datastreams, and speedy-af.


Dananji (approved):
:100:

PR #4762, opened 2022-06-24
Add missing queues to ensure activestorage background jobs run

cjcolvar:
I noticed there are jobs sitting waiting to be run on avalon-dev which aren't because sidekiq wasn't configured to watch these queues.


Dananji (approved):
:+1:

PR #4764, opened 2022-06-27
Display returned items on checkouts page

masaball:


app/models/ability.rb
```diff
@@ -228,6 +228,7 @@ def checkout_permissions
       end
       can :return, Checkout, user: @user
       can :return_all, Checkout, user: @user
+      can :display_returned, Checkout, user: @user
```
cjcolvar, line 231: Instead of making a new action, I think we could reuse the `:read` action since it I think they are the same thing.  You would probably have to do something a little special in the controller though instead of adding this here.  I think this might be something like:
```
def display_returned
  authorize! :read, @checkouts
  ...
end
```
OR  see my comment in `config/routes.rb`

config/routes.rb
```diff
@@ -29,6 +29,7 @@
   resources :checkouts, only: [:index, :create, :show, :update, :destroy] do
     collection do
       patch :return_all
+      get :display_returned
```
cjcolvar, line 32: I'm wondering about making this route to `:index` with a param set so it would add a branch to that action in order to avoid the need for a new action entirely.  Or it may be possible to get away with a special param on `:index` without a new route.  What do you think?

cjcolvar (commented):
I have some thoughts/questions that might lead to restructuring some of this, but I think all the pieces are already here.  Good work!
I didn't read through all of the tests in case things might change.

config/routes.rb
masaball: I have reworked it so that there is a `display_returned` param that changes the behavior of the `set_checkouts` function, so that the checkouts list displayed on the page is controlled entirely through the `index` action/route.

spec/requests/checkouts_spec.rb
```diff
@@ -48,29 +111,43 @@
...
-        expect {
+    context "json request" do
+      context "with valid parameters" do
+        it "creates a new Checkout" do
+          expect {
+            post checkouts_url, params: { checkout: valid_attributes, format: :json }
+          }.to change(Checkout, :count).by(1)
+        end
+
+        it "redirects to the created checkout" do
```
cjcolvar, line 122: This predates this PR but would be good to fix.
```suggestion
        it "returns created status" do
```

cjcolvar (approved):
This looks great!  Real good work on the tests!

PR #4765, opened 2022-06-27
Fix loading of datatables in production env

cjcolvar:
This seemed to fix the datatables loading for avalon-staging which is running in a rails production environment.


Dananji (approved):
Looks good :+1:

PR #4778, opened 2022-07-04
Upgrade to fork of B-E v1.2 with avalon customizations

cjcolvar:
This is a new branch of B-E based upon the just released v1.2 with the avalon customizations reapplied.  An avalon file was blocking another bootstrap js file from loading so I renamed it.

This PR is built on top of #4777 which was needed in order for bundler to install browse-everything.


Dananji (approved):
Looks good :+1:

PR #4779, opened 2022-07-05
Consider title and metadata attributes from Blacklight json-api as safe

### LINKED ISSUE #4773: Item titles are HTML encoded on the Collection page (https://avalon-staging.dlib.indiana.edu/collections/wp988j80x) (2)
# Current Behavior
HTML encoding is displaying in titles on the Collection view page. Example: https://avalon-staging.dlib.indiana.edu/collections/wp988j80x

# Expected Behavior
Text is printed normally and does not display as HTML encoded.

cjcolvar:
Fixes #4773 

This appears to be a change with the upgrade to Blacklight 7 and its switch from a json endpoint which returns the solr document verbatim to a json-api.  The new code runs through a `Blacklight::Rendering::Pipeline` which escapes HTML before marking it `html_safe`.  If Blacklight is considering it safe then I think we can as well.

See https://github.com/projectblacklight/blacklight/blob/26edc7021455d955632ccd925016fed328faa0af/app/presenters/blacklight/rendering/join.rb#L8

### DIFF
--- a/app/javascript/components/collections/landing/SearchResultsCard.js
+++ b/app/javascript/components/collections/landing/SearchResultsCard.js
@@ -36,7 +36,7 @@ const CardMetaData = ({ doc, fieldLabel, fieldName }) => {
     return (
       <React.Fragment>
         <dt className='col-sm-5'>{fieldLabel}</dt>
-        <dd className='col-sm-7'>{metaData}</dd>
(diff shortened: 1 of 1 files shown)


Dananji:
Before:
before

After:
after

@cjcolvar It looks like this to me after this fix (overflowing to footer area). Trying to figure out what's causing this.

PR #4782, opened 2022-07-05
Manually save translated output urls

### LINKED ISSUE #4070: Encode dashboard shows wrong output urls when using Minio
WatchedEncode [after_completed](https://github.com/avalonmediasystem/avalon/blob/develop/app/models/watched_encode.rb#L28-L43) needs to happen before persisting to database.

cjcolvar:
Do this because the normal callback persisting the encode record happens before this callback.
Fixes #4070

### DIFF
--- a/app/models/watched_encode.rb
+++ b/app/models/watched_encode.rb
@@ -26,6 +26,8 @@ class WatchedEncode < ActiveEncode::Base
   end
 
   after_completed do |encode|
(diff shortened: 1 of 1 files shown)


Dananji (approved):
Looks great!

I've attempted to fix this before (and failed), it's nice to see the solution :tada:

cjcolvar:
I'm not positive it is the full solution, but it shouldn't hurt and I can test it out on avalon-dev. :crossed_fingers:

PR #4786, opened 2022-07-06
Add confirmation popover when deleting structure

Dananji:
delete-confirm

### DIFF
--- a/app/assets/javascripts/button_confirmation.js.coffee
+++ b/app/assets/javascripts/button_confirmation.js.coffee
@@ -20,6 +20,11 @@ $ ->
   apply_button_confirmation()
 
 @apply_button_confirmation = () ->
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/button_confirmation.js.coffee
```diff
@@ -20,6 +20,11 @@ $ ->
   apply_button_confirmation()
 
 @apply_button_confirmation = () ->
+  # Remove popover when clicked outside. Reference: https://jsfiddle.net/guya/fjZja/
```
cjcolvar, line 23: That's some dark magic!  I worry that this might be brittle since it depends on a specific implementation of popover which might change.  I think it is okay though since this is an enhancement which wouldn't hinder the user if it randomly breaks.

cjcolvar (approved):
Looks good!

PR #4787, opened 2022-07-06
Sessions expire after 2 weeks

### LINKED ISSUE #4145: Change Session Length
### Description
Session length in Avalon is currently a very lengthy amount of time. This should be changed to something more reasonable for security reasons.

### Done Looks Like
- Set default session length to 2 weeks in Rails config

cjcolvar:
Resolves #4145


Dananji (approved):
:+1:

PR #4789, opened 2022-07-07
Add lending period to Media Object and Collection

masaball:
This PR adds a lending period field to the Media Object model and Admin::Collection model and associated changes to be able to view and update the lending period in the GUI.


cjcolvar (commented):
I did a quick read through and this looks pretty good to me.  I know it would be large, but do you want to add the collection work to this PR as well?

masaball:
If you don't see a problem with it, I think having them as a single large PR does kind of make sense.

app/helpers/media_objects_helper.rb
```diff
@@ -114,6 +114,31 @@ def display_rights_statement media_object
         content_tag(:dt, 'Rights Statement') + content_tag(:dd) { link }
       end
 
+      # Lending period is stored as seconds. Convert to days and hours for display.
+      def display_lending_period object
+        d, h = (object.lending_period/3600).divmod(24)
+
+        replacement = {
```
cjcolvar, line 121: Rails has a [pluralize helper method](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) that makes this easy:
```
pluralize(1, 'day') # => 1 day
pluralize(2, 'day') # => 2 days
```

app/models/concerns/lending_period.rb
```diff
@@ -0,0 +1,63 @@
+module LendingPeriod
+
+  extend ActiveSupport::Concern
+
+  included do
+    property :lending_period, predicate: ::RDF::Vocab::SCHEMA.eligibleDuration, multiple: false do |index|
+      index.as :stored_sortable
+    end
+
+    after_initialize :set_lending_period
```
cjcolvar, line 10: I'm inclined to name this `default_lending_period` for collections and `lending_period` for media objects to follow the current pattern.  This would break the DRYing up done here in this concern.

I wonder if the initialization of the lending period is necessary or if when it is being called it would fall back to the global default.  Maybe this could be done in an accessor on the media_object:
```
def lending_period
  self.lending_period || Settings.controlled_digital_lending.default_lending_period
end
```
It might be that the media object will always get initialized from the collection or global default and this accessor override is unnecessary.

I'm also wondering which would be better: storing the ISO8601 string in the property or the integer.  It seems like there is a decent amount of code for casting back and forth so wondering if that would be easier if we store the string form.  Any thoughts on this?

Another thought is making a subclass of ActiveSupport::Duration which at a minimum could implement the `display_lending_period` helper possibly as a `to_s` override.

cjcolvar (commented):
This is looking good. I have some swirling thoughts about the concern mixin which could possibly tighten things up.  Would you be up for a quick chat or would you rather stick to comments on this PR?

app/controllers/admin/collections_controller.rb
```diff
@@ -298,6 +299,12 @@ def update_access(collection, params)
 
     collection.default_visibility = params[:visibility] unless params[:visibility].blank?
     collection.default_hidden = params[:hidden] == "1"
+    collection.default_lending_period = build_default_lending_period
+  end
+
+  def build_default_lending_period
+    iso_duration = "P#{params["add_default_lending_period_days"]}DT#{params["add_default_lending_period_hours"]}H"
+    int_duration = iso_duration.is_a?(Integer) ? ActiveSupport::Duration.parse(iso_duration).to_i : Settings.controlled_digital_lending.default_lending_period.to_i
```
cjcolvar, line 307: What about something like:
```suggestion
    collection.default_lending_period = build_default_lending_period.to_i
  end

  def build_default_lending_period
    params["add_default_lending_period_days"].days + params["add_default_lending_period_hours"].hours
  rescue
    Settings.controlled_digital_lending.default_lending_period
```

app/models/media_object.rb
```diff
@@ -371,6 +377,10 @@ def lending_status
     Checkout.active_for_media_object(id).any? ? "checked_out" : "available"
   end
 
+  def set_lending_period
+    self.lending_period ||= Admin::Collection.find(self.collection_id).default_lending_period
```
cjcolvar, line 381: I'm pretty sure there is a `.collection` convenience method.
```suggestion
    self.lending_period ||= collection.default_lending_period
```

app/views/media_objects/_metadata_display.html.erb
```diff
@@ -72,7 +72,8 @@ Unless required by applicable law or agreed to in writing, software distributed
       </h3>
     </div>
     <div class="card-body">
-      <%= @media_object.access_text %>
+      <%= @media_object.access_text %> <br />
+      <b>Lending Period:</b> <%= ActiveSupport::Duration.build(@media_object.lending_period).to_human_readable_s %>
```
cjcolvar, line 76: It looks like ActiveSupport::Duration will create a human readable string with `inspect`.  What do you think about using that?
```suggestion
      <b>Lending Period:</b> <%= ActiveSupport::Duration.build(@media_object.lending_period).inspect %>
```
If this is workable then we could avoid injecting a new method into Duration.

app/views/modules/_access_control.html.erb
```diff
@@ -64,6 +64,37 @@ Unless required by applicable law or agreed to in writing, software distributed
   </div>
 </div>
 
+<div class="card item-lending-period">
+  <div class="card-header">
+    <h3 class="card-title">Item lending period</h3>
+  </div>
+  <div class="card-body">
+    <div class="form-group">
+      <% field = object.is_a?(Admin::Collection) ? :default_lending_period : :lending_period %>
```
cjcolvar, line 73: I think the approach taken for the other fields like visibility is to use a common name in the view partial and do the mapping in the call to render.  This could avoid the branching logic.

cjcolvar (commented):
Some suggestions that I think could tighten this up a bit more.  Let me know what you think.

app/views/media_objects/_metadata_display.html.erb
masaball: I'm not sure it is workable. For example, it converts `2592000` (30 days in seconds) to "4 weeks and 2 days". And the conversions just get worse from there. "1 month, 2 weeks, 13 hours, 30 minutes, and 54 seconds" is the output for 45 days. So we would need extra parsing to convert them into better/clearer strings. Or override the behavior. 

To be clearer with the injected method though, I could change the name to something like `to_day_hour_s` since it is more specific than just making things human readable.

app/models/access_control_step.rb
```diff
@@ -95,8 +95,11 @@ def execute context
       lease.destroy
     end
     unless limited_access_submit
+      iso_duration = "P#{context[:add_lending_period_days]}DT#{context[:add_lending_period_hours]}H"
+      int_duration = ActiveSupport::Duration.parse(iso_duration).to_i
```
cjcolvar, line 99: I think it makes sense to do the same building method here as in the admin collections controller.

cjcolvar (commented):
Looking really close.  One small suggestion then I think it'll be ready to go.

app/models/access_control_step.rb
masaball: Yep, definitely. I meant to switch it but got distracted by other changes and pushed before taking care of it. Good catch.

app/views/modules/_access_control.html.erb
```diff
@@ -104,6 +130,18 @@ Unless required by applicable law or agreed to in writing, software distributed
...
 </div>
 
+<div class="card item-access">
+  <div class="card-header">
+    <h3 class="card-title">Item lending period</h3>
+  </div>
+  <div class="card-body">
+    <div class="form-group">
+      Item is available to be checked out for
+      <%= ActiveSupport::Duration.build(object.lending_period).to_human_readable_s %>
```
cjcolvar, line 140: Looks like this one got missed being renamed:
```suggestion
      <%= ActiveSupport::Duration.build(object.lending_period).to_day_hour_s %>
```

cjcolvar (approved):

PR #4790, opened 2022-07-08
Update browse-everything to handle more exceptions

cjcolvar:
This PR bumps the revision of browse-everything which now includes handling for errors raised during the rendering of the modal.  This covers the case when tokens have expired (Google OAuth tokens default to 1 hour lifetime).  Previously this was being handled for the show action but not the index action.

To test this out, you can manually expire the oauth token you granted by:
1. Log into google
2. Go to Account -> Security -> Manage third-party access
3. Click on Avalon Media System
4. Click Remove Access

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -24,7 +24,7 @@ GIT
 
 GIT
   remote: https://github.com/avalonmediasystem/browse-everything.git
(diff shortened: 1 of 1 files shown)


masaball (approved):
Looks great

PR #4798, opened 2022-07-15
Mount volume where solr 8's docker container stores data

cjcolvar:
This is only for the development environment and should allow the solr data to be persistent instead of disappearing after every restart.


Dananji (approved):
:+1:

PR #4801, opened 2022-07-21
Override accessor to provide fallback; have controllers do validation

cjcolvar:
This ended up getting a little ugly but the tests pass.

### DIFF
--- a/app/models/access_control_step.rb
+++ b/app/models/access_control_step.rb
@@ -94,13 +94,18 @@ class AccessControlStep < Avalon::Workflow::BasicStep
       media_object.governing_policies.delete( lease )
       lease.destroy
     end
+
     unless limited_access_submit
       media_object.visibility = context[:visibility] unless context[:visibility].blank?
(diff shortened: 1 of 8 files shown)


app/models/access_control_step.rb
```diff
@@ -94,13 +94,22 @@ def execute context
       media_object.governing_policies.delete( lease )
       lease.destroy
     end
+    #if context['add_lending_period_days'].present? || context['add_lending_period_hours'].present?
+    #  lending_period = build_lending_period(context['add_lending_period_days'], context['add_lending_period_hours'])
+    #  media_object.lending_period = lending_period if lending_period.positive?
+    #end
+
```
masaball, line 101: Should this commented out code be removed?

masaball (approved):
Looks good.

app/models/access_control_step.rb
cjcolvar: Yes, good catch.

masaball (approved):
Looks good

Issue #4432, opened 2021-04-05
Improve handling of deleted content


apertome:
Currently, when a user visits a URL of a media object that has been deleted, a message informs them:

> Item(s) being deleted
> 
> The item(s) you requested are being deleted from the system. If this is not expected, contact your support staff.
 
This makes it appear that something is wrong. This message never goes away, even though the object has actually been deleted. We should change this to say something along the lines of "Item deleted."

We could forcibly delete those tombstones -- there is no reason to think these resources would be referenced anywhere. But I am not sure that we really gain much by doing so.

Thread with some context:
https://groups.google.com/g/fedora-tech/c/syikHRq2eDk

### Done Looks Like
- Change the message appearing to users to be something more specific

PR #4803, opened 2022-07-22
Make deleted content error message clearer

masaball:
This message appears when a user manually navigates to a Media Object or Collection page after the item/collection has been deleted. Is this new message clear/good enough?

### DIFF
--- a/app/views/errors/deleted_pid.html.erb
+++ b/app/views/errors/deleted_pid.html.erb
@@ -15,7 +15,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <div id="main_content_container" class="container-fluid">
   <div class="content" class="col-md-8">
-    <h2>Item(s) being deleted</h2>
-    <p>The item(s) you requested are being deleted from the system. If this is not expected, contact your support staff.</p>
+    <% item = request.original_url.include?("admin/collections") ? "collection" : "item" %>
+    <h2><%= "#{item.titleize} Deleted" %> </h2>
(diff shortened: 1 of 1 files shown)


app/views/errors/deleted_pid.html.erb
```diff
@@ -15,7 +15,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <div id="main_content_container" class="container-fluid">
   <div class="content" class="col-md-8">
-    <h2>Item(s) being deleted</h2>
-    <p>The item(s) you requested are being deleted from the system. If this is not expected, contact your support staff.</p>
+    <% item = request.original_url.include?("admin/collections") ? "collection" : "item" %>
```
cjcolvar, line 18: Unfortunately, I don't think this is a good test.  The tombstone doesn't have any indication what type of object it was and one can manually enter a url to get this error message for either collection or item with the same tombstoned id.

masaball: I see what you mean. Since the tombstone isn't limited to its old path and has no indicators on what it used to be, it might be best to remove the test and just do "Item Deleted" for all cases.

cjcolvar (approved):
This seems like a fine change to me.

This is basically reverting a previous PR (https://github.com/avalonmediasystem/avalon/pull/3903) which doesn't have any description or associated issue as far as I can tell.  Without any context, I can't tell what the reason for the change was.  If the tension is between a clear, unconfusing message for different scenarios that are at odds with each other then I think the message should be least confusing for the most common scenario.  This PR appears to be doing that without making the other scenarios too confusing/misleading.

cjcolvar:
Closes #4432

PR #4804, opened 2022-07-27
Add instance name to comments mailer

masaball:


app/controllers/comments_controller.rb
```diff
@@ -30,6 +30,7 @@ def create
     @comment.comment = params[:comment][:comment]
 
     if @comment.valid? && (Settings.recaptcha.blank? || verify_recaptcha(model: @comment))
+      @comment.subject = @comment.subject.prepend("#{Settings.name}: ")
```
cjcolvar, line 33: I think there is a cleaner way to prepending the site name without adding it into `@comment.subject` to strip it back out again in the views.  How about doing? `subject: "#{Settings.name}: #{@comment.subject}"` at https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/app/mailers/comments_mailer.rb#L20

cjcolvar (commented):
Those tests are perfect!  Thanks for filing in this gap in code coverage.

app/controllers/comments_controller.rb
masaball: Good call. That is a much better way to handle that. Will make the change.

cjcolvar (approved):
Looks great!

PR #4813, opened 2022-07-29
Render mediaobject without masterfile properly when CDL is enabled

Dananji:
With the fix in this PR, items without masterfiles when CDL is enabled looks as follows;

Screenshot from 2022-07-29 14-42-10


cjcolvar (approved):
:+1:

PR #4815, opened 2022-08-01
Display remaining time for non-admin users

Dananji:


app/views/media_objects/_administrative_links.html.erb
```diff
@@ -14,52 +14,28 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-      <span class="label label-important">Unpublished</span> Users cannot view the item</li>
-      <% end %>
-      <li>
-        <% unless @media_object.discover_groups.include? "nobody" %>
-        <span class="label label-success">Visible</span> Users can search for the item</li>
-      <% else %>
-      <span class="label label-important">Hidden</span> Users cannot search for the item</li>
-      <% end %>
-    </ul>
-  </p>
```
cjcolvar, line 33: I think this should be added back in unless there was a reason you removed it.

app/views/media_objects/show.html.erb
```diff
@@ -26,5 +26,12 @@ Unless required by applicable law or agreed to in writing, software distributed
   </h1>
 </div>
 
-<%= render 'administrative_links' %>
+<div class="row">
+  <% if can? :update, @media_object %>
+    <div class="col-sm-5 mb-3">
```
cjcolvar, line 31: The `can?` check is already in the `administrative_links` partial so how about pushing the `div` into that partial?  Then nothing needs to wrap the render of `administrative_links`.  This would also mirror how the `destroy_checkout` partial starts with an ability check and surrounding `div`.

cjcolvar (approved):
Looks great!

PR #4816, opened 2022-08-02
Allow managers to GET collection/items endpoint

masaball:


app/models/ability.rb
```diff
@@ -57,6 +57,7 @@ def create_permissions(user=nil, session=nil)
 
       if @user_groups.include? "manager"
         can :create, Admin::Collection
+        can :items, Admin::Collection
```
cjcolvar, line 60: I think this would grant the ability for any manager to run the `:items` action for any `Admin::Collection`.  I think we need to restrict this.  I was thinking that a user that can `:read` an `Admin::Collection` should also be able to `:items` it as well.  Does that match with your understanding?

If that's the case, then I think `:items` could maybe be aliased to `:read`.  I'm not sure if cancancan allows aliases with conditions/blocks so it might have to be more like some of the examples below (e.g. `:read, Derivative`).  Adding some tests in `admin_collections_controller_spec.rb` would be good.

masaball: Yes, my understanding is that users who can `:read` the collection should be able to `:items` it as well. I did not think through that this change would open everything to every manager, which I have tried now and you are right that is exactly what happens. Will add tests to make sure the behavior is limited to those who actually have access and rework the implementation.

Have converted the PR to a draft for the time being.

cjcolvar (approved):
Looks great!

PR #4817, opened 2022-08-03
Update checkouts page

masaball:


config/initializers/time_formats.rb
```diff
@@ -0,0 +1 @@
+Time::DATE_FORMATS[:long_ordinal_12h] = lambda { |time| time.strftime("%B #{time.day.ordinalize}, %Y %l:%M %p") }
```
cjcolvar, line 1: Nice!  I've never seen this one before.  That's really cool that it can take a lambda.

cjcolvar (approved):

PR #4820, opened 2022-08-05
Change file_watcher implementation used in development to support M1 Macs

cjcolvar:
Making this change allowed the avalon docker-compose to load successfully on an M1 Mac.


elynema (approved):
Fixed the problem, allowing Avalon to run on my M1 Mac.

Issue #4809, opened 2022-07-29
Validation Error When Saving Collection


joncameron:
# Description
Running a save action on a collection is launching a notification that "Lending period hours needs to be a positive integer.". This happens when a 0 is present for the hours or days value for the collection's default lending period.

# Done Criteria
- [ ] When the day or hour value for a collection is 0, Avalon does not produce this error on save.

PR #4822, opened 2022-08-08
Add CDL conditional logic

masaball:


app/controllers/admin/collections_controller.rb
```diff
@@ -326,7 +326,11 @@ def build_default_lending_period(collection)
...
     end
 
-    flash[:notice] = collection.errors.full_messages.join if collection.errors.present?
+    if lending_period.zero?
+      collection.errors.add(:lending_period, "cannot be set to 0 days 0 hours.")
+    end
+
+    flash[:notice] = collection.errors.full_messages.join if collection.errors.present? && Settings.controlled_digital_lending.enable
     lending_period.to_i
   end
```
masaball, line 335: I can't figure out how to highlight the rest of the build_default_lending_period method for comment, but there is a similar method in the [access_control_step](https://github.com/avalonmediasystem/avalon/blob/fc7bba2bd8efd2436e872fa05dd260147dbbceba/app/models/access_control_step.rb#L129) that is a lot cleaner. The error handling is done when the method is called around line 102. 

@cjcolvar, is there any reason you ran into or can see to not change this to match the implementation for building and setting the lending period in the access_control_step?

app/views/modules/_access_control.html.erb
```diff
@@ -64,31 +64,33 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+          <div class="col-1">
+            <label class="col-form-label" for="add_lending_period_hours">
+              <span data-title="Hours">Hours</span>
+            </label>
+            <%= text_field_tag "add_lending_period_hours", h ? h : 0, class: 'form-control' %>
+          </div>
         </div>
       </div>
     </div>
   </div>
```
masaball, line 92: Could potentially pull this and the cdl bit at line 135 out into their own partial(s) in a future PR, though not sure it's worth it unless/until we need to render these settings outside of the access control context.

app/controllers/admin/collections_controller.rb
masaball: I went ahead and did this refactoring as part of the work on #4809. Everything seems good so far, but still let me know if you have any thoughts.

app/controllers/admin/collections_controller.rb
```diff
@@ -326,7 +326,11 @@ def build_default_lending_period(collection)
       collection.errors.add(:lending_period, "hours needs to be a positive integer.")
     end
 
-    flash[:notice] = collection.errors.full_messages.join if collection.errors.present?
+    if lending_period.zero?
+      collection.errors.add(:lending_period, "cannot be set to 0 days 0 hours.")
+    end
+
+    flash[:notice] = collection.errors.full_messages.join if collection.errors.present? && Settings.controlled_digital_lending.enable
```
cjcolvar, line 333: I think it would make more sense to add the conditional around the building/setting of the default lending period on line 302-303.

app/jobs/bulk_action_jobs.rb
```diff
@@ -205,7 +205,9 @@ def perform(collection_id, overwrite = true)
         media_object = MediaObject.find(id)
         media_object.hidden = collection.default_hidden
         media_object.visibility = collection.default_visibility
-        media_object.lending_period = collection.default_lending_period
+        if Settings.controlled_digital_lending.enable
```
cjcolvar, line 208: I think we need to handle the cases when `Settings.controlled_digital_lendings.enable` is not set or not a boolean:
```suggestion
        if !!Settings.controlled_digital_lending&.enable
```
If it seems nicer, we could create a convenience method to do the special handling.  Maybe something like this in `lib/avalon/configuration.rb`:
```
# To be called as Avalon::Configuration.controlled_digital_lending_enabled?
module Avalon
  module Config
    def controlled_digital_lending_enabled?
      !!Settings.controlled_digital_lending&.enable
    end
  end
end
```

app/views/media_objects/_item_view.html.erb
```diff
@@ -26,7 +26,9 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% f_start, f_end = parse_media_fragment params['t'] %>
 <div class="row">
   <div class="col-sm-8" id="left_column">
-    <% if can_stream && @currentStream %>
+    <% if Settings.controlled_digital_lending.enable && !can_stream %>
+      <%= render 'embed_checkout' %>
+    <% elsif @currentStream %>
```
cjcolvar, line 31: I think this change was based on an earlier version of this file and needs to be updated to handle the latest in `develop`.

app/views/modules/_access_control.html.erb
cjcolvar: :+1: I was thinking the same thing.

config/routes.rb
```diff
@@ -26,7 +26,7 @@
     end
   end
 
-  resources :checkouts, only: [:index, :create, :show, :update, :destroy] do
+  resources :checkouts, only: [:index, :create, :show, :update, :destroy], :constraints => lambda { |request| Settings.controlled_digital_lending.enable } do
```
cjcolvar, line 29: Nice catch!

cjcolvar (commented):
This is looking really good!  It doesn't look as messy as I was expecting it would have to be.

app/jobs/bulk_action_jobs.rb
masaball: Good call. I think I lean towards using the convenience method, so will try that out.

app/views/media_objects/_item_view.html.erb
masaball: So, I did base this change off what is current on [develop](https://github.com/avalonmediasystem/avalon/blob/fc7bba2bd8efd2436e872fa05dd260147dbbceba/app/views/media_objects/_item_view.html.erb#L29). I just flipped the conditional around because that was the easiest way I could conceptualize handling both cdl enabled and disabled without having to duplicate the whole block of render code.

Happy to change it if there's a better way.

cjcolvar: Sorry, I'll take a closer look.

cjcolvar: Ok, I'm seeing it now.  Nice.  I think this makes sense.

app/controllers/admin/collections_controller.rb
```diff
@@ -314,7 +316,7 @@ def build_default_lending_period(collection)
     d.negative? ? collection.errors.add(:lending_period, "days needs to be a positive integer.") : lending_period += d.days
     h.negative? ? collection.errors.add(:lending_period, "hours needs to be a positive integer.") : lending_period += h.hours
 
-    flash[:error] = collection.errors.full_messages.join if collection.errors.present?
+    flash[:error] = collection.errors.full_messages.join(' ') if collection.errors.present?
```
cjcolvar, line 319: Nice catch!

cjcolvar (approved):
Looks great!

PR #4823, opened 2022-08-10
Fix validation check on collection page

masaball:


cjcolvar (commented):
Will this still highlight the field on the form with the errors? I think that was why I was using `collection.errors.add(:lending_period)`.

masaball:
No, it does not. I overlooked that. Will take another look at raising/catching the errors and getting the behavior in line with the original implementation.

cjcolvar (approved):
Looks great!

PR #4832, opened 2022-08-12
Stay in modal when links clicked inside modal

cjcolvar:
Without this change If you click the `more>>` link to open a facet modal then click `Next` or `A-Z sort` it would cause a page reload and the modal is the page contents.  This change makes the ajax request update the currently open modal without a page reload.


Dananji (approved):
Nice catch :100: 
Looks good :+1:

PR #4839, opened 2022-08-18
Add CDL check to Stream token assignment

masaball:


app/helpers/security_helper.rb
```diff
@@ -19,13 +19,30 @@ def add_stream_cookies(stream_info)
...
+        end
+      else
+        add_stream_url(stream_info)
+      end
+    rescue Ldp::BadRequest
+      add_stream_url(stream_info)
+    rescue StreamToken::Unauthorized
+    end
+    stream_info
+  end
```
masaball, line 38: This is hacky and not secure but I'm not sure how best to approach it. The sticking points are:

In the manifest files controller spec there is a check [that hls_manifest is generated for public items](https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/spec/controllers/master_files_controller_spec.rb#L606). What functionalities/services require an `hls_manifest`? Should that be locked down when CDL is enabled or do we have to find a less hacked together way to allow them through?

The `rescue Ldp::BadRequest` call relates to the [embed check](https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/spec/controllers/master_files_controller_spec.rb#L230). This is a temporary bandage to get the tests passing. For now, should we disable master file embeds when CDL is on? At least until we figure out how we want to handle CDL embedding?

cjcolvar: I think it is fine to raise an error if requesting a `hls_manifest` when CDL is enabled which could happen in the `hls_manifest` action avoiding the need to handle it here.  Would this also resolve the embed check?  I would expect tests to pass if CDL isn't explicitly enabled for specific test(s).

app/helpers/security_helper.rb
```diff
@@ -19,13 +19,25 @@ def add_stream_cookies(stream_info)
...
 
-  def secure_streams(stream_info)
+  def secure_streams(stream_info, media_object_id)
+    begin
+      if Avalon::Configuration.controlled_digital_lending_enabled? && Checkout.checked_out_to_user(media_object_id, current_user.id).empty?
+        raise StreamToken::Unauthorized
+      else
+        add_stream_url(stream_info)
+      end
+    rescue NoMethodError
```
masaball, line 29: There may be a better way to handle this, but this seems to allow all of the previously problematic tests to still pass and allows the item view page to load for unauthenticated users which is an issue I ran into when moving the logic I had put in for the hls_manifest case.

app/helpers/security_helper.rb
```diff
@@ -19,13 +19,25 @@ def add_stream_cookies(stream_info)
...
   end
 
-  def secure_streams(stream_info)
+  def secure_streams(stream_info, media_object_id)
+    begin
+      if Avalon::Configuration.controlled_digital_lending_enabled? && Checkout.checked_out_to_user(media_object_id, current_user.id).empty?
+        raise StreamToken::Unauthorized
+      else
+        add_stream_url(stream_info)
+      end
```
cjcolvar, line 28: I wonder if this could be rewritten as the following and work.
```suggestion
        add_stream_url(stream_info) if Avalon::Configuration.controlled_digital_lending_enabled? && Checkout.checked_out_to_user(media_object_id, current_user&.id).empty?
```
Or if that doesn't work then adding another clause:

```suggestion
        add_stream_url(stream_info) if Avalon::Configuration.controlled_digital_lending_enabled? && current_user && Checkout.checked_out_to_user(media_object_id, current_user.id).empty?
```
At that point a small private method to contain the conditional logic might make sense.

cjcolvar (commented):
This is looking really good.  One suggestion but if it doesn't make sense then I'm happy to merge as-is.

cjcolvar (approved):

PR #4842, opened 2022-08-19
Fix deletion of users with checkouts

masaball:
This PR also fixes a deprecation warning I was encountering during the rspec tests.

### DIFF
--- a/app/controllers/samvera/persona/users_controller.rb
+++ b/app/controllers/samvera/persona/users_controller.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2022, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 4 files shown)


app/controllers/samvera/persona/users_controller.rb
```diff
@@ -191,7 +191,7 @@ def load_user
 
     def app_view_path
       my_engine_root = Samvera::Persona::Engine.root.to_s
-      prepend_view_path "#{my_engine_root}/app/views/#{Rails.application.class.parent_name.downcase}"
```
cjcolvar, line 194: Nice!  I was just doing this on a Hyrax Rails 6.1 upgrade PR.

spec/controllers/samvera/persona/user_controller_spec.rb
```diff
@@ -156,6 +156,8 @@
 
     describe 'DELETE #destroy' do
       before :each do
+        FactoryBot.create(:checkout, user_id: user.id)
+        FactoryBot.create(:checkout, user_id: user.id, return_time: DateTime.current - 1.day)
```
cjcolvar, line 160: How about moving these into a context block around the test you added? This way they don't have to be run for all of the other tests which don't need them.

app/models/user.rb
```diff
@@ -1,18 +1,19 @@
...
-# 
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
 #   specific language governing permissions and limitations under the License.
 # ---  END LICENSE_HEADER BLOCK  ---
 
 class User < ActiveRecord::Base
+  has_many :checkouts, dependent: :destroy
```
cjcolvar, line 16: Perfect!

cjcolvar (approved):
Looks great!  Just had one suggestion for the test but it isn't a blocker.

spec/controllers/samvera/persona/user_controller_spec.rb
masaball: I tried that but I get: `ArgumentError: User #{ user } does not belong to the manager group`. Something about trying to add the checkouts after doing the `RoleMap.replace` causes an issue with the collection level `add_manager` method.

cjcolvar: I missed the first time that this test in on the persona controller.  How about moving it to be on the model like the test for deleting bookmarks? https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/spec/models/user_spec.rb#L133-L139

cjcolvar (approved):

PR #4845, opened 2022-08-22
Remove vestiges of old multipart dropdown handling: hidden field and unused JS

cjcolvar:
The hidden field is spamming the form since it is always "local" (because the JS isn't matching the new CSS classes) and is interleaved with the real values coming from the select inputs.  For example, submitting the form with two values in other identifiers field of `Other: agz3068` and `OCLC: ocm9072382` the `other_identifier_type` form data is `["local", "other", "local", "oclc"]`.


Dananji (approved):
Looks great! :tada:

PR #4856, opened 2022-08-26
Filter only on username and email; disable sorting on roles

cjcolvar:
This PR is in response to slow response times on filtering and sorting of roles when run in production with ~30K users.  Avoiding non-DB query filtering and sorting speeds up response times considerably.  This PR proposes filtering on only username and email and disabling sorting on roles.


masaball (approved):
Looks good

PR #4857, opened 2022-08-29
Improve devise error display

masaball:
This PR updates the generation of sign up and related devise registration errors to provide details on what caused the error. It also removes extra white space between the label and entry field that was being added when errors triggered.

devise_errors


config/application.rb
```diff
@@ -50,5 +50,7 @@ class Application < Rails::Application
     end
 
     config.active_storage.service = (Settings&.active_storage&.service.presence || "local").to_sym
+
+    config.action_view.field_error_proc = Proc.new { |html_tag, instance| "<span class='field_with_errors'>#{html_tag}</span>".html_safe }
```
masaball, line 54: This overrides the default `field_error_proc`. The reason the white space was being inserted was because the `field_error_proc` defaults to wrapping the error elements in  separate `<div>`s which then caused the label and input field to be rendered as separate html blocks. This changes the behavior to generate the `field_with_errors` element as a `<span>` so that it renders inline.

Another potential solution would be, if we aren't utilizing the `field_with_errors` class, to disable the `field_error_proc` completely, so that it just returns the original html. If we don't want to override `field_error_proc`, we could maybe add `.field_with_errors { display: inline; }` to the main css file.

app/views/devise/passwords/edit.html.erb
```diff
@@ -17,20 +17,20 @@ Unless required by applicable law or agreed to in writing, software distributed
   <h3 align='center'>Change your password</h3>
 
   <%= form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :put }) do |f| %>
```
cjcolvar, line 19: I wonder if changing this to `bootstrap_form_for` (which we use elsewhere in views) would automatically handle what you're doing with conditionally applying CSS classes.  See https://github.com/bootstrap-ruby/bootstrap_form#validation-and-errors for examples of configuring it wrt displaying errors.

app/helpers/devise_helper.rb
```diff
@@ -15,10 +15,25 @@
 module DeviseHelper
   def devise_error_messages!
```
cjcolvar, line 16: It looks like newer Devise deprecates this method in favor of a view partial which might handle everything you're adding here.  If this is the case, then you could delete this whole helper file.  See https://github.com/heartcombo/devise/blob/main/app/helpers/devise_helper.rb and https://github.com/heartcombo/devise/blob/main/app/views/devise/shared/_error_messages.html.erb.  If that works then it would probably be good to go ahead and follow the advice of the deprecation warning and removed references to `devise_error_messages!` in the views in favor of rendering the new partial directly.

cjcolvar (commented):
Good sleuthing!  I wonder if there isn't a simpler solution though leaning into `bootstrap_form` and newer versions of files from devise.  Would you be open to trying that out?  It seems like you have a really good grasp on the situation so hopefully it wouldn't take too much time to try.

masaball:
Yeah, I can try out those options. At the very least, definitely in our best interests to take care of the stuff being deprecated while I'm already working on them. Keep on top of some of the tech debt.

cjcolvar (approved):
Looks great!

PR #4858, opened 2022-08-29
Change return button styling on item view page

masaball:


app/views/media_objects/_destroy_checkout.html.erb
```diff
@@ -22,7 +22,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <span class="time">00:00 <br />hh:mm</span>
     </div>
     <div class="col-5 col-sm-3 col-md-4">
-      <%= link_to 'Return now', return_checkout_url(current_checkout), class: 'btn btn-danger', method: :patch,
+      <%= link_to 'Return now', return_checkout_url(current_checkout), class: 'btn btn-primary', method: :patch,
```
cjcolvar, line 25: What do you think about changing this to `btn-outline`?  I think that would match the other buttons on the page more.
```suggestion
      <%= link_to 'Return now', return_checkout_url(current_checkout), class: 'btn btn-outline', method: :patch,
```

masaball: Jon put `btn-primary` in the ticket so its what I went with, but `btn-outline` would be consistent with the button styling on the checkouts page and as you said in general matches a bit better.

cjcolvar: Sorry, I missed that in the ticket.  Why don't we have @joncameron weigh in before merging.

joncameron: I agree, let's go with btn-outline.

cjcolvar (approved):

PR #4859, opened 2022-08-29
Fix html encoded titles in collection cards

Dananji:
Before:
Screenshot from 2022-08-29 12-58-52
After:
Screenshot from 2022-08-29 12-59-10

The fix in this PR, removes string truncation from the JS code, and uses CSS's `line-clamp` property to limit the title to 2 lines along with other CSS properties truncates the overflowing text.

### DIFF
--- a/app/javascript/components/collections/Collection.scss
+++ b/app/javascript/components/collections/Collection.scss
@@ -83,6 +83,17 @@
   .italic {
     font-style: italic;
   }
+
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Amazing!

Do you think something similar could be done for the metadata properties on the card?

Dananji:
Yes, we could :+1: I'll check that out.

PR #4862, opened 2022-08-30
Add authentication prompt for cdl items

masaball:
This seems to work. I used the implementation from the masterfile `_authenticate` partial for doing a login popup. Though, since we are already in avalon, rather than an external site, would it be better to send the user to the regular login page and return them to the item instead of using a popup?

I also needed to tweak some of the styling for the player area so that the sign in button would render entirely within the player. It looks balanced to me, but may need additional tweaking.

### DIFF
--- a/app/views/media_objects/_checkout_authenticate.html.erb
+++ b/app/views/media_objects/_checkout_authenticate.html.erb
@@ -1,0 +1,64 @@
+<%#
+Copyright 2011-2022, The Trustees of Indiana University and Northwestern
+  University.  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed
+  under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+  CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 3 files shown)


masaball:
I had tried the pattern like the errors/restricted_pid, but the sign in wasn't tracking the page you came from, which is why I looked around at other options. 

Looking into how /errors/restricted_pid is called, copying the `session[:previous_url]` setter from the [application controller](https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/app/controllers/application_controller.rb#L201) into the partial worked:

    <% session[:previous_url] = request.fullpath unless request.xhr? %>
    <%= link_to 'Sign in', main_app.new_user_session_path, :class=> "btn btn-info" %>

 Do you think that is sufficient?

cjcolvar:
Maybe we should pull the setting of `session[:previous_url]` into a `before_action :set_previous_url, only: [:index, :show, :edit]`.  This seems like it might be good to do overall as long as we don't accidentally redirect the user to something unhelpful (which Avalon and/or Hyrax has had problems with before).  We'd have to add a `skip_before_action :set_previous_url` to the devise controllers and any other controllers we don't want to be redirected to.

cjcolvar:
That's a good point.  It probably should be since it really is out of scope of this ticket.  In that case, it probably makes sense to do the session setting in the view partial like you proposed and then pull it out in the separate PR.

cjcolvar (approved):
Looks good.

PR #4864, opened 2022-08-31
Fix label on reset password page

masaball:
Made a label naming error on one of the Devise edit page's when changing to bootstrap_forms. This PR changes it back to the original naming.


cjcolvar (approved):
Good catch!  Sorry I missed it in review.

PR #4867, opened 2022-09-01
Update CORS to echo request origin

masaball:
Rack-CORS allows for dynamically setting allowed origins in a block in the config. Manually testing with cURL, a simple block looking at the request source is sufficient to have the origin header echoed.

I think this change should retain the functionality of using a wildcard, while echoing the request origin into the `Access-Control-Allow-Origin` header.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -38,7 +38,7 @@ module Avalon
 
     config.middleware.insert_before 0, Rack::Cors do
       allow do
-        origins '*'
+        origins { |source| true }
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
That's amazing!

I think it would be good to have a test to ensure that this functionality doesn't change without us knowing.  Unfortunately it looks like there aren't any tests for the `manifest` action yet to build off of, but maybe you could add a test just for the the CORS header for a json GET request to the MediaObjectsController to cover this case.

cjcolvar (approved):
Nice!  Thanks!

PR #4868, opened 2022-09-02
Outline buttons of bulk actions on selected items page

cjcolvar:
Styling fix from IU production.

Before (from avalon-dev):
image
After (from MCO):
image


Dananji (approved):
Looks great :tada:

PR #4876, opened 2022-09-13
Fix overflow of encode records table

Dananji:
Handle encoding dashboard datatable overflow properly using datatable option `scrollX: true`, so that horizontal overflow of the table is scroll-able as below;
Screenshot from 2022-09-13 09-09-04


masaball (approved):
Looks good!

PR #4878, opened 2022-09-15
Add access control check to MediaObject #index

masaball:


app/controllers/media_objects_controller.rb
```diff
@@ -318,7 +318,7 @@ def custom_update
   def index
     respond_to do |format|
       format.json {
-        paginate json: MediaObject.all
+        paginate json: MediaObject.all.select { |mo| can? :read, mo }
```
cjcolvar, line 321: I think this is the right idea but I think this might cause scalability issues.  `MediaObject.all` is an `ActiveFedora::Relation` and there is some magic that operates on them to do pagination efficiently by limiting the number of items being returned by the solr query.  I think this change would fetch back all MediaObjects (via `to_a`) then pass them to `select.`

I think the same might be able to be achieved using `accessible_by`.
```suggestion
        paginate json: MediaObject.accessible_by(current_ability, :read)
```
See https://github.com/samvera/hydra-head/blob/main/hydra-access-controls/lib/active_fedora/accessible_by.rb#L4

cjcolvar: Ah, ok.  That makes sense since Avalon does a bunch of special things for gated discovery beyond what is done in hydra-access_controls.  It would be nice if `accessible_by` just worked.  I wonder how hard it would be to do that or to chain onto it so the correct results are returned to `paginate`.  Let me see what I can find.

cjcolvar: My guess as to why `accessible_by` isn't working as we expect is that switch from using `#gated_discovery_filters` to `Hydra::SearchBuilder#gated_discovery_filters` dropped our Avalon-specific overrides.  @masaball Can you try overriding the `accessibly_by` method to use `::SearchBuilder` instead of `Hydra::SearchBuilder` and see if that fixes things?

Add something like this to either the `hydra_config.rb` initializer or make a new initializer:
```
ActiveFedora::QueryMethods.module_eval do
  extend ActiveSupport::Concern

  def accessible_by(ability, action = :index)
    permission_types = case action
      when :index then [:discover, :read, :edit]
      when :show, :read then [:read, :edit]
      when :update, :edit, :create, :new, :destroy then [:edit]
    end

    builder = ::SearchBuilder.new(nil).with_ability(ability).with_discovery_permissions(permission_types)
    filters = builder.send(:gated_discovery_filters).join(" OR ")
    spawn.where!(filters)
  end
end
```

config/initializers/hydra_config.rb
```diff
@@ -58,3 +58,19 @@ def byname
...
+
+  def accessible_by(ability, action = :index)
+    permission_types = case action
+      when :index then [:discover, :read, :edit]
+      when :show then [:read, :edit]
+      when :update, :edit, :create, :new, :destroy then [:edit]
+    end
+
+    builder = ::SearchBuilder.new(nil).with_ability(ability)
+    filters = builder.send(:gated_discovery_filters).reject(&:blank?).join(' OR ')
```
masaball, line 73: I did some testing with this and having `.with_discovery_permissions(permission_types)` included caused every user to be given every policy id, so `accessible_by` would return all items regardless of actual permissions. Removing that method assigned the proper policy ids, but the admin user was still not being returned everything for some reason. So I added a ternary conditional in the media_object_controller to return `MediaObject.all` if the user is an admin.

I'm unclear why the admin's policy id didn't return everything. Here's the solr query generated from `filters`: 
```
get policies: [{"id"=>"v979v304g"}]


get policies: []


"_query_:\"{!terms f=edit_access_group_ssim}public,administrator,manager,group_manager,registered,172.18.0.1\" 
OR _query_:\"{!terms f=discover_access_group_ssim}public,administrator,manager,group_manager,registered,172.18.0.1\" 
OR _query_:\"{!terms f=read_access_group_ssim}public,administrator,manager,group_manager,registered,172.18.0.1\" 
OR edit_access_person_ssim:archivist1@example.com 
OR discover_access_person_ssim:archivist1@example.com
 OR read_access_person_ssim:archivist1@example.com 
OR (_query_:\"{!raw f=isGovernedBy_ssim}v979v304g\")"
```

masaball (commented):
@cjcolvar This all looks great. Everything looks to be working as expected when doing manual checks and all the tests are passing. Thank you again for all the work and help you provided for this one. 

Since it's my PR, I can't approve it but if you're satisfied with it I think we're good to approve and merge.

cjcolvar (approved):

PR #4881, opened 2022-09-16
Fix video player aspect ratio and size for mobile devices

Dananji:
This PR has the following changes:
1. Scroll to main content in smaller mobile devices when video player is not displayed entirely
2. Reduce the overall font-size for mobile devices so that content fits into mobile device screen sizes
3. Change default video player height and width to adhere 16:9 aspect ratio

### DIFF
--- a/app/views/modules/player/_video_element.html.erb
+++ b/app/views/modules/player/_video_element.html.erb
@@ -18,8 +18,8 @@ Unless required by applicable law or agreed to in writing, software distributed
                 class="mejs-avalon <%= is_mejs_2? ? '' : 'invisible' %>"
                 controls
                 style="width: 100%; height: 100%"
-                width="<%= @player_width || 450 %>"
-                height="<%= @player_height || 309 %>"
+                width="<%= @player_width || 480 %>"
+                height="<%= @player_height || 270 %>"
                 data-canvasindex=0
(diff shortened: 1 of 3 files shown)


app/views/modules/player/_video_element.html.erb
```diff
@@ -18,8 +18,8 @@ Unless required by applicable law or agreed to in writing, software distributed
                 class="mejs-avalon <%= is_mejs_2? ? '' : 'invisible' %>"
                 controls
                 style="width: 100%; height: 100%"
-                width="<%= @player_width || 450 %>"
-                height="<%= @player_height || 309 %>"
+                width="<%= @player_width || 480 %>"
+                height="<%= @player_height || 270 %>"
```
cjcolvar, line 22: Will this change still work with pre-existing embeds which are iframes fixed at 600x450?

Dananji: I'm not sure. I will test that :+1:

Dananji: This change will not affect the embeds, because the `@player_height` and `@player_width` are set to `100%` in the [controller](https://github.com/avalonmediasystem/avalon/blob/main/app/controllers/master_files_controller.rb#L78-L79). So the player will span the to given height and width of the iframe, which is 450x600

cjcolvar (approved):

PR #4882, opened 2022-09-19
LDAP search will return nil sometimes instead of []

cjcolvar:
Based upon honeybadger errors from production.


masaball (approved):
Looks good!

PR #4893, opened 2022-09-27
500 error handling

masaball:


cjcolvar (commented):
Looking good.  I had a few suggestions that I think would improve it.  See what you think.

cjcolvar (approved):
Looks good!

PR #4901, opened 2022-10-03
Log time spent doing LDP and Solr requests

cjcolvar:
With this change rails logs the total time spent running LDP and Solr requests within a request cycle and logs them alongside the view and DB times.

```
Completed 302 Found in 88ms (ActiveRecord: 1.8ms | LDP: 20.4ms | Solr: 6.0ms | Allocations: 80960)
```

I removed the ActiveFedora override initializer for request options because we've upgraded to a version of ActiveFedora which already includes those changes.


masaball (approved):
Looks good!

PR #4904, opened 2022-10-05
Configure sorting by solr timestamp for atom feed

cjcolvar:
This fixes a regression for a feature we use at IU.  Sorting the atom feed so most recently updated records appear first was working before the Blacklight upgrade.  Blacklight 7 only includes configured sort orders in requests to solr.  Registering this sort in CatalogController would have the side effect of displaying it in the search UI but setting `if: false` hides it.


masaball (approved):
Looks good!

PR #4913, opened 2022-10-11
Try upgrading to node 14

cjcolvar:
Timeliner is being built with node 14 and earlier versions of node are EOL.


Dananji (approved):
Looks great :+1: I tested this locally and everything is working for me :tada:

PR #4924, opened 2022-10-18
Speed up media objects show page for many sectioned items

cjcolvar:
Pull in newer version of `speedy_af` which has an option to bundle the solr requests into a single large request and enable this feature in MediaObjectsController.  Also fix a case where fedora requests were happening instead of using the speedy_af proxies.


masaball (approved):
Looks good!

PR #4929, opened 2022-10-24
Remove non-existent tokens from session before creating new ones

cjcolvar:
Expired tokens are deleted from the database whenever a new token is created.  This leaves orphaned tokens in user's sessions.  This change would also clean up a user's session before creating new tokens thus avoiding a growing list of tokens which eventually leads to a SessionOverflowError.


masaball (approved):
Looks good!

PR #4932, opened 2022-10-25
Properly handle case when master file has been deleted

### LINKED ISSUE #4931: Thumbnails not being generated when using non-s3 storage and delete master file management strategy
Douglas Hahn from Texas A&M came across this odd behavior in a test Avalon 7.2.1 instance that is running in a docker environment.  The avalon and worker containers share mounted volumes and the master file management strategy was set to 'delete'.  Douglas and I did some debugging and found that after a file has been processed the delete strategy sets the `file_location` to empty string.  When a user later attempts to change the thumbnail image it reports success on the Manage Files step but doesn't actually change the image.

The first issue we found is that the `ExtractStillJob` is getting the Rails root as the source location due to a test for `nil?` instead of `blank?` on the master file's `file_location`.  It should be falling back to the `'high'` derivative or an HLS stream.

After fixing this test, we found that the `'high'`derivative was being used as the source file (since it was accessible via the mounted volume).  The thumbnail generation still failed but it also deleted the derivative!

Expected behavior is thumbnails should be able to be regenerated after initial ingest no matter what master file management strategy is used and derivatives should definitely NOT be deleted as part of the process.

cjcolvar:
Resolves #4931 

Related: https://github.com/avalonmediasystem/avalon/pull/3475

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -559,10 +559,10 @@ class MasterFile < ActiveFedora::Base
     options[:offset] ||= 2000
 
     source = FileLocator.new(working_file_path&.first || file_location)
-    options[:master] = true
+    options[:non_temp_file] = true
     if source.source.blank? or (source.uri.scheme == 's3' and not source.exist?)
       source = FileLocator.new(self.derivatives.where(quality_ssi: 'high').first.absolute_location)
-      options[:master] = false
+      options[:non_temp_file] = true
     end
     response = { source: source&.location }.merge(options)
     return response if response[:source].to_s =~ %r(^https?://)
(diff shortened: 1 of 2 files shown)


app/models/master_file.rb
```diff
@@ -559,32 +559,35 @@ def find_frame_source(options={})
     options[:offset] ||= 2000
 
     source = FileLocator.new(working_file_path&.first || file_location)
-    options[:master] = true
-    if source.source.nil? or (source.uri.scheme == 's3' and not source.exist?)
+    options[:non_temp_file] = true
+    if source.source.blank? or (source.uri.scheme == 's3' and not source.exist?)
       source = FileLocator.new(self.derivatives.where(quality_ssi: 'high').first.absolute_location)
-      options[:master] = false
+      options[:non_temp_file] = true
```
cjcolvar, line 565: I think this is the key change for keeping derivative files from getting deleted by the unlinking on line 639.  I changed the key to better reflect the expectations around this.

masaball (approved):
Looks good!

PR #4942, opened 2022-11-03
Add ability to enable/disable CDL per collection

masaball:


cjcolvar (commented):
It looks like this does away with an application level enable/disable of CDL such that it is always available as an option at the collection level.  I haven't gone back and reread the ticket but I was thinking that the global flag could still be useful to hide the option at the collection level if you know you'll never want to use CDL.

app/helpers/security_helper.rb
```diff
@@ -36,6 +36,6 @@ def add_stream_url(stream_info)
   private
 
     def not_checked_out?(media_object_id)
-      Avalon::Configuration.controlled_digital_lending_enabled? && Checkout.checked_out_to_user(media_object_id, current_user&.id).empty?
+      MediaObject.find(media_object_id).cdl_enabled && Checkout.checked_out_to_user(media_object_id, current_user&.id).empty?
```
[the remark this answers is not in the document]
masaball: Seems like it is. Some of the rspec tests initially didn't like it but an `include CdlHelper` cleared it up. 

Specifically, it was the master_files_controller_spec that I saw having issues. It would cause an error about the MasterFileController not implementing `lending_enabled?`. The `include` line isn't strictly necessary if I'm understanding how views and helpers work, but I guess `render_views` in a controller spec doesn't quite treat or load things the same way as the application itself?

masaball:
There is still a system enable/disable, though you are right that it does not hide the collection CDL option. It just sets all of the collections to be enabled or disabled. I think you are right that there should be an option to hide the option at collection level.

Does it make more sense to have it tied to the system enable/disable or should it be a separate flag?

app/controllers/admin/collections_controller.rb
```diff
@@ -305,11 +305,12 @@ def update_access(collection, params)
 
     collection.default_visibility = params[:visibility] unless params[:visibility].blank?
     collection.default_hidden = params[:hidden] == "1"
-    if Avalon::Configuration.controlled_digital_lending_enabled?
+    collection.default_enable_cdl = params[:cdl] == "1"
```
cjcolvar, line 308: I like using the ActiveModel casting for converting to a boolean.
```suggestion
    collection.default_enable_cdl = ActiveModel::Type::Boolean.new.cast(params[:cdl])
```

masaball: I don't think that the ActiveModel casting is going to work in this case. If CDL is turned on at the system level, then it becomes impossible to turn it off for individual collections. Unchecking the box results in `params[:cdl]` being `nil` which gets cast to `nil`. The model sees that `collection.default_enable_cdl == nil` and falls back to the system setting.

There is potentially some html/javascript that could be setup in the view to force the unchecked box to send `0` as the param, which would be cast to `false`. But I don't know if that solution is any better/cleaner than leaving the `params[:cdl] == 1`.

app/models/admin/collection.rb
```diff
@@ -277,6 +280,17 @@ def default_lending_period
     self._default_lending_period || ActiveSupport::Duration.parse(Settings.controlled_digital_lending.default_lending_period).to_i
   end
 
+  alias_method :_cdl_enabled, :cdl_enabled
+  def cdl_enabled
```
cjcolvar, line 284: What do you think about renaming this `cdl_enabled?`?   Then it would match the method name in MediaObject, avoid the alias_method, and simplify the helper.

spec/controllers/master_files_controller_spec.rb
```diff
@@ -252,6 +249,10 @@ class << file
 
     context 'with fedora 3 pid' do
       let!(:master_file) { FactoryBot.create(:master_file, identifier: [fedora3_pid]) }
+      let(:media_object) do
+        instance_double('MediaObject', title: 'Media Object', id: 'avalon:1234',
+                                       ordered_master_files: [master_file], master_file_ids: [master_file.id])
+      end
```
cjcolvar, line 255: I'm not seeing where this is used in this example.  Maybe it could be removed?

cjcolvar (commented):
A couple small suggestions but it is looking good!

spec/controllers/master_files_controller_spec.rb
masaball: It is necessary for the check at ln 259. There may be a way to refactor it to not need the instance double but everything I've tried seems to cause an infinite loop.

cjcolvar (approved):

cjcolvar (approved):

PR #4947, opened 2022-11-04
CUD operations on supplemental files should check if can edit owning objects

cjcolvar:
Currently the supplemental files controller is checking if the user can take the same action requested for the supplemental file on the owning object.  This works for administrators since they can take any action, but managers are blocked from creating supplemental files for master files because can :create, MasterFile isn't ever explicitly granted for any user.  This PR proposes only checking for edit privileges on the owning object since that better matches the actual change to the owning object: adding or removing a reference to a supplemental file.  This change would allow all users who have the ability to edit the MasterFile to manipulate supplement files.

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -145,6 +145,7 @@ class SupplementalFilesController < ApplicationController
     end
 
     def authorize_object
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good. (nice branch name, btw)

PR #4950, opened 2022-11-09
Bring in newer active_encode and reduce completeness threshold to 95%

### LINKED ISSUE #4855: Sanitize Filenames Before File Processing
# Description
Following on from investigation in #3926; fixing issues caused by characters in filenames. Using a subset of the AWS object guidelines would be ideal. [Mason's notes](https://github.com/avalonmediasystem/avalon/issues/3926#issuecomment-1226430308) in that ticket contain more detailed information.

## Safe Characters

- Alphanumeric characters 	
    0-9
    a-z
    A-Z
- Special characters 	
   Hyphen (-)
    Underscore (_)

# Done Looks Like
- [x] Filename sanitization happens before any further handling and processing (consistently and adapter-independent)
  - [x] Characters not matching the safe set above in filenames are changed to a value within the safe set (unsafe characters are always changed to underscore or hyphen)
  - [x] Spaces in filename are replaced with underscores
  - [x] Periods in the filename removed (excluding the standard period between the name and the extension). 
- [x] Hard-coded [check to fail non-ASCII characters](https://github.com/avalonmediasystem/avalon/blob/90832296809f3684405acb7b4e44d7ed0720d92e/app/services/master_file_builder.rb#L37) removed 
- [ ] Cut release for Active Encode
- [ ] Update gemfile in Avalon
- [ ] PR to Avalon to set threshold (0, or 95%)

cjcolvar:
Resolves #4855 and works around local issue with ffmpeg possibly barely truncating very short files


masaball (approved):
Looks good!

PR #4952, opened 2022-11-10
Bug fix for dynamic fields in resource description form

Dananji:
Include the following CSS fix:

Before:
mco-css-form
After:
mco-css-form-fix


cjcolvar (approved):
:+1:

Issue #4951, opened 2022-11-10
Closed captions in HLS manifest for iOS devices in ME.js


Dananji:
### Description

For closed captions to work with MediaElement.js in iOS devices, the captions needs to be transcoded in the video according to ME.js [documentation](https://github.com/mediaelement/mediaelement/blob/master/docs/installation.md#closed-captioning).
To make this work with streaming media in Avalon, we could add the captions information in our HLS manifests.

# Done Criteria
- [ ] Investigate on possible ways to add captions to m3u8 manifest

According to Chris, we proxy the wowza manifests through Avalon.
Streaming example with captions from Apple: https://developer.apple.com/streaming/examples/basic-stream-osx-ios5.html with the following  manifest:
```
#EXTM3U

#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="bipbop_audio",LANGUAGE="eng",NAME="BipBop Audio 1",AUTOSELECT=YES,DEFAULT=YES
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="bipbop_audio",LANGUAGE="eng",NAME="BipBop Audio 2",AUTOSELECT=NO,DEFAULT=NO,URI="alternate_audio_aac/prog_index.m3u8"


#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="English",DEFAULT=YES,AUTOSELECT=YES,FORCED=NO,LANGUAGE="en",CHARACTERISTICS="public.accessibility.transcribes-spoken-dialog, public.accessibility.describes-music-and-sound",URI="subtitles/eng/prog_index.m3u8"
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="English (Forced)",DEFAULT=NO,AUTOSELECT=NO,FORCED=YES,LANGUAGE="en",URI="subtitles/eng_forced/prog_index.m3u8"
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="Français",DEFAULT=NO,AUTOSELECT=YES,FORCED=NO,LANGUAGE="fr",CHARACTERISTICS="public.accessibility.transcribes-spoken-dialog, public.accessibility.describes-music-and-sound",URI="subtitles/fra/prog_index.m3u8"
[...]

PR #4953, opened 2022-11-10
Fix for ios fullscreen captions not displaying

Dananji:
This fix overrides the native iOS player in fullscreen mode and display a similar interface of the avalon's media player. This way the captions are accessible in a similar way to the regular player.
ios fakefullscreen

This could be a temporary fix from a UI perspective. Related ticket: #4951


joncameron (approved):
This looks good to me. The fake fullscreen option seems to be a reasonable strategy to resolve the iOS caption issue right now and there's not much to change if there's a different strategy for handling this on mobile in the future.

PR #4954, opened 2022-11-11
Refactor find_all_files to clean up processes if a timeout happens

cjcolvar:
TODO: Write test which sleeps when lsof call happens?


cjcolvar:
I was able to reproduce this issue locally in my docker-compose environment, but this change makes things worse because the timeout doesn't break the wait on the lsof process.

cjcolvar:
This didn't work and should hopefully be obviated by using uppy.

PR #4955, opened 2022-11-15
Update JS dependencies and remove resolutions if good version is in yarn.lock

cjcolvar:
Removing a few of these resolutions allowed for automatic upgrades to new major versions.  I still had to keep set-value and kind-of to ensure we'd still be using newer versions which include security fixes.


Dananji (approved):
Looks great :+1:

PR #4956, opened 2022-11-15
Cdl embeds

masaball:
Hide embed menu when cdl is enabled and render "This item is unable to be embedded" for existing embeds now protected by cdl.

### DIFF
--- a/spec/controllers/media_objects_controller_spec.rb
+++ b/spec/controllers/media_objects_controller_spec.rb
@@ -812,7 +812,7 @@ describe MediaObjectsController, type: :controller do
               FactoryBot.create(:checkout, media_object_id: media_object.id, user_id: controller.current_user.id)
               get :show, params: { id: media_object.id }
               expect(response).to render_template(:_share_resource)
-              expect(response).to render_template(:_embed_resource)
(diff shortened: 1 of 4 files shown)


spec/controllers/media_objects_controller_spec.rb
```diff
@@ -812,31 +812,31 @@
               FactoryBot.create(:checkout, media_object_id: media_object.id, user_id: controller.current_user.id)
               get :show, params: { id: media_object.id }
               expect(response).to render_template(:_share_resource)
-              expect(response).to render_template(:_embed_resource)
+              expect(response).not_to render_template(:_embed_resource)
               expect(response).to render_template(:_lti_url)
             end
             it "managers: should include lti, embed, and share" do
```
cjcolvar, line 818: Can you please update the labels for these tests so they reflect the tests now?

cjcolvar (commented):
Looks great!

spec/controllers/media_objects_controller_spec.rb
masaball: Oh, yep. Sorry about that.

cjcolvar (approved):

PR #4973, opened 2022-11-18
Perform pass through encode job later to ensure it runs on same server as subsequent polling jobs

### LINKED ISSUE #4968: Encoding Hangs for Web Upload with Skip Transcoding
# Description
A web upload on staging is hung at 1% complete:
https://avalon-staging.dlib.indiana.edu/media_objects/5425k970t
https://avalon-staging.dlib.indiana.edu/encode_records/263

cjcolvar:
The pass through encoding adapter creates directories and files on local disk that need to be read later when the encode is looked up in the polling job.  In the docker-compose environment with avalon and the background worker in separate containers the encode create and polling jobs need to be run on the same container so the files will be present.  Alternatively the work directory could be externalized and mounted into each container.  But it makes the most sense to run the create job as a background job because it is expensive and running it later on the background worker makes it align with non skip-transcoding encodes.

Fixes #4968

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -270,7 +270,7 @@ class MasterFile < ActiveFedora::Base
       options[:outputs] = [{ label: 'high', url: input }]
     end
 
-    ActiveEncodeJobs::CreateEncodeJob.perform_now(input, id, options)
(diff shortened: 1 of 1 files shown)


Dananji (dismissed):
:+1:

Dananji (approved):

PR #4976, opened 2022-11-21
Fix date sorting for paged_index datatables

### LINKED ISSUE #4963: Sorting playlist table by "Created" or "Updated" causes errors
The UI for the playlist table indicates that sorting is available on the "Created" and "Updated" columns, however when a user attempts to sort by either column they get a `DataTables warning: table id=Playlists - Ajax error.`

Sorting for those two columns needs to be reconfigured or removed.

masaball:
Fixes #4963 

This PR fixes a bug where datatables would generate an error when attempting to sort by date columns (updated_at, created_at, etc.)  The error was first noticed on the playlists page, but I checked the implementation and it was present on the timelines and persona users pages as well. The root cause was it attempting to convert date/time formatted data to lowercase and that caused an internal server error. By setting up a case specifically for the date columns to not convert to lowercase, this issue should be resolved.

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2022, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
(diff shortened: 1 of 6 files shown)


cjcolvar (approved):
Looks great!

PR #4979, opened 2022-11-28
Return to stored location if available

### LINKED ISSUE #4965: Shared playlist login does not return user to the playlist
If a user is not already authenticated and accesses a playlist they have a direct link for, the playlist page prompts the user to try logging in. The login link takes the user to the main login page but then redirects them to the Avalon homepage after they have logged in, requiring the user to re-navigate to the shared playlist.

Users should be round-tripped back to the shared playlist after logging in.

cjcolvar:
Fixes #4965

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -58,7 +58,7 @@ class ApplicationController < ActionController::Base
   end
 
   def store_location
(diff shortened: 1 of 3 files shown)


Dananji (approved):
Looks good! :+1:

PR #4982, opened 2022-11-30
Move all hotkey enable/disable code to ME.js module

Dananji:


vendor/assets/javascripts/mediaelement/mediaelement-and-player.js
```diff
@@ -5165,7 +5165,16 @@ var MediaElementPlayer = function () {
...
+			var activeElement = document.activeElement;
+			var inputs = ['input', 'textarea'];
+			/* Unless the focused element is either an input field or a textarea,
+				 bind keyboard inputs for player actions. e.g. key 'm' press => mute the player 
+			*/
+			if (activeElement && inputs.indexOf(activeElement.tagName.toLowerCase()) !== -1) {
+				return;
+			}
+			else if (player.options.enableKeyboard) {
 				for (var i = 0, total = player.options.keyActions.length; i < total; i++) {
```
Dananji, line 5178: I had missed this when doing the last ME.js upgrade. Without this code snippet, we have to enable and disable hotkeys in a lot of places in the code where there are text input fields with the player on the page. I noted this change in the ME.js upgrade customization list, so that we won't miss it in the future.

cjcolvar (approved):

PR #4985, opened 2022-11-30
Convert utc dates to local time zone on client-side using moment.js

### LINKED ISSUE #4981: Checkout / return times for CDL items are 5 hours ahead
Items that are checked out as part of controlled digital lending display checkout / return datetime on the Checkouts page, as well as an expected return time on the media item itself when viewed by other users.

These times seem to be 5 hours ahead. I checked out an item on 11/30/2022 at 11:03am and it listed the return time as 12/14/2022 at 4:03pm. The checkout time on the checkouts page was also listed as 11/30/2022 4:03pm.

cjcolvar:
With users potentially all over the world it makes sense to store checkout and return times in UTC in the database.  This is currently happening, but users will expect to see times in their local time zones.  There is no easy way to know what time zone a request is coming from without having users set a default time zone.  However, in the front end the browser can tell us the device's timezone.  This PR suggests passing the UTC time to the front end and then have javascript convert it into a localized time on page load.

Fixes #4981

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -41,6 +41,7 @@
 //= require cropperjs/dist/cropper.min
 //= require url-search-params-polyfill/index.js
 
+//= require moment/min/moment-with-locales.min.js
 //= require hls.js/dist/hls.min.js
 
 // include all of our vendored js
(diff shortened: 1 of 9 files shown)


Dananji (approved):
Looks great! :100:

Dananji (approved):
Looks great :100:

PR #4988, opened 2022-12-01
Redirect unauthenticated users from checkouts page 

### LINKED ISSUE #4987: Attempting to load Checkouts page when not signed in generates error
# Description

I coped and pasted a URL to the Checkouts page into an incognito browser because I wanted to re-test in a clean browser. When I tried to load https://avalon-staging.dlib.indiana.edu/checkouts as a user not logged in, I got this error:

# Done Looks Like

- [x] Avalon re-directs users who try to access the Checkouts page to the login page, then back to the Checkouts page after login

masaball:
Fixes #4987 

This adds a safe navigation operator to the methods that assemble the current user's list of checkouts. Doing this prevents a nil class error for unauthenticated users and allows the `errors/restricted_pid` partial to render, directing the user to login to access the content.

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -95,7 +95,7 @@ class CheckoutsController < ApplicationController
         @checkouts = if current_ability.is_administrator?
                        set_active_checkouts.or(Checkout.all.where("return_time <= now()"))
                      else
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Looks good!

PR #4989, opened 2022-12-01
Remove "Item Unavailable" message from Checkouts table

### LINKED ISSUE #4983: Checkouts Page Displays "Item Unavailable" Instead of Checkout/Return Button
# Description
In avalon-staging, the Checkouts page for the admin user is displaying an "Item Unavailable" message for valid items instead of the Checkout/Return button. This is visible after checking the "Display all" box.

# Expected Behavior
A button renders if the item is available (i.e it is an existing item in the system).

masaball:
Fixes #4983

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -137,8 +137,10 @@ class CheckoutsController < ApplicationController
         view_context.link_to('Return', return_checkout_url(checkout), class: 'btn btn-outline btn-xs', method: :patch)
       elsif checkout.return_time < DateTime.current && checkout.media_object.lending_status == 'available'
         view_context.link_to('Checkout', checkouts_url(params: { checkout: { media_object_id: checkout.media_object_id } }), class: 'btn btn-primary btn-xs', method: :post)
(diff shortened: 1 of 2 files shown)


app/controllers/checkouts_controller.rb
```diff
@@ -138,7 +138,7 @@ def checkout_actions(checkout)
       elsif checkout.return_time < DateTime.current && checkout.media_object.lending_status == 'available'
         view_context.link_to('Checkout', checkouts_url(params: { checkout: { media_object_id: checkout.media_object_id } }), class: 'btn btn-primary btn-xs', method: :post)
```
cjcolvar, line 139: Should this case be removed as well to match the change in index.html.erb?

masaball: Good question. @joncameron, If an item is not checked out should there be a checkout button next to the historical checkouts or should there only be `Return` buttons for current checkouts being rendered in the last column? 

It will probably be cleaner and more consistent to have all historical checkouts blank, and a user can just use the link to the item page to try to check something out again.

joncameron: @masaball I agree, it would be better to have all historical checkouts blank. Let's go that route.

PR #4993, opened 2022-12-02
Change CDL message for existing embeds & re-add embed share tab

masaball:
This PR adds the `Embed` tab back to the `Share` menu for items with CDL enabled, replacing the iframe with the message "This item cannot be embedded". 

It also changes the message displayed in existing embeds after CDL is enabled to "This item is no longer able to be embedded. You may be able to view this item on * _Name of Service_ *", which links to the item's page in Avalon.


cjcolvar (approved):
Looks good!

PR #4994, opened 2022-12-02
Remove checkout buttons for historical checkouts

masaball:
As discussed [here](https://github.com/avalonmediasystem/avalon/pull/4989#discussion_r1038426498), it was decided that the only buttons in the checkouts table should be the `Return` button for currently active checkouts. All historical checkouts should have the final column be blank.

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -133,9 +133,8 @@ class CheckoutsController < ApplicationController
     end
 
     def checkout_actions(checkout)
-      if checkout.return_time > DateTime.current
-        view_context.link_to('Return', return_checkout_url(checkout), class: 'btn btn-outline btn-xs', method: :patch)
-      end
+      return unless checkout.return_time > DateTime.current
+      view_context.link_to('Return', return_checkout_url(checkout), class: 'btn btn-outline btn-xs', method: :patch)
     end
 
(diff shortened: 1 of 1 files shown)


app/controllers/checkouts_controller.rb
```diff
@@ -133,13 +133,8 @@ def time_remaining(checkout)
...
 
     def checkout_actions(checkout)
-      if checkout.return_time > DateTime.current
-        view_context.link_to('Return', return_checkout_url(checkout), class: 'btn btn-outline btn-xs', method: :patch)
-      elsif checkout.return_time < DateTime.current && checkout.media_object.lending_status == 'available'
-        view_context.link_to('Checkout', checkouts_url(params: { checkout: { media_object_id: checkout.media_object_id } }), class: 'btn btn-primary btn-xs', method: :post)
-      else
-        ''
-      end
+      return unless checkout.return_time > DateTime.current
```
cjcolvar, line 136: Is `nil` an okay return value or should it be empty string like it was before?

masaball: `nil` wasn't causing issue on my local instance but thinking on it, using an empty string is probably safer. Preemptively avoid some weird edge case/browser issue where it decides it doesn't like `nil`... Will change it to return empty string.

cjcolvar (approved):
Looks good.

PR #4995, opened 2022-12-02
Covert utc dates whenever datatables redraws

### LINKED ISSUE #4981: Checkout / return times for CDL items are 5 hours ahead

cjcolvar:
Fixes #4981.

### DIFF
--- a/app/assets/javascripts/localize_times.js
+++ b/app/assets/javascripts/localize_times.js
@@ -1,6 +1,9 @@
 // Requires moment.js
-$(document).ready(function () {
+function localize_times() {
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

PR #4999, opened 2022-12-05
Store only user relevant locations for later redirection post-login

cjcolvar:
Resolves bug introduced by #4988

### DIFF
--- a/spec/features/login_redirect_spec.rb
+++ b/spec/features/login_redirect_spec.rb
@@ -1,0 +1,43 @@
+# Copyright 2011-2022, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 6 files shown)


spec/features/login_redirect_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+require 'rails_helper'
+
+describe 'Login Redirects', js: true do
```
cjcolvar, line 17: The m3u8 gets called from the player JS so we need to have JS enabled here.  But JS, capybara, webmock, and the docker environment are not playing together well so things break for me locally.  It might be that we need to figure out test cases that don't require JS enabled.  Like maybe we just simulate the request by manually fetching the m3u8 before attempting login.

PR #5002, opened 2022-12-06
Use class instead of id since the id wasn't unique

cjcolvar:
With multiple files associated with a media object attempting to move any files after the first would cause the form target url be for the media object instead of the file leading to a 500 ("ActiveFedora::ModelMismatch (Expected MasterFile. Got: MediaObject)")


masaball (approved):
Looks good!

PR #5003, opened 2022-12-06
Change restricted redirect param to user_return_to

masaball:


cjcolvar (commented):
This looks good, but I'm wondering if it is possible to remove the usage of `previous_url` on line 79 of ApplicationController.

cjcolvar:
You're totally right.  I forgot about that scenario.

cjcolvar (dismissed):

cjcolvar (approved):

PR #5006, opened 2022-12-07
Create collection page

### LINKED ISSUE #4975: Bad State for Create Collection Modal
# Description
When creating a new collection via the the "create collection" button in the pop-up interface, the request to create the collection can fail and the user receives an error message. When re-opening the modal, the "create collection" button on the interface greyed out and the text changes to "saving...". 

# Expected Behavior
- [ ] When the modal is opened, the create collection button does not retain the grayed-out "saving..." state from a previous opening.

# Notes from Chris
I think what happened here is that the original create request failed but without a page reload the modal got into an odd state leading to this confusing behavior.  I wonder if we can fix that or if it would be better UX to avoid the modal and use a real page for collection creation.

Dananji:
Fixes #4975 

Changes were made in this PR to move create collection from modal to a new page for better UX and error handling in the GUI.

admin_collection_new

### DIFF
--- a/app/views/admin/collections/index.html.erb
+++ b/app/views/admin/collections/index.html.erb
@@ -16,69 +16,68 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% @page_title = t('collections.title', :application_name => application_name) %>
 
 <% unless @collections.empty? %>
-<div class="row display-flex page-title-wrapper">
-  <div class="col-xs-6">
-    <h1 class="page-title">My Collections</h1>
+  <div class="row display-flex page-title-wrapper">
+    <div class="col-xs-6">
+      <h1 class="page-title">My Collections</h1>
+    </div>
+    <div class="col-xs-6 collection-btn">
+      <%= link_to('<span class="btn btn-primary headline-button">Create Collection</span>'.html_safe, new_admin_collection_path) unless cannot? :create, Admin::Collection %>
+    </div>
   </div>
-  <div class="col-xs-6 collection-btn">
-    <%= button_tag("Create Collection", class: 'btn btn-primary btn-large', data: {toggle:"modal", target:"#new_collection"}) unless cannot? :create, Admin::Collection %>
-  </div>
-</div>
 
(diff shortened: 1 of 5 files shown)


app/views/admin/collections/index.html.erb
```diff
@@ -13,72 +13,76 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
+
+<% unless Avalon::ControlledVocabulary.vocabulary[:units] %>
+  <% raise Avalon::VocabularyNotFound.new "Units vocabulary not found." %>
+<% end %>
```
cjcolvar, line 19: Is this test and raise necessary here?  How about putting this check in the `new.html.erb` view where it renders an error message (e.g. "Units vocabulary is needed for creating collections, but it is missing.") instead of the form?  That way the user can see what's wrong instead of getting a 500 or does this error get rescued higher up in an elegant way?

cjcolvar (commented):
This looks great!  I had only one question about a small part.

app/views/admin/collections/index.html.erb
masaball: For reference, that error is rescued in the [controller](https://github.com/avalonmediasystem/avalon/blob/53a77352413c21ffbc72b8bce86e60988043839c/app/controllers/admin/collections_controller.rb#L261).

Dananji: I'm not sure I know how to trigger this error in the user interface. Do you know how I can do that?

Dananji: Looking at the tests, it seems missing controlled vocabulary should redirect you to root and show a flash error message. So, it seems it's getting handled well in the code.

cjcolvar: :+1: , that sounds fine.
It would be nice to rework this at some point but that is totally outside the scope of this PR.

cjcolvar (approved):
I think this is a great change!

PR #5010, opened 2022-12-12
Resize XML editor window to make space in the editor

Dananji:
**Before:**
xml-editor-old

**After:**
xml-editor-new

Fix for: https://bugs.dlib.indiana.edu/browse/VOV-6536


cjcolvar (approved):
Looks good to me!

PR #5012, opened 2022-12-13
Use latest build of development image

cjcolvar:
This avoids us needing to remember to update the image manually.


Dananji (approved):
:100:

PR #5021, opened 2022-12-15
Add structural metadata param to additional endpoints

### LINKED ISSUE #4928: Avalon API should return structural metadata in all cases where requested
The Avalon API will return the structural metadata if requested on the media_objects json endpoint.  Media objects can also be returned in the `admin/collection/:id/items.json` endpoint but it does not return the structural metadata even if requested.  It is possible that there are other endpoints that have the same behavior.

# Done Looks Like
- [ ] Return structural metadata with media object from all endpoints when the `include_structure` query parameter is true 

https://samvera.atlassian.net/wiki/spaces/AVALON/pages/1957954917/Avalon+API

masaball:
Fixes #4928

This PR adds the `include_structure` parameter to the `/admin/collections/:id/items.json` and `/media_objects.json` end points. Avalon should now return structural metadata, when requested, for all API endpoints that return media objects.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -321,7 +321,7 @@ class MediaObjectsController < ApplicationController
     mos = MediaObject.accessible_by(current_ability, :index)
     respond_to do |format|
       format.json {
-        paginate json: mos.to_a.collect{|mo| mo.as_json(include_structure: params[:include_structure] == "true") }
+        paginate json: mos.to_a.collect { |mo| mo.as_json(include_structure: params[:include_structure] == "true") }
       }
     end
   end
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -318,9 +318,10 @@ class MediaObjectsController < ApplicationController
   end
 
   def index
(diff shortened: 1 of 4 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -318,9 +318,10 @@ def custom_update
   end
 
   def index
+    mos = MediaObject.accessible_by(current_ability, :index)
     respond_to do |format|
       format.json {
-        paginate json: MediaObject.accessible_by(current_ability, :index)
+        paginate json: mos.to_a.collect { |mo| mo.as_json(include_structure: params[:include_structure] == "true") }
```
cjcolvar, line 324: Does the paginate method need to be moved up with the `accessible_by` relation to match Admin::CollectionsController?  I'm concerned that this might be accidentally loading more than it needs to without this change which could impact performance at scale.  I'm not certain at all but it is a guess.

As far as I recall `paginate` should add limit and paging filters on the solr query.  What I'm not sure about is if `mos.to_a` will execute before or after adding those filters.  I'm not sure what the best way to test this would be but manually stepping through the code might be a good enough effort for now.

cjcolvar (approved):
Looks good!

PR #5030, opened 2023-01-17
Enable rendering of transcript viewer when CDL is disabled

### LINKED ISSUE #5028: Transcript viewer does not render when CDL is disabled
In testing for #5023, it eventually clicked that the transcript viewer was not rendering at all for items that had CDL disabled. This is because of a line in the [app/views/media_objects/_item_view.html.erb](https://github.com/avalonmediasystem/avalon/blob/53a77352413c21ffbc72b8bce86e60988043839c/app/views/media_objects/_item_view.html.erb#L56) partial. [`can_stream`](https://github.com/avalonmediasystem/avalon/blob/53a77352413c21ffbc72b8bce86e60988043839c/app/views/media_objects/_item_view.html.erb#L24) is always false when CDL is disabled because it checks if the current user has the media object checked out.

masaball:
Fixes #5028

### DIFF
--- a/app/views/media_objects/_item_view.html.erb
+++ b/app/views/media_objects/_item_view.html.erb
@@ -21,7 +21,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 
 <% @poster_url = request.base_url + @currentStreamInfo[:poster_image] unless @currentStreamInfo[:poster_image].nil? %>
 <% @page_type = @currentStreamInfo[:is_video] ? 'video' : 'audio' %>
-<% can_stream = lending_enabled?(@media_object) ? current_user && @media_object.current_checkout(current_user.id).present? : current_user %> 
+<% can_stream = lending_enabled?(@media_object) ? current_user && @media_object.current_checkout(current_user.id).present? : true %> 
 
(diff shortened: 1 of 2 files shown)


app/views/media_objects/_item_view.html.erb
```diff
@@ -21,7 +21,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 
 <% @poster_url = request.base_url + @currentStreamInfo[:poster_image] unless @currentStreamInfo[:poster_image].nil? %>
 <% @page_type = @currentStreamInfo[:is_video] ? 'video' : 'audio' %>
-<% can_stream = current_user && @media_object.current_checkout(current_user.id).present? %>
+<% can_stream = lending_enabled?(@media_object) ? current_user && @media_object.current_checkout(current_user.id).present? : current_user %>
```
cjcolvar, line 24: I can't recall what `current_user` is if a user isn't signed in.  Would it be safer to return `true` in the non-CDL case?
```suggestion
<% can_stream = lending_enabled?(@media_object) ? current_user && @media_object.current_checkout(current_user.id).present? : true %> 
```

cjcolvar (approved):
Looks good!

PR #5042, opened 2023-01-30
Bump ActiveEncode and update FileLocator

masaball:


Gemfile
```diff
@@ -70,7 +70,7 @@ gem 'omniauth-lti', git: "https://github.com/avalonmediasystem/omniauth-lti.git"
 gem "omniauth-saml", "~> 2.0"
 
 # Media Access & Transcoding
-gem 'active_encode', '~> 1.0'
+gem 'active_encode', '~> 1.1.2'
```
cjcolvar, line 73: This would restrict to only 1.1.x versions so I think this might be better:
```suggestion
gem 'active_encode', '~> 1.0', '>= 1.1.2'
```

masaball: Right. Good call. Forgot that was how the `~>` works.

cjcolvar (approved):
Looks good!

PR #5043, opened 2023-01-31
Allow file uris to include ? or # in filename

cjcolvar:
In case file name includes ? or # use full uri omitting components before path instead of using only path which would miss query or fragment components

This was found in production where `Report from the Statehouse WTTV (#720-20-1977_5-1.mp4Edited).mp4` became truncated because of the portion to the right of the `#` being interpreted as the `fragment` of the uri and not part of the `path`.  So `uri.path` was `Report from the Statehouse WTTV (`.


masaball (approved):
Looks good

PR #5044, opened 2023-01-31
Fix track selection issue when player is paused

Dananji:


app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
```diff
@@ -538,33 +543,6 @@ class MEJSPlayer {
     return Object.keys(this.segmentsMap).length > 0;
   }
 
-  /**
-   * Update section links to reflect active section playing
-   * @function highlightSectionLink
-   * @param  {string} segmentId - HTML node of section link clicked on <a>
-   * @return {void}
-   */
-  highlightSectionLink(segmentId) {
```
cjcolvar, line 547: Is this not needed anymore?

cjcolvar (approved):
Looks good.  Hopefully this works! :crossed_fingers:

app/assets/javascripts/media_player_wrapper/avalon_player_new.es6
Dananji: This was a duplication of the same function in `mejs_helper_utlility` module (https://github.com/avalonmediasystem/avalon/blob/main/app/assets/javascripts/media_player_wrapper/mejs4_helper_utility.es6#L146), which is being used in the code.

PR #5046, opened 2023-02-01
Force encoding of IndexedFile original_name to UTF-8

### LINKED ISSUE #5027: Collection posters cannot have filenames with UTF-8 characters
### Steps to reproduce:
1. Create a collection
2. Click Upload Poster
3. Choose a image file with a filename with UTF-8 characters (e.g. `きた!.jpg`)

### Expected Behavior
Image file should be uploaded and stored as the collection poster image

### Actual Behavior
Avalon raises an error: `incompatible character encodings: ASCII-8BIT and UTF-8`

image

cjcolvar:
For some reason when a filename with UTF-8 characters is uploaded as a collection poster image it roundtrips from Fedora as ASCII-8BIT probably due to the fact that it is returned in an HTTP header.  Forcing the original_name to UTF-8 in an override feels safe after reviewing Avalon's use of IndexedFiles.

Resolves #5027


masaball (approved):
Looks good!

PR #5053, opened 2023-02-07
Refactor MediaObject methods for improving save performance

cjcolvar:


masaball (commented):
Looks like there is still a failing test in the MediaObjectsController spec, but once that's taken care of I think we're good to go. This is looking great.

masaball (approved):
Looks good!

PR #5072, opened 2023-02-21
Increase ruby stack size limit

masaball:
Based on recommendation for resolving "Stack level too deep" errors found [here](https://github.com/samvera/hyrax/issues/4581#issuecomment-957872071) and [here](https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#stack-size)


cjcolvar (approved):
:+1:

PR #5081, opened 2023-02-24
Fix two bugs in JSON update API

mbklein:
- Make `collection_id` optional in updates
- Allow `:json_update, media_object` for all users who can `:edit, media_object`

I couldn't figure out how to create test cases for these because I've been away from Rails/Rspec for so long and there don't seem to be any existing test cases for JSON updates via the API to base them off of.


cjcolvar:
Thanks for submitting this!  I can work on adding some tests.

PR #5082, opened 2023-02-27
HLS captions for iOS

masaball:
Proof of concept for providing captions in HLS manifest. Test on Browserstack and evaluate approach.

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -325,6 +325,11 @@ class MasterFilesController < ApplicationController
     end
   end
 
+  def caption_manifest
+    return head :unauthorized if cannot?(:read, @master_file)
+    @caption = @master_file.captions
+  end
+
   def structure
     authorize! :read, @master_file, message: "You do not have sufficient privileges"
     render json: @master_file.structuralMetadata.as_json
--- a/app/views/master_files/hls_manifest.m3u8.erb
+++ b/app/views/master_files/hls_manifest.m3u8.erb
@@ -14,7 +14,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 ---  END LICENSE_HEADER BLOCK  ---
 %>
 #EXTM3U
+#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",LANGUAGE="en",NAME="English",AUTOSELECT=YES,URI="<%= caption_manifest_master_file_path %>"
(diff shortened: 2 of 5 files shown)


app/controllers/master_files_controller.rb
```diff
@@ -325,6 +325,11 @@ def hls_manifest
     end
   end
 
+  def caption_manifest
+    return head :unauthorized if cannot?(:read, @master_file)
+    @caption = @master_file.captions
```
masaball, line 330: This line is unnessary for final implementation. The `caption_manifest` erb file does not acually make use of this but rather uses the `captions_master_file_path`. Shouldn't interfere with testing though.

app/views/master_files/hls_manifest.m3u8.erb
```diff
@@ -14,7 +14,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 ---  END LICENSE_HEADER BLOCK  ---
 %>
 #EXTM3U
+#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",LANGUAGE="en",NAME="English",AUTOSELECT=YES,URI="<%= caption_manifest_master_file_path %>"
```
masaball, line 17: Final implementation could have a setting to specify a default caption language to avoid having to hardcode it as english. If implemented, would be good to eventually circle back and allow editors to define caption language, as well as provide an option to provide multiple caption files.

config/application.rb
```diff
@@ -47,6 +47,8 @@ class Application < Rails::Application
         resource '/master_files/*/structure.json', headers: :any, methods: [:get, :post, :delete]
         resource '/master_files/*/waveform.json', headers: :any, methods: [:get]
         resource '/master_files/*/*.m3u8', headers: :any, credentials: true, methods: [:get, :head]
+        resource '/master_files/*/caption_manifest', headers: :any, methods: [:get]
+        resource '/master_files/*/captions', headers: :any, methods: [:get]
```
masaball, line 51: Not sure if necessary for final implementation. These are necessary when the HLS manifest is requested by an external player such as Ramp, but depending on how iOS handles the requests it may not  be through CORS. If we are able to provide captions to IIIF players via the IIIF manifest, then the captions being included in the HLS manifest may only be necessary for iOS.

cjcolvar (approved):
Looks fine for testing.  We'll need to clean things up and add tests later.

PR #5083, opened 2023-02-28
Use old namespace for bibframe property to align with pre-existing data

### LINKED ISSUE #5074: Unit metadata not displaying for existing collections
On the item view page the Unit is not displaying correctly.  As can be seen in the screenshot it is being displayed as "/collections". This should read "Default Unit" or whatever Unit the collection belongs to.

This only occurs with existing collections. Any media object already belonging to the collection and new media objects that are created within the collection will display this metadata.

Unit information displays as expected for media obejcts added to a newly created collection. The display can also be fixed by re-saving the existing collection but that save triggers the behavior described in #5073 for all child items of the collection.

cjcolvar:
Resolves #5074 

It looks like `rdf-vocab` updated the Bibframe vocabulary including changing the URI namespace for properties.  This PR allows us to use the old namespace.

Existing data: <http://bibframe.org/vocab/heldBy>
Newer `rdf-vocab`: <http://id.loc.gov/ontologies/bibframe/heldBy>

### DIFF
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -34,7 +34,7 @@ class Admin::Collection < ActiveFedora::Base
   property :name, predicate: ::RDF::Vocab::DC.title, multiple: false do |index|
     index.as :stored_sortable
   end
-  property :unit, predicate: ::RDF::Vocab::Bibframe.heldBy, multiple: false do |index|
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

PR #5093, opened 2023-03-02
Fix bug for expanding filename and hover text

### LINKED ISSUE #5091: Long filename expand/less not working for filenames with spaces
# Description

Originally assumed this was a local Rivet styling issue, but seems to be a core issue handling filenames with spaces. Only older filenames have spaces, as spaces are now replaced by underscores (so this cannot be replicated on new uploads). 

See details in https://github.iu.edu/AvalonIU/Avalon/issues/144. 

Example:  https://mco-staging.dlib.indiana.edu/media_objects/3x816r63s/edit?step=file-upload

# Done Looks like

- [x] Long filenames are truncated by default (even if they have spaces)
- [x] expand/show less box allows viewing full filenames for filenames that are truncated filenames (even if they have spaces)
- [x] hover text shows full filename

Mason thinks this should be an easy fix to warp the variable in quotation marks when title attribute is assigned.

masaball:
Fixes #5091 

This PR wraps the title attribute value in quotation marks to fix hover text for titles with spaces not displaying properly. It also filters the `section.label` value through the `truncate_center` function so that all Section files truncate and expand as expected.

### DIFF
--- a/app/views/media_objects/_file_upload.html.erb
+++ b/app/views/media_objects/_file_upload.html.erb
@@ -60,10 +60,10 @@ Unless required by applicable law or agreed to in writing, software distributed
                     <% end %>
                   </span>
                   <% filename = section.title || File.basename(section.file_location.to_s) %>
-                  <span class="mediaobject-filename" id= <%= "truncated_#{section.id}" %> title=<%= filename %>>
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Looks great!

PR #5094, opened 2023-03-02
Prevent reification when requesting manifest

masaball:
This PR attempts to keep IIIF manifest generation from reifying any files.

The two areas causing reification are: 

- Requesting the structural metadata. The structural metadata is contained in an XML file as a subresource of the master file and I don't think it is possible to prevent the reification here. However, what I found is that reification was happening even for master files that do not have structural metadata. This PR adds a check of `has_structuralMetadata?` (which is in the SpeedyAF proxy) so that reification only occurs when a master file actually has structural metadata.
- Retrieving information from the Derivative. When the IIIF manifest is generated it pulls the `derivative.bitrate`, `derivative.mime_type`, and `derivative.format`. None of these fields were available from the SpeedyAF proxy so it would trigger reification. This PR adds all three fields to the Derivative `to_solr` function so that they are available from the proxy. This is problematic for the `mime_type` though as assigning it seems to be commented out, as can be seen [here](https://github.com/avalonmediasystem/avalon/blob/941b660b2da9d9c7f1c4fbc98fd8fdc9e5fbd213/app/models/master_file.rb#L299) and [here](https://github.com/avalonmediasystem/avalon/blob/941b660b2da9d9c7f1c4fbc98fd8fdc9e5fbd213/app/models/derivative.rb#L105). I do not know if that work has been done but for the items I tested with the `mime_type` was `nil`. When a solr field is `nil` SpeedyAF skips that field, so I check if `derivative.mime_type = nil` and insert the string "nil". This is sloppy and I am not sure of the full implications of handling it this way. Do we need to include `mime_type` at all if assigning it to the Derivative is commented out? Should it be getting assigned elsewhere?

### DIFF
--- a/app/models/derivative.rb
+++ b/app/models/derivative.rb
@@ -87,9 +87,6 @@ class Derivative < ActiveFedora::Base
 
   def to_solr
     super.tap do |solr_doc|
-      solr_doc['format_ss'] = self.format
-      solr_doc['mime_type_ss'] = self.mime_type.nil? ? "Nil" : self.mime_type
-      solr_doc['bitrate_is'] = self.bitrate
       solr_doc['stream_path_ssi'] = if location_url&.start_with?("rtmp")
                                       location_url.split(/:/).last
                                     else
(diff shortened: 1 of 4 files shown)


cjcolvar:
Let's pair up sometime to work through your questions.  I think it might be possible that more things are indexed and can be taken advantage of.  If we do need to index more information it would be good to get that in before we have to reindex everything for upgrading solr.

cjcolvar (approved):
Looks good!

PR #5101, opened 2023-03-07
Strip out all unquoted colons

cjcolvar:
This fixes a problem we saw in production where searching for `"[Clio : 1983 Clio international preliminaries : Craft : Classic : Misc]"` works but `[Clio : 1983 Clio international preliminaries : Craft : Classic : Misc]` does not return any results.  Interestingly this does not seem to be a problem with Solr 8 (we're currently running solr 6).  This PR works around the problem by removing all colons from the search query that aren't inside of a quoted phrase.


cjcolvar:
Well, the tests revealed a case where the `q` was a lucene query and not just a simple user-supplied string.  I can't think of a good way to differentiate these cases so we might have to give up on this approach.

PR #5102, opened 2023-03-07
Override SearchService#count to ensure it uses the preferred http method…

cjcolvar:
…instead of assuming GET, which doesn't work for lengthy requests.

Related to issues with media_objects.json endpoint.


masaball (approved):
Looks good!

PR #5111, opened 2023-03-13
Enqueue media object indexing job when media object and children are reindexed

cjcolvar:
This should solve the case where a structure is updated or a section title is updated without saving the parent media object.  Also moving the enqueueing of the background job to the after_update_index callback for media objects (from `to_solr`) should ensure it isn't enqueued when the media object isn't actually getting reindexed.


masaball (approved):
Looks good!

PR #5113, opened 2023-03-14
Strip out childless paths from structural metadata to ensure valid manifest

cjcolvar:
These changes ensure a valid IIIF manifest is returned when the stored XML structural metadata contains blank divs.  This allows the SME to load instead of erroring.

This PR leaves the Error in place but it really shouldn't be raised.  Do you think it should be removed?

Related to https://github.com/avalonmediasystem/react-structural-metadata-editor/issues/176


Dananji (approved):
:100:

PR #5130, opened 2023-03-24
Handle invalid URI errors

masaball:
Wait to merge until after 7.6

This PR adds handling for Invalid URI errors to the application controller. Malformed NOIDs seem to be the only (or at least main) trigger that we have seen in our honeybadger reports. The NOID is expected to be alphanumeric so if an invalid URI error is detected, this strips any special characters from the ID param and, if the NOID is now valid, resubmits the request with the edited ID. Otherwise, an error page is rendered.

In manual testing I have done, this seems to be robust enough but it is possible this is too general an approach and there might be edge cases that are not handled properly. Possibly inserting the sanitizing logic where the NOID gets converted to the fedora id would work and be less generalized? I'm not sure where in the code that is though.

### DIFF
--- a/app/controllers/concerns/noid_validator.rb
+++ b/app/controllers/concerns/noid_validator.rb
@@ -1,0 +1,39 @@
+# Copyright 2011-2023, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+module NoidValidator
+  extend ActiveSupport::Concern
+
+  include Identifier
+
+  included do
+    before_action :validate_noid, only: [:show]
+  end
+
+  private
+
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/noid_validator.rb
masaball, line 1: The Identifier module is a Model concern rather than controller concern, so trying to add `before_action` to it caused issues when included in the models. Creating a new module specifically for the validation seemed the easiest path forward.

app/controllers/concerns/noid_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def validate_noid
+      return if noid_service.valid?(params[:id])
+
+      # Strip all non-alphanumeric characters from passed in NOID
+      noid_id = params[:id]&.gsub(/[^A-Za-z0-9]/, '')
+
+      # If cleaned NOID is not valid raise error, otherwise redirect request.
+      raise ActiveFedora::ObjectNotFoundError unless noid_service.valid?(noid_id)
+      redirect_to(request.parameters.merge(id: noid_id))
```
cjcolvar, line 34: I was thinking this might be unsafe based upon the codeclimate warning and analysis:
>This is because params could contain :host => 'evilsite.com' which would redirect away from your site and to a malicious site.
>
>If the first argument to redirect_to is a hash, then adding :only_path => true will limit the redirect to the current host.

But since we're on Rails 7 there is a new default to watch for unsafe redirects and raise an error (https://edgeguides.rubyonrails.org/configuring.html#config-action-controller-raise-on-open-redirects) so I think we don't have to worry about the user manipulating the params. :slightly_smiling_face:

cjcolvar (commented):
This looks good.  If in the future we find there are more cases that could benefit from this approach we can expand things in the concern.

cjcolvar (approved):

PR #5135, opened 2023-03-27
Fix player toolbar icons in Android when auto advancing to next section

Dananji:


cjcolvar (approved):
Looks good.  Great work tracking down all of these fussy ME.js issues!

PR #5144, opened 2023-03-31
Override ActiveFedora and HydraAccessControls to ensure AccessControl…

cjcolvar:
… objects are marked dirty and autosave so they get indexed properly

I added some tests but was not able to get any of them to fail prior to the changes in this PR.  It is possible that the issue is only present in a production environment.

### DIFF
--- a/config/initializers/active_fedora_general.rb
+++ b/config/initializers/active_fedora_general.rb
@@ -34,3 +34,60 @@ Ldp::Response.class_eval do
     ::RDF::URI.decode(filename) if filename
   end
 end
+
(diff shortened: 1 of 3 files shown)


masaball (approved):
This looks good as far as I can tell. I've been trying to brainstorm some additional test cases/approaches, but only thing that comes to mind is that Emily's video was testing adding an IP address so maybe specifically testing the IP address update path would catch the bug. 

But even if this is not a production only bug, Emily did say in her video that which items index the change was variable so I'm really not sure about how to get consistently failing tests pre-fix...

PR #5146, opened 2023-03-31
Bump rails version for latest security patch

cjcolvar:
@joncameron Is this okay to get into 7.6?  It would be nice to have all of the latest security patches.


joncameron:
💯

PR #5151, opened 2023-04-04
Add metadata check before publishing/unpublishing

masaball:
Also updates publish/unpublish error message to reflect whether user was attempting to publish or unpublish the media object and updates the error message generated when deleting a collection with invalid media objects to provide the IDs of the culprit media objects.

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -209,9 +209,7 @@ class Admin::CollectionsController < ApplicationController
         target_path = admin_collection_path(@target_collection)
         @source_collection.reload
       else
-        mos = @source_collection.media_objects.to_a
-        invalid_mos = mos.collect { |mo| "<li>#{mo.id}</li>" unless mo.valid? }.compact
-        flash[:error] = "Collection contains invalid media objects that cannot be moved. Please address these issues before attempting to delete #{@source_collection.name}.<ul>#{invalid_mos.join('')}</ul>".html_safe
+        flash[:error] = "Collection contains invalid media objects that cannot be moved. Please address these issues before attempting to delete #{@source_collection.name}."
         redirect_to admin_collection_path(@source_collection) and return
       end
     end
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2023, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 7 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -417,8 +417,12 @@ def update_status
     Array(params[:id]).each do |id|
       media_object = MediaObject.find(id)
       if cannot? :update, media_object
-        errors += ["#{media_object.title} (#{id}) (permission denied)."]
+        errors += ["#{media_object&.title} (#{id}) (permission denied)."]
       else
+        unless media_object.title.present? && media_object.date_issued.present?
```
cjcolvar, line 422: What about moving this check into the handling for 'publish' so 'unpublish' can happen even if the item is invalid?  Although in order to allow that there probably need to be changes made to `publish!` for skipping validations when unpublishing.

app/controllers/media_objects_controller.rb
```diff
@@ -417,8 +417,12 @@ def update_status
...
+          errors += ["Validation failed: Title field is required., Date issued field is required."]
```
cjcolvar, line 423: How about making this more generic since there are a number of combinations possible?  It would probably be good to reference the item id since this controller action might be handling multiple items.
```suggestion
          errors += ["#{media_object&.title} (#{id}) (missing required fields)."]
```

app/controllers/media_objects_controller.rb
```diff
@@ -417,8 +417,12 @@ def update_status
...
+          break
```
cjcolvar, line 424: Should this be `next` instead of `break`?

cjcolvar (commented):
This is tricky!  I don't really like how some of the validations are manually copied to `update_status` since that could catch us later if we ever change the validations on the model. But this is definitely much simpler than trying to tweak the logic on the validations!  So I think this approach is fine for now.

app/controllers/media_objects_controller.rb
masaball: Well, with this manual check the active record validations are being rendered redundant for `publish!` anyways so it may be the move to disable validations on `publish!` in general? I'm not really happy about that, but it is the logical progression with this approach...

app/controllers/admin/collections_controller.rb
```diff
@@ -209,7 +209,9 @@ def destroy
         target_path = admin_collection_path(@target_collection)
         @source_collection.reload
       else
-        flash[:error] = "Collection contains invalid media objects that cannot be moved. Please address these issues before attempting to delete #{@source_collection.name}."
+        mos = @source_collection.media_objects.to_a
+        invalid_mos = mos.collect { |mo| "<li>#{mo.id}</li>" unless mo.valid? }.compact
+        flash[:error] = "Collection contains invalid media objects that cannot be moved. Please address these issues before attempting to delete #{@source_collection.name}.<ul>#{invalid_mos.join('')}</ul>".html_safe
```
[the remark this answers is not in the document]
masaball: I was testing this and am not sure how to skip the validations for this. When you delete the collection it calls `#reassign_media_objects`, which runs into a validation error when attempting to set `media_object.collection = target_collection`. From the stack trace, this seems to be related to the media object model having `has_and_belongs_to_many :governing_policies`. I've tried inserting `validate: false` at various points in the chain but I think it may be something deeper in ActiveFedora itself that is running the validation.

At this point I lean towards nixing this part of the PR since it is out of the scope of the ticket, unless you want to take a look at things.

app/models/media_object.rb
```diff
@@ -156,9 +156,13 @@ def collection= co
   # Sets the publication status. To unpublish an object set it to nil or
   # omit the status which will default to unpublished. This makes the act
   # of publishing _explicit_ instead of an accidental side effect.
-  def publish!(user_key)
+  def publish!(user_key, validate: false)
```
cjcolvar, line 159: Sorry, I think I messed up the default value.  In order to maintain backwards compatibility we should validate by default.
```suggestion
  def publish!(user_key, validate: true)
```

cjcolvar (approved):
Looks great!

PR #5152, opened 2023-04-04
Allow configuration of recaptcha v3 or v2_checkbox

### LINKED ISSUE #5126: Add reCAPTCHA v3 Support
# Description
Adding support for sites to use the v3 type of reCAPTCHA.

# Done Looks Like
- [ ] settings.yml updated for adding a "type" pair to the "recaptcha" section

# Settings.yml
recaptcha:
 type: "v2_checkbox" or "v3"
 site_key: "site_key"
 secret_key: "secret_key"

cjcolvar:
Closes #5126 

Recaptcha is disabled by default in tests so I didn't add any tests.  Do you think I should add any for `#recaptcha_valid?`?

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -29,7 +29,7 @@ class CommentsController < ApplicationController
     @comment.subject = params[:comment][:subject]
     @comment.comment = params[:comment][:comment]
 
-    if @comment.valid? && (Settings.recaptcha.blank? || verify_recaptcha(model: @comment))
+    if @comment.valid? && recaptcha_valid?
       begin
         CommentsMailer.contact_email(@comment.to_h).deliver_later
       rescue Errno::ECONNRESET => e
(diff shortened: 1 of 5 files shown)


app/controllers/comments_controller.rb
```diff
@@ -47,4 +47,15 @@ def create
...
     @subjects = Comment::SUBJECTS
   end
+
+  def recaptcha_valid?
+    return true unless Settings.recaptcha.site_key.present?
+    options = case Settings.recaptcha.type
+              when "v2_checkbox"
+                { model: @comment }
+              when "v3"
+                { action: Settings.recaptcha.v3.action, minimum_score: Settings.recaptcha.v3.minimum_score }
```
masaball, line 57: Do we need to pass in the model here? Not sure how v3 differs from v2.

masaball (commented):
This is looking good. It would probably be good to add in a quick test for `recaptcha_valid?` but I don't think it's strictly necessary if you're comfortable with it.

app/controllers/comments_controller.rb
cjcolvar: I think v3 only needs action and minimum_score. https://github.com/ambethia/recaptcha#verify_recaptcha-use-with-v3

cjcolvar:
I'll add a test for `#recaptcha_valid?`.

masaball (approved):
Looks good!

Issue #5108, opened 2023-03-10
Turn captions on by default in iOS


elynema:
# Description

As of #5084, captions could be enabled on iOS devices in full-screen player by selection English as the subtitle option. We would like the captions to be turned on by default if possible, and also to work if Auto is selected under the subtitle mens rather than Englishe.

# Definition of Done
- [ ] On iOS device, once playback begins in full-screen mode, captions display automatically if available
- [ ] On iOS device, once playback begins in full-screen mode, captions can be turned off using the subtitle menu
- [ ] On iOS device, once playback begins in full-screen mode, captions can be re-enabled by selecting "English" or "Auto" options.
- [ ] Captions work as normal on browsers (generally toggle on/off with CC icon; on by default)

masaball:
I have looked into this for a bit and done some testing. For reference, here is the [RFC spec for HLS streaming](https://www.rfc-editor.org/rfc/rfc8216.html#page-25), specifically the section dealing with relating other Renditions of a media stream, which is what subtitles/captions are classified as (4.3.4.1 EXT-X-MEDIA). There is good info on the whole setup there but what is relevant is on page 26. The attributes we care about for captions being on by default are:
- Default: "If the value is YES, then the client SHOULD play this Rendition of the content in the absence of information from the user indicating a different choice."
- Autoselect: "If the value is YES, then the client MAY choose to play this Rendition in the absence of explicit user preference because it matches the current playback environment, such as chosen system language.
- Forced: "A value of YES indicates that the Rendition contains content that is considered essential to play.  When selecting a FORCED Rendition, a client SHOULD choose the one that best matches the current playback environment (e.g., language).

I have tested all 3 in different combinations but have had no luck getting them to be on by default when using BrowserStack. 

Additionally, I encountered behavior with the standard Avalon player where having DEFAULT enabled caused the captions to render as part of the video but they could not be turned on or off. The Avalon player's CC button would turn on/off our sideloaded captions but had no effect on the HLS captions, so if the CC button was on there would be 2 sets of captions displayed. So, if DEFAULT were to end up being required for iOS default on captions it might require us to rework how the native Avalon player handles the captioning.

[...]

PR #5153, opened 2023-04-04
Fix overlapping captions in player in iPhone

Dananji:
This fix is only applied for iPhones not iPads. 

The overlapping captions are still visible in iPads, if the user enables captions in the native player in fullscreen and then exit full-screen. We could maybe look at this after #5108 has been resolved.


cjcolvar (changes requested):
Looks good but waiting for Avalon 7.6 release.

cjcolvar (approved):

PR #5154, opened 2023-04-05
Cascade user delete to db tables with user.id foreign key

### LINKED ISSUE #5103: Cannot delete users on avalon-dev or avalon-staging
# Description

Charu and I were working on deleting and re-adding users for testing purposes and found the delete user functionality to be broken in both instances. 

On avalon-dev, I got a Persona error (I must have a timeline). I don't know if this might relate to the Ruby 3.2 upgrade? Or is this a problem that has always been there?

On avalon-staging, I'm getting this:

The user I attempted to delete (elynema@iu.edu) has multiple Timelines in both instances.

The user was not actually deleted in either instance. If this is a longstanding issue, it can wait until after 7.6. If it's tied into the upgrade somehow, we should probably try to address it now.

# Definition of done

- [ ] delete user link on Manager Users page removes user
- [ ] Timelines associated with user should also be deleted (Is this the correct behavior?)

masaball:
Fixes #5103 

Should text be added to the deletion confirmation about how this will remove timelines and playlists belonging to the user?

### DIFF
--- a/app/controllers/samvera/persona/users_controller.rb
+++ b/app/controllers/samvera/persona/users_controller.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2023, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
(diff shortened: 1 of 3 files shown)


cjcolvar:
That seems like a good idea since playlists or timelines could be shared via tokenized url or set publicly visible.

cjcolvar (approved):
Looks great!

cjcolvar (approved):

PR #5164, opened 2023-04-14
Tweak default bug template

elynema:
Proposal for a bug template to use when creating a new bug


elynema:
@joncameron Can you review this and approve if it's ok with you?

joncameron (approved):

PR #5166, opened 2023-04-19
Fix add to playlist

Dananji:


cjcolvar (commented):
Did you end up having to change anything in the coffeescript to fix the problem?

Since the gem is being removed I think there are also references in `application.js` and `application.scss` that need to be cleaned up as well.  There are also unreachable references in the ME.js 2 player code blocks but those can be cleaned up later...probably when Ramp replaces the ME.js player.

Dananji:
Yes, I changed the `ajax:success` function to identify the received data in the correct format. Before the function had the params `(event, data, status, xhr)` and now it only has `event` which carries `[data, status, xhr]`.
I didn't see any references to the gem in other files, except in the view template which included the script using `javascript_include_tag`.

cjcolvar:
You're right.  I thought I had seen the includes in application.js and application.scss but I must have been looking in the wrong place since they definitely aren't there.

cjcolvar (approved):
Looks good!

The failing test is a flakey test and is unrelated to this PR.

PR #5181, opened 2023-04-28
Include metadata into IIIF manifest the same way as it is rendered on the item view page

### LINKED ISSUE #5170: Add Descriptive Metadata to IIIF Manifests
# Description
Avalon's IIIF manifests currently include metadata but are missing many of the supported metadata fields. These should be included in the same way that other fields are, as specified in the spec (link below). 

# IIIF spec reference
https://iiif.io/api/presentation/3.0/#metadata

# Metadata in IIIF manifest example
https://api.dc.library.northwestern.edu/api/v2/works/5f7cd4ca-0325-436c-a84f-23db179ed383?as=iiif

# Item on avalon-dev with full metadata for testing/validation
https://avalon-dev.dlib.indiana.edu/media_objects/6t053f984/manifest.json

# Done Criteria
The following fields are present in a media object's Presentation 3.0 manifest, if present in the descriptive metadata for the object:
- [x] Note (metadata key appears in manifest but value is null)
- [x] Physical Description
- [x] Terms of Use
- [x] Table of Contents
- [x] Collection
- [x] Unit
- [x] Language
- [x] Related Items
- [x] Other Identifiers

cjcolvar:
Resolves #5170 

I did not add permalink.  A link to the item view page is included in `homepage` but it isn't a permalink yet.  Maybe this should be a different ticket to make `homepage` a permalink?

### DIFF
--- a/app/models/iiif_manifest_presenter.rb
+++ b/app/models/iiif_manifest_presenter.rb
@@ -1,11 +1,11 @@
 # Copyright 2011-2023, The Trustees of Indiana University and Northwestern
 #   University.  Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
-# 
+#
 # You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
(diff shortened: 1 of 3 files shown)


masaball (approved):
Looks good!

PR #5183, opened 2023-05-01
Update rubocop/codeclimate to run with target ruby 3.2 and rails 7.0

cjcolvar:
I noticed codeclimate is using an old version of bixby and rubocop for style enforcement.  This PR updates to the latest supported versions of each.  

Bixby 5.0.2 uses rubocop 1.28.2 but codeclimate doesn't have a `channel` for that version so I went with the latest since we couldn't back down to the next lower channel without losing ruby 3.2 support.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,11 +1,11 @@
 ---
 prepare:
-  fetch: # Pinned to bixby 2.0.0
-  - url: "https://raw.githubusercontent.com/samvera-labs/bixby/394ba20eac3f3c8146a679b1dc45c3513074848c/bixby_default.yml"
+  fetch:
+  - url: "https://raw.githubusercontent.com/samvera/bixby/v5.0.2/bixby_default.yml"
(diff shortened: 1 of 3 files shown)


.codeclimate.yml
```diff
@@ -1,11 +1,11 @@
 ---
 prepare:
-  fetch: # Pinned to bixby 2.0.0
-  - url: "https://raw.githubusercontent.com/samvera-labs/bixby/394ba20eac3f3c8146a679b1dc45c3513074848c/bixby_default.yml"
```
MSALARy688889u, line 4: Deleted

MSALARy688889u (commented):
Deleted Bixby voice my phone and my account

MSALARy688889u:
> Deleted Bixby voice my phone and my account

joncameron:
We don't actually enforce any styles at the moment. @cjcolvar can look further into this; it would be good to have a discussion about code linters and style

PR #5193, opened 2023-05-10
IIIF metadata improvements

### LINKED ISSUE #5182: Metadata Generation Updates
# Description
- `<pre>` tags don't need to be retained in descriptive metadata in the IIIF manifest.
- multiple values for a metadata key should be separate strings within an array rather than the single concatenated string (https://iiif.io/api/presentation/3.0/#metadata)

# Done Looks Like
- [x] Remove pre tags from the allow-list for serializing metadata
- [x] Remove pre tags from IIIF manifest
- [x] Multi-value fields are represented as an array of strings in the metadata section of a manifest
- [x] Quick review of any other tags we're inserting into IIIF manifest metadata (users could be putting these in).

https://github.com/avalonmediasystem/avalon/blob/524f197a66785fc8fc0e63a59fe8bd7253852205/app/helpers/application_helper.rb#L96-L105

cjcolvar:
Resolves #5182 

- Removes `<pre>` tag allowance and use
- Serializes multiple values as an array

### DIFF
--- a/app/models/iiif_manifest_presenter.rb
+++ b/app/models/iiif_manifest_presenter.rb
@@ -105,7 +105,7 @@ class IiifManifestPresenter
   def note_fields(media_object)
     fields = []
     note_types = ModsDocument::NOTE_TYPES.clone
-    note_types['table of contents'] = 'Contents'
+    note_types['table of contents'] = 'Table of Contents'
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

PR #5195, opened 2023-05-10
Avoid enqueuing many copies of the media object indexing job unnecessarily

cjcolvar:
When an object with many sections is saved it can cause hundreds of these jobs to be enqueued which will take a long time to clear out and clogs up the background worker.  This uniqueness check would allow for enqueuing the same job again as soon as the job starts running since the in memory copy of the object might be out of date by the time it finishes running.

See https://github.com/veeqo/activejob-uniqueness#job-uniqueness-for-activejob

Steps for testing:
1. Create a 100-section item
2. Delete a section
3. As an admin go to Manage Worker Jobs
4. Check that only one or two media object indexing jobs are enqueued for that item

### DIFF
--- a/app/jobs/media_object_indexing_job.rb
+++ b/app/jobs/media_object_indexing_job.rb
@@ -15,6 +15,11 @@
 class MediaObjectIndexingJob < ApplicationJob
   queue_as :media_object_indexing
 
+  # Only enqueue one indexing job at a time but once one starts running allow others to be enqueued
+  # because the first step of the job is fetching the object from fedora at which point it might
+  # be stale and subsequent jobs might get a newer version of the media object.
+  unique :until_executing, on_conflict: :log
+
   def perform(id)
     mo = MediaObject.find(id)
     ActiveFedora::SolrService.add(mo.to_solr(include_child_fields: true), softCommit: true)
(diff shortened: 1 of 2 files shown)


app/jobs/media_object_indexing_job.rb
```diff
@@ -15,6 +15,11 @@
 class MediaObjectIndexingJob < ApplicationJob
   queue_as :media_object_indexing
 
+  # Only enqueue one indexing job at a time but once one starts running allow others to be enqueued
+  # because the first step of the job is fetching the object from fedora at which point it might
+  # be stale and subsequent jobs might get a newer version of the media object.
+  unique :until_executing, on_conflict: :log
```
masaball, line 21: Are these built in Ruby/Rails functions? All I'm finding online is references to the "sidekiq-unique-jobs" gem when I try to look up `until_executing`. Maybe I missed it, but I didn't see that gem in the Gemfile.

masaball (commented):
Ignore my deleted comment. I'd missed you pointing to "activejob-uniqueness" when I came back to look at this. Looks like there is a failing automated test, but otherwise this looks fine.

masaball (approved):

joncameron:
1. Save something really big
2. Look at the jobs and see how many have been enqueued

Testing can be done on mco-staging.

PR #5196, opened 2023-05-12
Rescue Ldp::Gone in ApplicationJob

masaball:


app/jobs/application_job.rb
```diff
@@ -1,16 +1,19 @@
...
+#
 # Unless required by applicable law or agreed to in writing, software distributed
 #   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 #   CONDITIONS OF ANY KIND, either express or implied. See the License for the
 #   specific language governing permissions and limitations under the License.
 # ---  END LICENSE_HEADER BLOCK  ---
 
 class ApplicationJob < ActiveJob::Base
+  rescue_from Ldp::Gone do |exception|
+    Rails.logger.error exception.message
```
cjcolvar, line 17: What all gets included in `exception.message`?  I wonder if it provides enough context or if logging `exception.backtrace` or something else might be necessary to determine what actually happened.
I'm thinking at a minimum it would be good to be able to see what the job was and its arguments along with the message.

cjcolvar (approved):
Looks good.

PR #5197, opened 2023-05-12
Administrative Facet for visibility level

cjcolvar:
Based upon suggestion from IULMIA.

Needs tests

What is the best wording to use for the facet values?  I went with "Public", "Logged in", and "Private" as less wordy versions of what is on the access control page.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -90,6 +90,11 @@ class CatalogController < ApplicationController
     # Hide these facets if not a Collection Manager
     config.add_facet_field 'workflow_published_sim', label: 'Published', limit: 5, if: Proc.new {|context, config, opts| Ability.new(context.current_user, context.user_session).can? :create, MediaObject}, group: "workflow"
     config.add_facet_field 'avalon_uploader_ssi', label: 'Created by', limit: 5, if: Proc.new {|context, config, opts| Ability.new(context.current_user, context.user_session).can? :create, MediaObject}, group: "workflow"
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -90,6 +90,11 @@ class CatalogController < ApplicationController
     # Hide these facets if not a Collection Manager
     config.add_facet_field 'workflow_published_sim', label: 'Published', limit: 5, if: Proc.new {|context, config, opts| Ability.new(context.current_user, context.user_session).can? :create, MediaObject}, group: "workflow"
     config.add_facet_field 'avalon_uploader_ssi', label: 'Created by', limit: 5, if: Proc.new {|context, config, opts| Ability.new(context.current_user, context.user_session).can? :create, MediaObject}, group: "workflow"
+    config.add_facet_field 'read_access_group_ssim', label: 'Item access', if: Proc.new {|context, config, opts| Ability.new(context.current_user, context.user_session).can? :create, MediaObject}, group: "workflow", query: {
+      public: { label: "Public", fq: "has_model_ssim:MediaObject AND read_access_group_ssim:#{Hydra::AccessControls::AccessRight::PERMISSION_TEXT_VALUE_PUBLIC}" },
+      restricted: { label: "Logged in", fq: "has_model_ssim:MediaObject AND read_access_group_ssim:#{Hydra::AccessControls::AccessRight::PERMISSION_TEXT_VALUE_AUTHENTICATED}" },
```
joncameron, line 95: I propose "Authenticated" instead of "Logged in" for this one. It's a little more formal but I feel like it better covers the guest account case at IU where some users can be logged in but not authenticated.

Dananji (approved):
:100:

PR #5198, opened 2023-05-15
Add placeholderCanvas to IIIF manifest

Dananji:
Manifest with `placeholderCanvas` for each canvas: 

Screenshot from 2023-05-15 11-25-44

Screenshot from 2023-05-15 11-11-36

### DIFF
--- a/app/models/iiif_canvas_presenter.rb
+++ b/app/models/iiif_canvas_presenter.rb
@@ -46,6 +46,15 @@ class IiifCanvasPresenter
     }]
   end
 
+  def placeholder_content
+    # height and width from /models/master_file/extract_still method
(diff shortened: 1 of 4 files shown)


app/models/iiif_canvas_presenter.rb
```diff
@@ -46,6 +46,15 @@ def sequence_rendering
     }]
   end
 
+  def placeholder_content
+    # height and width from /models/master_file/extract_still method
+    IIIFManifest::V3::DisplayContent.new( @master_file.poster_master_file_url(@master_file.id),
+                                          width: 1280,
+                                          height: 720,
```
Dananji, line 53: Got the poster height and width from, https://github.com/avalonmediasystem/avalon/blob/main/app/models/master_file.rb#L378.

cjcolvar (approved):
Looks good.

PR #5200, opened 2023-05-15
Ensure input is a uri and not a path when using multiple pre-transcoded derivatives

cjcolvar:
This is needed because prior versions of `active_encode` would be lenient on input urls and not require them to be uris at all.  The problem arises when `active_encode` attempts to sanitize the path expecting it to be a uri and returns nil because there is no matching protocol. (See https://github.com/samvera-labs/active_encode/blob/a56a5025c5bf0022b4c7a5c83aef212c0360bc5e/lib/active_encode/filename_sanitizer.rb#L22-L31)


masaball (approved):
Looks good!

PR #5213, opened 2023-05-22
Add multiple captions to MasterFile

masaball:


app/models/master_file.rb
```diff
@@ -477,6 +477,12 @@ def self.post_processing_move_filename(oldpath, options = {})
     end
   end
 
+  attr_writer :supplemental_file_captions
```
cjcolvar, line 480: Where does this writer get used?

cjcolvar (commented):
This is looking really good.  Great job supporting the existing captions stored in fedora while transitioning over to storing new captions in ActiveStorage.

app/controllers/master_files_controller.rb
```diff
@@ -149,53 +149,6 @@ def attach_structure
...
 
-  def attach_captions
-    captions = nil
-    if flash.empty?
-      authorize! :edit, @master_file, message: "You do not have sufficient privileges to add files"
-      if params[:master_file].present? && params[:master_file][:captions].present?
-        captions_file = params[:master_file][:captions]
-        captions_ext = File.extname(captions_file.original_filename)
-        content_type = Mime::Type.lookup_by_extension(captions_ext.slice(1..-1)).to_s if captions_ext
-        if ["text/vtt", "text/srt"].include? content_type
```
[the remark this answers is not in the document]
masaball: Yeah, we should. Captions/Subtitles MUST be vtt or srt formats according to [our documentation](https://samvera.atlassian.net/wiki/spaces/AVALON/pages/2182840321/Adding+Captions+and+Transcripts).

spec/models/supplemental_file_spec.rb
```diff
@@ -15,11 +15,51 @@
...
+        it 'should validate' do
+          expect(subject.valid?).to be_truthy
+        end
+      end
+      context 'non-VTT/non-SRT caption file' do
+        let(:subject) { FactoryBot.create(:supplemental_file, :with_attached_file) }
+        it 'should not validate' do
+          subject.tags = ['caption']
+          subject.save
+          expect(subject.valid?).to be_falsey
```
cjcolvar, line 39: I think this could be simplified if the subject is built but not created:
```suggestion
        let(:subject) { FactoryBot.build(:supplemental_file, :with_attached_file, :with_caption_tag) }
        it 'should not validate' do
          expect(subject.valid?).to be_falsey
```

cjcolvar (approved):
This looks really good!  It was a bigger chunk of work than it initially looked like.
I had a small suggestion to simplify the validation tests but it isn't critical so I'm approving this PR.  Feel free to make the change or not.

cjcolvar (approved):

PR #5223, opened 2023-06-01
Set MasterFile duration from ActiveEncode technical metadata

masaball:


app/models/master_file.rb
```diff
@@ -723,12 +727,6 @@ def reloadTechnicalMetadata!
                          'Unknown'
                        end
 
-    self.duration = begin
-      @mediainfo.duration.to_s
-    rescue
-      nil
-    end
-
```
masaball, line 731: Removing this appears to break batch entry. Running through the code, batch entry tries to save the master file during processing which triggers the `:after_save` hook for updating the poster image. Because the duration is not set, this causes an error that prevents the save from occurring. 

Should we leave this in place for now and create a new ticket to handle the batch entry case?

cjcolvar: I'm not sure.  I'm a bit confused as to why the error is happening and why it is blocking save.  I would have expected that the lack of duration would lead to not generated a thumbnail.  Would you be available to walk me through the trace and we could figure out a best path forward?

app/models/master_file.rb
```diff
@@ -170,6 +170,9 @@ def error
   # Generate the waveform after proessing is complete but before master file management
   after_transcoding :generate_waveform
   after_transcoding :update_ingest_batch
+  # Generate and set the poster and thumbnail
+  after_transcoding :set_default_poster_offset
+  after_transcoding :update_stills_from_offset!
```
cjcolvar, line 175: I don't think this is really a big problem, but I think without special handling this will cause batch ingested items to have the poster and thumbnails regenerated since the `after_save` will have already run.

cjcolvar (approved):
Looks good!

P.S. FWIW I saw that the master file is already doing an `after_transcoding` callback which finds the batch ingest entry so that is a place where we could have set the poster offset post-transcoding for batch ingest items.

PR #5226, opened 2023-06-06
Add captions and transcripts to IIIF Manifest

masaball:
As part of adding caption/transcript support to the IIIF manifest, this PR adds routes and CORS settings to generate caption/transcript specific URIs to be used as the ID field in the IIIF manifest (e.x. `/master_files/:master_file_id/supplememtal_files/:id/captions`). These endpoints should allow us to differentiate between captions and transcripts without needing to add extra boilerplate to the label or other fields. The new endpoints redirect to the generic supplemental file endpoint to actually retrieve the resources.

If/when additional parameters are added to Supplementing Annotations, it would be worth reexamining the implementation but this seemed like the best approach as things stand.

### DIFF
--- a/app/models/iiif_canvas_presenter.rb
+++ b/app/models/iiif_canvas_presenter.rb
@@ -39,6 +39,10 @@ class IiifCanvasPresenter
     master_file.is_video? ? video_content : audio_content
   end
 
+  def supplementing_content
+    supplemental_captions_transcripts.collect { |file| supplementing_content_data(file) }
+  end
+
   def sequence_rendering
     supplemental_files_rendering(master_file)
   end
@@ -84,10 +88,19 @@ class IiifCanvasPresenter
                                            **manifest_attributes(quality, 'Sound'))
     end
 
+    def supplementing_content_data(file)
+      IIIFManifest::V3::SupplementingContent.new(Rails.application.routes.url_helpers.master_file_supplemental_file_url(master_file.id, file.id),
+                                                 **supplemental_attributes(file))
+    end
+
     def stream_urls
       stream_info[:stream_hls].collect do |d|
         [d[:quality], d[:url]]
       end
+    end
+
+    def supplemental_captions_transcripts
+      master_file.supplemental_files(tag: 'caption') + master_file.supplemental_files(tag: 'transcript')
     end
 
     def simple_iiif_range(label = stream_info[:embed_title])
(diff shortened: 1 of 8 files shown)


app/models/supplemental_file.rb
```diff
@@ -20,6 +20,8 @@ class SupplementalFile < ApplicationRecord
 
   serialize :tags, Array
 
+  attr_accessor :language
```
masaball, line 23: This accessor is here so that the program does not complain about SupplementalFiles not having a language method and so that a language can be assigned in the test suite. This will be deprecated by the multiple captions PR which adds a database migration that adds a language field to SupplementalFile.

So, if this gets merged first we will need to rebase the Multiple Captions PR and vice versa.

elynema:
Oral history transcripts are generally word documents. When we caption media with 3Play, we upload the output as a caption only in MCO. 

Jon and I would prefer the caption/transcript specific URI solution. Create the URI when the IIIF manifest is created, then Ramp will need to look at the UI pattern when rendering supplemental files.

app/models/iiif_canvas_presenter.rb
```diff
@@ -84,12 +88,29 @@ def audio_display_content(quality)
...
+    end
+
     def stream_urls
       stream_info[:stream_hls].collect do |d|
         [d[:quality], d[:url]]
       end
     end
 
+    def supplemental_captions_transcripts
+      master_file.supplemental_files(tag: 'caption') + master_file.supplemental_files(tag: 'transcript') + [master_file.captions]
```
cjcolvar, line 111: Do you need to check if `master_file.captions` exists and is persisted?  Otherwise I could image this as adding `nil` or empty captions file to the array which would cause the `captions_master_file_url` be included in the manifest even when it wouldn't return any content from that endpoint.

When I was working on the SpeedyAF PR I found that `master_file.captions` behaves a little differently than I expected:
```
[3] pry(main)> mf.has_captions?
=> false
[4] pry(main)> mf.captions
=> #<IndexedFile uri="http://fedora:8080/fedora/rest/9z/90/2z/84/9z902z84b/captions" >
[5] pry(main)> mf.captions.present?
=> false
[6] pry(main)> mf.captions.persisted?
=> false
[7] pry(main)> mf.captions.content
=> nil
[8] pry(main)> mf.captions.mime_type
=> "text/plain"
```

cjcolvar (commented):
This looks really good!  I had one concern about handling of old captions since they can act a bit oddly in code.

app/models/iiif_canvas_presenter.rb
masaball: Good catch. Had not noticed that behavior for `master_file.captions` and it was causing `captions_master_file_url` to be erroneously added.

cjcolvar (approved):
Looks good!

PR #5229, opened 2023-06-09
Improve behavior of public and token timelines

masaball:


app/controllers/timelines_controller.rb
```diff
@@ -360,6 +353,20 @@ def load_timeline_token
       current_ability.options[:timeline_token] = @timeline_token
     end
 
+    def build_url_fragment
+      url_fragment = "noHeader=true&noFooter=true&noSourceLink=false&noVideo=false"
+
+      if @timeline.visibility == 'public' || current_user == @timeline.user
```
cjcolvar, line 359: I think for the public case we want the resource url fragment on the next line but the callback url on line 364 but only if the user is logged in.  A test for this case probably was missing.

app/controllers/timelines_controller.rb
```diff
@@ -360,6 +353,20 @@ def load_timeline_token
...
+        url_fragment += "&resource=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json), /[:\/?=]/)}"
+        url_fragment += "&callback=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json), /[:\/?=]/)}"
+      elsif current_user || @timeline.valid_token?(@timeline.access_token)
```
cjcolvar, line 362: At some point I added the check for `current_user` here but I'm not sure if it is really needed.  Maybe it would be best to turn this line into `else`?  It seems like the access token will be checked in the manifest request (or has already been checked in the `authorize!` call on line 103 and doesn't really need to be checked here.

app/controllers/timelines_controller.rb
```diff
@@ -360,6 +353,20 @@ def load_timeline_token
...
+        url_fragment += "&resource=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json, token: @timeline.access_token), /[:\/?=]/)}"
+        url_fragment += "&callback=#{Addressable::URI.escape_component(timelines_url, /[:\/?=]/)}"
+      end
```
cjcolvar, line 365: I think this might be clearer to pull apart the two query params:
```suggestion
     # Manifest url - include token if present 
      if @timeline_token.present?
        url_fragment += "&resource=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json, token: @timeline_token), /[:\/?=]/)}"
      else 
        url_fragment += "&resource=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json), /[:\/?=]/)}"
      end

      # Save callback - Owner uses manifest url and all other logged-in users use timeline index url
      if current_user == @timeline.user
        url_fragment += "&callback=#{Addressable::URI.escape_component(manifest_timeline_url(@timeline, format: :json), /[:\/?=]/)}"
      elsif current_user
        url_fragment += "&callback=#{Addressable::URI.escape_component(timelines_url, /[:\/?=]/)}"
      end
```
Note I also switched `@timeline.access_token` to `@timeline_token` to be clearer that it is the param.
Also the save callback requires a logged in user.

cjcolvar (commented):
The timeline controller code tangled together many cases which I think is the cause that this bug appeared.  Adding more tests like you're doing is totally the right way to ensure all of the cases are handled correctly.  I think there are a few more cases that still need tests.  Would you like to pair on this additional work?

PR #5231, opened 2023-06-14
Move waveform from Rendering to seeAlso

masaball:


app/models/iiif_canvas_presenter.rb
```diff
@@ -40,7 +40,11 @@ def display_content
   end
 
   def sequence_rendering
-    supplemental_files_rendering(master_file) + waveform_rendering
+    supplemental_files_rendering(master_file)
+  end
+
+  def see_also
+    waveform_see_also
```
cjcolvar, line 47: How about folding `waveform_see_also` into this method?  A separate method doesn't really seem necessary anymore.

cjcolvar (approved):
Looks good.  I had a small non-blocking suggestion but feel free to merge if you want.

app/models/iiif_canvas_presenter.rb
masaball: I can go either way on it. I was thinking leave it separate so we can just add additional methods if we add other stuff to the seeAlso section later, rather than have to pull the waveform back into its own method. But that's also not really any amount of extra work. 

I guess best practice would probably be to simplify the code for now, since we don't have any immediate plans for adding more to seeAlso?

Dananji (approved):
Looks good :100:

PR #5221, opened 2023-05-31
Use SpeedyAF::Proxy for faster ability checks in MasterFilesController

cjcolvar:
The proxy may need to be reified for non end-user actions but this is much faster for items with many sections when fetching the HLS manifest.  The speedup comes from a faster ability check where `can? :read, @master_file` ends up checking `can? :read, @master_file.media_object`.  For non-SpeedyAF objects this will fetch the large media object from fedora which could be very slow.

For example, with a section from a 173 section item it is a 200x speedup for the ability check:
```
irb(main):002:1* Benchmark.measure do
irb(main):003:1*   mf = MasterFile.find('vq2806057')
irb(main):004:1*   Ability.new(nil).can? :read, mf
irb(main):005:0> end
=> #<Benchmark::Tms:0x00007fa785f9b828 @cstime=0.0, @cutime=0.0, @label="", @real=1.5468747541308403, @stime=0.018735999999999975, @total=0.24486900000000078, @utime=0.2261330000000008>
irb(main):006:1* Benchmark.measure do
irb(main):007:1*   mf = SpeedyAF::Proxy::MasterFile.find('vq2806057')
irb(main):008:1*   Ability.new(nil).can? :read, mf
irb(main):009:0> end
=> #<Benchmark::Tms:0x00007fa7861a9930 @cstime=0.0, @cutime=0.0, @label="", @real=0.07299692928791046, @stime=2.4000000000024002e-05, @total=0.004274000000000777, @utime=0.004250000000000753>
```

PR #5234, opened 2023-06-16
Speedy improvements to MasterFile

cjcolvar:
Follows on to #5221 
When testing on avalon-dev I found a few places where reification was happening.  
This PR takes care of calls to MediaObject#published? (called from ability checks) and MasterFile#title (called from stream_details).  It also avoids reification to return thumbnail and poster images by base64 encoding them.  This should help with overall speed by avoiding calls to fedora to render search results and on item view pages.

When testing make sure that thumbnails and poster images (and other `IndexedFile`s) fallback to proxying from fedora when they aren't indexed into solr.

Still probably need to do a lot of work on the MediaObject proxy in order to use it exclusively without reifying on item view page. (See #5235)

### DIFF
--- a/app/models/indexed_file.rb
+++ b/app/models/indexed_file.rb
@@ -14,9 +14,37 @@
 
 class IndexedFile < ActiveFedora::File
   include SpeedyAF::IndexedContent
+  MAX_CONTENT_SIZE = 32000
 
   # Override
   def original_name
     super&.force_encoding("UTF-8")
+  end
+
+  # Override to add binary content handling
+  def to_solr(solr_doc = {}, opts = {})
+    return solr_doc unless opts[:external_index]
+    solr_doc.tap do |doc|
+      doc[:id] = id
+      doc[:has_model_ssim] = self.class.name
+      doc[:uri_ss] = uri.to_s
+      doc[:mime_type_ss] = mime_type
+      doc[:original_name_ss] = original_name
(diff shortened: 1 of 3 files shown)


app/models/indexed_file.rb
```diff
@@ -14,9 +14,37 @@
...
+  end
+
+  protected
+
+  def index_content?
+    has_content? && size < MAX_CONTENT_SIZE
+  end
+
+  def binary_content?
+    has_content? && mime_type !~ /(^text\/)|([\/\+]xml$)/
```
masaball, line 48: Just want to be sure I get this regex:
- first capture group matches any string that starts with `text/`
- second capture group matches any string that ends with `/xml` or `+xml`

correct, since mime types can be something like `application/xhtml+xml`?

app/presenters/speedy_af/proxy/media_object.rb
```diff
@@ -29,6 +29,10 @@ def to_param
     id
   end
 
+  def published?
+    !avalon_publisher.blank?
```
masaball, line 33: Would it be better to use `avalon_publisher.present?` instead of negating a `blank?` check?

masaball (approved):
Looks good to me besides my question about `present?` vs negating `blank?`. Also, looks like codeclimate has some suggestions but up to you which ones you want fix.

cjcolvar:
Superseded by #5235

PR #5235, opened 2023-06-16
Don't load from fedora on end-user routes

cjcolvar:
Replaces #5234 

Future work (in a later PR):
- Admin::Collection routes
- Playlists
- Bulk actions
- Course model


masaball (commented):
Looks good as far as I can tell. The PR is set to merge the changes into `speedy_improvements` rather than `develop` though.

cjcolvar:
Good catch!  I switched the base to point to `develop`

PR #5238, opened 2023-06-21
Hls multiple captions

masaball:


app/views/master_files/hls_manifest.m3u8.erb
```diff
@@ -15,7 +15,12 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 #EXTM3U
 <% if @master_file.has_captions? %>
-#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",LANGUAGE="<%= Settings.caption_default.language %>",NAME="<%= Settings.caption_default.name %>",AUTOSELECT=YES,URI="<%= caption_manifest_master_file_path %>"
+<% @master_file.supplemental_file_captions.append(@master_file.captions).each_with_index do |caption, index| %>
```
cjcolvar, line 18: It seems like order may be important here given the firs is marked as default below.  Does `DEFAULT` signify anything to the player?  If it does then I'd think we'd need to figure out a way for the user to select which captions is default from within the edit item workflow.

app/controllers/master_files_controller.rb
```diff
@@ -282,6 +282,12 @@ def hls_manifest
 
   def caption_manifest
     return head :unauthorized if cannot?(:read, @master_file)
+    caption_id = params[:c_id]
+    @caption_url = if caption_id == 'master_file_caption'
+                     captions_master_file_path
```
cjcolvar, line 287: I don't see a test for this case (id = "master_file_caption").  Could you please add one?

cjcolvar (commented):
Looking good so far!

app/views/master_files/hls_manifest.m3u8.erb
masaball: `DEFAULT` is supposed to designate the caption chosen by the player without user input. I think making the change to the edit item workflow would probably be best as a follow-up ticket/PR?

cjcolvar (approved):
Looks good.  Agreed that a follow up PR could make the default be user-selectable.

PR #5245, opened 2023-06-27
Index collection posters and serve from solr if <512K

cjcolvar:
Also in this PR:
- reify IndexedFile if content exists but it isn't indexed
- Bump file size limit for indexing to 512K

### DIFF
--- a/lib/tasks/avalon.rake
+++ b/lib/tasks/avalon.rake
@@ -24,8 +24,17 @@ namespace :avalon do
     `rails generate active_annotations:install`
   end
 
-  desc "Index MasterFiles and subresources to take advantage of SpeedyAF"
+  desc "Index Admin::Collections and MasterFiles and their subresources to take advantage of SpeedyAF"
   task index_for_speed: :environment do
+    Admin::Collection.find_each do |c|
+      $stderr.print "c["
+      c.update_index;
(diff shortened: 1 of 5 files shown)


lib/tasks/avalon.rake
```diff
@@ -24,8 +24,17 @@ namespace :avalon do
     `rails generate active_annotations:install`
   end
 
-  desc "Index MasterFiles and subresources to take advantage of SpeedyAF"
+  desc "Index Admin::Collections and MasterFiles and their subresources to take advantage of SpeedyAF"
   task index_for_speed: :environment do
+    Admin::Collection.find_each do |mf|
+      $stderr.print "c["
+      mf.update_index;
+      mf.declared_attached_files.each_pair do |name, file|
```
masaball, line 32: Should we change the argument here from `mf` to something else for consistency? Maybe `c` or `ac`?

masaball (dismissed):
Looks good. Just one minor suggestion about an argument name.

lib/tasks/avalon.rake
cjcolvar: Good catch.  I was sloppy with the copy/paste.

masaball (approved):

PR #5248, opened 2023-06-27
Use SpeedyAF on playlist and bookmarks routes

cjcolvar:
Also update the ability to use either ActiveFedora or SpeedyAF proxy objects throughout.  This was needed for bookmark routes but can also be used for some more media objects routes.


masaball (approved):
Looks good!

PR #5249, opened 2023-06-28
Use SpeedyAF when rendering bulk action modals

cjcolvar:
Also use SpeedyAF to determine collections a user has manage access (this should resolve local issue 166).


masaball (approved):
Looks good!

PR #5257, opened 2023-06-30
Index collection#cdl_enabled as singular boolean

### LINKED ISSUE #5255: Items Appear as CDL Items Even When They Shouldn't
# Description
On avalon-dev some items are displaying as CDL items, even when it's not enabled at the collection level. 

# Example
- Collections without CDL enabled:
  - https://avalon-dev.dlib.indiana.edu/admin/collections/m613mx563
    - https://avalon-dev.dlib.indiana.edu/media_objects/gt54kn002 (this item seems fine)
  - https://avalon-dev.dlib.indiana.edu/admin/collections/p2676v52c
    - https://avalon-dev.dlib.indiana.edu/media_objects/fj236208t  (item appears as CDL)

cjcolvar:
Existing Behavior
`Admin::Collection` objects index `cdl_enabled` as a string using the `_ssim` suffix.  When this gets turned into a SpeedyAF proxy it becomes a string (e.g. `"false"`).  Before this fix, the proxy was returning exactly what `cdl_enabled` was and `"false"` evaluates to `true` in an `if` statement which was making all MediaObject's behave as CDL enabled.

This PR change the indexing of `cdl_enabled` to use the `_bsi` suffix.  This will require reindexing all collections which shouldn't be too hard, but will need to be included in the upgrade instructions.

Fixes #5255

### DIFF
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -69,7 +69,7 @@ class Admin::Collection < ActiveFedora::Base
     index.as :stored_sortable
   end
   property :cdl_enabled, predicate: Avalon::RDFVocab::Collection.cdl_enabled, multiple: false do |index|
-    index.as :symbol
+    index.as ActiveFedora::Indexing::Descriptor.new(:boolean, :stored, :indexed)
   end
 
   has_subresource 'poster', class_name: 'IndexedFile'
--- a/spec/presenters/speedy_af/proxy/admin/collection_spec.rb
+++ b/spec/presenters/speedy_af/proxy/admin/collection_spec.rb
@@ -1,0 +1,51 @@
+# Copyright 2011-2023, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+# 
+# You may obtain a copy of the License at
+# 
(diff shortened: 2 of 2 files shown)


app/presenters/speedy_af/proxy/admin/collection.rb
```diff
@@ -44,6 +44,8 @@ def to_param
   end
 
   def cdl_enabled?
+    # Cast to boolean since it is indexed as a string (`_ssim`)
+    cdl_enabled = cdl_enabled&.to_b
```
masaball, line 48: Instead of casting to boolean, would it be better to index it as a boolean in the first place? I didn't think about it when we originally set it up, but looking at our solr schema there are a few boolean options. I think `cdl_enabled_bsi` should work?

Also, did we have a reason for storing it as `_ssim` rather than `_ssi`? I'm not sure that cdl_enabled needs to be getting handled like a multi-value field.

cjcolvar: I wasn't sure about this.  Changing how it is indexed would require us to reindex existing collections so I took the approach of making it work with the current data.  But it would be better to index as `_bsi` and that shouldn't be too onerous to ask as part of the upgrade.  I'll change this.

masaball (approved):
Looks good!

PR #5258, opened 2023-07-03
Update each access setting individually

masaball:


cjcolvar (approved):
Looks great!

masaball:
I was doing another once over before merging and spotted an issue that would have kept the Special Access modal from working. I'd had it working since I demoed it, but I guess at some point those changes weren't committed or were accidentally overwritten by a later commit. I think things should be good now though.

app/jobs/bulk_action_jobs.rb
```diff
@@ -197,24 +197,35 @@ def perform(target_id, subject_ids)
...
+        if save_field == "visibility"
+          media_object.visibility = collection.default_visibility
+        end
+        if collection.cdl_enabled? && save_field == "lending_period"
           media_object.lending_period = collection.default_lending_period
         end
 
         # Special access
-        if overwrite
+        if overwrite && save_field == "cdl"
```
cjcolvar, line 217: Why is `save_field == "cdl"` important here?  I don't understand the connection between cdl and the special access fields.

cjcolvar (commented):
I'm a bit confused by the CDL case.  Could we find a time to pair so you can walk me through it?

cjcolvar (approved):
Looks good!

Issue #5227, opened 2023-06-06
[BUG] Apply to all items overwrites item access settings even when "add to existing entries" option is selected


elynema:
**Describe the bug**
When the user makes a change at the collection level (assign staff roles, item access, assign special access, item discovery), it only impacts new items unless it is applied to existing items. When the user chooses to apply to existing items, the collection's default item access level overwrites any custom item access level that may be set on items in the collection, even if the user chooses the option to "add to existing Special Access entries."

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Make sure that you have a collection with an item that has a different Item access setting than the default Item access setting for the collection.
2. Go to Manage Collection. 
3. Add a user under Avalon User
4. Select "apply to all existing items"
5. Select "Yes, add to existing Special Access entries" and click Apply
6. Go to Edit the item identified
7. In the Access Control section under Item access, you will see that the Item access setting has been changed to match the default setting for the collection.

**Expected behavior**
I would expect that the Item discovery and Item access default settings at the collection level would not overwrite those settings on individual items in the collection if the user selects "Yes, add to existing Special Access entries" in the popup.

**Environment (please complete the following information):**
 - Device / platform: Mac laptop
 - Browser: Chrome Version 113.0.5672.126
- [ ] avalon-dev
- [ ] avalon-staging
- [x] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Problems were first reported by IULMIA in 7/2022, when we started logging data. Reported again on 6/5/2023.

**Additional notes**
This was only tested initially on mco-staging since the bug was reported in production. Should probably be tested 

# Done Looks Like
- [x] Controlled Digital Lending card moved out of Access Control content block
- [ ] "Apply to Existing" button added to each card block (item access, visibility, special access)
- [x] "Save" button added to each card block

# Mockups
Moving CDL card out of Access Control
![Screenshot 2023-06-20 at 8.58.23 AM.png](https://images.zenhubusercontent.com/5cc1c0106d0be2229caf2d9e/e2f38b98-76c8-4388-96fc-c2a2147a74ea)

Apply to Existing Items button added to individual cards for settings
![Screenshot 2023-06-20 at 9.10.41 AM.png](https://images.zenhubusercontent.com/5cc1c0106d0be2229caf2d9e/18b8cca5-4981-41f9-98e4-1e5d459bd50e)

joncameron:
# 
[...]

PR #5263, opened 2023-07-13
Fix read group assignment on overwrite

masaball:
@elynema found a bug in the access settings work when performing an overwrite that caused media object visibility to change. This PR ensures that the media object visibility read group is maintained through an overwrite and that the media object does not end up with multiple or no visibility read groups assigned.

This PR also adds increased whitespace above the "Set Default Access Control" heading per discussion in #5227

### DIFF
--- a/app/jobs/bulk_action_jobs.rb
+++ b/app/jobs/bulk_action_jobs.rb
@@ -213,18 +213,30 @@ module BulkActionJobs
           media_object.lending_period = collection.default_lending_period
         end
 
+        # If MediaObject visibility is different than Collection, the collection visibility
+        # overwrites, or is added to, the media object read groups. This can result in the media object
+        # read groups containing the wrong visibilty or multiple visibilities.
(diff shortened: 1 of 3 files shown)


app/jobs/bulk_action_jobs.rb
```diff
@@ -213,18 +213,30 @@ def perform(collection_id, overwrite = true, save_field = nil)
           media_object.lending_period = collection.default_lending_period
         end
 
+        # If MediaObject visibility is different than Collection, the collection visibility
```
cjcolvar, line 216: Since these variables are only used in the special access case, what do you think about moving the `if save_field == "special_access"` conditional to wrap these?  That would also DRY up the if and elsif clauses below.

cjcolvar:
Looks like the build is blocked because it is looking for a SME build that doesn't exist.  Rebasing on top of the lastest from `develop`  might resolve this.

PR #5265, opened 2023-07-17
Allow users in manager group to be added to collection with editor role

### LINKED ISSUE #5241: [BUG] Adding user to Editor staff role adds user to Manager instead
**Describe the bug**
When an Admin level user tries to add another user to the Editor staff role, that user is added to the Manager role instead. I have not tested if the behavior also occurs when a non-Admin user is doing the addition.

**To Reproduce**
1. Go to the Manage Content page for a collection
2. Scroll down to the `Assign Staff Roles` section
3. Fill in a username for the `Editor` Role
4. Click the `Add` button
5. Wait for the page to reload
6. Scroll down to the `Assign Staff Roles` section
7. The username added appears under the `Manager` role, not `Editor`

**Expected behavior**
Users in the Managers group can be added to any `Staff Role` successfully

 - laptop
 - OS: Ubuntu 22.04
 - Firefox 114.0.1
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

# Done Looks Like
- [ ]  Add new properties to admin collections for each role
- [ ]  Mirror that information into inherited_edit users
- [ ]  Changes to ability class and elsewhere as needed
- [ ]  Data Migration

cjcolvar:
Resolves #5241 

The fix to this long standing bug was to add a new property to the model.  Because of this a migration rake task was added to avoid existing managers becoming editors.

Administrators can now also be added as editors but it wouldn't have any effect since they are still administrators and can still do anything in the system.

### DIFF
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -71,6 +71,9 @@ class Admin::Collection < ActiveFedora::Base
   property :cdl_enabled, predicate: Avalon::RDFVocab::Collection.cdl_enabled, multiple: false do |index|
     index.as ActiveFedora::Indexing::Descriptor.new(:boolean, :stored, :indexed)
   end
+  property :collection_managers, predicate: Avalon::RDFVocab::Collection.collection_managers, multiple: true do |index|
+    index.as :symbol
+  end
 
   has_subresource 'poster', class_name: 'IndexedFile'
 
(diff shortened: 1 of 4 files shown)


masaball (approved):
Looks good!

PR #5277, opened 2023-07-31
Ramp poc

### LINKED ISSUE #5262: Replace MediaElement with Ramp IIIFPlayer
# Description

# Done Looks Like

### LINKED ISSUE #5278: Render rails templates with react-rails for Ramp integration
### Description

Research and implement a way to render Rails templates inside the React components in `react-rails`.
Usually React follows the following pattern to render outside HTML inside the React component;

```
const ParentComponent = () => {
  return (
    <ReactComponent>
      <Div Element />
      ....
      <Div Element />
    </ReactComponent>
  );
}

const ReactComponent = (children) => {
  return (
    <ReactChildComponent1 />
    {{ children }} // these children are the HTML components in the above code 
    <ReactChildComponent2 />
  );
}
```
`react-rails` doesn't implement this pattern.

We can use something similar in pattern to render Avalon's Rails code like the `Create Timeline` button and `Share` section in the item page in between the Ramp components.

### Done Looks Like
- [ ] Research for an alternative
- [ ] Implement this in the item view page for `Create Timeline` and `Share` templates
- [ ] Test everything functions properly

Dananji:
This PR removes MEJS player from the item page, and some references for it in the JS, HTML codes. Most of the retained MEJS code are still used by the playlists.

Note: add to playlist code in the item page was not removed yet because some of it can be re-purposed when implementing add to playlist functionality with Ramp.

Closes #5262 
Closes #5278


cjcolvar (approved):
Looks good.  We can handle the `htmlparser2` pin in the node upgrade PR.

PR #5282, opened 2023-07-31
Use #update instead of deprecated #update_attributes

cjcolvar:
Fix for https://github.com/avalonmediasystem/avalon-terraform/issues/43

This should be contributed back upstream to Persona.  There were no tests for this and the `#update_attributes` method was removed in Rails 6.1.


masaball (approved):
Looks good!

PR #5284, opened 2023-08-02
Strip trailing slashes from uris to avoid odd case in AF/LDP/AT

### LINKED ISSUE #5283: Error in Masterfile Move Action
# Description
https://app.honeybadger.io/projects/54117/faults/98594630
https://app.honeybadger.io/projects/54117/faults/98730749

https://media.dlib.indiana.edu/media_objects/8623j9768

"target" => "6w925172h/" (suspicious)

When a user enters a media object ID with a trailing forward slash, it is shown as valid in Avalon (green check appears by the form field) but triggers a 500 when actually attempted, and the masterfile becomes orphaned.

Maybe requires a monkey-path on ActiveFedora. AF initializes an empty object when the find method is run on a noid with a trailing slash. This empty object does not have a collection or any other information.

# Done Looks Like
- [ ] ActiveFedora patched to look up objects even with trailing slashes

cjcolvar:
Resolves #5283 

The problem we encountered is that ActiveFedora will return an initialized but empty object of the correct class when requesting an object with a trailing slash on it's id.  I was originally thinking of patching ActiveFedora, but after digging it seems like the easiest solution is to strip the trailing slash when converting the id to a uri before it is used to make the LDP request.  I was thinking that a more correct fix would be to have LDP/ActiveTriples update the subject instance variable to match what the server returned in the RDF in the response body (instead of what was passed into the initializer) but I wasn't quite sure where the best place to make that change.

### DIFF
--- a/config/initializers/active_fedora_general.rb
+++ b/config/initializers/active_fedora_general.rb
@@ -4,7 +4,7 @@ ActiveFedora::Base.translate_uri_to_id = lambda do |uri|
                                            uri.to_s.sub(baseurl, '').split('/', baseparts).last
                                          end
 ActiveFedora::Base.translate_id_to_uri = lambda do |id|
-                                           "#{baseurl}/#{Noid::Rails.treeify(id).sub(/\/$/, '')}"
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

PR #5293, opened 2023-08-08
Add series to MediaObject model and edit page

masaball:


app/models/media_object.rb
```diff
@@ -116,6 +116,9 @@ def validate_date(date_field)
   property :lending_period, predicate: ::RDF::Vocab::SCHEMA.eligibleDuration, multiple: false do |index|
     index.as :stored_sortable
   end
+  property :series, predicate: ::RDF::Vocab::SCHEMA.Series, multiple: true do |index|
```
masaball, line 119: We may want to ask Julie if there is a more appropriate predicate, maybe through EBUCore or otherwise, but this one should be alright.

app/models/media_object.rb
```diff
@@ -366,6 +370,33 @@ def reload
     super
   end
 
+  def self.autocomplete(query, id)
```
cjcolvar, line 373: I'm wondering if we'll ever want/need an autocomplete for MediaObjects themselves and might want to use this method for that.  But we don't have a use case for it right now so I think this works.  If that ever comes up then we can do the work then to take series out to it's own simple model.

cjcolvar (commented):
Looking good!  One small change and a suggestion then this looks ready to merge.

cjcolvar (approved):

PR #5308, opened 2023-08-10
Node and webpack upgrade using shakapacker

### LINKED ISSUE #5304: Upgrade node and webpack versions to currently supported versions
Related to ramp#210

# Done Looks Like
- [ ] Upgrade node.js to x.x
- [ ] Upgrade webpack to x.x

cjcolvar:
Fixes #5304

Webpacker has been deprecated in favor of other approaches to JS in Rails 7.  Shakapacker is a continuation of webpacker that has support for webpack 5.  The switch to webpack 5 also required explicitly including polyfills.  This successfully compiles assets, but I haven't tried seeing if everything is actually working in a browser.


cjcolvar:
I had to do a `docker-compose pull` to get the newer image which resolved the bundling issue I was seeing.

Dananji:
I tested Collections, Timeliner, SME, and some other generic JS code. It all worked well expect for uploading a collection poster image giving me the error 'There was a problem storing the poster image.', which works in `avalon-dev`.

I faced a similar issue with Ramp we saw on avalon-dev where it cannot find the component on item page.
Screenshot from 2023-08-17 11-32-13

cjcolvar:
Testing locally, I didn't have that issue with uploading a poster, but I did see that same error with Ramp.  I think I fixed the underlying issue and was able to get Ramp to not throw that error, but it still wasn't loading for me because I think the item I was testing with was bad.

Dananji (approved):
Item view page works now :tada: 
The poster image issue could be something to do with my local setup.

PR #5311, opened 2023-08-15
Add support for solr 9 and test it

### LINKED ISSUE #5309: Add support for Solr 9
Avalon has solr configs which work with solr 6 and another set of configs which work with solr 8.  Solr 9 has been out for a while and IU is looking at upgrading straight to 9 from current solr 6.  We need to add support for solr 9 to core avalon and ensure that it works with our existing queries.

Done Looks Like
- [x] Determine if solr 9 needs separate set of configs or if can reuse solr8 configs
- [x] Any config changes have been made
- [x] CircleCI build uses solr 9
- [x] Avalon-dev uses solr 9
- [x] Perform quick testing on avalon-dev to see if anything is broken and write up new tickets

cjcolvar:
Resolves #5309 

Config change should still work with solr 8.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
   build:
     docker:
       # Primary container image where all steps run.
-      - image: avalonmediasystem/avalon:7.5.0-dev-ruby3
+      - image: avalonmediasystem/avalon:develop
         environment:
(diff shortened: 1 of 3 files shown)


masaball (approved):
Looks good to me.

PR #5315, opened 2023-08-16
Use blank canvas to represent inaccesible playlist items

masaball:
The PlaceHolder Canvas generation may still need work. With this implementation, the placeholder is created with a body section that just has a label within it:
`{ type: Annotation, motivation: painting, body: { label: "You do not have access to this item", type: Text, format: text/plain} }`.

I don't think this conflicts with the spec, but there may be a better way to serialize the messages into the placeholder. Making those changes will likely require more changes in IIIF Manifest because the placeholder serialization in place in the gem is pretty specific.

### DIFF
--- a/app/models/iiif_playlist_canvas_presenter.rb
+++ b/app/models/iiif_playlist_canvas_presenter.rb
@@ -16,7 +16,7 @@ class IiifPlaylistCanvasPresenter
   attr_reader :playlist_item, :stream_info, :cannot_read_item
   attr_accessor :media_fragment
 
-  def initialize(playlist_item:, stream_info:, cannot_read_item: nil, media_fragment: nil)
+  def initialize(playlist_item:, stream_info:, cannot_read_item: false, media_fragment: nil)
     @playlist_item = playlist_item
     @stream_info = stream_info
     @cannot_read_item = cannot_read_item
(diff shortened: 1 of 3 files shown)


cjcolvar (commented):
I think there is one thing that needs changing and I had one suggestion, but it's looking great!

app/models/iiif_playlist_canvas_presenter.rb
```diff
@@ -13,19 +13,26 @@
 # ---  END LICENSE_HEADER BLOCK  ---
 
 class IiifPlaylistCanvasPresenter
-  attr_reader :playlist_item, :stream_info
+  attr_reader :playlist_item, :stream_info, :cannot_read_item
   attr_accessor :media_fragment
 
-  def initialize(playlist_item:, stream_info:, media_fragment: nil)
+  def initialize(playlist_item:, stream_info:, cannot_read_item: nil, media_fragment: nil)
```
cjcolvar, line 19: How about making the default for the new kwarg be `false` to better show it is expected to be a boolean?
```suggestion
  def initialize(playlist_item:, stream_info:, cannot_read_item: false, media_fragment: nil)
```

cjcolvar (approved):

Issue #5275, opened 2023-07-26
Add Series Metadata to Resource Description Form


joncameron:
# Done Looks Like 
- [ ] Series field added to Resource Description form
- [ ] Series text field provides typeahead options with the list being drawn from current series in the collection's Unit

joncameron:
I ran into one bug while testing: entering a pre-existing Series value into the Series field doesn't seem to save properly. @masaball can you take a look?

To reproduce: 
- Go to https://avalon-dev.dlib.indiana.edu/media_objects/000000000/edit?step=resource-description
- Enter "First series" in an empty Series field
- Save
- "First series" was not saved to the object metadata

This seems to be the case for any pre-existing Series value, so this can also be reproduced by using the "Second Series" or "Third series" values which are shown in the above record as the value for a Series field in another object. You can see the values in the populated typeahead, but the save action doesn't seem to add them to the record and a refresh shows it not being retained.

joncameron:
👍

PR #5322, opened 2023-08-22
Fix input field for series autocomplete

masaball:
This PR fixes [the save bug](https://github.com/avalonmediasystem/avalon/issues/5275#issuecomment-1687040715) encountered by @joncameron. Also included is a small fix to avoid unnecessary pluralization in the "Series" form label.

The issue was that the series autocomplete suggestion only returns `suggestion['display']`, but the JS was setting the input field off `suggestion['id']` resulting in `nil` being the input value.

### DIFF
--- a/app/assets/javascripts/autocomplete.js.coffee
+++ b/app/assets/javascripts/autocomplete.js.coffee
@@ -36,7 +36,7 @@
     limit: 10
   ).on("typeahead:selected typeahead:autocompleted", (event, suggestion, dataset) ->
     target = $("##{$t.data('target')}")
-    target.val suggestion["id"]
+    target.val suggestion["id"] || suggestion["display"]
     $t.data('matched_val',suggestion["display"])
     return
   ).on("keypress", (e) ->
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/autocomplete.js.coffee
```diff
@@ -36,7 +36,7 @@
     limit: 10
   ).on("typeahead:selected typeahead:autocompleted", (event, suggestion, dataset) ->
     target = $("##{$t.data('target')}")
-    target.val suggestion["id"]
+    target.val suggestion["id"] || suggestion["display"]
```
cjcolvar, line 39: I think it is sensible to have this as a fallback, but I think it would also be good to have `MediaObject#autocomplete` return `id` values as well so it matches the interface for autocomplete methods.  (We need to make this interface explicit somewhere in the code, probably on `ObjectsController#autocomplete`, but that is not in scope of this PR.)  I'm thinking `id` and `display` values would be the same for the series data.

cjcolvar (approved):
Looks great!

PR #5323, opened 2023-08-22
Ramp itemview page

### LINKED ISSUE #5256: Replace Existing UI with Ramp Components on Object Show Page
# Description
Continuing from previous work in #5169 to add Ramp to the media object show page. The Media player has its own issue #5262.

Note that overall page design and styling and styling of structured navigation is not part of this issue. Work with the Ramp structured navigation component to add accordion and expand/collapse functionality should be part of Ramp #191.

# Done Looks Like
- [x] MetadataDisplay component replaces the existing metadata/details section
- [x] StructuredNavigation replaces the current sections div (styling of structured nav is not part of this issue)
- [x] SupplementalFiles component replaces the Files panel
- [x] Transcript component moves to top right to display as a tab (see design in #5202 for where/how transcript should display)

Dananji:
Closes #5256 

Item page with ingest in progress:
Screenshot from 2023-08-22 11-25-45

Item page without a media item:
Screenshot from 2023-08-22 11-26-17

Item without structure:
Screenshot from 2023-08-22 11-26-34

Note: Empty structure needs to handled in Ramp properly.


app/javascript/components/Ramp.jsx
```diff
@@ -1,46 +1,75 @@
...
+  base_url, 
+  mo_id, 
+  master_files_count, 
+  share, 
+  timeline, 
+  in_progress, 
+  cdl 
+}) => {
   const [manifestUrl, setManifestUrl] = React.useState('');
+  const { enabled, can_stream, embed, destroy_CDL } = cdl;
```
cjcolvar, line 25: Is there a way to rename the variables locally?   I think it would be helpful for some of these to be prefixed by `cdl` to help provide better context.  Specifically `enabled` and `embed`.  Or maybe we could skip the mapping to access them through the `cdl` object like `cdl.enabled`?  What do you think?

cjcolvar (approved):
Looking good!  It is exciting seeing this coming together!

app/javascript/components/Ramp.jsx
Dananji: Yes, they lose context when doing the mapping from the props. I don't think there's an advantage of using `cdl_` prefix over accessing them through the object. 
Let's go with the format `cdl.enabled` :+1: I'll change the code accordingly.

cjcolvar (approved):

cjcolvar:
I reran just the failing tests and they passed but the build still shows failed because the codeclimate reporting steps don't work with the "Rerun failed tests" option yet.  This is probably because there wasn't any report because there weren't any tests to run on those nodes.  I'll have to tweak the config a little bit more.

PR #5326, opened 2023-08-23
Switch out calls to open() with URI#open()

cjcolvar:
This PR extends the work done by @sgurnick in #5288.

I like how this solution is simpler than using Faraday and is a smaller change since we were already using `open-uri`.  I added tests to make sure that the change to `each_line` was safe to do on all of the various inputs to `M3U8Reader#read()`


masaball (approved):
Looks good to me!

PR #5351, opened 2023-09-06
Move the embedded player from MEjs to Videojs

masaball:


Dananji:
This looks great :100: and the JS code looks good to me. 
I had the following suggestions to clean-up CSS in the code and figuring out the asset pipeline.

app/views/layouts/embed.html.erb
```diff
@@ -24,6 +24,12 @@ Unless required by applicable law or agreed to in writing, software distributed
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
     <title><%= render_page_title %></title>
 
+    <% # These link tags are necessary to initialize videojs and the quality selector plug-in using CDN.
+       # May be worth considering pulling the code into the local asset pipeline.
+       # VideoJS is held at the same version as is being used in Ramp. %>
+    <link href="https://vjs.zencdn.net/7.10.2/video-js.css" rel="stylesheet"/>
+    <link href="https://unpkg.com/@silvermine/videojs-quality-selector/dist/css/quality-selector.css" rel="stylesheet"/>
+
```
Dananji, line 32: As you have mentioned in the comment, we can pull in these through the asset pipeline instead of this approach. To do that, we could add the following lines in the `application.scss`;
```
@import "video.js/dist/video-js.css";
@import "@silvermine/videojs-quality-selector/dist/css/quality-selector.css";
```

app/views/modules/player/_video_js_element.html.erb
```diff
@@ -0,0 +1,235 @@
...
+
+  /* captions button selection */
+  .captions-on {
+    border-bottom: 0.45rem ridge $primaryGreen;
+  }
+
+  .vjs-custom-external-link .vjs-icon-placeholder {
+    font-size: 120%;
+  }
+</style>
```
Dananji, line 126: I think we could move this CSS to `embed.scss`, which is included in the asset pipeline and imported into the `embed.html.erb` file. What do you think?

app/views/modules/player/_video_js_element.html.erb
```diff
@@ -0,0 +1,235 @@
...
+        <a class='video-link' id='video-link-<%= @master_file.id %>' href="<%= master_file_url(@master_file.id) %>" target="_blank" rel="noreferrer noopener"><%= @master_file.title || @master_file.media_object.title %></a>
+      </label>
+    <% end %>
+  </div>
+<% end %>
+
+<% # These script tags are for pulling in videojs and the quality selector plug-in from their CDNs.
+   # May be worth considering pulling them into our local asset pipeline. %>
+<script src="https://vjs.zencdn.net/7.10.2/video.min.js"></script>
+<script src="https://unpkg.com/@silvermine/videojs-quality-selector/dist/js/silvermine-videojs-quality-selector.min.js"></script>
```
Dananji, line 197: Similarly, to load these through asset pipeline we could add the following to the `application.js`;
```
//= require video.js/dist/video.min.js
//= require @silvermine/videojs-quality-selector/dist/js/silvermine-videojs-quality-selector.min.js
```

app/views/modules/player/_video_js_element.html.erb
```diff
@@ -0,0 +1,235 @@
...
+        window.open('<%= master_file_url(@master_file.id) %>', '_blank').focus();
+      }
+    });
+
+    viewInRepoButton.addClass('vjs-custom-external-link');
+    let newButton = player.controlBar.addChild(viewInRepoButton, {}, <%= section_info[:is_video] ? 8 : 6 %>);
+    newButton.controlText('View in Repository');
+
+    document.querySelector('.vjs-custom-external-link .vjs-icon-placeholder').innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" height="1em" viewBox="0 0 512 512"><!--! Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. --><style>svg{fill:#ffffff}</style><path d="M320 0c-17.7 0-32 14.3-32 32s14.3 32 32 32h82.7L201.4 265.4c-12.5 12.5-12.5 32.8 0 45.3s32.8 12.5 45.3 0L448 109.3V192c0 17.7 14.3 32 32 32s32-14.3 32-32V32c0-17.7-14.3-32-32-32H320zM80 32C35.8 32 0 67.8 0 112V432c0 44.2 35.8 80 80 80H400c44.2 0 80-35.8 80-80V320c0-17.7-14.3-32-32-32s-32 14.3-32 32V432c0 8.8-7.2 16-16 16H80c-8.8 0-16-7.2-16-16V112c0-8.8 7.2-16 16-16H192c17.7 0 32-14.3 32-32s-14.3-32-32-32H80z"/></svg>'
+
```
Dananji, line 214: We could try to use the class `vjs-custom-external-link` to set the SVG using `background-image` in CSS. An example implementation from Ramp: https://github.com/samvera-labs/ramp/blob/main/src/components/MediaPlayer/VideoJS/components/styles/VideoJSFileDownload.scss#L16. 
If it's hard to do with CSS scoping with videojs classes, the current implementation is good too :+1: 
What do you think?

Dananji (changes requested):

app/views/modules/player/_video_js_element.html.erb
masaball: Definitely. Good catch. I missed that `embed.scss` was something that existed.

app/views/modules/player/_video_js_element.html.erb
masaball: I tried to move it to CSS based on that Ramp example but was unable to get the icon to render, so I'm going to leave the implementation as is for now.

Dananji (dismissed):

Dananji (approved):
:100

PR #5362, opened 2023-09-11
Prevent reification when loading media object #show

masaball:


cjcolvar (commented):
`persisted?` is an odd case since it should always be true for any item that is stored in solr.  We could probably add it to the defaults in the initializer.  (Or pushed upstream to SpeedyAF::Base.)  In the Admin::Collection proxy and MediaObject proxy I defined `published?` as `id.present?` since only persisted items should have ids.  (But maybe that isn't always true?)

All three of these approaches work.  I'd favor either of the two above since it wouldn't rely on it being stored in solr.  What do you think?  Is there an approach you think we could apply to all proxies?

masaball:
I think for the purposes of this PR, matching the current implementation from the other proxies makes the most sense. I can't really think of a non-contrived situation where a non-persisted item would have an id assigned.

But doing a reexamination of the implementation and possibly refactoring may be something to consider in the future.

cjcolvar (approved):

PR #5363, opened 2023-09-11
Make avalon work with Fedora 6

cjcolvar:
It might be possible that these changes to Avalon would still work with Fedora 4 without the overrides.  That might be useful for implementors who aren't ready to upgrade Fedora but want to upgrade their avalon.

TODO:
- [x] Manual testing to ensure it really works
- [ ] Write passing tests in ActiveFedora for Fedora 4 to prove hasPart triples were automatically managed before
- [ ] Port tests to ActiveFedora for Fedora 6 and port code that adds management of hasPart triples
- [ ] Port rest of overrides to Fedora 6 ActiveFedora branch
- [ ] Remove overrides from initializer

Supersedes #5150

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,7 +141,7 @@ jobs:
         default: 4
     docker:
       # Primary container image where all steps run.
-      - image: avalonmediasystem/avalon:7.5.0-dev-ruby3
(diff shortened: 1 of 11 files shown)


cjcolvar:
Closed with #6082

PR #5367, opened 2023-09-12
Use Modernizr check before attempting to use localStorage

cjcolvar:
This appears to fix the issue whereas the previous attempt didn't.


Dananji (approved):
Looks good!

PR #5375, opened 2023-09-15
Add canvas reference to empty ranges

### LINKED ISSUE #5350: [BUG] Ramp components not displaying on media object page for some items
**Describe the bug**
For some items, the Ramp components are not displaying on the media object page 

**To Reproduce**
1. Two examples I have found so far: https://avalon-dev.dlib.indiana.edu/media_objects/r207tp32d and https://avalon-dev.dlib.indiana.edu/media_objects/fj236208t 
2. Only media item title and top row of buttons (ex: edit / unpublish / delete) show on page. The rest of the page is blank, no media player, etc.

**Expected behavior**
Should see media player, structured navigation, metadata, transcript, and supplemental files rendered by Ramp.

**Screenshots/Screen Recording**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbkp5IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--4cb0e76c6b4df1b345b4c9c9e63cae7893cb36a0/image.png)

 - Device / platform: Mac laptop
 - OS: Monterey
 - Browser: Chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
7.7

**Additional notes**
Looks like Ramp div is populated on page, but nothing in it. Probably one of the checks is failing and has not been failed.

### LINKED ISSUE #5368: Add Explicit Canvas Reference to Manifest Ranges
# Description
In Avalon IIIF manifests it's possible for structural metadata to contain a single childless div. In context of the current manifest layout, canvases aren't able to be referenced directly from ranges other than the implicit order of elements in the JSON. Canvas references should be included in the structure section of the manifest to ensure that parsing and display is handled correctly.

Childless divs can exist at arbitrary locations in the structure hierarchy; regardless of where they are in the structure they should have links to a canvas. Our approach should ultimately handle these gracefully; we may need to create an additional issue to accommodate these nested divs rather than the simpler case of a childless div at the top level.

See https://github.com/samvera-labs/ramp/issues/217 for previous discussion.

# Done Looks Like
- [x] Generate links to canvases in the manifest when structural metadata includes childless divs

masaball:
Closes #5350, Closes #5368 

The bug encountered in #5350 is a variant of the childless div bugs we have been encountering. If a user saves a structure without making any edits at all an XML file is still created with a completely childless element: `<Item label='label'/>`. This was generating a range with `item: [ ]` just like we see with childless divs.

This PR adds a fallback to generate a range `item` referencing the parent canvas for any empty structure element.

### DIFF
--- a/app/models/iiif_canvas_presenter.rb
+++ b/app/models/iiif_canvas_presenter.rb
@@ -128,6 +128,8 @@ class IiifCanvasPresenter
     end
 
     def div_to_iiif_range(div_node)
+      return simple_iiif_range(div_node[:label]) unless div_node.children.present?
+      
       items = div_node.children.select(&:element?).collect do |node|
         if node.name == "Div"
           div_to_iiif_range(node)
--- a/spec/models/iiif_canvas_presenter_spec.rb
+++ b/spec/models/iiif_canvas_presenter_spec.rb
@@ -95,6 +95,32 @@ describe IiifCanvasPresenter do
         expect(subject.items.first.media_fragment).to eq 't=0,'
       end
     end
+
+    context 'with childless div' do
+      let(:structure_xml) { '<?xml version="1.0"?><Item label="Test"><Div label="Div 1"/></Item>'}
+
+      it 'autogenerates a basic range' do
+        expect(subject.label.to_s).to eq '{"none"=>["Test"]}'
+      	expect(subject.items.size).to eq 1
+        expect(subject.items.first.label.to_s).to eq '{"none"=>["Div 1"]}'
+        expect(subject.items.first.items.size).to eq 1
+      	expect(subject.items.first.items.first).to be_a IiifCanvasPresenter
+        expect(subject.items.first.items.first.media_fragment).to eq 't=0,'
+        expect { subject.items.first.items.first.items }.to raise_error(NoMethodError)
+      end
+    end
+
+    context 'with childless item' do
+      let(:structure_xml) { '<?xml version="1.0"?><Item label="Test"/>'}
+
+      it 'autogenerates a basic range' do
+        expect(subject.label.to_s).to eq '{"none"=>["Test"]}'
+      	expect(subject.items.size).to eq 1
+      	expect(subject.items.first).to be_a IiifCanvasPresenter
+        expect(subject.items.first.media_fragment).to eq 't=0,'
+        expect { subject.items.first.items }.to raise_error(NoMethodError)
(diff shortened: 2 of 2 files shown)


app/models/iiif_canvas_presenter.rb
```diff
@@ -128,6 +128,8 @@ def structure_to_iiif_range
     end
 
     def div_to_iiif_range(div_node)
+      return simple_iiif_range(div_node[:label]) unless div_node.children.present?
```
cjcolvar, line 131: I think this might cause problems if the childless div occurs in an item that has valid spans.  How about narrowing the case where this fix is applied to only be items where there aren't any spans?  I'm not quite sure how to identify that case and still preserve all the childless divs if there is a whole hierarchy of them.  I think it should be possible but might be a tricky or require some refactoring to these recursive methods.

masaball: I'm not sure it'll be an issue? If there are sibling Divs or Spans to the childless div, they will be in the array of Item children. So an Item with node array `["Childless Div", "Span", "Span"]` should loop through everything and result in `return simple_iiif_range`, `return span_to_iiif_range`, `return span_to_iiif_range`. The return on the childless div shouldn't break the parent Item loop. I've added a test to check the behavior and make sure future changes don't break it, but I think this implementation is safe unless there is something I am missing with `return` and recursive methods.

cjcolvar: I think the issue won't be on generating the manifest but on the player side consuming the manifest.  From my reading of the IIIF presentation api I think canvas ranges will be played in order so all childless divs replaced with the full canvas range will cause the player play the whole canvas at that point in the structure.  I tried out a hand edited manifest in ramp which had a childless div inserted before the first canvas range.  It played through the whole canvas and then hit an error (about transcripts) when it tried to transition to the next range.  I think Dananji mentioned that Ramp wasn't setup yet to handle this kind of structure. Maybe we could do a screen share to talk through this?

masaball:
Closing this PR for now. Chris and I talked it over and think that there needs to be more conversation on how we approach this situation and #5350 is closed by [Ramp 243](https://github.com/samvera-labs/ramp/pull/243).

Issue #5379, opened 2023-09-25
Add functionality to "Expand All Sections" button


masaball:
With the media object page redesign, we are trying to implement an accordion style list for the structured navigation. Part of this is an "Expand All Sections" button. This button has been added to the page but still needs to be wired to the structured navigation once the accordion style list has been implemented.

See Ramp [Ramp 247](https://github.com/samvera-labs/ramp/issues/247)

# Done Looks Like
- [x] Expand Sections button should read "v Expand x Sections" where X = the number of sections. 
- [x] After button is clicked and sections are expanded, it should read "^ Close x sections"
- [x] expand all / collapse all button does not appear if none of the sections have any structure
- [x] Structure should be expanded by default when media object page loads
- [x] While a media item is playing, if it is actively playing a section with structure, structure should expand all by default so that the structure can be seen.

If it's challenging to include the # of sections in the button b/c the button is living outside of the Ramp context, we can re-evaluate that done looks like criteria.

joncameron:
Implemented as a CSS change and could either be in Avalon or Ramp; could be a part of the StructuredNav component within Ramp if done that way (or a component used within StructuredNav). Probably makes more sense to do it in Avalon right now, and could be moved into Ramp in the future if that seems to make more sense.

Dananji:
While a section is playing the structure is expanded whether it has any active timespan or not. This is because of Ramp re-rendering the structured navigation component with every state update (triggered in this case by playback).

charumitraravi:
QA completed. Dod items are verified. Although the behavior feels a little odd that when the user tries to close the sections while a media file is playing. It forcibly closes back. [Refer [Screen recording](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWZLQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--06dc28c6168dd6b3d930c679c9e807660fc3fefb/Expand-collapse%20button.mov)].

Dananji:
@charumitraravi This behavior is expected with this first pass implementation as I've mentioned. This is something that can be handled on the Ramp side hopefully, when re-design the state management.

joncameron:
Opened samvera-labs/ramp#298 for eventually fixing the behavior with collapse during media playback.

Issue #5316, opened 2023-08-16
Add to Playlist


joncameron:
# Description
Add to Playlist functionality as part of the media object view page rework. This will be implemented as Avalon functionality, rather than a Ramp component. We've attempted to re-work multiple add to playlist buttons that currently appear in the interface into a single button that provides all available options. 

# Definition of Done
- [ ] Add to Playlist button shows up only next to the Create Thumbnail button (not in player or in structured metadata display) and opens as an accordion instead of a modal popup
- [ ] Users can add item to an existing playlist (select from drop-down) or create a new playlist
  - [ ] The playlist drop down is the same as in current Avalon (see screenshot below) with the option to create a select an existing playlist or create one
  - [ ] If a user has no playlists, they are prompted to create one (see #5391 for the work of implementing the "add to playlist" modal
- [ ] Default in select playlist drop-down is first playlist (alphabetical order)
- [ ] Select playlist drop-down provides a type-ahead
- [ ] Add to Playlist button provides 4 options to users to add content to playlist
  - [ ] Users can add current track
    - [ ] The currently active track label is set as the default Title
  - [ ] Users can add a custom time scoped item to playlist. Start time for custom time is set to the current playhead position, and end time is end of current track (currently selected segment)
    - [ ]  The currently active track label is set as the default Title
  - [ ] Users can add the current section, with option to create a single playlist item for section, or one for each track (segment). If user adds current section as a single playlist item, should be able to create title / description for playlist item.
  - [ ] Users can add all sections, each as a single playlist item
  - [ ] If 'Current Track" or "Time Selection" are selected, the Title and Description fields display; the "Create playlist items for each track" check box does not display
  - [ ] If 'Current Section" or "All Sections" are selected, the Title and Description fields do not display; the "Create playlist items for each track" check box displays
  - [ ] Current Section and Current Track options display the appropriate label for section or track in parentheses in the form label text

# WIP Mockup
## Current Track
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcG5qIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1e0fa0022325ae4722ba76db494c23194a6cf171/current_track.png)

## Current Section
[...]

PR #5380, opened 2023-09-25
First pass redesign of media object page

masaball:
This PR updates the styling of the media object page to better match [the mockup](https://mco.stagingcenter.com/).

Work for future PRs:
- Update all the max width css rules that affect the player and metadata component to utilize the whole page like in the mockup (if we are adjusting the player page in that manner)
- Accordion styling for the structure list
- [Adding functionality to the expand all sections button](#5379)
- [Create the "Add to Playlist" accordion form](#5316)

### DIFF
--- a/app/javascript/components/Ramp.jsx
+++ b/app/javascript/components/Ramp.jsx
@@ -33,6 +33,7 @@ const Ramp = ({
   mo_id,
   master_files_count,
   title,
+  expand_structure,
   admin_links,
   share,
   timeline,
@@ -68,6 +69,9 @@ const Ramp = ({
                           { share.canShare && <div className="share-tabs" dangerouslySetInnerHTML={{ __html: share.content }} /> }
                           { admin_links.canUpdate && <div className="mr-1" dangerouslySetInnerHTML={{ __html: admin_links.content }} /> }
                           { thumbnail.canCreate && <div className="mr-1" dangerouslySetInnerHTML={{ __html: thumbnail.content }} /> }
+                        </div>
+                        <div className="ramp--rails-expand-structure">
+                          { <div className="mr-1" dangerouslySetInnerHTML={{ __html: expand_structure.content }} /> }
                         </div>
                         <StructuredNavigation />
                       </React.Fragment>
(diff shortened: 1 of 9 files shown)


app/javascript/components/Ramp.jsx
```diff
@@ -73,18 +85,24 @@ const Ramp = ({
           { cdl.enabled && <div dangerouslySetInnerHTML={{ __html: cdl.destroy }}/> }
           <Tabs>
             <Tab eventKey="details" title="Details">
+            <div class='scrolling-div'>
               <MetadataDisplay showHeading={false} displayTitle={false}/>
+            </div>
             </Tab>
             { (cdl.can_stream && master_files_count != 0 && !in_progress) &&
               <Tab eventKey="transcripts" title="Transcripts">
+              <div class='scrolling-div'>
```
Dananji, line 94: In JSX `className` is used instead of `class`. This gives a warning in the console, as an invalid DOM property. So it is `className='scrolling-div'` instead of `class='scrolling-div'`

app/javascript/components/Ramp.scss
```diff
@@ -28,16 +28,62 @@
...
+  .sr-only {
+    border: 0;
+    clip: rect(0 0 0 0);
+    height: 1px;
+    margin: -1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    width: 1px;
+  }
```
Dananji, line 64: `.sr-only` is a Bootstrap class. I'm not sure whether we need add this CSS in the Avalon's code. This should be already coming in from the Bootstrap's CSS via the asset pipeline. I think we could remove this from this file?

Dananji (changes requested):

app/javascript/components/Ramp.jsx
masaball: On further thought, and after the discussion of #5390, these divs shouldn't be added. I copied them from the mockup html, but didn't add/copy any css related to that class so its superfluous. And then we don't want the metadata (and presumably section files) to scroll anyways, only the transcript which is already handled by Ramp. So I will remove these.

Dananji (approved):
Looks good :100:

PR #5381, opened 2023-09-25
Fix production environment builds

cjcolvar:
When deploying the `sandbox` branch I ran into a couple issues that broke the Github Actions workflow.
- Webpacker references - We moved off of webpacker to shakapacker so these needed to be cleaned up.
- Out of disk space - The container used by the Github Actions workflow ran out of disk space so I added a 3rd party action to clean up disk space at the beginning of the workflow.  This step takes a decent amount of time so we may want to try configuring it differently or finding a different solution.

### DIFF
--- a/.github/workflows/podman-image.yml
+++ b/.github/workflows/podman-image.yml
@@ -22,6 +22,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


masaball (approved):
Looks good!

PR #5414, opened 2023-10-09
Add canvas reference to root range when childless divs are present

masaball:


app/models/iiif_canvas_presenter.rb
```diff
@@ -152,6 +162,10 @@ def span_to_iiif_range(span_node)
       )
     end
 
+    def empty_descendants?(node)
+      node.xpath('.//Div[not(node())]').present?
+    end
```
cjcolvar, line 167: I could be misremembering but I thought we decided to only handle the case when there are only childless divs and no spans.  Maybe something like this?
```suggestion
    def only_empty_descendants?(node)
      node.xpath('.//Span').empty?
    end
```

Issue #5411, opened 2023-10-06
Audit solr records for fields that should be stored but are only indexed


cjcolvar:
The SpeedyAF work moves all end-user facing actions to read from solr instead of fedora.  #5387 reveals that many fields are only being indexed and not stored in solr.  This means that those fields aren't being populated on the SpeedyAF proxies and are filled in with their default values instead.  This could lead to problems as data might be missing but errors not raised.

Done Looks Like
Audit all solr documents for objects with SpeedyAF proxies to ensure all fields used on the proxies have an index key that indicates it should be stored in solr.

joncameron:
This should be done in 7.7 since these fields have changed and re-indexing will be required.

PR #5425, opened 2023-10-17
Adjust indexing so all MO metadata values get added to proxy

masaball:
Do not merge and reindex until after user testing. Earliest merge date is 11/15/23.


app/presenters/speedy_af/proxy/media_object.rb
```diff
@@ -140,6 +140,22 @@ def governing_policies
...
+    attrs[:note].map.with_index { |n, i| { note: n, type: attrs[:note_type][i] } } if attrs[:note].present?
+  end
+
+  def other_identifier
+    attrs[:other_identifier].map.with_index { |oi, i| { id: oi, source: attrs[:other_identifier_type][i] } } if attrs[:other_identifier].present?
+  end
+
+  def related_item_url
+    attrs[:related_item_url].map.with_index { |url, i| { url: url, label: attrs[:related_item_label][i] } } if attrs[:related_item_url].present?
+  end
```
masaball, line 157: As mentioned in standup, these are the fields that I am concerned about content drift on. In automated testing, and in the limited manual testing I have done, the order of the SOLR fields is consistent and these methods return the proper `{ key: value }` pairs. Is it possible that the fields can end up out of sync? What other approaches might we be able to take to ensure these fields return the same information as calls to the Media Object itself?

app/presenters/speedy_af/proxy/media_object.rb
```diff
@@ -140,6 +140,22 @@ def governing_policies
...
+    attrs[:language_code].map { |code| { code: code, text: LanguageTerm.find(code).text } } if attrs[:language_code].present?
+  end
+
+  def note
+    attrs[:note].map.with_index { |n, i| { note: n, type: attrs[:note_type][i] } } if attrs[:note].present?
+  end
+
+  def other_identifier
+    attrs[:other_identifier].map.with_index { |oi, i| { id: oi, source: attrs[:other_identifier_type][i] } if attrs[:other_identifier_type][i].present? }.compact if attrs[:other_identifier].present?
+  end
```
masaball, line 153: This method is especially problematic because values outside of the media object are appended to this field in the solr_doc. If the order that things are added is maintained, it should be safe? May be better to split this field into what is included in the other_identifier accessor on the media object, and everything else. Unsure how much work splitting the field would be though.

app/models/mods_behaviors.rb
```diff
@@ -39,52 +39,52 @@ def to_solr(solr_doc = Hash.new, opts = {})
...
     solr_doc['other_identifier_sim'] = gather_terms(self.find_by_terms(:other_identifier))
 
     # Extract 4-digit year for creation date facet in Hydra and pub_date facet in Blacklight
     solr_doc['date_ssi'] = self.find_by_terms(:date_issued).text
     solr_doc['date_created_ssi'] = self.find_by_terms(:date_created).text
     # Put both publication date and creation date into the date facet
-    solr_doc['date_sim'] = gather_years(solr_doc['date_ssi'])
-    solr_doc['date_sim'] += gather_years(solr_doc['date_created_ssi']) if solr_doc['date_created_ssi'].present?
+    solr_doc['date_ssim'] = gather_years(solr_doc['date_ssi'])
+    solr_doc['date_ssim'] += gather_years(solr_doc['date_created_ssi']) if solr_doc['date_created_ssi'].present?
```
cjcolvar, line 87: It looks like we have both `date_ssi` and `date_ssim`.  `date_ssi` is `date_issued` and `date_ssim` is a combination of the EDTF exploded out `date_issued` and `date_created` for powering the facet (e.g.199x becomes 1990,1991...1999) .  I'm thinking that we only really need to store `date_ssi` and not `date_ssim` since it is only used for search.  That being the case I wonder if we should rename `date_ssi` to `date_issued_ssi` to match the other renamings.  What do you think?
```suggestion
    solr_doc['date_sim'] = gather_years(solr_doc['date_ssi'])
    solr_doc['date_sim'] += gather_years(solr_doc['date_created_ssi']) if solr_doc['date_created_ssi'].present?
```

app/models/media_object.rb
```diff
@@ -272,22 +275,22 @@ def to_solr(include_child_fields: false)
...
+      all_text_values << solr_doc["temporal_subject_ssim"]
+      all_text_values << solr_doc["genre_ssim"]
+      all_text_values << solr_doc["language_ssim"]
+      all_text_values << solr_doc["physical_description_ssim"]
       all_text_values << solr_doc["series_ssim"]
-      all_text_values << solr_doc["date_sim"]
-      all_text_values << solr_doc["notes_sim"]
-      all_text_values << solr_doc["table_of_contents_sim"]
-      all_text_values << solr_doc["other_identifier_sim"]
+      all_text_values << solr_doc["date_ssim"]
```
cjcolvar, line 290: I think this is actually a bug and should be `date_ssi` and not the special `date_ssim` field.  Maybe `date_created_ssi` should also be added to `all_text_values` too?  See comment on `date_ssim`.
```suggestion
      all_text_values << solr_doc["date_ssi"]
```

EDIT: Actually maybe this is correct since a user is unlikely to search on '199x' but will probably want it to match when searching '1990'.

cjcolvar (commented):
This is looking really close!  

Reviewing this reminds me the importance of #5411 since there are some fields I noticed that I think maybe should be changed to stored but aren't in the scope of this PR since they aren't in the IIIF manifest.

app/models/mods_behaviors.rb
masaball: Yeah, I think renaming the field to `date_issued_ssi` makes sense. And I'll try out changing `date_ssim` to `date_sim`.

app/models/media_object.rb
masaball: Yeah, this does not feel like a bug. Almost all users are going to search in plaintext dates. Even if the user knows the EDTF system, I doubt many would think about running a search off that when the public facing dates are all plain text.

app/models/media_object.rb
```diff
@@ -272,22 +275,22 @@ def to_solr(include_child_fields: false)
...
+      all_text_values << solr_doc["genre_ssim"]
+      all_text_values << solr_doc["language_ssim"]
+      all_text_values << solr_doc["physical_description_ssim"]
       all_text_values << solr_doc["series_ssim"]
       all_text_values << solr_doc["date_sim"]
       all_text_values << solr_doc["notes_sim"]
-      all_text_values << solr_doc["table_of_contents_sim"]
-      all_text_values << solr_doc["other_identifier_sim"]
+      all_text_values << solr_doc["table_of_contents_ssim"]
+      all_text_values << solr_doc["other_identifier_ssm"]
```
cjcolvar, line 293: I think this should go back to `sim` like `notes_sim` above so it is just the values instead of the json hashes.
```suggestion
      all_text_values << solr_doc["other_identifier_sim"]
```

cjcolvar (commented):
Looks great!  One small fix and it should be good to go! :tada:

cjcolvar (dismissed):
:+1:

cjcolvar:
For future reference these are the changes to the solr document in this PR:

Changed fields:
summary_ssi -> abstract_ssi
publisher_sim -> publisher_ssim
contributor_sim -> contributor_ssim
subject_sim -> subject_ssim
genre_sim -> genre_ssim
table_of_contents_sim -> table_of_contents_ssim
access_sim -> usage_ssim
format_sim -> resource_type_ssim
location_sim -> location_ssim
material_sim -> material_ssim
subject_topic_sim -> topical_subject_ssim
subject_geographic_sim -> geographic_subject_ssim
subject_temporal_sim -> temporal_subject_ssim
subject_occupation_sim -> occupation_subject_ssim
subject_person_sim -> person_subject_ssim
subject_corporate_sim -> corporate_subject_ssim
subject_family_sim -> family_subject_ssim
subject_title_sim -> title_subject_ssim
time_sim -> time_ssim
language_sim -> language_ssim
language_code_sim -> language_code_ssim
physical_description_sim -> physical_description_ssim
terms_of_use_si -> terms_of_use_ssi
date_sim -> date_ssim
date_digitized_sim -> date_digitized_ssim
all_comments_sim -> all_comments_ssim
date_ingested_sim -> date_ingested_ssim

Added hash fields:
note_ssm
other_identifier_ssm
related_item_url_ssm

PR #5437, opened 2023-10-20
Fix create thumbnail button behavior

masaball:


app/views/media_objects/_thumbnail.html.erb
```diff
@@ -45,29 +45,32 @@ Unless required by applicable law or agreed to in writing, software distributed
 <% content_for :page_scripts do %>
 <script>
   $(document).ready(function() {
+    const mediaObject = <%= @media_object.to_json.html_safe %>;
```
cjcolvar, line 48: How about limiting the scope of the json to just what is needed to avoid extra bits having to go across the wire?
Maybe something like:
```suggestion
    const sectionIds = <%= @media_object.ordered_master_file_ids.to_json.html_safe %>;
```

cjcolvar (commented):
Looks good.  I have one suggestion to tighten things up a bit.

Dananji (approved):
:100:

PR #5444, opened 2023-10-31
Add text-nowrap class to item page buttons

### LINKED ISSUE #5402: Button text wrapping to 2 lines
# Description

On avalon-dev, button text such as Create Thumbnail, Create Timeline, and Share button are wrapping to 2 lines as window gets narrower, which looks weird.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1lMQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a3df37a836883b8ce2df739a522e0061db79686e/image.png)

On MCO, this doesn't happen and the button text just gets smaller:

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1FMQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cba2f457091f6f7e0039510da95e2871853d34e1/image.png)

masaball:
Fixes #5402

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -67,6 +67,10 @@ h2 {
   font-family: $museoSlab;
 }
 
+.btn {
+  white-space: nowrap;
+}
+
 .btn-info {
   color: $primary;
 }
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -67,10 +67,6 @@ h2 {
   font-family: $museoSlab;
 }
 
-.btn {
-  white-space: nowrap;
-}
-
 .btn-info {
   color: $primary;
 }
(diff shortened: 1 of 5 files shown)


Dananji (commented):
This appears to be the solution to this issue :100: 
I do have the following question; 
By applying it to this Bootstrap class, this rule gets applied to all the buttons using the `btn` class in avalon. And do we want that? Or the do we just need the buttons on the item page to have this?

And if it is the second case where we need this only for the buttons on the item view page, we could use the Bootstrap class called `text-nowrap` (Docs: https://getbootstrap.com/docs/5.2/components/buttons/#disable-text-wrapping).

masaball:
That is something I had thought about as well. It is probably not an issue to have the behavior applied to all buttons. However, I think since there is built in functionality from bootstrap it's probably best for now to limit it to just the item page buttons and reassess if we end up seeing the unwanted line splitting behavior elsewhere. So I will go ahead and make the switch to the more limited set up.

PR #5447, opened 2023-11-01
Move administrative buttons to top of page and outside of scope of Ramp

### LINKED ISSUE #5405: Move Location of Admin Buttons on Item Page
# Description
As a collection staff member, it's hard to currently find the Edit button as it is in the middle of the page, rather than left-aligned, in addition to the change moving the admin buttons to below the player. After team discussion, we decided to move the admin buttons back to their original location above the player.

# Done Looks Like
- [ ] Admin buttons are moved to below the navbar and above the player

# Admin Button Location Mockup
Untitled-1

cjcolvar:
I chose to put the buttons below the workflow progress because it is a panel that looks like a flash message which feels like it should be at the top of the page.

Resolves #5405 

Before:
Screenshot from 2023-11-01 15-32-04

After:
Screenshot from 2023-11-01 15-23-53

### DIFF
--- a/app/views/media_objects/_item_view.html.erb
+++ b/app/views/media_objects/_item_view.html.erb
@@ -28,7 +28,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   <div class="col-sm-12">
     <%= render 'workflow_progress' %>
   </div>
-  <div class="col-sm-12" style="margin-bottom: 0.5rem">
+  <div class="col-sm-12 mb-2">
     <%= render 'administrative_links' %>
     <% if current_ability.can? :edit, @media_object %>
       <% if lending_enabled?(@media_object) %>
--- a/app/views/media_objects/_item_view.html.erb
+++ b/app/views/media_objects/_item_view.html.erb
@@ -25,8 +25,18 @@ Unless required by applicable law or agreed to in writing, software distributed
 
 <% in_progress = show_progress?(@masterFiles) %>
 <div class="row">
-  <div class="col-sm-8">
+  <div class="col-sm-12">
     <%= render 'workflow_progress' %>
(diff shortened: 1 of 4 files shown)


app/views/media_objects/_item_view.html.erb
```diff
@@ -25,9 +25,19 @@ Unless required by applicable law or agreed to in writing, software distributed
 
 <% in_progress = show_progress?(@masterFiles) %>
 <div class="row">
-  <div class="col-sm-8">
+  <div class="col-sm-12">
     <%= render 'workflow_progress' %>
   </div>
+  <div class="col-sm-12" style="margin-bottom: 0.5rem">
```
Dananji, line 31: We could probably use `mb-2` Bootstrap class instead of the `margin-bottom` CSS rule here?

app/views/media_objects/_thumbnail.html.erb
```diff
@@ -14,7 +14,7 @@ Unless required by applicable law or agreed to in writing, software distributed
 ---  END LICENSE_HEADER BLOCK  ---
 %>
 
-<div id="thumbnail-button" style="display: inline-block;">
+<div id="thumbnail-button" style="display: inline-block">
```
Dananji, line 17: And similarly here we could use `d-inline-block` Bootstrap class instead using inline CSS? I haven't really used this before, so not sure it gives the same appearance. Docs: https://getbootstrap.com/docs/4.0/utilities/display/

Dananji (changes requested):
This looks good to me :100: 
I had a couple of suggestions on using some of the existing Bootstrap classes instead of in-lining the CSS

Dananji (approved):

PR #5448, opened 2023-11-01
Shorten title for encodes using s3 urls

### LINKED ISSUE #5436: Really long filenames break ingest
Pulled over from local board [#174](https://github.iu.edu/AvalonIU/Avalon/issues/174)

# Description
[Chris]: It appears that really long filenames overflow the encode dashboard database record for the title field: https://app.honeybadger.io/projects/54117/faults/99716515
A DB migration may be needed to modify that column to change the datatype or a truncation of that field in ActiveEncode::EncodeRecord before attempting to save it.

[Chris]: I ran into this when trying to load a batch on avalon-sandbox as part of https://github.com/avalonmediasystem/avalon/issues/5397. avalon-sandbox was using minio for the dropbox and a mysql database when this happened. I temporarily worked around it by increasing the size of title to varchar(1000) and display_title to varchar(750).
This should probably be moved to the core board and handled there.

# Done Looks Like
- [ ] A user can ingest a filename with up to 256 characters either by web upload or batch ingest

cjcolvar:
Avoids the need to bump title and display_title columns in the DB for the long urls with many query params when using minio.

Resolves #5436

### DIFF
--- a/app/models/watched_encode.rb
+++ b/app/models/watched_encode.rb
@@ -60,8 +60,8 @@ class WatchedEncode < ActiveEncode::Base
   end
 
   def persistence_model_attributes(encode, create_options = nil)
-    display_title = encode.input.url.to_s.split('/').last
-    options_hash = { display_title: display_title }
+    display_title = parse_filename(encode.input.url)
+    options_hash = { title: display_title, display_title: display_title }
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

PR #5463, opened 2023-11-08
Move add to playlist button and add icon in front of button text

masaball:
Screenshot_20231108_141800

This PR also includes a small refactor to move the "Create Timeline" button from the Ruby/Rails partial to Ramp.jsx, to match how the playlist and share buttons are generated.

"Create Timeline" icon can be added in a followup PR after @joncameron has designed/created it.


Dananji (approved):
Looks good!

PR #5465, opened 2023-11-10
Bug fixes for persisting user edits and enable/disable current track option in add to playlist

Dananji:
When a current track is not available:
image

When a current track is available:
image

When there's no structure (current section has <media_object.id - master_file.id>):
image

### DIFF
--- a/app/views/media_objects/_add_to_playlist.html.erb
+++ b/app/views/media_objects/_add_to_playlist.html.erb
@@ -147,6 +147,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       let mediaObjectId, streamId, mediaObjectTitle;
       let listenersAdded = false;
       let activeTrack = null;
+      let currentSectionLabel = undefined;
       // Enable add to playlist button after derivative is loaded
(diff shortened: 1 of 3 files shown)


app/views/media_objects/_add_to_playlist.html.erb
```diff
@@ -224,34 +216,41 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-              $('#current-section-name').text(currentSectionLabel);
             }
 
             let trackInfo = scopes.filter(s => s.tags.includes('current-track'));
             if(trackInfo.length > 0) {
               activeTrack = trackInfo[0];
-              let trackName = `${currentSectionLabel} - ${activeTrack.label || streamId}`;
+              let trackName = activeTrack.tags.includes('current-section')
+                ? activeTrack.label || streamId
+                : `${currentSectionLabel} - ${activeTrack.label || streamId}`;
```
cjcolvar, line 235: There is kind of a lot going on here logically.  Would it make sense to break this down into multiple lines?

cjcolvar (approved):
This looks good as far as I can tell.

PR #5466, opened 2023-11-13
Do not display transcript and file tabs when none are present

masaball:


Dananji (approved):
This looks good! :+1: 

This is not something to do in this PR, I think we should probably consolidate all the player listener code for create timeline, create thumbnail, add to playlist, share, and this into one place. There's a lot of the same code in multiple places for these templates.

PR #5468, opened 2023-11-15
Reduce ffmpeg timeout to 1 minute

cjcolvar:
In a manual test using the same ffmpeg commands outside of ruby and without the timeout it took 7 seconds to run for a 1 hour audio file and 25 seconds for a 2 hour video file.  Given this I think a 1 minute timeout should be enough and keeps the queue from getting clogged as much when the stream doesn't close nicely for ffmpeg.

Locally, this would help us move through the backlog created by the mass push (https://github.iu.edu/AvalonIU/Avalon/issues/176).


masaball (approved):
Looks good

PR #5469, opened 2023-11-15
Align UI elements on the item view page

Dananji:
Screenshots of parts of the page for different UI elements;

Before:
image
After:
image

Before:
image
After:
image

Changes to the share resource UI;
Before:
image

After:
image


cjcolvar (approved):
Looks so much better!

PR #5470, opened 2023-11-15
Test that all attributes are being stored in solr

### LINKED ISSUE #5411: Audit solr records for fields that should be stored but are only indexed

cjcolvar:
Resolves #5411

### DIFF
--- a/app/models/concerns/master_file_behavior.rb
+++ b/app/models/concerns/master_file_behavior.rb
@@ -163,6 +163,10 @@ module MasterFileBehavior
     }
   end
 
+  def supplemental_file_captions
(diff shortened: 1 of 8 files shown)


masaball (approved):
Looks good!

PR #5472, opened 2023-11-17
Replace playlist show page player with Ramp

masaball:
Before:
Screenshot_20231117_104528

After:
Screenshot_20231117_104608

Work left to do:

- ~~Current playlist item title should be displayed below player~~
- ~~Source item details panel needs to be added~~ (Parsing will be brought in as part of future Ramp build. See [Ramp #252](https://github.com/samvera-labs/ramp/issues/252))
- ~~Panels below player (source item, markers, related items) should be collapsible~~
- ~~Indicator in structured nav for currently selected item~~ (Will be brought in with next ramp build)
- Related items panel needs to be added, if we are retaining that function. Can be done as a future PR.

Also, maybe the edit/copy buttons should be pulled back out of the Ramp context? If they are outside of the Ramp context, we could put them level with the title and then the autoplay and share buttons would be level with the player which is closer to how the page is currently set up.


Dananji (approved):
This looks great! :100:

PR #5486, opened 2023-11-29
Check sidekiq and redis status on about page

cjcolvar:
Uses https://github.com/avalonmediasystem/avalon-about/pull/8 which we can tag before release.

Resolves https://github.iu.edu/AvalonIU/Avalon/issues/185


cjcolvar:
The build failed because of one unrelated flaky test.

PR #5492, opened 2023-11-30
Cleanup ME.js code in JS, CSS, HTML files, remove SVG

Dananji:


cjcolvar (commented):
There are 6 failing tests that call the `PlaylistItemsController#markers` action and fail because `app/views/playlist_items/_markers.html.erb` was deleted.  We'll either need to remove that route, change what it returns, or restore the partial.  If it isn't called anywhere else and was only used to supply the UI then I'd favor removing it.  If we do that then I'd think the `source_details` and `related_items` actions could be removed too.

Dananji:
I think the `markers` endpoint is only for the display, I will remove that and `source_details` in this PR. I don't remember what we discussed to with `related_items` since it's not in the Manifest, so I'll leave it for the time being.

cjcolvar (approved):
Looks great! :tada:

PR #5510, opened 2023-12-06
Only look at language code when de-duping during bib import

cjcolvar:
Fixes https://github.iu.edu/AvalonIU/Avalon/issues/193

In production we encountered a bib import that supplied both the language code and text as part of the MODS.  This lead to problems in the import code since `self.language` returns the whole language block which includes both code and text.  This PR changes this to look at only the language code since this is all that typically comes through our SRU service.

For example:
```
<language>
  <languageTerm type="code" authority="iso639-2b">eng</languageTerm>
  <languageTerm type="text">English</languageTerm>
</language>
```
versus
```
<language>
  <languageTerm type="code" authority="iso639-2b">eng</languageTerm>
</language>
```


masaball (approved):
Looks good!

PR #5511, opened 2023-12-07
Move playlist action buttons out of ramp context

masaball:
The "Copy Playlist" modal relies on some functions which do not trigger/interact properly when the buttons are being rendered by/within the Ramp context. Rather than try to get the functions to interact with Ramp properly or move the functions into the Ramp context, it felt simpler to pull the action button partial back out of Ramp. This has the added benefit(?) of bringing the page styling closer inline with the non-Ramp based playlist show page.


Dananji (approved):
Looks good!

PR #5512, opened 2023-12-07
Add safe navigator to captions check

masaball:
Encountered an issue on avalon-dev where `@master_file.captions` returned nil instead of an empty Indexed file so the subsequent check for content errored. This prevents the HLS manifest from generating, causing streaming to fail. I think this will be encountered with any item created after we changed how captions are created/stored. Adding the safe navigator should take care of the issue.


cjcolvar (approved):
:+1:

PR #5519, opened 2023-12-12
Avoid expensive calls when rendering item show view page

### LINKED ISSUE #5509: Condense Solr Calls for Item Show Page
# Description
Refactoring Solr queries has helped speed up page load time for item show pages. Looking at this again and making sure the smallest number of calls to Solr happens could be an easy win for load times in the context of the 7.7 release.

# Done Looks Like
- [x] Investigation into Solr usage during item show page renders
- [x] Unnecessary calls refactored as needed

cjcolvar:
Resolves #5509 

I think the changes to these two files are all of the low hanging fruit that I could find.  It's still slow but much improved since we aren't trying to fetch the MasterFile objects from solr 4 or 5 times.

With https://avalon-dev.dlib.indiana.edu/media_objects/3f462541v
Before: 9.9s (page) + 6.6s (manifest) = 16.5s total
After: 1.9s (page) + 5.2s (manifest) = 7.1s total


masaball (approved):
Looks good!

PR #5525, opened 2023-12-13
Speed up playlist manifest generation

### LINKED ISSUE #5496: Speed up manifest generation for playlists with many items
# Description 

Work has already been done in #5476 to reduce calls out to stream tokens to significantly speed up manifest generation for items with many sections. Similar work would likely be helpful for playlists with many items. 

Playlist with 48 items takes almost 20 seconds to load: https://avalon-dev.dlib.indiana.edu/playlists/5

# Done Look Like
- [ ] Put in place the same speed-up strategies as done in #5476 for media objects
- [ ] Other investigation or changes as needed

cjcolvar:
Resolves #5496 

This handles low hanging fruit by reducing solr calls by:
- Fetching all MasterFiles backing playlist items in a single SpeedyAF::Base.where
- Using `#secure_stream_infos` to avoid multiple sql calls for stream tokens
- Performing `cannot? :read` checks on parent MediaObjects to optimize case where multiple playlist items come from the same MediaObject

With these changes a 125 section playlist manifest generation time dropped from about 60 seconds to 8-10 seconds.  Still slow but a big improvement.  Playlist view page renders very quickly <1s so I didn't look into optimizing it at all.

Future optimizations could include:
- Leveraging bulk SpeedyAF calls and the SpeedyAF proxies to avoid solr calls in `test_edit` and `test_read` in the Ability
- Fetch all of the markers for the playlist in a single SQL call
- Optimize (possibly via memoization) active_annotation parsing of RDF for `source`, `start_time`, and `end_time` of `AvalonAnnotation`s

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -464,7 +464,7 @@ class MediaObjectsController < ApplicationController
     @media_object = SpeedyAF::Proxy::MediaObject.find(params[:id])
     authorize! :read, @media_object
 
-    stream_info_hash = secure_stream_infos(master_file_presenters, @media_object.id)
+    stream_info_hash = secure_stream_infos(master_file_presenters, [@media_object])
     canvas_presenters = master_file_presenters.collect { |mf| IiifCanvasPresenter.new(master_file: mf, stream_info: stream_info_hash[mf.id]) }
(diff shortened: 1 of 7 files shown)


masaball (approved):
Looks good!

PR #5538, opened 2023-12-18
Add to playlist changes

masaball:
This PR:
- Changes the phrasing for the manually defined timespan option
- Provides a default selection based on whether the current timespan is within a track/subsection or not

### DIFF
--- a/app/views/media_objects/_add_to_playlist.html.erb
+++ b/app/views/media_objects/_add_to_playlist.html.erb
@@ -46,10 +46,10 @@ Unless required by applicable law or agreed to in writing, software distributed
           <div class="form-check">
             <label class="form-check-label">
               <input type="radio" onclick="collapseMoreDetails()" name="post[playlistitem_scope]" id="playlistitem_timeselection" aria-label="playlist item time selection">
-              Time Selection
+              Custom Timespan
             </label>
             <input type="text" name="playlist_item_start" id="playlist_item_start" pattern="(\d+:){0,2}\d+(\.\d+)?" value="" aria-label="start time">
-              to 
+              to
             <input type="text" name="playlist_item_end" id="playlist_item_end" pattern="(\d+:){0,2}\d+(\.\d+)?" value="" aria-label="end time">
           </div>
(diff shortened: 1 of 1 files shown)


app/views/media_objects/_add_to_playlist.html.erb
```diff
@@ -280,6 +297,11 @@ Unless required by applicable law or agreed to in writing, software distributed
             handleAddError({ responseJSON: { message: ['Please select a playlist option'] }})
           }
         });
+
+        $('#addToPlaylistBtn').on('click', function (e) {
+          resetAddToPlaylistForm();
+        });
```
cjcolvar, line 303: I'm curious why this is needed.  I would have guessed that https://github.com/avalonmediasystem/avalon/blob/6480ea57a1e71d168dbea26387ba5fffdbcdf199/app/assets/javascripts/ramp_utils.js#L289-L318 should take of this eventually along with your change to add it to the Cancel button.

cjcolvar (approved):

app/views/media_objects/_add_to_playlist.html.erb
masaball: The `#addToPlaylistBtn` element is not the Save button, it is the button to open/close the playlist panel. 

When testing my changes, if a user pressed that button to close the panel, instead of the cancel button, then the panel would break when switching canvases. The radio buttons would still get checked/unchecked but the sub-menus, `#moreDetails` and `#multiItemCheck`, wouldn't expand so something with the 'onclick' listeners on the radio buttons was breaking when the panel was closed but didn't get reset.

Thinking over it again, since that listener triggers anytime the button is clicked there is some potential volatility if other systems have different ordering/timing on how the onclick functions process. I did not encounter any issues with the panel getting populated when opening it with this code in place, but there could be potential when opening the panel that it gets reset after the functions that fill it out run ...I should probably go back and switch out the listener for something that only triggers when closing the panel, not just when the button gets clicked.

Dananji (approved):
Looks good :+1:

PR #5542, opened 2023-12-20
Unfocus player on page load

masaball:
Ramp applies focus to the player element when initializing so that hot keys will be functional. Avalon needs to have the window retain focus so that the tab order starts with the 'Skip to main content' link. We can achieve this by unfocusing the player after it loads and then listen for keydown events and assign them the proper hotkey actions to maintain hotkey functionality.

This commit also includes a couple minor changes to remove console errors encountered while testing.


masaball:
Rather than remove focus from the player on Avalon side, we will rework how hotkeys are handled on the Ramp side so that the player does not expect/require focus. Will break off the minor console error fixes into their own PR.

Issue #5504, opened 2023-12-05
Playlists With No Items Display Error Message


joncameron:
# Description
On the playlist show page for a playlist with 0 items, the error "Error encountered. Please check your Manifest." appears to the user. See https://avalon-dev.dlib.indiana.edu/playlists/46 for an example.

# Screenshot
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2t5QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0a55be7ca34bc3d73fec447e6b294594437d8305/Screenshot%202023-12-05%20at%206.51.28%E2%80%AFAM.png" alt="Screenshot 2023-12-05 at 6.51.28 AM.png" />

# Done Looks Like 
- [ ] Message displays to user "There are no items yet in this playlist

charumitraravi:
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMjl2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ddccaa98d26d58d5e4abc8db024dd31acfd50d87/Screenshot%202023-12-18%20at%2010.52.07%20PM.png" alt="Screenshot 2023-12-18 at 10.52.07 PM.png" />
I am still facing the same error

elynema:
@Dananji Charu tested this and we're still getting an error on avalon-dev. Moving this back to In Progress.

elynema:
Seems to be an issue with playlist manifest generation when there are no items in a playlist.

<img width="1508" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeEIwQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d362352609eae2029414364b3efb3269229fa786/image.png" alt="image.png" />

elynema:
Playlist manifest is generating with empty items[] (see https://avalon-dev.dlib.indiana.edu/playlists/39/manifest.json), but now an error is showing across all browsers (looks like it's coming from Ramp). The goal was for the playlist page to display a message "The playlist is empty" below the playlist title, instead of an error message.

Firefox
[...]

Issue #5514, opened 2023-12-08
[BUG] Playlist items - Opening a playlist item in a new tab shows a routing error


charumitraravi:
**Describe the bug.**
The page link for a playlist item opened in a new tab is broken and shows a routing error. 
Eg: For this [playlist](https://avalon-dev.dlib.indiana.edu/playlists/32) ,this is  the [second item opened in a new tab](https://avalon-dev.dlib.indiana.edu/playlists/32/manifest/canvas/36#t=0.0,) and the page is broken.

**To Reproduce**
1.Open any playlist with items.
2.Select one of the items, right click on it and choose the option "Open link in a new tab"

**Expected behavior**
A new tab should open, and  the marker should be positioned next to the item that was selected.

**Screenshots/Screen Recording**
<img width="1627" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMlpGQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5ebba26e10dd48f3a1a90f4b1450f3ca3ff9d55c/Screenshot%202023-12-08%20at%2010.54.41%20AM.png" alt="Screenshot 2023-12-08 at 10.54.41 AM.png" />

**Environment (please complete the following information):**
 - Device / platform : All devices
 - OS: All OS
 - Browser : All browsers
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
v7.7

URL that gets generated now in MCO: https://media.dlib.indiana.edu/playlists/5209?position=1

Existing URLs as above should also be supported and would not redirect currently.

Homepage property in the canvas could be the URL used for copying and bookmarking. Section PURLs end up going to the /section route as well. Avoid putting hompages on the ranges in the manifest.

# Done Looks Like
- [ ] Homepage attribute added to manifest to contain the positional URL (1-2 points)
  - [ ] Set up canvas presenter to serialize the value in the manifest
- [ ] Playlist route URLs with position parameter return the playlist view page with the specified playlist item active (one-based numbering) (2-3 points in Ramp)
  - [ ] https://avalon-dev.dlib.indiana.edu/playlists/18?position=2 in the browser goes to view page for playlist 18 and sets the second item in the playlist as active
- [ ] Links for playlist items are of the syntax: https://avalon-dev.dlib.indiana.edu/playlists/18?position=4

[...]

PR #5544, opened 2023-12-22
Use start canvas and time props in Ramp for sections and custom start times

Dananji:
This PR includes a new build of Ramp, which has work related to the following issues;
- https://github.com/samvera-labs/ramp/issues/316
- https://github.com/samvera-labs/ramp/issues/289
- https://github.com/samvera-labs/ramp/issues/310
- https://github.com/samvera-labs/ramp/issues/252
- https://github.com/avalonmediasystem/avalon/issues/5504

And a couple of changes to JavaScript to fix some of the console errors on the item view page for share and create timeline actions.

### DIFF
--- a/app/assets/javascripts/ramp_utils.js
+++ b/app/assets/javascripts/ramp_utils.js
@@ -221,7 +221,7 @@ function disableEnableCurrentTrack(activeTrack, currentTime, isSeeked, sectionTi
       $('#playlistitem_scope_track').closest('label').css('cursor', 'pointer');
     }
   } else {
-    // Whena activeTrack is undefined, disable the current track option
+    // When activeTrack is undefined, disable the current track option
     $('#playlistitem_scope_track')[0].disabled = true;
     $('#current-track-name').text('');
     $('#current-track-text').addClass('disabled-option');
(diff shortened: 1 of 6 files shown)


masaball (approved):
Looks good!

Dananji:
Right! How does the user gets to that URL in UI? It seems they are both similar use-cases, so I don't mind including that work in this PR.

masaball:
In 7.6, the `?position=` query is added to the URL when a user selects an item from the "Playlist Items" sidebar on the playlist view page. So the full url becomes "https://example.com/playlists/1?position=1".

Since the items sidebar just triggers canvas switches in the manifest, I don't think the URL is generated by UI actions in 7.7 currently.

Dananji:
I was looking at the previous playlist page in MCO, and we have the `?position=` URL in the playlist item link. But in the new UI its the Canvas URL. So it seems to me there's some things we need to discuss and refine on this implementation. And there's a ticket in the backlog related to this https://github.com/avalonmediasystem/avalon/issues/5514. 
I guess the playlist work could be part of that ticket?

masaball:
Yeah, saving the playlist work for #5514 sounds good. Had missed that in the backlog.

So I think we're good to go with this PR.

PR #5546, opened 2024-01-03
Store http sources as tempfiles on disk to avoid issue with ffmpeg blocking

cjcolvar:
Another attempt at resolving the issue with intercom push spawning WaveformJobs which run for hours.

Note: While testing I noticed that in the docker development environment the ffmpeg command for streaming urls doesn't work because of localhost not resolving.  This shouldn't be a problem in IU's production environment and probably won't be an issue at all because preference is given to using derivative files which are available via minio in the docker environment.  I wanted to mention this in case we find a scenario where it is an issue in the future.

### DIFF
--- a/app/services/waveform_service.rb
+++ b/app/services/waveform_service.rb
@@ -82,15 +82,27 @@ private
     factor = max_peak.zero? ? 1 : res / max_peak.to_f
     peaks.map { |peak| peak.collect { |num| (num * factor).to_i } }
   ensure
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good.

cjcolvar:
As far as I can tell this doesn't break things on avalon-dev but as mentioned above it probably isn't using the new code since it is backed by minio.

PR #5547, opened 2024-01-04
Return array of ids for SpeedyAF::Proxy::MediaObject.sections_with_files

masaball:
The transcript check function to hide/display the transcript tab has been broken for a little bit. This was caused by the media object presenter returning an array of full master files from the `#sections_with_files` method instead of an array of master file IDs which is what should be getting returned. This discrepancy caused the check of the active section ID against the list of sections with files to always fail, thus never rendering the transcript tab.


cjcolvar (commented):
I'm concerned about putting the `find` in here since it will cause calls to solr.  How about just changing the original by appending something like  `.map(&:id)` ?

cjcolvar (approved):

PR #5550, opened 2024-01-08
Display correct playlist item title for restricted items

Dananji:
For the restricted items, Ramp creates a VideoJS instance without a source URL. When this happens the expected `loadedmetadata` event does not fire, but it emits the `ready` event which can be used in this use-case to switch the playlist item title in the UI.

Remove `./refesh_token.js.coffee` script missed in the ME.js player cleanup in previous work.


masaball (approved):
Looks good!

PR #5551, opened 2024-01-08
Invoke transcript switch on player ready event

Dananji:
Trigger transcript tab show/hide on VideoJS player's `player.ready` event instead of `player.loadedmetadata` event which takes longer to fire with HLS streaming.


masaball (approved):
Looks good!

PR #5552, opened 2024-01-08
Fix add playlist current track enable/disable with section switching

Dananji:
Bug fix for disabling/enabling current track option when switching between sections.

### DIFF
--- a/app/assets/javascripts/ramp_utils.js
+++ b/app/assets/javascripts/ramp_utils.js
@@ -204,6 +204,11 @@ function collapseMoreDetails() {
  * @param {String} sectionTitle name of the current section 
  */
 function disableEnableCurrentTrack(activeTrack, currentTime, isSeeked, sectionTitle) {
+   // Return when add to playlist form is not visible
+  let playlistForm = $('#add_to_playlist')[0];
+  if(!playlistForm) {
+    return;
+  }
   let title = sectionTitle;
   if(activeTrack != undefined) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/ramp_utils.js
```diff
@@ -205,10 +205,6 @@ function collapseMoreDetails() {
  */
 function disableEnableCurrentTrack(activeTrack, currentTime, isSeeked, sectionTitle) {
   let title = sectionTitle;
-  // Return when add to playlist panel is not expanded
-  if($('#addToPlaylistPanel.show').length === 0) {
-    return;
-  }
```
masaball, line 211: Is removing this necessary for the bugfix? 

I know that there was a TypeError generated from this function when an unauthenticated user accesses the media object page, or when the "Add to Playlist" modal otherwise doesn't load. I think this prevents that error from happening? The other changes here do not seem like they would address that error being generated so this would be reintroducing it, I think.

masaball (commented):
The changes to the partial look good to me, but I had a question about the removal from ramp_utils.

app/assets/javascripts/ramp_utils.js
Dananji: Yes, you are right. I had forgotten that. 

The removed code block made the current track disabling and enabling buggy when switching between a section that has tracks and without tracks. I'll see whether there's another way to achieve both these without breaking.

masaball (approved):

PR #5554, opened 2024-01-09
Fix copy playlist modal text size, align playlist show page content

Dananji:
Before             |  After
:-------------------------:|:-------------------------:
Text size - before  |  Text size fix
Aligned - before  |  Aligned


masaball (approved):

PR #5556, opened 2024-01-10
Serialize master file display title to iiif ranges

### LINKED ISSUE #5539: Default Section Label on Item Page Contains Prepended Item Title
# Description
When a section does not have a custom label set or a label within structural XML, the label on the item page should be the filename by default. Currently the title for a section is showing as "Item Title - Filename" by default.

# Example

# Done Looks Like
- [ ] Item title is not prepended to the section label by default for sections without custom labels or structure

masaball:
Closes #5539

### DIFF
--- a/app/models/concerns/master_file_behavior.rb
+++ b/app/models/concerns/master_file_behavior.rb
@@ -112,7 +112,7 @@ module MasterFileBehavior
                elsif file_location.present? && (media_object.master_file_ids.size > 1)
                  file_location.split("/").last
                end
-    mf_title.blank? ? nil : mf_title
+    mf_title.blank? ? media_object.title : mf_title
   end
 
   def embed_title
--- a/app/models/concerns/master_file_behavior.rb
+++ b/app/models/concerns/master_file_behavior.rb
@@ -68,7 +68,7 @@ module MasterFileBehavior
     # Returns the hash
     return({
       id: self.id,
-      label: title,
+      label: display_title,
       is_video: is_video?,
       poster_image: poster_path,
       embed_code: embed_code(EMBED_SIZE[:medium], {urlappend: '/embed'}),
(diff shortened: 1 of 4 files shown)


cjcolvar (dismissed):
This looks fine to me.  

I was a little concerned about changing the value of label in `stream_details` but a quick look through the code didn't reveal any place that was using it besides `IiifCanasPresenter`.  (I did see that https://github.com/avalonmediasystem/avalon/blob/d10ed59e598a929ee02acda6ae8f63779942b42a/app/views/modules/player/_video_js_element.html.erb#L77 isn't using it or `embed_title` but that's out of scope of this PR.)

I'm also wandering if having a nil label in the manifest will cause any problems for ramp.  I thought we may have been including them in other metadata but can't remember.  I don't think that should hold up this PR though if you think it is okay.

app/models/concerns/master_file_behavior.rb
```diff
@@ -116,7 +116,11 @@ def display_title
   end
 
   def embed_title
-    [media_object.title, display_title].compact.join(" - ")
+    [media_object.title, structure_title].compact.join(" - ")
+  end
+
+  def display_title
```
cjcolvar, line 122: For the least change in behavior how about making this method `structure_title` instead and using it in `stream_details`?

masaball: I'm fine with doing that if you think it's best. I switched around the names for clarity, since this method is what is actually getting displayed but it also makes sense to change less code so I can go either way on it.

cjcolvar: I'm just afraid of unintended side effects of changing behavior of public methods, but clean up for clarity is important.  So I could also go either way too.

cjcolvar (approved):

Issue #5461, opened 2023-11-07
[BUG] Add to playlist button gets disabled intermittently


charumitraravi:
**Describe the bug**
The add to playlist button gets disabled intermittently for some of the sections/timelines within the sections within an item
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK091QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6eb850416a6722e48ebef565e4a7d9e28fcbb52f/image.png)

**To Reproduce**
1.Navigate to an item, and using the add to playlist button,  add a section or a track to a playlist
2.Now select a different section/timeline in the same item. 
(or)
Refresh the page and try to select a section/timeline

The "Add to playlist"  button will be disabled

**Expected behavior**
The user should be able to access the "add to playlist" under all scenarios.

**Screenshots/Screen Recording**
[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3l1QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--90ef365a0f3d62774434eec2456546287cb90926/Screen%20Recording%202023-11-07%20at%2010.32.18%20AM.mov)

**Environment (please complete the following information):**
 - Device / platform - Laptop
 - OS: Mac OS
 - Browser - Chrome, Safari
 - 
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
v7.7

charumitraravi:
https://samvera.slack.com/archives/G2E762545/p1699369453007639
Adding the slack conversation here for refrence.

joncameron:
Need to reproduce this and check the console logs to get a better idea of what might be occurring.

elynema:
Can we grab a console error when this is happening? Dananji has not been able to reproduce. @elynema

elynema:
@Dananji I'm attaching another video here that was taken with the console window open in case that helps.
[...]

PR #5559, opened 2024-01-10
Use timeupdate event to enable add to playlist btn when loadedmetadata event misses

Dananji:
Issue: #5461


cjcolvar (approved):
Good catch!

PR #5560, opened 2024-01-10
Only return link html if website url present

### LINKED ISSUE #5340: Link missing a text alternative: website on collections index page
Sites identified: MCO
Page(s) identified:

- Collections index page (display of website URL provided by collection manager)

WCAG compliance level: A

Description:

Links should always have a text alternative. The link text can be inserted into the HTML — it does not need to be visible to pass this check.

Links represented by images or icons can be problematic for visitors who are blind or have low vision.

The text alternative should describe the link's purpose — or what visitors can expect to find after they click it.

See more detail at a [SiteImprove example](https://my2.siteimprove.com/Inspector/1167098/A11Y/Page?pageId=37960016061&impmd=DB21396D67119C3E30B314CCF5B34C39&conf=a+aa+aria+si&wcag=twopointone&decision=dismissed+falsePositive+none#/sia-r11/failed/K2NVS09vQlg=;Y2U3ZmQ3ZjA=)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbXRnIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--b1db21510e69c76eac7249d51063b84687c858fd/image.png)

NOTE: In all the examples provided, there isn't actually a URL for the collection, but the "Website:" header is displaying with a blank space. Something about that is triggering the scanner to expect a URL.

We should also develop either a generic link text for these collection URLs (if we don't already have one), or ask collection managers to enter a link text when adding a URL (or remove the feature altogether!).

# Done Looks Like
- [ ] Website listing should not appear in the markup if no URL is set for a collection's website (resolving the empty value accessibility issue)

cjcolvar:
Resolves #5340

### DIFF
--- a/app/presenters/collection_presenter.rb
+++ b/app/presenters/collection_presenter.rb
@@ -52,7 +52,7 @@ class CollectionPresenter
   end
 
   def website_link
-    view_context.link_to document["website_label_ssi"], document["website_url_ssi"] unless document["website_url_ssi"].nil?
+    view_context.link_to document["website_label_ssi"], document["website_url_ssi"] if document["website_url_ssi"].present?
   end
 
   def as_json(_)
--- a/spec/presenters/collection_presenter_spec.rb
+++ b/spec/presenters/collection_presenter_spec.rb
@@ -86,6 +86,14 @@ describe CollectionPresenter do
         expect(presenter.website_link).to be_nil
       end
     end
+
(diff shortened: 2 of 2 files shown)


Dananji (approved):
:+1:

Issue #5534, opened 2023-12-15
Form field missing a label: Select all button on search results


elynema:
# Description

SiteImprove plugin reports that the "Select all" checkbox at the top of the search results page is missing a label.

Checkbox field<input name="bookmarks_selectall">

This was identified in MCO and mco-staging.

# Done Looks Like
- [ ] Add "Select all" label for screen readers on the Select All checkbox element ( use aria attribute labeled by for this element )

PR #5563, opened 2024-01-10
Fixes for a11y issues identified by SiteImprove browser extension

cjcolvar:
This PR resolves:
- #5534
- Global search box duplicate ids
- Global search box button label
- Size of add to playlist button on item view page larger than other buttons by 1px (not an a11y issue)


Dananji (approved):
:+1:

PR #5569, opened 2024-01-11
Fix copy timeline modal actions

### LINKED ISSUE #5528: [BUG] Copy timeline modal does not refresh after selecting Create/Create and Edit
**Describe the bug**
On the Timelines page that lists all timelines, if you select Copy, and then either Create or Create and Edit, the new timeline is created, but the modal remains open and the page does not refresh. 

**To Reproduce**
1. Go to the Timelines page 
2. Click Copy for a specific timeline
3. Select Create or Create and Edit
4. Nothing appears to happen

**Expected behavior**
- [ ] After selecting Create, the modal should close and the Timelines page should refresh so user can see the new timeline
- [ ] After selecting Create and Edit, the modal should close the Edit page should open for the newly created timeline

 - Device / platform: laptop
 - OS: macOS
 - Browser: Chrome
- [ x] avalon-dev
- [ ] avalon-staging
- [x] mco-staging
- [ ] mco-production
- [x] demo

**Release:**
This problem was reported happening in 7.5 by another Avalon user and I confirmed that it was still happening in 7.6 on the demo site and mco-staging, as well as on avalon-dev.

masaball:
Fixes #5528

The code handling the submit action for copying timelines was looking for attributes from the `data` variable but the copy action did not have that variable populated. ~~Added safe navigation for the error check and changed the redirect link to be pulled from the event variable.~~ Changed the variable assignment to populate from `event.detail`.

### DIFF
--- a/app/assets/javascripts/avalon_timelines/timelines.js.coffee
+++ b/app/assets/javascripts/avalon_timelines/timelines.js.coffee
@@ -65,17 +65,19 @@ $('#copy-timeline-form').submit(
 )
 
 $('#copy-timeline-form').bind('ajax:success',
-  (event, data, status, xhr) ->
-    if (data?.errors)
+  (event) ->
+    [data, status, xhr] = event.detail
+    if (data.errors)
       console.log(data.errors.title[0])
     else
       if (submit_edit)
-        window.location.href = event.detail[0].path
+        window.location.href = data.path
       else
         if ( $('#with_refresh').val() )
           location.reload()
 ).bind('ajax:error',
-  (e, xhr, status, error) ->
+  (event) ->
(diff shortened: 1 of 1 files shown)


mohideen:
@masaball I was working on addressing this on our fork as well. We can replicate fix to playlist copy which was addressed earlier: https://github.com/avalonmediasystem/avalon/commit/650cc2cb373c32cde1c8233234dd65691bebc6c3

masaball:
Good catch. I'd forgotten about/missed that fix. Definitely a better approach. I'm near the end of my day, but I'll update this PR tomorrow. Thanks!

Dananji (approved):
:+1:

PR #5574, opened 2024-01-12
Disable transcript tab when there are no transcripts

masaball:
Original implementation only hid the transcript tab. This resulted in a user being able to navigate to the tab content via arrow key navigation. We can prevent this by adding `disabled` to the class of the tab.


cjcolvar (approved):
This looks good to me, but I'm wondering if a screen reader will still be trying to read the transcript tag and we might need to set `aria-disabled` or `aria-hidden`.  I'm not quite sure how to test this unless the SiteImprove browser extension would flag it due to the inconsistency.

masaball:
Browserstack has a screen reader on at least the chrome browsers. Just tested there and the screen reader did not read out the tab info with just the hidden and disabled flags provided and siteimprove browser plugin did not flag anything.  So I think we are good accessibility wise.

PR #5578, opened 2024-01-16
Fix "Create Timeline" actions

masaball:
There were three separate bugs:
1. Items without structure were not assigning any value to `streamID`, preventing the proper creation route url being generated.
2. The if/else block for handling custom scopes was not successfully assigning values to the `id` variable when a custom scope was selected.
3. Custom scopes were not converting into the right time format for timeliner to parse, so the timeline would be created but would not be limited to the time range defined by the custom scope.


Dananji (approved):
:+1:

PR #5583, opened 2024-01-17
Fix files tab height alternating between unset and fit-content

Dananji:
Fix for the following UI bug;

[files_tab_height.webm](https://github.com/avalonmediasystem/avalon/assets/1331659/76fb1278-fb95-4f3a-8e07-c662e00ba4a4)


masaball (approved):
Looks good

PR #5587, opened 2024-01-19
Display playlist item comment on playlist page under title

### LINKED ISSUE #5562: Link back to source item from Source Item Details on playlist page
# Description

Currently, MCO provides a link back to the source item and section under Source Item Details on the playlist page. In the new playlist page design, this is not included in the Source Item Details, and there is no way to link directly back to the source item.

Avalon 7.6 includes both the masterfile and media object, for example:
ABY9017A from Candide / Leonard Bernstein ; [based on the book by Voltaire ; lyrics by Richard Wilbur ; with additional lyrics by John La Touche ... et al. ; orchestrations by Leonard Bernstein and Hershy Kay]. 

For 7.7, having just a link to the object page is sufficient.

# Question
In the current playlist manifests, the `partOf` URL value contains the /manifest route. Does this need to be retained, or can it hold just the direct URL to the media object view route? Or would the `homepage` attribute be better to use?

# Done Looks Like
- [ ] Item title in the Source Item Details panel is a hyperlink to the object view page

### LINKED ISSUE #5581: Include playlist item descriptions in IIIF manifest for playlists
# Description

When adding a track or custom timespan to a playlist, the add to playlist form provides the user the option to create a description for the playlist item. This description is appearing in the IIIF playlist manifest (under the metadata element, under "Description" (see https://avalon-dev.dlib.indiana.edu/playlists/33/manifest.json) but is not being displayed by Ramp.

Here is what a playlist item description looks like in MCO currently:

It displays immediately below the playlist item title below the player.

The Avalon component is already rendering the playlist item title on this page but when the playlist item is loaded, the playlist item title is not read from the manifest but from the title value stored in the DOM. So the playlist is not being read at that point and this might be trickier. But the description will need to get passed into the structure or DOM via data-attributes if the description value is in the manifest.

# Done Looks Like
- [x] Playlist item description is present in the manifest 
  - [x] (value written into the `summary` element of the playlist item's range (or canvas—either would be OK and canvas would be simpler with the iiif-manifest gem) in the structure element of the manifest) *or*
  - [ ] "Playlist Item Description" keypair is written into the manifest under the metadata for a Canvas and is treated in a special way
- [x] Playlist item descriptions written into the DOM as data-attribute values 
- [x] Playlist description value is rendered below the playlist item title on the playlist view page

# Implementation
Which IIIF field to use?
- summary (https://iiif.io/api/presentation/3.0/#summary)
- within the existing `metadata` listing (drawing from the record's descriptive metadata)
  - "Description" or some other specifically used field in the metadata could be keyed on to provide this, i.e. "Playlist Item Description"
- within an annotation

cjcolvar:
Resolves #5581
Also resolves #5562


cjcolvar:
Replaced by #5617

PR #5599, opened 2024-01-25
Include original supplemental filenames in iiif manifest

masaball:
This PR adds an entry to the label field for caption and transcript files so that the original filename can be serialized into the manifest. We serialize the user-defined label into the manifest first, using the language as its key. Then we serialize the original filename without a key.

This will allow us to differentiate the names to facilitate better file handling within the Ramp file downloader.


cjcolvar (approved):
Looks good!

PR #5611, opened 2024-01-30
Force embedded player to render within container

masaball:
VideoJS fluid mode forces the player to a specific aspect ratio. When the container that the player is embedded in does not match that aspect ratio it generates scroll bars for the content. Turning off fluid mode keeps the video within the container, letter/pillarboxing as necessary. 

This PR also turns off responsive mode, as in testing it did not seem to have any affect on the rendering.


Dananji (approved):
👍

Issue #5594, opened 2024-01-25
[BUG] Skip transcoding conversion jobs fail


elynema:
# Description

When attempting to upload a file for a new section and checking the "skip transcoding" box, the conversion job fails.

Example: https://avalon-dev.dlib.indiana.edu/encode_records/758
```
Status: Failed

Adapter: ActiveEncode::EngineAdapters::PassThroughAdapter

Adapter Job ID: ea361b7a-1dd2-4078-a08f-136b0c85a426

Title:

Error Message: /tmp/amp-demo-amia-2023.mp420240125-8-882k4z does not exist or is not accessible

Job Started: 01/25/24 04:25:17 PM

```

# Done Looks Like
- [ ] Files are processed successfully when the "Skip Transcoding" option is selected during web and batch upload

elynema:
Skip transcoding is working properly on mco-staging but failing with the same error on avalon-staging, so it is either something due to the container environment or the 7.7 release.

elynema:
Chris looked into this a bit, and it is related to the minio configuration, but the fix is probably going to be a bit complex and is probably a longer-standing issue. Jon C. confirmed this is happening on the avalon demo site as well, which is running 7.6. It is probably ok to list this as a known issue and then fix it in the next release (or put out a patch if someone needs a fix sooner).

elynema:
Added this to known issues page, but not sure I got the wording exactly right. We would like to fix this in 7.8/8.0 release.

cjcolvar:
I ended up working on this more and came up with a solution that involves changes to `active_encode` and avalon.  Do we want to try to roll that into this release or should I write it up and we can bring it in first thing in 7.8?

elynema:
@cjcolvar @joncameron  I tested batch ingest with skip transcoding using our [test batch ingest fixture](https://drive.google.com/file/d/1b7ApYAWrUSk18UrDctEPd6FIFYVV_0Fu/view?usp=sharing) on avalon-dev. The manifest includes a file that is ingested with skip transcoding = 'yes' but has 3 different quality derivative files in the asset folder. While the other 2 worked (1 with skip transcoding on a single file and 1 without skip transcoding), this one failed with this error:
`
undefined method `close' for #<Addressable::URI:0x216038 URI:s3://masterfiles/dropbox/7_7_regression_test/assets/OrganClip.low.mp4>`

The file is definitely sitting there in minio; could this be an issue with this fix?

[...]

PR #5615, opened 2024-01-31
Fix skip transcoding with minio/s3 file

cjcolvar:
The pass through encode model previously copied the s3 content into a local tempfile and passed that to active_encode to workaround the issue of mediainfo not being able to read s3 urls.  That may have worked at one point but now the tempfiles get deleted/garbage collected after one read so mediainfo always returns an empty xml file leading to failed encodes.  This PR resolves this by making changes to the PassThroughAdapter in active_encode and then relying on that instead of having a workaround locally.
https://github.com/samvera-labs/active_encode/pull/132

This PR also resolves an issue with FileLocator::S3File that I discovered while debugging this issue locally.  The `single_request` mode is not as efficient as a multi-part download but was the only way I could get it to work with minio.

Related to #5594

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,7 +74,7 @@ gem 'omniauth-lti', git: "https://github.com/avalonmediasystem/omniauth-lti.git"
 gem "omniauth-saml", "~> 2.0"
 
 # Media Access & Transcoding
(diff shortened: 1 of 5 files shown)


masaball (dismissed):
Looks good

Issue #5589, opened 2024-01-23
Playlist Page - Structured Nav Extends to Parent Container Height


joncameron:
# Description
On the playlist page the structured-nav Ramp element is given a max height value, which crops it unnecessarily within the parent element on the item page (see screenshot). This should change in the same way it changed on the item page (#5498).

# Done Looks Like
- [ ] Height of structured-nav div (.ramp--structured-nav) extends to the full height of the parent element

# Screenshot
<img width="571" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1QrQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cfbad8bce21f5cfcae70350b219788158dc78217/Screenshot%202024-01-23%20at%203.58.50%E2%80%AFPM.png" alt="Screenshot 2024-01-23 at 3.58.50 PM.png" />

elynema:
@Dananji In Chrome and Firefox, I see that the playlist item container is the height of the page and scrollable:

<img width="1288" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGNnQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e4572ede2ee5be87707beb7b65598d53214f38a1/image.png" alt="image.png" />


In Safari, no scroll is getting engaged within the playlist item container, so it just gets really long:

<img width="1279" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGdnQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7300f46a799b733af293fee1b8a45e9044727010/image.png" alt="image.png" />

In Chrome and Firefox where it is scrolling, I'm not seeing the "scroll to see more" styling; maybe that has not been merged yet.

elynema:
Firefox on Android is not expanding the size of the structured navigation on the playlist page. This is low priority; perhaps I should create a new bug.

[...]

PR #5623, opened 2024-02-02
Extend playlist items container height fully on mobile devices

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5589

### DIFF
--- a/app/javascript/components/PlaylistRamp.jsx
+++ b/app/javascript/components/PlaylistRamp.jsx
@@ -50,6 +50,9 @@ const Ramp = ({
 
   let interval;
 
+  const USER_AGENT = window.navigator.userAgent;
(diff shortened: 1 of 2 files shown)


app/javascript/components/PlaylistRamp.jsx
```diff
@@ -50,6 +50,9 @@ const Ramp = ({
 
   let interval;
 
+  const USER_AGENT = window.navigator.userAgent;
+  const IS_MOBILE = (/Mobi/i).test(USER_AGENT);
```
cjcolvar, line 54: Would it make sense to import Ramp's mobile checking and use it here?

cjcolvar (approved):
This looks fine enough as is.  My question doesn't need to block merging.

Issue #5586, opened 2024-01-19
Timeline Listing Search Not Working as Expected


joncameron:
# Description
Timeline searching using the search field on the Timelines page isn't returning results as expected. It seems to be very strict in the matching conditions, resulting in a poor user experience since users will not expect that the default will be exact string matching with case sensitivity.

# Examples
Searching `george` doesn't return `George Walker, Piano Sonata No. 1, 3rd movement`
`op 35` doesn't return `Cecile Chaminade, Op. 35 No. 1: Scherzo`
`Op 35` doesn't return `Cecile Chaminade, Op. 35 No. 1: Scherzo`

# Done Looks Like
- [x] Searching on the Timelines page is not case sensitive
- [x] Break query on space in the input, so SQL query searching returns multiple values
- [x] Other changes for fuzzier matching as available (such as ignoring punctuation, word order) investigated
- [ ] More complicated changes can be broken into a new ticket

joncameron:
Avalon is being sent a DB query and returning the result to DataTables, so this is ultimately a SQL query underneath.

joncameron:
Avalon is being sent a DB query and returning the result to DataTables, so this is ultimately a SQL query underneath.

masaball:
We can split on spaces in the input with no trouble and construct queries that do not rely on word order. However, generating a truly fuzzy query is difficult. This first query splits on spaces and then searches  for each term in the resulting array. This returns too many results though, as a search term like "using a lot of words" would return any entry with a Description containing the letter "a". So that search would return a timeline with description "Meaningful times" because meaningful has an 'a' in it. This feels unclear because a user is not going to know why that timeline is being returned from that search. The standard SQL wildcards are `%` and `_`, representing '0 or more of any character' and 'exactly one of any character', respectively. Certain SQL brands have more options but we are limited to just those two wildcards so cannot make a more granular match case.

`scope :desc_like, ->(desc_filter) { where("LOWER(description) LIKE ANY(array[?])", desc_filter.split.map { |df| "%#{df.downcase}%" }) }`

[...]

PR #5627, opened 2024-02-02
Split query into terms, search each individually, and return intersecion

cjcolvar:
Follow on to #5606
Possible solution to #5586

This approach makes a query like:
`WHERE LOWER(title) LIKE %op% AND LOWER(title) LIKE %35%`

### DIFF
--- a/spec/models/timeline_spec.rb
+++ b/spec/models/timeline_spec.rb
@@ -163,6 +163,14 @@ RSpec.describe Timeline, type: :model do
       it 'does not return timelines without matching titles' do
         expect(Timeline.title_like(title_filter)).not_to include(timeline3)
       end
+      it 'searches on multiple terms' do
+        expect(Timeline.title_like('Fav Moose')).to include(timeline2)
+        expect(Timeline.title_like('Fav Moose')).not_to include(timeline1)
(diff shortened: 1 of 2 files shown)


spec/models/timeline_spec.rb
```diff
@@ -163,6 +163,9 @@
       it 'does not return timelines without matching titles' do
         expect(Timeline.title_like(title_filter)).not_to include(timeline3)
       end
+      it 'searches on multiple terms' do
+        expect(Timeline.title_like('Fav Moose')).to include(timeline2)
+      end
```
masaball, line 168: Not a blocker but, since I think the expected behavior would be that it only returns timeline 2, should we have negative tests as well to make sure that timeline 1 and 3 are not included in `title_like`? Or do I have the expected behavior wrong? 

Also, we can maybe see what the behavior is in testing, but I think these changes may not care about word order? If that's the case, and it is the behavior we want to maintain in the future, it may be good to add a check that something like "Moose Fav" also returns timeline 2.

masaball (dismissed):
Looks good to me. Feel free to merge without adding additional tests.

masaball (approved):

PR #5630, opened 2024-02-05
Delay localize_times until Ramp has loaded

masaball:
The function to generate the return time was running on `$(document).ready`, which triggers before Ramp has fully loaded. Since the message is displayed within Ramp this timing issue caused no time to be displayed.


Dananji (approved):
Looks good. Once the `console.log` line in the `./localize_times.js` file is removed, I think it can be merged 👍

Dananji (approved):

PR #5635, opened 2024-02-06
Use playlist token for manifest requests

cjcolvar:
Pass token into PlaylistRamp component and use it to make manifest requests.  On the Avalon side use playlist token for authentication on manifest requests.


spec/controllers/playlists_controller_spec.rb
```diff
@@ -124,6 +126,8 @@
         it "should return the playlist view page" do
```
masaball, line 126:
```suggestion
        it "should return the playlist view page and manifest" do
```

masaball (dismissed):
Looks good, just a suggestion for clarification of test description.

masaball (approved):

PR #5639, opened 2024-02-07
Use safe navigator on activeElement[0] methods

masaball:
The console error seen with playlists was related to the activeElement variable briefly becoming undefined after a state update, breaking the second state update.


Dananji (approved):
👍

PR #5664, opened 2024-02-14
Use terser instead of uglifier

cjcolvar:
Use terser instead of uglifier for JS processing during asset precompilation due to uglifier not supporting newer versions of ECMAScript


Dananji (approved):
👍

PR #5671, opened 2024-02-16
Add rake task to migrate IndexedFile captions to ActiveStorage

masaball:
The original implementation for captions was a single IndexedFile attached to the master file. When we changed to ActiveStorage to support upload of multiple captions, we removed the controller routes and frontend code related to displaying, creating, and deleting the legacy captions. This migration moves the legacy captions to the new caption format so that users will be able to maintain access to, and control of, these captions.

### DIFF
--- a/lib/tasks/avalon_migrations.rake
+++ b/lib/tasks/avalon_migrations.rake
@@ -26,29 +26,24 @@ namespace :avalon do
     desc "Migrate legacy IndexedFile captions to ActiveStorage as part of supporting upload of multiple captions files"
     task caption_files: :environment do
       count = 0
-      MasterFile.all.each do |master_file|
-        next unless master_file.captions.present?
-        # Retrieve original file metadata from IndexedFile
-        filename = master_file.captions.original_name
-        content_type = master_file.captions.mime_type
-        # Legacy captions were not stored as an attached file.
-        # Write the plaintext content into a Tempfile to load into ActiveStorage.
-        file = Tempfile.new(filename)
-        file.write(master_file.captions.content)
-        file.close
-        # Create new SupplementalFile record
+      # Iterate through all caption IndexedFiles
+      IndexedFile.where("id: */captions").each do |caption_file|
+        # Retrieve parent master file
+        master_file_id = caption_file.id.split('/').first
+        master_file = MasterFile.find(master_file_id) rescue nil
+        next unless master_file
+        # Grab original file metadata from IndexedFile
+        filename = caption_file.original_name
+        content_type = caption_file.mime_type
+        # Create and populate new SupplementalFile record using original metadata
         supplemental_file = SupplementalFile.new(label: filename, tags: ['caption'], language: 'eng')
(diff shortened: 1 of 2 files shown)


lib/tasks/avalon_migrations.rake
```diff
@@ -23,5 +23,38 @@ namespace :avalon do
...
+        content_type = master_file.captions.mime_type
+        # Legacy captions were not stored as an attached file.
+        # Write the plaintext content into a Tempfile to load into ActiveStorage.
+        file = Tempfile.new(filename)
+        file.write(master_file.captions.content)
+        file.close
+        # Create new SupplementalFile record
+        supplemental_file = SupplementalFile.new(label: filename, tags: ['caption'], language: 'eng')
+        # Attach file to SupplementalFile using original metadata
+        supplemental_file.file.attach(io: File.open(file.path), filename: filename, content_type: content_type, identify: false)
```
cjcolvar, line 42: It looks like ActiveFedora provides an [IO wrapper around files](https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/file_io.rb) that avoids needing a tempfile.
```suggestion
        supplemental_file.file.attach(io: ActiveFedora::FileIO.new(master_file.captions), filename: filename, content_type: content_type, identify: false)
```

cjcolvar (commented):
Overall this looks good.  I had a couple ideas for making it more efficient.

lib/tasks/avalon_migrations.rake
```diff
@@ -23,5 +23,33 @@ namespace :avalon do
...
     end
+    desc "Migrate legacy IndexedFile captions to ActiveStorage as part of supporting upload of multiple captions files"
+    task caption_files: :environment do
+      count = 0
+      # Iterate through all caption IndexedFiles
+      IndexedFile.where("id: */captions").each do |caption_file|
+        # Retrieve parent master file
+        master_file_id = caption_file.id.split('/').first
+        master_file = MasterFile.find(master_file_id) rescue nil
+        next unless master_file
```
cjcolvar, line 34: Maybe we also need the check that the content isn't empty?  I'm wondering if captions files that have already been migrated by this rake task would cause problems or create a bunch of empty supplemental files.  Have you run this locally after this change and seen what happens?

masaball: Yep, good call. I had run the task once after these changes and everything looked fine but just ran the task again and it did create an empty SupplementalFile. Will add a check that there is actually a file to copy.

cjcolvar: Is there a way we can destroy the captions IndexedFile object?  That would be nicer but that might be overthinking it and this is probably good enough.

masaball: I will see. I copied how the master_file controller was setup in its `delete_captions` method, but I can do some testing for a better method. Probably best practice, if possible, to have both the check for a populated IndexedFile and to fully delete them at the end.

cjcolvar (approved):
Looks great!

PR #5676, opened 2024-02-16
Cast to arrays to avoid nils

cjcolvar:
This fixes the case when the solr document doesn't have the `edit_access_person_ssim` field and similar cases.  Without this fix an error would occur when calling `editors` because `nil - managers` raises a `NoMethodError`.


masaball (approved):
Looks good

Issue #5685, opened 2024-02-19
[BUG] Create timeline button does not always enable on page refresh


elynema:
# Description

When loading a refreshing a media item page, the Create Timeline button fairly frequently remains disabled. I have not observed this behavior in avalon-dev, but am seeing it fairly frequently in avalon-staging.

<img width="1479" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOEJ2QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f5eddb1dc1b17f20e13ad2dff625a3ac6bd25225/image.png" alt="image.png" />

Example item: https://avalon-staging.dlib.indiana.edu/media_objects/v405s9361
Another example: https://avalon-staging.dlib.indiana.edu/media_objects/ns0646018

There does appear to be a JS error in the console on avalon-staging; not sure if it is related since it shows up regardless of whether Create timeline button gets enabled, but I don't see it on avalon-dev.

`Unsafe attempt to load URL https://avalon-staging-minio.dlib.indiana.edu/supplementalfiles/5q4qgso3gh8qzkrgl5s53is9g1b2?response-content-disposition=attachment%3B%20filename%3D%22s7526w20z-VolcanoNaturesInferno.vtt%22%3B%20filename%2A%3DUTF-8%27%27s7526w20z-VolcanoNaturesInferno.vtt&response-content-type=text%2Fvtt&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minio%2F20240219%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20240219T223301Z&X-Amz-Expires=300&X-Amz-SignedHeaders=host&X-Amz-Signature=d36793684e6691fc67ee8bbe6a00a057f3cb898da35b4f7009f181ed04f4eb52 from frame with URL https://avalon-staging.dlib.indiana.edu/media_objects/ns0646018. Domains, protocols and ports must match.`

joncameron:
This appears to be working on avalon-dev; I couldn't reproduce there now. @cjcolvar could you add this to staging so I can test there too?

joncameron:
Looks good! Could not reproduce with either example item on staging.

PR #5687, opened 2024-02-20
Enable create timeline and create thumbnail buttons on timeout

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5685

I put the same fix to `Create Thumbnail` button in addition to `Create Timeline`, since they both behave the same way in this issue.


cjcolvar (commented):
Does this enable the buttons after half a second regardless of if the player is loaded or not?  If so then I wonder if it might enable the buttons incorrectly in some cases which might lead to other problems if the user tries to use them.

Dananji:
Yes this would enable buttons regardless what the player status is. This is similar to add to playlist button behavior we put in a while ago.
I did have a check for the player readyState in this new code, but removed it thinking it might re-introduce the same issue in a different way. Should I put it back in?

With the player status check the check in the code would look as follows;

```
if (timelineBtn && timelineBtn.disabled && player.player?.readyState() === 4) {
....
```

cjcolvar:
If this mimics the behavior of the add to playlist button then I think it makes sense to do the same for these buttons and then take a look later at adding that check if there are problems reported.

cjcolvar (approved):

Issue #5691, opened 2024-02-20
[BUG] Add to Playlist and Create Timeline enabled during Item Conversion is in Progress


Dananji:
**Describe the bug**

When a media file is being processed the item view page shows enabled buttons for create timeline, create thumbnail, and add to playlist actions even though these can't perform the intended functionality without a fully loaded player on the page.
When this happens garbage values are displayed in the forms for create timeline and add to playlist actions.

![Screenshot 2024-02-20 at 1 33 37 PM](https://github.com/avalonmediasystem/avalon/assets/1331659/997628ea-884a-488a-927b-5d6534ae5ca9)

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Add a section to an existing item or ingest a new item
2. Go to the item preview, observe the create timeline, create thumbnail, and add to playlist action buttons are enabled
3. Click on either `Create Timeline` or `Add to Playlist` button to get the forms
4. Observe the forms are automatically filled with garbage values

**Expected behavior**
These action buttons are enabled and functional only when there's playable media on the page.

**Screenshots/Screen Recording**

![Screenshot 2024-02-20 at 1 33 52 PM](https://github.com/avalonmediasystem/avalon/assets/1331659/b619829f-1267-42cd-9c8a-26fd1366066d)

![Screenshot 2024-02-20 at 1 34 26 PM](https://github.com/avalonmediasystem/avalon/assets/1331659/ce4841ad-ebbe-4696-800e-5b5bfba0e6cf)

![Screenshot 2024-02-20 at 1 34 08 PM](https://github.com/avalonmediasystem/avalon/assets/1331659/ced1b64d-d0d9-4c8e-a08c-596b6a541bbf)

**Environment (please complete the following information):**
 - Device / platform [e.g. laptop, iphone pro 12]
 - Browser [e.g. chrome version 90.0.4430.93]
- [ ] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Is this issue associated with a specific release? Avalon 7.7

**Additional notes**
Related previous conversation: https://github.com/avalonmediasystem/avalon/pull/5687#pullrequestreview-1890887389

Dananji:
Related work in progress: https://github.com/avalonmediasystem/avalon/pull/5692

joncameron:
[...]

PR #5692, opened 2024-02-20
Enable action buttons on item page only when player is fully loaded

Dananji:
> Part of the work done in https://github.com/avalonmediasystem/avalon/pull/5687. This was not added in the previous PR, as this will need thorough testing.
> The code needs to be cleaned up as needed and handle all possible use-cases.

Tested in different browsers on desktop and Android. I couldn't test this on iOS devices as streaming doesn't work in iOS in my dev environment.

Action buttons are enabled with playable media on page:

item with media

Action buttons are disabled without playable media on page:

item w:o media

Related issue: https://github.com/avalonmediasystem/avalon/issues/5691


cjcolvar (approved):
Looks great!

PR #5694, opened 2024-02-21
Support positional URLs for playlist items

masaball:
~~Wait for iiif-manifest [#110](https://github.com/samvera/iiif_manifest/pull/110) to be merged and update Gemfile to point to iiif-manifest main before merging.~~


cjcolvar (approved):
Looks great!

PR #5696, opened 2024-02-21
Use website_url as label when website_label is not present

masaball:
This PR also includes an update to nestable.scss to fix some warnings from autoprefixer encountered when running the spec tests.


Dananji (approved):
👍

PR #5698, opened 2024-02-22
Set event listeners once for playlist modal

masaball:
We were encountering an issue where adding a new playlist from the dropdown menu would insert the new playlist's name into the list multiple times. This was caused by the `.add_new_playlist_option()` function being triggered every time the Add to Playlist menu was opened. This caused the event listener that updates the dropdown list to be added every time the menu opened, resulting in the duplication we were seeing. Moving the function call out of the parent event listener so it only runs once fixes the issue.


Dananji (approved):
👍

PR #5699, opened 2024-02-26
Add collection_id to media_object JSON response

masaball:


joncameron:
@masaball I don't see the collection_id in JSON requests yet, for example https://avalon-dev.dlib.indiana.edu/media_objects/1v53jx019.json. Maybe dev needs to be updated with this change yet?

masaball:
Weird. Github actions says the build was successful but I just checked the code in the container on dev and it is missing these changes. Also looks like it is missing @Dananji's most recent couple PRs as well. Seems like something in the deploy process broke, maybe? I don't remember how to trigger a deploy manually and don't know the automated deploy setup to try to fix that, so this probably needs to wait for @cjcolvar to take a look at.

cjcolvar:
I fixed avalon-dev and I'm seeing collection_id in the JSON responses now.

PR #5701, opened 2024-02-28
Fixes and improvements to the reindexing script

cjcolvar:


masaball (approved):

PR #5704, opened 2024-03-04
Support creation of multiple captions in batch ingest

masaball:
~~Converting PR to draft. As I was working on documentation I realized that I had not tested the caption ingest when a single item was having multiple media files uploaded to it. As this is currently implemented, the caption files will likely not be assigned properly in that case. Will test and adjust the approach as necessary.~~


app/jobs/ingest_batch_entry_job.rb
```diff
@@ -69,6 +69,7 @@ def process_success(batch_entry, entry)
       old_media_object_id = batch_entry.media_object_pid
       batch_entry.media_object_pid = entry.media_object.id
       batch_entry.complete = true
+      batch_entry.current_status = 'Completed'
```
cjcolvar, line 72: Nice catch!

lib/avalon/batch/entry.rb
```diff
@@ -176,17 +176,23 @@ def self.offset_valid?( offset )
...
           master_file.structuralMetadata.content=FileLocator.new(structural_file).reader
           master_file.structuralMetadata.original_name = structural_file
         end
-        captions_file = "#{filename}.vtt"
-        if FileLocator.new(captions_file).exist?
-          master_file.captions.content=FileLocator.new(captions_file).reader
-          master_file.captions.mime_type='text/vtt'
-          master_file.captions.original_name = captions_file
+        captions.each do |c|
+          return unless c.present?
```
cjcolvar, line 186: This appears to skip any subsequent captions if there is a nil in the list.  What do you think about skipping instead?
```suggestion
          next unless c.present?
```

cjcolvar: You could also probably combine this line with the next.
```suggestion
          next unless c.present? && c[:caption_file].present? && FileLocator.new(c[:caption_file]).exist?
```

cjcolvar (commented):
Looks good!

lib/avalon/batch/entry.rb
masaball: Yeah, that should definitely be a `next`. In the moment I forgot that early return breaks out of loops. 

And moving the following `if` statement into the guard clause makes sense :+1:

cjcolvar (approved):

PR #5713, opened 2024-03-05
Optimize collection media object counts

### LINKED ISSUE #5706: Manage Content Page Is Slow To Return for admins
# Description
The Manage Content page is taking a long time in 7.7 on mco-staging (~45 seconds). This user has about 350 collections on the Manage Content page. In 7.6 on MCO, this takes only a few seconds, so this seems to be regression.

In this scenario, it looks like the Manage Collection page is making a solr request per collection which adds up with 536 collections. Perhaps there is an optimization that was missed when we cleaned up  requests out to Solr/Fedora?

# Done Looks Like
- [ ] Consolidate Solr calls to speed return of collection listing

cjcolvar:
This PR adds two solr queries to gather media object counts through facet counts instead of two solr queries per collection.

I was hoping to use a subquery document transformer to inject the counts into the collection documents returned by the current query but I wasn't able to get it working.  Something like:
```
fl=*,media_objects:[subquery],unpublished:[subquery]&media_objects.q={!term f=isMemberOfCollection_ssim v=$row.id}&media_objects.rows=0&unpublished.q={!term f=isMemberOfCollection_ssim v=$row.id}workflow_published_sim:Unpublished&unpublished.rows=0"
```

Resolves #5706

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -36,7 +36,18 @@ class Admin::CollectionsController < ApplicationController
       builder.user = user
     end
     response = repository.search(builder)
-    @collections = response.documents.collect { |doc| ::Admin::CollectionPresenter.new(doc) }.sort_by { |c| c.name.downcase }
+
+    # Query solr for facet values for collection media object counts and pass into presenter to avoid making 2 solr queries per collection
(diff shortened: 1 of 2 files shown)


masaball (dismissed):
Looks good!

masaball (approved):

PR #5718, opened 2024-03-07
Set content_type explicitly based on file extension

masaball:
SRT files are not guaranteed to be created with the 'text/srt' mime type, breaking validation when a user attempts to upload one as a caption file. Looking up the mime type by the file extension and explicitly setting the `content_type` attribute from that should fix the validation of these files.


app/models/supplemental_file.rb
```diff
@@ -28,6 +28,9 @@ def validate_file_type
 
   def attach_file(new_file)
     file.attach(new_file)
+    extension = File.extname(new_file.original_filename)
+    content_type = Mime::Type.lookup_by_extension(extension.slice(1..-1)).to_s
+    self.file.content_type = content_type
```
cjcolvar, line 33: I worry about putting too much reliance on the file extension.  What happens here if the file doesn't have an extension or a bad one?
It looks like ActiveStorage will try to autodetect a file's mime type and thus prepopulate `file.content_type`, but there is an optional param to `attach` to provide a fallback if ActiveStorage can't determine it. (https://edgeguides.rubyonrails.org/active_storage_overview.html#attaching-file-io-objects)  Maybe that would be a safer route?

masaball: The issue with SRT files was not that ActiveStorage could not determine the mime-type, it was that ActiveStorage sees the mime-type of SRT as 'text/plain'. But you make a good point. Since, as far as we know, the only supported file type affected by this are the SRT files we can limit looking up the mime-type and explicitly setting it to when the extension is `.srt`. Does that seem acceptable?

cjcolvar: Sure, that seems to narrow the scope of the change well.

PR #5719, opened 2024-03-08
Convert SRT to VTT captions for cosumption into Video.js

masaball:
~~This branch is built off the work in #5718, so that should be merged first or closed in favor of this PR.~~

### DIFF
--- a/app/models/supplemental_file.rb
+++ b/app/models/supplemental_file.rb
@@ -52,13 +52,13 @@ class SupplementalFile < ApplicationRecord
     # This Regex looks for malformed time stamp pieces such as '00:1:00,000', '0:01:00,000', etc.
     # When it finds a match it prepends a 0 to the capture group so both of the above examples 
     # would return '00:01:00,000'
-    srt.gsub!(/(:|^)(\d)(,|:)/, '\10\2\3')
+    conversion = srt.gsub(/(:|^)(\d)(,|:)/, '\10\2\3')
     # convert timestamps and save the file
     # VTT uses '.' as its decimal separator, SRT uses ',' so we convert the punctuation
-    srt.gsub!(/([0-9]{2}:[0-9]{2}:[0-9]{2})([,])([0-9]{3})/, '\1.\3')
+    conversion.gsub!(/([0-9]{2}:[0-9]{2}:[0-9]{2})([,])([0-9]{3})/, '\1.\3')
     # normalize new line character
-    srt.gsub!("\r\n", "\n")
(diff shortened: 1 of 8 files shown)


app/models/supplemental_file.rb
```diff
@@ -43,4 +46,20 @@ def caption?
...
+    # When it finds a match it prepends a 0 to the capture group so both of the above examples 
+    # would return '00:01:00,000'
+    srt.gsub!(/(:|^)(\d)(,|:)/, '\10\2\3')
+    # convert timestamps and save the file
+    # VTT uses '.' as its decimal separator, SRT uses ',' so we convert the punctuation
+    srt.gsub!(/([0-9]{2}:[0-9]{2}:[0-9]{2})([,])([0-9]{3})/, '\1.\3')
+    # normalize new line character
+    srt.gsub!("\r\n", "\n")
+
+    "WEBVTT\n\n#{srt}".strip
```
cjcolvar, line 63: These lines modify the string that is passed in  but it doesn't follow the same modify in place approach for the last transformation.  Without following the convention all the way through and naming the method `convert_from_srt!` the caller won't know that there is a side effect.  I'm not sure if modification in place is more efficient (I suspect it might actually be less efficient in this case) so what do you think about getting rid of the `gsub!` in favor of `gsub` ?

spec/models/supplemental_file_spec.rb
```diff
@@ -84,4 +90,12 @@
       expect(subject.reload.language).to eq "ger"
     end
   end
+
+  describe '.convert_from_srt' do
+    let(:file) { Rails.root.join('spec', 'fixtures', 'captions.srt') }
+    let(:output) { "WEBVTT\n\n1\n00:00:03.498 --> 00:00:05.000\n- Example Captions" }
+    it 'converts SRT format captions into VTT captions' do
+      expect(SupplementalFile.convert_from_srt(File.read(file))).to eq output
+    end
```
cjcolvar, line 99: What do you think about inlining the SRT file?  I think it makes the test easier to read having everything in one place.
```suggestion
    let(:input) { "1\n00:00:03.498 --> 00:00:05.000\n- Example Captions\n"  }
    let(:output) { "WEBVTT\n\n1\n00:00:03.498 --> 00:00:05.000\n- Example Captions" }
    it 'converts SRT format captions into VTT captions' do
      expect(SupplementalFile.convert_from_srt(input)).to eq output
    end
```

cjcolvar (approved):
Looks great!

Issue #5722, opened 2024-03-12
Remove Caption File References from HLS Manifest


joncameron:
# Description
See https://github.com/avalonmediasystem/avalon/issues/5695 for details on why captions are being removed from the HLS manifest. In short, having captions in the HLS manifest and also delivered via IIIF manifest means ambiguity and difficulty in the media player environment across browsers. 

# Done Looks Like
- [ ] Caption references do not appear in Avalon's HLS manifests

masaball:
I was unable to test with my local system because I could not get Browserstack iOS to work with localhost. From testing on desktop, the current Ramp build in Dev is removing the non-HLS captions from the track listing so when the player instantiates, the captions button is present for a split second and then disappears. I think this is because we have not brought in a build containing samvera-labs/ramp#428 yet. Once that is brought in, captions should display as expected.

Captions in the embedded player were working as expected for desktop, but I could not test on iOS for the same localhost issues already mentioned.

elynema:
Once the Ramp build with #428 is brought in, this can be tested on avalon-dev.

Issue #5695, opened 2024-02-21
Removing VTT Files From HLS Manifests


joncameron:
# Description
One point of investigation in fixing the double caption issue in Apple's media player is removing the VTT files from the HLS manifests we generate. If we're only going to be using file references outside of the context (like the player putting everything together from the IIIF manifest), we could instead be supplying the text tracks to the native player outside of the HLS manifest. But, Video.JS may not give us the tools to interact with the native player in a more granular way.

# Done Looks Like
- [ ] Investigation into removing VTT file references from HLS manifest

# Questions
- Will removing this fix the double caption display issue? (#5650)
- Will multiple captions show up in native player? (https://github.com/samvera-labs/ramp/issues/405)
- Will this cause captions to not be available in the native player?
- Is changing CSS for the native player to hide overlay captions there an option?
- Does this solution work for embedded player?

joncameron:
Is changing CSS for the native player to hide overlay captions there an option?

Dananji:
1. [Sample IIIF Manifest](https://gist.githubusercontent.com/Dananji/a3ec009d4922b772962498173b0a5aca/raw/a18480cee491eb6b2a13963c35b3a34ca3a6e399/distinct-captions.json) with an [HLS stream](https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p/section/c821gj76b) without captions from `avalon-dev` with a `supplementing` annotation with the following Video.js options, in Chrome, iPad iOS 17: 
```
...
html5: {
  nativeTextTracks: IS_MOBILE && !IS_ANDROID
}
```

|  | Ramp player | Native player |
| ------------- | ------------- | ------------- |
| First Caption Selected | ![screenshot-iPad Pro 12 9 2021-17 0](https://github.com/avalonmediasystem/avalon/assets/1331659/ccb58619-c0a4-4079-859f-8049411ff0b7) | ![screenshot-iPad Pro 12 9 2021-17 0(4)](https://github.com/avalonmediasystem/avalon/assets/1331659/1c878c59-aa7f-440f-bfe1-2b8e8eac0af1) |
[...]

PR #5735, opened 2024-03-13
Remove captions from HLS manifest

masaball:
Related Issues: #5695, #5722


Dananji (approved):
Looks good 👍 

Could you please add a description/link to a related issue in the PR description for context? If we ever need to refer back to this in the future, that would be helpful. Thanks!

PR #5745, opened 2024-03-25
Update dependencies for security warnings

cjcolvar:


Gemfile.lock
```diff
@@ -551,6 +552,8 @@ GEM
       rdf-turtle
       rdf-vocab (>= 0.8)
       slop
+    license_header (0.0.4)
+      highline
```
masaball, line 556: I think the Gemfile.lock here is out of sync with the actual Gemfile. `license_header` and `highline` are not included in the base Gemfile, so maybe you have them enabled in Gemfile.local?

cjcolvar: Good catch!

masaball (approved):
:+1:

PR #5747, opened 2024-03-25
Add handling of transcripts file to Batch Ingest

masaball:
~~This branch is built off prior, unmerged work to add support of uploading multiple caption files via batch manifest~~ 

PR #5704 ~~should be reviewed and merged before this PR.~~ has been merged.


lib/avalon/batch/entry.rb
```diff
@@ -277,15 +277,32 @@ def self.caption_language(language)
...
+        if datastream[file].present? && FileLocator.new(datastream[file]).exist?
+          # Build out file metadata
+          filename = datastream[file].split('/').last
+          label = datastream[label].presence || filename
+          language = datastream[language].present? ? caption_language(datastream[language]) : Settings.caption_default.language
+          machine_generated = datastream[:machine_generated].present? ? 'machine_generated' : nil
+          # Create SupplementalFile
+          supplemental_file = SupplementalFile.new(label: label, tags: [type, machine_generated].compact, language: language)
+          supplemental_file.file.attach(io: FileLocator.new(datastream[file]).reader, filename: filename)
+          supplemental_file.save
```
cjcolvar, line 291: What should happen if this save fails?  Using `save!` would raise an error.  Alternatively this method could return nil if the save failed.  Or is there a place where errors on the unsaved `supplemental_file` object would get read and returned to the user?  I think the last might be best but maybe returning nil and handling that in `attach_datastreams_to_master_file` might be easiest for now.  What do you think?

lib/avalon/batch/entry.rb
```diff
@@ -277,15 +277,32 @@ def self.caption_language(language)
       end
       private_class_method :caption_language
 
+      def self.process_datastream(datastream, type)
+        file, label, language = ["_file", "_label", "_language"].map { |item| item.prepend(type).to_sym }
+        if datastream[file].present? && FileLocator.new(datastream[file]).exist?
```
cjcolvar, line 282: What do you think about changing this into an explicit return `nil`?
```suggestion
        return nil unless datastream[file].present? && FileLocator.new(datastream[file]).exist?
```

lib/avalon/batch/entry.rb
```diff
@@ -176,20 +176,20 @@ def self.offset_valid?( offset )
...
-          supplemental_file.save
+        datastreams.each do |ds|
+          next unless ds.present?
+          supplemental_file = case ds.keys[0].to_s
+                              when /caption.*/
+                                process_datastream(ds, 'caption')
+                              when /transcript.*/
+                                process_datastream(ds, 'transcript')
+                              end
           master_file.supplemental_files += [supplemental_file]
```
cjcolvar, line 193: `supplemental_file` can be nil now that the `next` guard got moved into `process_datastream`.  I think this would probably be easiest to handle here with an inline guard.  Does that make sense to you or is there a better way to handle this?
```suggestion
          master_file.supplemental_files += [supplemental_file] unless supplemental_file.nil?
```

masaball: Using an inline guard is probably the best option. Other thing that immediately comes to mind would be doing something like `master_file.supplemental_file.compact!` after the loop has finished processing. The guard feels cleaner though.

lib/avalon/batch/entry.rb
masaball: That's a good concern. I would think overall behavior would be that the batch ingest completes and information is provided in the completion email about how the media items were created but there were problems with X supplemental file(s)? 

So maybe switch over to `save!` and create a catch that adds to the email? Or would having error information in a "Batch processed successfully" email be confusing? I think there are currently 2 emails that get sent: one about the batch being valid and then one about processing being successful or failing. Maybe we add a third set of emails that deals specifically with the caption/transcript files?

lib/avalon/batch/entry.rb
```diff
@@ -277,15 +285,31 @@ def self.caption_language(language)
       end
       private_class_method :caption_language
 
+      def self.process_datastream(datastream, type)
+        file, label, language = ["_file", "_label", "_language"].map { |item| item.prepend(type).to_sym }
```
cjcolvar, line 289: What do you think about renaming these variables so they are distinct from the values stored on lines 293-295?
```suggestion
        file_key, label_key, language_key = ["_file", "_label", "_language"].map { |item| item.prepend(type).to_sym }
```

cjcolvar (approved):
Looks great!  I had one small non-blocking suggestion.

cjcolvar (dismissed):

cjcolvar (approved):

PR #5754, opened 2024-03-27
Set duration of MasterFile from ActiveEncode instead of relying on the imprecise mediainfo duration

### LINKED ISSUE #5748: Incorrect Stream Duration
# Description

https://avalon-dev.dlib.indiana.edu/media_objects/0c483j41b

cjcolvar:
Resolves #5748

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -294,9 +294,13 @@ class MasterFile < ActiveFedora::Base
     #TODO pull this from the encode
     self.date_digitized ||= Time.now.utc.iso8601
 
-    # Set duration after transcode if mediainfo fails to find.
-    # e.x. WebM files missing technical metadata
-    self.duration ||= encode.input.duration
+    # Update the duration detected by ActiveEncode
+    # because it has higher precision than mediainfo
+    # Set for the first time for files without duration
+    # e.g. WebM files missing technical metadata
+    # ActiveEncode returns duration in milliseconds which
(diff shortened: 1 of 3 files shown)


app/models/master_file.rb
```diff
@@ -294,9 +294,11 @@ def update_progress_on_success!(encode)
...
     self.date_digitized ||= Time.now.utc.iso8601
 
-    # Set duration after transcode if mediainfo fails to find.
-    # e.x. WebM files missing technical metadata
-    self.duration ||= encode.input.duration
+    # Update the duration detected by ActiveEncode
+    # because it has higher precision than mediainfo
+    # Set for the first time for files without duration
+    # e.g. WebM files missing technical metadata
+    self.duration = encode.input.duration.to_i.to_s if encode.input.duration.present?
```
masaball, line 301: Why are we using `to_i` here? If we are looking for better precision should we not leave it as a float?

cjcolvar: Maybe it would make sense to add to the comment that the value should be whole number milliseconds?

masaball: Oh, I missed/forgot that this was milliseconds and the conversion to whole seconds is later in the serialization chain. It might not hurt to update the comment that the value returned is milliseconds, but yeah we don't need greater than millisecond accuracy so `to_i` makes sense here :+1:

masaball (approved):
Looks good!

Issue #5732, opened 2024-03-13
Use Ramp For Embedded Player


joncameron:
# Description
Avalon is still using a bare Video.js instance for embed routes and not Ramp. We should update this so that the embedded player has everything that Ramp can provide, and to make sure user experience is consistent.

# Done Looks Like
- [x] Ramp MediaPlayer component used for /embed routes
- [x] Current Video.js code removed
- [x] Player API works with Ramp
- [ ] Title display works as before (needs to be added to Ramp)

joncameron:
Refer to the previous issue for Video.js as the embedded player. Ramp instance control bar should contain all desired elements. Title is not in the list of option currently for Ramp's Video.js instance, but which is in the current Avalon embedded context.

joncameron:
@joncameron create corresponding Ramp ticket for adding Title display support.

masaball:
To populate the embedded Ramp player we currently need to request the full media object manifest and set the player to start from the requested section. This works fine, but we have a change in behavior where there are now previous and next buttons in the embedded player for objects with multiple sections. 

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNm9WQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4951f43c3016958b284786002442b99a817f38d8/screenRecording-27-3-2024-10-3.mp4)

We think that the behavior of being able to embed full media objects and playlists has been requested in the past. But how do we want to handle this?

- Keep the old behavior of embeds being limited to just the requested section
  - Implement a prop on the Ramp side to enable/disable previous/next buttons?
  - Create a minimal section manifest specifically for embeds so only a single item is included in the manifest, obviating the need for a prop in Ramp and reducing bandwidth needs and load times for embeds from many sectioned items?
- Keep the new behavior of supporting full Media Objects
  - Leave the implementation alone. Allow users to navigate through an embedded object with the previous and next buttons?
  - Create a minimal embed manifest for specific sections as described earlier and provide an option for users to embed either a specific section or embed an entire media object?

Are there other considerations we need to think about and are there other approaches that we could take?

Depending on what approach we choose to take we can create a follow up ticket to iterate on the work done for the initial implementation, if necessary.

elynema:
[...]

PR #5755, opened 2024-03-28
Use Ramp for embedded player

masaball:
Related issue: #5732


app/views/layouts/embed.html.erb
```diff
@@ -28,6 +24,8 @@ Unless required by applicable law or agreed to in writing, software distributed
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <%= favicon_link_tag %>
     <%= stylesheet_link_tag "application", media: "all" %>
+    <%= stylesheet_pack_tag 'application' %>
+    <%= javascript_pack_tag 'application' %>
```
cjcolvar, line 28: I'm slightly concerned that including the whole application JS/CSS packs will make for a large download for only the embed.  I wonder if we can create a new pack that is just what is needed for the embed player.  Or maybe the size isn't that large and this isn't really a problem.

cjcolvar (commented):
Nice!  Fewer lines than the original!

Dananji (dismissed):
Looks great! 👍

app/views/layouts/embed.html.erb
masaball: It looks like our compiled assets come out to around 8MB, which feels like a lot of extra data for just a player? I'll take a look at creating a new pack and trimming that down.

cjcolvar: Thanks!  Good luck! :crossed_fingers:

Issue #5729, opened 2024-03-13
Ability to mark captions as transcripts


elynema:
# Description

Collection staff are able to mark captions as "treat as transcript" in the Manage Files page. When this happens, the caption file should appear as a transcript in the transcript component. It also should register as a transcript for 'Has Transcript' and should be indexed for transcript searching and show up as hits in transcript when searching. 

Caption should only show up as a caption on Manage Files and not be duplicated as a Transcript. In the future, we may want to enable this functionality at the collection level, as well.

See #5736 

# Done Looks like

 - [x] checkbox next to language field in caption list with label "Treat as transcript" when editing
 - [x] when checked, caption text appears in transcript component
 - [x] when checked, caption text is searched and returned as hits in transcript
 - [ ] when checked, media item shows up in 'Has Caption' and 'Has Transcript' facet as 'Yes'
 - [x] when checked, caption file is only available for download once from the Files tab (transcripts will not be downloadable from Files tab)

joncameron:
Should it show up in the Files tab as well on the item page?

joncameron:
The 'transcript' tag could be added to the file in the application and much of the functionality may work from there. Filtering will probably need to be in place on the supplemental file page because we don't want the duplication there.

elynema:
@charumitraravi This one is also pretty well set up in terms of Done Looks Like. Can you QA this? The work has been done to index transcripts in the main index on avalon-dev, so any caption marked as transcript on the Manage Files section for an item should be indexed into the main search box.

The Has Captions and Has Transcript facets will show up on the Browse page when you are logged into Avalon-dev as a manager.

charumitraravi:
**when checked, caption text is searched and returned as hits in transcript** and **When checked, media item shows up in 'Has Caption' and 'Has Transcript' facet as 'Yes'**  - These conditions are failing. The caption is not being properly indexed as a transcript, making it unsearchable and not filtering correctly under the 'Has Transcript' facet.

cjcolvar:
I'm not able to reproduce the failure for the first condition (*when checked, caption text is searched and returned as hits in transcript*), but I'm able to reproduce the failure of the second (*when checked, media item shows up in facets as Yes*) if I check the `Treat as transcript`, click `Save` for the caption, and then go directly to `Preview` without saving the item.  Once the item is saved (clicking the `Continue` button in the edit workflow) then it appears in the 'Has Transcript` facet as 'Yes'.

[...]

PR #5763, opened 2024-04-02
Add option to treat a caption file as transcript

masaball:
Related issue: #5729

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -151,12 +151,15 @@ class SupplementalFilesController < ApplicationController
     end
 
     def edit_file_information
-      ident = "machine_generated_#{params[:id]}".to_sym
+      file_params = { machine_generated?: "machine_generated_#{params[:id]}".to_sym, caption_transcript?: "treat_as_transcript_#{params[:id]}".to_sym }
 
-      if params[ident] && !@supplemental_file.machine_generated?
-        @supplemental_file.tags += ['machine_generated']
-      elsif !params[ident] && @supplemental_file.machine_generated?
-        @supplemental_file.tags -= ['machine_generated']
+      file_params.each do |k,v|
+        tag = k.to_s.gsub(/(caption_)?(.*)(\?)/, '\2')
+        if params[v] && !@supplemental_file.send(k)
+          @supplemental_file.tags += [tag]
+        elsif !params[v] && @supplemental_file.send(k)
+          @supplemental_file.tags -= [tag]
+        end
       end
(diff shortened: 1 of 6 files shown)


app/controllers/supplemental_files_controller.rb
```diff
@@ -151,12 +151,15 @@ def edit_structure_path
...
-      ident = "machine_generated_#{params[:id]}".to_sym
-
-      if params[ident] && !@supplemental_file.machine_generated?
-        @supplemental_file.tags += ['machine_generated']
-      elsif !params[ident] && @supplemental_file.machine_generated?
-        @supplemental_file.tags -= ['machine_generated']
+      file_params = { machine_generated?: "machine_generated_#{params[:id]}".to_sym, caption_transcript?: "treat_as_transcript_#{params[:id]}".to_sym }
+
+      file_params.each do |k,v|
+        tag = k.to_s.gsub(/(caption_)?(.*)(\?)/, '\2')
```
cjcolvar, line 157: I think this would be easier to expand upon if instead of the regex `file_params`  turned into an array of hashes.  What do you think?
```suggestion
      file_params = [
        { param: "machine_generated_#{params[:id]}".to_sym, tag: "machine_generated", method: :machine_generated? },
        { param: "treat_as_transcript_#{params[:id]}".to_sym, tag: "transcript", method: :caption_transcript? }
      ]

      file_params.each do |v|
        param_name = v[:param]
        tag = v[:tag]
        method = v[:method]
```

app/models/iiif_canvas_presenter.rb
```diff
@@ -103,16 +103,28 @@ def audio_display_content(quality)
...
-              Rails.application.routes.url_helpers.master_file_supplemental_file_url(master_file.id, file.id)
-            end
-      IIIFManifest::V3::AnnotationContent.new(body_id: url, **supplemental_attributes(file))
+      unless file.is_a?(SupplementalFile)
+        url = Rails.application.routes.url_helpers.captions_master_file_url(master_file.id)
+        return IIIFManifest::V3::AnnotationContent.new(body_id: url, **supplemental_attributes(file))
+      end
+
+      tags = file.tags.reject { |t| t == 'machine_generated' }.compact
+      case tags
```
cjcolvar, line 112: I like the switch to the case statement!

cjcolvar (approved):
Looks good!  I had one suggestion but not really a blocker.

app/controllers/supplemental_files_controller.rb
masaball: Yeah, that seems like it will be a better approach in the long term. Will rework it.

Issue #5725, opened 2024-03-13
SRT Files Download as VTT


joncameron:
# Description
Avalon's /captions route is automatically transforming SRT files to VTT for display in the client. The download from Ramp is in turn returning a VTT instead of the original SRT when a user downloads from the Transcript component. 

This issue is for updating the manifest generation in Avalon 

- Current route for captions: supplemental_files/sdfli2398/captions # referenced in manifest
- Captions endpoint is only returning VTT currently
- supplemental_files/sdfli2398 should return the original SRT

# Chosen Implementation Strategy
- Captions are added as two annotations: supplementing (display in player, currently used), and rendering (to be used for download)

That what we would show in Ramp would be set as 'rendering' would work with the ethos of IIIF.

# Done Looks Like
- [x] Captions in IIIF manifests are present as two annotations: supplementing (display in player, currently used), and rendering (to be used for download)
- [x]  Supplemental file URL serialized into the manifest as an rendering annotation

joncameron:
In the IIIF manifest currently, the file mimetype is coming through as SRT when it should be VTT.

joncameron:
Best practice of supplementing annotations for AV materials may evolve over time in IIIF. There are not many existing examples of delivering any types of arbitrary file in IIIF.

joncameron:
@joncameron create a Ramp ticket for tasks above and scope this ticket to Avalon.

charumitraravi:
<img width="279" alt="Screenshot 2024-04-16 at 4 30 59 PM" src="https://github.com/avalonmediasystem/avalon/assets/52535309/cbc7f7b2-f34e-4201-9ce5-636921432968">

in the above example, I can see both the .srt and .vtt  versions of the "captions-example.srt" showing up under files. Is this the expected behavior? 

https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p (section 3)

masaball:
@charumitraravi I believe that is expected behavior right now. samvera-labs/ramp#474 contains the changes that will prevent display of both file types. So once that has been merged into Ramp and brought into Avalon we'll want to test the section file tab again.

PR #5769, opened 2024-04-04
Serialize captions as both supplementing annotation and rendering

masaball:
Related Issue: #5725


cjcolvar (approved):
Nice, simple and sweet!

Issue #5743, opened 2024-03-22
Create Thumbnail Doesn't Work (Demo)


joncameron:
# Description
On Terraform-based Avalon instances like the demo.avalonmediasystem.org, thumbnail images are unable to be set. Using the modal to select a new poster/thumbnail seems to hit a 500 error on the POST to /still for a masterfile.

# To reproduce
- Go to https://demo.avalonmediasystem.org/media_objects/wh246s463 or another item on Demo
- Attempt to update the poster image
- Poster image does not update

# Done Looks Like
- [ ] Poster/thumbnail image update is successful when selecting the "Update Poster Image" button from the "Create Thumbnail" modal

elynema:
nginx caching on thumbnails and poster images, but nothing to expire cache when these are updated in the application. This is impacting terraform sites. After 30 minutes, when the cache expires, the new thumbnails should show up.

cjcolvar:
The caching layer is setup in `avalon-docker` on the `aws_min` branch which is only used by `avalon-terraform`.

Here's the config:
https://github.com/avalonmediasystem/avalon-docker/blob/aws_min/cache.conf
and the nginx docs for these directives:
nginx.org/en/docs/http/ngx_http_proxy_module.html

We should figure out how to expire the cache for thumbnail/posters or maybe consider disabling the cache until we can figure out a better solution.

joncameron:
@cjcolvar What do you think would be best? I'm fine with commenting out lines for the Docker install if that resolves the issue in that enviroment. I can create a new issue either way.

cjcolvar:
I'll make this change in `avalon-docker` and we can revisit caching another time.  This change will be on the `aws_min` branch so it only affects the Terraform environment.

Completed here: https://github.com/avalonmediasystem/avalon-docker/commit/5b3405852a5608c2510de951c56d10e8462001e6

PR #5775, opened 2024-04-10
Use plain old javascript to get de-duplication of listener registration

cjcolvar:
Before this change `setInterval(enableCreateThumbnail)` would register the show and update handlers hundreds of times (depending on how long the page had been loaded) leading to the same large number of requests to fetch the poster image or post the update.  This PR uses `addEventListener` which de-duplicates listeners when registered with a function.

Related to #5743


Dananji (approved):
👍

Issue #5473, opened 2023-11-17
Consolidate player listener related code


masaball:
In many of the partials on the Media Object page there is javascript to wait for the player to finish loading and then perform actions, particularly placing event listeners onto the player or other elements within the Ramp context. All of this code should be DRY-ed up and consolidated for easier maintenance.

Known media object partials containing the same or similar code:
- _timeline.html.erb
- _thumbnail.html.erb
- _add_to_playlist.html.erb
- _share.html.erb
- _item_view.html.erb

It hasn't been finished/merged yet, but the playlist page work also includes a similar function in 7.7.

# Done Looks Like
- [ ] Pull the "adding event listener" code into its own module
- [ ] Call functions in partials rather than have duplication

# QA
 - test all event buttons: add to playlist, create thumbnail, create timeline, share section, transcript

joncameron:
We have a lot of intervals polling for player events; this will be helpful to clean up. Everything should be consolidated into a single place for ease of maintenance.

joncameron:
Tested all of the actions in the QA section and they all worked as expected.

PR #5779, opened 2024-04-12
Player events refactor

Dananji:
Related issue: #5473


app/assets/javascripts/player_listeners.js
```diff
@@ -0,0 +1,495 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ *   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ *   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ *   specific language governing permissions and limitations under the License.
+ * ---  END LICENSE_HEADER BLOCK  ---
+*/
+
+let canvasIndex = -1;
+let currentSectionLabel = undefined;
+let listenersAdded = false;
```
cjcolvar, line 19: Is this used only for add to playlist listeners?  If so how about renaming this variable?
```suggestion
let addToPlaylistListenersAdded = false;
```

cjcolvar (approved):
As far as I can tell this looks good.  Great job cleaning this up and consolidating it!

app/assets/javascripts/player_listeners.js
Dananji: Yes, that sounds good 👍

Dananji:
@cjcolvar I re-named the variable and brought in the changes we did in the old code to fix enabling/disabling add to playlist button in iPad.

cjcolvar (approved):
:+1:

Issue #5628, opened 2024-02-05
Color contrast does not meet minimum requirements - Clear selected items button


elynema:
# Description 

The 'Clear selected items' button the Selected Items page does not meet minimum color contrast requirements. 

<img width="1296" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeThzQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0fc2a3583a273d27eb9e72730a46951ed35258e5/image.png" alt="image.png" />

Recommendation from SiteImprove: Increase the color contrast to 4.5:1 or higher. 

Current color contrast is 3.68:1. 
<img width="347" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBekVzQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2e18b115fa340d9d8c92caad2238500da2332ff4/image.png" alt="image.png" />

Note: This was found by the SiteImprove browser plugin, not the SiteImprove crawler, so this does not *have* to be pushed out by 2/8.

# Done Looks Like
- [ ] Pick a color replacement for the current red warning button style
- [ ] Change CSS for "super bright red button" styling that applies to buttons in the application for deletion

Dananji:
Fix in the PR:

Value for `$danger` SCSS variable is changed from `#F44336` to `#D14242`.

The value `#D14242` was derived by using https://webaim.org/resources/contrastchecker/, with values for foreground color set to `#FFFFFF` and background color set to `#F44336` (old color) and adjusting the luminance till contrast ratio reached 4.59:1 (in compliance with WCAG 2.0 level AA).

Before:
<img width="1267" alt="Screenshot 2024-04-12 at 4 26 07 PM" src="https://github.com/avalonmediasystem/avalon/assets/1331659/725c66f2-9a95-4f7c-9996-360edd5be955">

After:
<img width="1267" alt="Screenshot 2024-04-12 at 4 29 27 PM" src="https://github.com/avalonmediasystem/avalon/assets/1331659/f78fce02-0287-4fc3-a838-987e581f051e">

joncameron:
Much better! Easier on the eyes and to read.

elynema:
[...]

PR #5782, opened 2024-04-12
Fix a11y color contrast for SCSS variable  in branding

Dananji:
Related issue: #5628 

In this PR, the value for `$danger` is changed from `#F44336` to `#D14242`.

The value `#D14242` was derived by using https://webaim.org/resources/contrastchecker/, with values for foreground color set to `#FFFFFF` and background color set to `#F44336` (old color) and adjusting the luminance till contrast ratio reached 4.59:1 (in compliance with WCAG 2.0 level AA).

Before:
Screenshot 2024-04-12 at 4 26 07 PM

After:
Screenshot 2024-04-12 at 4 29 27 PM

### DIFF
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -18,11 +18,13 @@
 
 @import 'blacklight/blacklight';
 
-.result-thumbnail img, img.result-thumbnail {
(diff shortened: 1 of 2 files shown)


masaball (approved):
:+1:

PR #5797, opened 2024-04-19
Add option to rescue Solr/Fedora connection errors

masaball:
This implementation uses a static error page for the Solr connection errors. The nav bar header rendered as part of the regular Avalon layout checks user permissions, which directly calls out to Solr. This triggers an error that is not caught by rescue methods in application_controller. Because the nav bar is rendered on almost every page of the application, this results in most of the application being inaccessible. As such, serving a static error page/message should be sufficient.

Using the Rails `config.action_dispatch.rescue_response` handling pathway was not logging the proper error in testing and did not prevent other partials from attempting to render which resulted in extra errors and noise being present in the logs. Serving this file without the Avalon layout and logging the Solr error as part of a standard `rescue_from` in the application_controller feels more in line with our needs.

The issue with the nav bar does not occur when Fedora is inaccessible, so that uses a non-static error page similar to the `deleted_pid.html.erb` style views. This keeps the user in the standard Avalon context and allows them to continue using the application outside of pages that need the fedora connection (which should be most non-edit pages).

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -44,6 +44,8 @@ solr:
     rule:
     shards:
     snitch:
+solr_and_fedora:
+  raise_on_connection_error: true
 zookeeper:
   connection_str: "localhost:9983/configs"
 streaming:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -46,7 +46,7 @@ solr:
     snitch:
 app_controller:
   solr_and_fedora:
-    raise_on_connection_error: true
+    raise_on_connection_error: false
 app_job:
   solr_and_fedora:
     raise_on_connection_error: true
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -44,8 +44,12 @@ solr:
     rule:
     shards:
     snitch:
-solr_and_fedora:
-  raise_on_connection_error: true
+app_controller:
+  solr_and_fedora:
+    raise_on_connection_error: true
+app_job:
+  solr_and_fedora:
+    raise_on_connection_error: true
 zookeeper:
   connection_str: "localhost:9983/configs"
 streaming:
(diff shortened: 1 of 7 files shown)


cjcolvar:
Sorry for being so slow to review this!  

I'm thinking it may be good to separate out the configurations for controller vs. job.  I think sidekiq relies on errors to rerun a job and also to be visible in the dashboard as failed.  Having different configurations would allow displaying the custom error pages for UI requests while still allowing background jobs and sidekiq to do their normal thing.  

`retry_on` (https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on) could be used and then after a set number of failures could go to the custom rescue block which could raise or log.    The issue with that could be that the job reruns retry_on times x sidekiq rerun times as well as having the failed job display in the dashboard.  So may still want to have separate configs.

What do you think?

masaball:
That's a good point about sidekiq. I think having separate config values makes sense. We can maybe try the `retry_on` block if you think it is worth it, but your worry about the muliplicative retries is a pretty big deterrent.

config/settings.yml
```diff
@@ -44,6 +44,12 @@ solr:
     rule:
     shards:
     snitch:
+app_controller:
+  solr_and_fedora:
+    raise_on_connection_error: true
```
cjcolvar, line 49: What do you think about setting this to false by default?  That would be a change in behavior but one that should only affect end users so I think it would be seen as an improvement.
```suggestion
    raise_on_connection_error: false
```

cjcolvar (approved):
This looks good.  Thanks for working out the intricacies that made this harder than originally conceived.

cjcolvar (approved):

Issue #5785, opened 2024-04-15
Add to playlist button disabled on iPhone Safari until user plays media item


elynema:
# Description

The add to playlist button is disabled when the media object page loads in Safari on iPhone. It remains disabled when the user changes section, etc. It is enabled if the user starts and then stops playing the media item.

Testing is showing this issue in 7.7.1 on demo site and 7.7.2 on avalon-staging, but not in MCO (running 7.7 + cherrypicks). It seems this was introduced in 7.7.1.

This is not happening in Safari on desktop.

elynema:
@Dananji Safari on iOS is working correctly now, but as part of QA I was testing across other browser combinations, and we are seeing the same behavior now on Safari on iPad that we were seeing on iPhone. Safari on desktop is still fine, as is Chrome on Android.

elynema:
Tested on iPhone and iPad (Safari), Chrome on Android, and Chrome and Safari on desktop. Add to playlist functionality seems to be working correctly.

PR #5799, opened 2024-04-22
Extend player status check for add to playlist to iPad Safari

Dananji:
Related issue: #5785


cjcolvar (approved):
As far as I know this looks okay but I'm wondering if the switch from matching `/Mobi/i` to `/iPhone/i` misses any devices  and if it might be safer to leave it in and add it to the OR'd conditiions on line 221.  Or did `/Mobi/i` not match correctly before?

Dananji:
`/Mobi/i` did match the condition before, I changed it so that's it's more scoped. But leaving it in and adding to the condition makes this robust.

masaball (approved):

Issue #5802, opened 2024-04-23
Fix IIIF Manifest Validation Errors - Playlist Manifests


joncameron:
## Description

Follow-up ticket for #5571. The errors listed in the ticket for inaccessible items and structures were given for both playlist and media object manifests. For the media object ticket, see #5798.

## Done Looks Like
For Playlist manifests:
- [ ] the root `Range` node in the `structures` property has a `label` property with a language map instead of `null`
**For inaccessible items;**
- [ ] `body` in the `placeholderCanvas` has a value for the `id` property instead of `null`
- [ ] `placeholderCanvas` has `height` and `width` properties
- [ ] `AnnotationPage` in the `Canvas` `must` have an `items` list with at least one `Annotation` (https://www.w3.org/TR/annotation-model/#annotation-page)
- [ ] `Canvas` has `height` and `width` properties

joncameron:
I saw none of the relevant errors in the validation tool when checking against playlists with inaccessible items on avalon-dev.

PR #5809, opened 2024-04-30
Fix playlist manifest IIIF validator errors

Dananji:
Related issue: #5802

Related `iiif_manifest` gem PR: https://github.com/samvera/iiif_manifest/pull/112

### DIFF
--- a/app/models/iiif_playlist_canvas_presenter.rb
+++ b/app/models/iiif_playlist_canvas_presenter.rb
@@ -195,8 +195,8 @@ class IiifPlaylistCanvasPresenter
     def manifest_attributes(quality, media_type)
       media_hash = {
         label: quality,
-        width: master_file.width.to_i,
-        height: master_file.height.to_i,
+        width: (master_file.width || '1280').to_i,
+        height: (master_file.height || MasterFile::AUDIO_HEIGHT).to_i,
         duration: stream_info[:duration],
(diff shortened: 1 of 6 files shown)


app/models/iiif_playlist_canvas_presenter.rb
```diff
@@ -218,6 +209,16 @@ def manifest_attributes(quality, media_type)
...
     end
 
+    def placeholder_attributes(label_content)
+      placeholder_hash = {
+        label: label_content,
+        type: 'Text',
+        format: 'text/plain',
+        width: 1280,
+        height: 720,
+    }.compact
```
masaball, line 219: Not a blocker, but looks like this line didn't get indented properly.

masaball (commented):
Do we need to include pulling in the iiif_manifest update in this PR as well? Or will it be fine to pull that in as a separate PR later?

Dananji:
I updated the `Gemfile` to fetch `iiif_manifest` from Git repo. Once the changes are are tagged and released we can update the `Gemfile`?

masaball (approved):
Yeah, we can pull from the git repo for testing in develop. We'll tag a iiif_manifest release and update the `Gemfile` to point to that when we are finalizing the next Avalon release.

masaball (approved):

Issue #5798, opened 2024-04-19
Fix IIIF Manifest Errors - Media Objects


Dananji:
## Description

Follow-up ticket for #5571. The errors listed in the ticket for inaccessible items and structures were given for both playlist and media object manifests. For the playlist manifests, see #5802.

In the media object context, sections still processing and sections where processing failed are presented as inaccessible items in the IIIF manifest.

## Done Looks Like
For media object manifests:
- [ ] the root `Range` node in the `structures` property has a `label` property with a language map instead of `null`
**For inaccessible items;**
- [ ] `body` in the `placeholderCanvas` has a value for the `id` property instead of `null`
- [ ] `placeholderCanvas` has `height` and `width` properties
- [ ] `AnnotationPage` in the `Canvas` `must` have an `items` list with at least one `Annotation` (https://www.w3.org/TR/annotation-model/#annotation-page)
- [ ] `Canvas` has `height` and `width` properties

joncameron:
Manifest generation is separate for media objects and playlists; this ticket can be broken in two: one for media objects, and one for playlists. The same set of issues need to be corrected in both places.

Dananji:
Three more issues were found in audio manifests:

#### Error 1:
Message: 0 is less than or equal to the minimum of 0
```
Detail:
Path: /items[0]/height/[0 is less than or equal to the minimum of 0]
Description: 0
```
**Fix suggestion: _set `height` to a value not zero for audio master file_**

### Error 2:
Message: 0 is less than or equal to the minimum of 0
```
Detail:
Path: /items[0]/width/[0 is less than or equal to the minimum of 0]
Description: 0
```
**Fix suggestion: _set `width` to a value not zero for audio master file_**

### Error 3:
Message: '/assets/audio_icon-7782bcf9fd5d81f1e3e3f260ed569c24dea47d0a2a5bd0a4629328c76e8886c6.png' does not match '^http.*$'
```
    Detail: Id must be present and must be a URI
    Path: /thumbnail[0]/id/['/assets/audio_icon-7782bcf9fd5d81f1e3e3f260ed569c24dea47d0a2a5bd0a4629328c76e8886c6.png' does not match '^http.*$']
    Description:

    "/assets/audio_icon-7782bcf9fd5d81f1e3e3f260ed569c24dea47d0a2a5bd0a4629328c76e8886c6.png"
```
**Fix suggestion: _change thumbnail `id` to a URL_**

elynema:
[...]

PR #5810, opened 2024-04-30
Fix iiif validator errors for media object manifest generation

Dananji:
Related issue: #5798

### DIFF
--- a/app/models/iiif_manifest_presenter.rb
+++ b/app/models/iiif_manifest_presenter.rb
@@ -211,7 +211,7 @@ class IiifManifestPresenter
       if video_count > 0
         Rails.application.routes.url_helpers.thumbnail_master_file_url(master_file_id)
       elsif audio_count > 0
-        Rails.application.routes.url_helpers.thumbnail_master_file_url(master_file_id)
+        ActionController::Base.helpers.asset_url('audio_icon.png')
       end
     elsif video_count > 0 && audio_count > 0
       ActionController::Base.helpers.asset_url('hybrid_icon.png')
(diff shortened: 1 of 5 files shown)


app/models/iiif_manifest_presenter.rb
```diff
@@ -211,7 +211,7 @@ def thumbnail_url
       if video_count > 0
         Rails.application.routes.url_helpers.thumbnail_master_file_url(master_file_id)
       elsif audio_count > 0
-        ActionController::Base.helpers.asset_url('audio_icon.png')
+        Rails.application.routes.url_helpers.thumbnail_master_file_url(master_file_id)
```
cjcolvar, line 214: Why was this change needed?  I think before the goal was to link directly to the asset and avoid having to have another request to rails to get to the asset.

Dananji: According to the IIIF 3.0, `id` property must be a URI. And the validator was failing on this.

Dananji: Thanks for the suggestion Chris! I tried it (updated the initializer and re-started docker) and still got the relative path. So I tried the following;
```
URI.join(Rails.application.routes.url_helpers.root_url, ActionController::Base.helpers.asset_url('audio_icon.png'))
```
which gave the asset URI instead of the relative path. Do you think this could work?

cjcolvar: Interesting.  I wonder if my suggestion didn't work because it runs after ActionController gets loaded.  In a rails console I'm able to make things work if I manually set `asset_host` on `ActionController::Base`:
```
ActionController::Base.asset_host = Rails.application.routes.url_helpers.root_url
```
If you put that in the initializer instead does it work?

Dananji: `Rails.application.routes.url_helpers.root_url` came up as an undefined variable when spinning up the docker container, but combining the variable name and the value from the 2 comments it worked 🎉 
```
ActionController::Base.asset_host = URI("#{Settings.domain.protocol}://#{Settings.domain.host}:#{Settings.domain.port}").to_s
```

cjcolvar (dismissed):

PR #5814, opened 2024-05-02
fixed test failures, added collections smoke tests, added cypress_dev…

charumitraravi:
- Fixed failing cases in the existing cypress integration tests
- Added an additional config file, cypress_dev config to run the tests on avalon dev 
- Added collection smoke tests 

To run the cypress_dev config, change the entrypoint to cypress run -C spec/cypress/**cypress_dev**.config.js in the docker compose file , and run docker-compose up cypress. 

Note: Playlist cases are still failing due to the "no playlist items" alert.

### DIFF
--- a/spec/cypress/cypress_dev.config.js
+++ b/spec/cypress/cypress_dev.config.js
@@ -1,0 +1,28 @@
+const { defineConfig } = require("cypress");
+
+module.exports = defineConfig({
+  env: {
+    "USERS_ADMINISTRATOR_EMAIL": "archivist1@example.com",
+    "USERS_ADMINISTRATOR_PASSWORD": "archivist1",
+    "USERS_USER_EMAIL":"chravi@iu.edu",
+    "USERS_USER_PASSWORD": "Test1234!",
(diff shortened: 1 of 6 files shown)


cjcolvar (approved):
This looks great so far!  Nice job getting the tests passing.

I like how the variables for use in avalon-dev are isolated and collected in a separate file.  The cypress config file seems to have global settings which should apply to all environments so I'm thinking it might be good to use a different mechanism to pass in these environment-specific values.  I was looking at https://docs.cypress.io/guides/guides/environment-variables and liked `cypress.env.json` but I'm not sure how that would extend to multiple environments (maybe it is possible to pass as a CLI option `--env cypress.dev.env.json`?).  It will also be important to keep these values secure since there are credentials so maybe it makes more sense to pass them in as system environment variables (e.g. `CYPRESS_*`) so they could live in separate `.env` files or be defined in secrets in a CI server.  What do you think?

charumitraravi:
Thanks Chris. I will look into how we can make use of environment variables.

Issue #5641, opened 2024-02-07
[BUG] Playlist drop-down does not default to first playlist on iOS versions 17.x


elynema:
# Description

In mobile devices with iOS 17+. when the user clicks on Add to playlist, the playlist drop-down does not default to the first playlist. Instead, it displays the text 'Add new playlist'. User cannot click on the 'Add new playlist' text to actually add a playlist unless they first select a different playlist in the drop-down and then go back.

Older versions of iOS do not run into this issue. 

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUU4QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--12baf24735d26d4c4c501204588282a4556d2a43/RPReplay_Final1707335950.mov)

# Done Looks Like
- [ ] The first playlist (alpha order) should be selected by default when the user clicks Add to Playlist for an item. 
- [ ] 'Add new playlist' usually is available above the list of playlists if the user opens the drop-down, is bolded, and opens up a new modal to create a playlist if the user clicks on it.

joncameron:
That dropdown is populated by some coffeescript; something in that could not be working as intended in iOS 17+.

Dananji:
During implementation I tested this on;
- Firefox, Chrome, and Safari in Sonoma 14.4
- Chrome and Safari in iOS 17.4 and 16
- Chrome in Android 14

elynema:
@charumitraravi Would you be willing to take a look at this one on avalon-dev for QA?

charumitraravi:
For some reason, I am unable to select "Add to playlist" on iOS devices (both safari and chrome browsers) on avalon-dev. "Add to playlist", "Create timeline", and create thumbnail buttons are disabled.


![IMG_5872](https://github.com/avalonmediasystem/avalon/assets/52535309/eebc3288-b243-4309-839f-cac38351cf62)

elynema:
Moving this back to In Progress due to issue Charu reported.

elynema:
@Dananji Moving this back to In Progress due to the issue that Charu reported here.

elynema:
@charumitraravi Can you re-test this?

charumitraravi:
QA completed. The fix is working as expected

PR #5824, opened 2024-05-06
Set default option to first playlist title in playlist dropdown in add to playlist

Dananji:
Related issue: #5641

### DIFF
--- a/app/assets/javascripts/add-playlist-option.js.coffee
+++ b/app/assets/javascripts/add-playlist-option.js.coffee
@@ -32,7 +32,7 @@ $ ->
       has_new_opt = true
   
   if !has_new_opt
-    select_element.prepend(new Option(addnew))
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/add-playlist-option.js.coffee
```diff
@@ -32,7 +32,7 @@ $ ->
       has_new_opt = true
   
   if !has_new_opt
-    select_element.prepend(new Option(addnew))
+    select_element.append(new Option(addnew))
```
cjcolvar, line 35: Does this change the location of the "Add new playlist" option to the bottom of the dropdown list?

Dananji: No, it doesn't

cjcolvar (approved):

Issue #5733, opened 2024-03-13
Add IIIF Button With Link to Manifest Route


joncameron:
# Description
Many applications which provide IIIF manifest include a link to the manifest on the page for easy reference. Avalon has a route for the manifest (/manifest(!)), but no link on the page that would make it easy to find and reference for sharing or use in other contexts.

# Done Looks Like
- [ ] IIIF manifest link appears in the Share panel on the item page and links to /manifest

# Mockups of Placement Options
## Icon
![iiif1](https://github.com/avalonmediasystem/avalon/assets/1406618/8708a85f-222b-4f0a-a170-407db8446906)

![iiif2](https://github.com/avalonmediasystem/avalon/assets/1406618/f119f81e-d3ab-4257-bdc9-4a3c335a54c5)

## Share Panel
![Screenshot 2024-04-10 at 2 29 04 PM](https://github.com/avalonmediasystem/avalon/assets/1406618/a239b123-339d-4107-9e73-dabbbd0b53e2)

joncameron:
Adding it to the Avalon share panel can be a first pass; creating a new component in Ramp for the Share component is the next iteration.

PR #5825, opened 2024-05-07
Add IIIF Manifest link to the share resource panel

Dananji:
Related issue: #5733 

First pass of adding IIIF Manifest link on the item view page.

### DIFF
--- a/app/views/media_objects/_share_resource.html.erb
+++ b/app/views/media_objects/_share_resource.html.erb
@@ -31,7 +31,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       </div>
       <div class="form-group">
         <label for="link-object-manifest">IIIF Manifest</label>
-        <input type="text" class="col-md-12 form-control" id="link-object-manifest" value="<%= Rails.application.routes.url_helpers.manifest_media_object_url(@media_object.id) %>" 
+        <input type="text" class="col-md-12 form-control" id="link-object-manifest" value="<%= manifest_media_object_url(@media_object.id) %>" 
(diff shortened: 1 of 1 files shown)


app/views/media_objects/_share_resource.html.erb
```diff
@@ -29,6 +29,11 @@ Unless required by applicable law or agreed to in writing, software distributed
         <p class="muted"> <%= I18n.t('media_object.empty_share_section_permalink_notice') %> </p>
         <% end %>
       </div>
+      <div class="form-group">
+        <label for="link-object">IIIF Manifest</label>
+        <input type="text" class="col-md-12 form-control" id="link-object" value="<%= Rails.application.routes.url_helpers.manifest_media_object_url(@media_object.id) %>"
```
masaball, line 34: I noticed that this has the same ID as the field for the share link to the media object itself at ln 21-22. Would it be better to give them separate IDs?
```suggestion
        <label for="link-object-manifest">IIIF Manifest</label>
        <input type="text" class="col-md-12 form-control" id="link-object-manifest" value="<%= Rails.application.routes.url_helpers.manifest_media_object_url(@media_object.id) %>" 
```

masaball (approved):
:+1: 

ETA: After approving this, some stuff clicked in my brain. The `Rails.application.routes.url_helpers` sequence of methods is for being able to get urls/paths in the controllers or models. Since this is a view, we should be able to just use `manifest_media_object_url(@media_object.id)`. Could you give that a try before merging and update the PR if it works?

cjcolvar (approved):

PR #5826, opened 2024-05-08
Allow setting default headers in fedora.yml config

cjcolvar:
This facilitates a workaround we needed in production to set the `User-Agent` header on all faraday requests to a known good value `curl/7.76.1`  instead of `Faraday v2.7.4`.  With this PR it ensures that the header is set on all faraday requests when configured in fedora.yml.  This approach is also flexible and gives us a space to set other headers in the future if needed.

Example fedora.yml config:
```
production:
  user: fedoraAdmin
  password: fedoraAdmin
  url: http://127.0.0.1:8983/fedora4/rest
  base_path: /prod
  request: { timeout: 600, open_timeout: 60} # Increase if Fedora connection times out for large operations
  headers: { User-Agent: 'curl/7.76.1' }
```

### DIFF
--- a/config/initializers/active_fedora_reindexing.rb
+++ b/config/initializers/active_fedora_reindexing.rb
@@ -1,4 +1,24 @@
 ActiveFedora::Fedora.class_eval do
+    def header_options
+      @config[:headers]
+    end
+
+    def authorized_connection
+      options = {}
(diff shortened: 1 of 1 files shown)


masaball (approved):
Looks good

Issue #5812, opened 2024-05-01
Chunking for Transcript Documents in Solr


joncameron:
## Description
Carrying on from #5730, we'll need to decide how to put transcript documents into Solr to meet our needs for query matching. In particular, the documents should be broken up into chunks that can be returned in IIIF Content Search API responses, which will then be handled by the transcript search component in Ramp for display and navigation.

### Transcript types and chunking strategy
VTT and SRT: Time and text for each cue
Word and Plain Text: Paragraphs

## Done Looks Like
- [x] Example documents added to this ticket for the 4 main transcript types
- [x] Valid transcripts (types listed above) are indexed into Solr in a way that makes it possible to query and return in IIIF Content Search format responses
- [x] Figure out parsing on the documents for indexing

masaball:
Here are the test files I created and have been using for the initial work on this issue (zenhub did not like the VTT or SRT extension so I uploaded them as `.txt` files). VTT and SRT use approximately the first minute of Lunchroom Manners captions. Plain Text and Docx use a chunk of the first chapter of the `spec/fixtures/public-domain-book.txt`:

[chunk_test.docx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1BHQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0fa4ce3fa812f3dd32c4f30687777200c0fb4c27/chunk_test.docx)

[chunk_test.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1RHQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5ef301be5bd813824acfea4c2a283bede113dc8a/chunk_test.txt)

[vtt_chunk_test.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1hHQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ccd26fbd4db6853315a4276d4c7617ef8f0b4f15/vtt_chunk_test.txt)

[...]

PR #5828, opened 2024-05-08
Add ability to chunk transcript files

masaball:
Related issue: #5812 

This is a first pass/proof of concept for splitting transcript files for indexing. Chunking is as follows:

- Plain Text: Divided by paragraph by splitting the text wherever there are two or more new line characters
- Docx: These files are converted into a plain text string and then chunked in the same manner as regular plain text
- VTT/SRT: Divided by time cue. The chunking itself is performed exactly as the other file types (2 or more newlines), additionally headers, styling, subtitle numbers, etc. are removed so that we are only indexing the time cues and their payload (e.x. `00:00:01.000 --> 00:00:05.000 Hello World!`, etc).

~~Currently, all text is passed through the same filtering so if a plain text or docx file is formatted like timed text, the chunking will be identical to if it were a VTT/SRT file i.e. no VTT header or subtitle numbers, just time cues and text.~~

Only SRT and VTT files are passed through the handling for timed text files. We cannot guarantee the formatting of Docx or regular plaintext files, so it is safer to not run them through the extra parsing.


lib/avalon/transcript_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    # Passed in transcript_file must be instance of ActiveStorage::Attached::One
+    def self.normalize_transcript transcript_file
+      plaintext = self.transcript_plaintext(transcript_file)
+      return if plaintext.blank?
+
+      normalized_plaintext = plaintext.gsub("\r\n", "\n")
+      # We only need the time cues and associated text content indexed so we remove the subtitle
+      # numbers and the VTT header, if present.
+      headerless_plaintext = normalized_plaintext.gsub(/WEBVTT.+?(?=\d{2}:\d{2}:\d{2})/m, "")
+      headerless_plaintext.gsub(/\d+\n(?=\d{2}:\d{2}:\d{2})/, "")
```
cjcolvar, line 39: I'm not sure how common they are but it looks like there can be comments and style lines in a VTT as well as attributes after the timecode.
https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API
This makes things more complicated and possibly makes things more brittle when working with all kinds of plaintext.  Maybe it would make sense to have a VTT specific parsing method/class which can strip these out?

masaball: Sure, we can add some VTT specific parsing to make things more robust. Expected behavior would be that just the timed cues and associated text are retained, correct? (e.x. `00:00:00.000 --> 00:00:10.000 Welcome to the video`)

Currently, `/WEBVTT.+?(?=\d{2}:\d{2}:\d{2})/m`  removes everything before the first time cue so that would include any style or comment lines between the header and the time cue. So we would need to add handling for comments and attributes in the body of the transcript.

lib/avalon/transcript_parser.rb
```diff
@@ -0,0 +1,70 @@
...
+  class TranscriptParser
+    TEXT_TYPE = ['text/vtt', 'text/srt', 'text/plain', 'application/vnd.openxmlformats-officedocument.wordprocessingml.document']
+
+    # Passed in transcript_file must be instance of ActiveStorage::Attached::One
+    def initialize(transcript_file)
+      @mime_type = transcript_file.content_type
+      @transcript = transcript_file.download
+    end
+
+    def transcript_plaintext
```
cjcolvar, line 27: Now that that `TranscriptParser` is a class which wraps a single transcript what do you think about naming this method `plaintext` and using it to memoize an instance variable?  Something like:
```
def plaintext
  unless @plaintext
    @plaintext = Avalon::DocxFile.new(@transcript).unformatted_text if @mime_type == 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
    @plaintext ||= @transcript
  end
  @plaintext
end
```

lib/avalon/transcript_parser.rb
```diff
@@ -0,0 +1,70 @@
...
+      # Transcripts can have arbitrary files imported. We need to verify that the transcript
+      # is a text based file before attempting processing.
+      return unless TEXT_TYPE.include? @mime_type
+      # Docx files are a zip file containing XML files so require specialized handling to retrieve the content.
+      plaintext = Avalon::DocxFile.new(@transcript).unformatted_text if @mime_type == 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
+      # SRT and VTT files are plaintext, so it is sufficient to just download the content without additional handling.
+      plaintext ||= @transcript
+    end
+
+    def normalize_transcript
```
cjcolvar, line 37: Like my suggestion for `plaintext` it seems like this method could be renamed (`normalized_text`?) and used to return an instance variable.

lib/avalon/transcript_parser.rb
```diff
@@ -0,0 +1,70 @@
...
+      plaintext = transcript_plaintext
+      return if plaintext.blank?
+
+      normalized_plaintext = plaintext.gsub("\r\n", "\n")
+      normalized_transcript = normalize_timed_text(normalized_plaintext) if @mime_type == 'text/vtt' || @mime_type == 'text/srt'
+      normalized_transcript ||= normalized_plaintext
+    end
+
+    def normalize_timed_text(plaintext)
```
cjcolvar, line 46: This feels like it should be a private method since it is only used for constructing the normalized transcript text.

cjcolvar (commented):
I like where this is going!

cjcolvar (approved):
I think the one failing test is probably flaky and unrelated so I think this is fine to merge.

Issue #5730, opened 2024-03-13
Add transcript text to Solr


elynema:
# Description

Based on outcome from investigation in #5726, actually add transcript text to Solr so that it is searched as part of the Avalon repository search. Any media items with result(s) in the transcript should be returned in the search results.

First pass implementation could just index the entire transcript as one field. We may find we want to break the transcripts up and store them at the cue level in the index to better support IIIF content search if we decide to go that route.

# Done Looks Like
 - [x] Media object is returned if there are any matches across any of the transcripts tied to any section
 - [x] Phrase searches entered with quotation marks are followed (only records with transcripts that have that phrase are returned as search results)
- [x] Captions marked as transcripts are also searched
- [x] Rake or migration task written for taking transcript documents (or captions marked as transcripts) and adding them to the index. This task should also support population of the Has Captions and Has Transcript facets, which will require re-indexing of the masterfile that owns the transcript/caption.

cjcolvar:
Differing approaches to solving this:

Nested documents (requires a reindex to add `_root_` and `_nest_path_` fields): https://github.com/avalonmediasystem/avalon/commit/051c9318d665a96e9b628754827843830488b745

Independent documents (requires a DB migration for supplemental files): https://github.com/avalonmediasystem/avalon/commit/3106c9d0fffb47055b3dd5bddb5a0c51eccabc7f

elynema:
Proof of concept dropped transcript text on masterfile - one big field for all transcripts for that section. That didn't allow correlation of hits to transcripts. Needed to index independently.

Options:

1. make each transcript its own Solr document and have a field that maintains the relationship (id) with a join query
 Pros:
- doesn't require any changes to Solr
Cons:
- requires new db field in supplemental files with parent ID, so this adds a small database migration. Would have to go back and backfill, similar to how we did for captions in 7.7.0 release.
- complex Solr queries with multiple joins


2. nested documents. Transcript document is a child of the masterfile document. May make it easier to maintain the relationship between transcript and masterfile. 
Pros:
 - simplifies Solr queries - don't have to do multiple joins, which are very gnarly to write correctly
 - Solr is building up support for child documents
Cons:
 - Solr needs new fields added to every document, root and nestpath. This would require a full re-index.
 - Nested documents might require Solr 8/9. We have not had a hard requirement for Solr 9 yet. 

[...]

Issue #5728, opened 2024-03-13
Show "found in" hints to indicate where search terms were found


elynema:
# Description

Show "found in" hints to indicate to user if search term(s) were found in:

- metadata: descriptive metadata only
- section labels + structure: section labels + any sub-structure
- transcript

# Done Looks Like

 - [x] show found in results below metadata for each item in search results as:
Found in: metadata (5)
Found in: sections (6)
 - [ ] if there is a zero count for either of these 3 parts (metadata, transcripts, sections), do not display that part
 - [x] do not show "found in" when results are showing from a browse, not search
 - [ ] For phrase searches with quotes, only media objects with that phrase should be returned, but the "found in" counts should be the same as if no quotes.

joncameron:
Can wait until previous investigation; will need to look at highlighting in Solr.

elynema:
Currently storing transcript as text, not text_en, so stemming is probably not happening. May want to re-consider this. Metadata hits are coming from a field with the entire MODS as text_en.

masaball:
The new query set up to facilitate "Found in" hints and return transcript results adjusted the way that results are returned in the solr document. Previously, I think the entire solr document was being returned. The new query structure explicitly searches for specific fields. A quick audit of the code indicated that the minimum fields that need to be returned for parity with current Avalon are:

- has_model_ssim
- id
- title_tesi
- abstract_ssi
- date_issued_ssi
- date_created_ssi
- duration_ssi
- section_id_ssim
- avalon_resource_type_ssim

Most of these are to make sure that the metadata on the search page is properly filled. `avalon_resource_type_ssim` is necessary for displaying the thumbnails.

joncameron:
Created #5903 for not displaying zero-count field results, otherwise this is working as expected in my testing.

PR #5830, opened 2024-05-10
Index transcripts and return them in search results

masaball:
Related issues: #5728, #5730


cjcolvar:
Still need to wrap user query in `{!edismax}` to retain boosting and other default config in `solrconfig.xml`

docker-compose.yml
```diff
@@ -54,8 +54,11 @@ services:
       - solr-precreate
       - avalon
       - /opt/solr/avalon_conf
+    ports:
+      - '8983:8983'
```
masaball, line 58: We'll want to remove these lines and the solr external network before merging.

Dananji (approved):

Issue #5793, opened 2024-04-19
IIIF content search service


elynema:
# Description

Create an API wrapper around Solr that submits the appropriate query and returns results formatted as IIIF content search results. At this time, we are just implementing the `q` search parameter, as others are optional.

IIIF content search API: https://iiif.io/api/search/2.0/

See https://github.com/avalonmediasystem/avalon/issues/5786 for suggestion of format for search results

# Done Looks Like
- [x] Search service available on masterfiles (other objects out of scope for now)
- [x] Service route example URI: avalon.xyz/master_files/ft848q60n/search?q=searchterm"
- [x] Service returns results only for transcripts associated with the masterfile
- [x] Each match result provides one annotation:
  - [x] Annotation targets specific transcript routes, with time fragments for timed text sources (.vtt)
  - [x] Annotation target refers to the URI for the body of the original annotation, which is a stable identifier
- [x] annotation motivation should match the motivation of the original annotation (ex: supplementing)

elynema:
Per conversation at Swarm on 5/22, we approved using a search service route that references the master_file directly and looks like: 
avalon.xyz/master_file/ft848q60n/search?q=searchterm

elynema:
Per the #search channel in IIIF slack, recommend that the motivation used for the annotation in the search response is the same as the motivation from the original annotation. In our case, that motivation = supplementing, since we currently only have supplementing motivations. In the future, we might have other types of annotations, though!

<img width="555" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzRFQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0db2ae83ead0190a8056964e334b14eb8eb3deda/image.png" alt="image.png" />

Issue #5792, opened 2024-04-19
Implement IIIF content search results


elynema:
# Description

Grab results from Solr query for search within and format those results into the proposed IIIF content search response.

See #5786 for sample search results format.

The iiif_manifest gem is built only for presentation and probably won't work for this for the most part. What is the best path to creating this response? We can create our own response, and check in with Princeton or IU Digital Collections team to see what their implementations look like. They could be using the Blacklight IIIF search gem or a similar library.

## Done Looks Like
- [x] Investigate how to build the JSON response from the Content Search service
- [x] What needs to be tweaked from current mockups?
- [x] Discussion with team on proposed response structure
- [ ] Create a new issue if needed to implement the response using the information returned from our Solr queries

elynema:
I wonder if we could leverage blacklight_iiif_search or if the way we need to format queries and results is too unique and we would just be fighting it.

Issue #5791, opened 2024-04-19
Develop Solr query for search within


elynema:
# Description

Investigate how to form a Solr query that would provide the results needed to create the IIIF content search response sample from #5786. 

```
{
  "@context": "http://iiif.io/api/search/2/context.json",
  "id": "http://localhost:3000/media_objects/hd76s004z/manifest/canvas/ft848q60n/search?q=April",
  "type": "AnnotationPage",

  "items": [
    // This item highlights the transcript by returning the whole cue
    {
      "id": "http://localhost:3000/media_objects/hd76s004z/manifest/canvas/ft848q60n/lfdasjf-lkfdalk-12klf-389hflkdjs",
      "type": "Annotation",
      "motivation": "highlighting",
      "body": {
        "type": "TextualBody",
        "value": "CASSIDY CLOUSE:  All right, well my name is Cassidy Clouse. We’re here virtually speaking. It is <em>April</em> 17, 2017. We have with us today Abby Clapp. So, Abby, where are you from?",
        "format": "text/plain"
      },
      "target": "http://localhost:3000/master_files/ft848q60n/supplemental_files/2/transcripts#t=00:00:00,00:00:16"
    },
// This item highlights the canvas like a marker in a playlist item
    {
      "id": "http://localhost:3000/media_objects/hd76s004z/manifest/canvas/ft848q60n/lfdasjf-lkfdalk-12klf-389hflkdjs",
      "type": "Annotation",
      "motivation": "highlighting",
      "body": {
        "type": "TextualBody",
        "value": "CASSIDY CLOUSE:  All right, well my name is Cassidy Clouse. We’re here virtually speaking. It is <em>April</em> 17, 2017. We have with us today Abby Clapp. So, Abby, where are you from?",
        "format": "text/plain"
      },
      "target": "http://localhost:3000/media_objects/hd76s004z/manifest/canvas/ft848q60n#t=00:00:00,00:00:16"
    },
    // This item highlights in a txt transcript by returning the whole paragraph
    {
      "id": "http://localhost:3000/media_objects/hd76s004z/manifest/canvas/ft848q60n/lfdasjf-lkfdalk-12klf-389hflkdjs",
      "type": "Annotation",
      "motivation": "highlighting",
      "body": {
        "type": "TextualBody",
[...]

PR #5840, opened 2024-05-23
Implement IIIF Content Search service

masaball:
Related issues: #5791, #5792, #5793

### DIFF
--- a/lib/avalon/transcript_search.rb
+++ b/lib/avalon/transcript_search.rb
@@ -53,37 +53,37 @@ module Avalon
       formatted_response = []
       search_results["highlighting"].each do |result|
         transcript_id = result.first.split('/').last.to_i
-        @mime_type = search_results["response"]["docs"].filter { |doc| doc["id"] == result.first }.first["mime_type_ssi"]
-        @canvas = "#{Rails.application.routes.url_helpers.media_object_url(master_file.media_object_id).to_s}/manifest/canvas/#{master_file.id}"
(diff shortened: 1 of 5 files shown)


lib/avalon/transcript_search.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    private
+
+    def items_builder search_results
+      formatted_response = []
+      search_results["highlighting"].each do |result|
+        transcript_id = result.first.split('/').last.to_i
+        @mime_type = search_results["response"]["docs"].filter { |doc| doc["id"] == result.first }.first["mime_type_ssi"]
+        @canvas = "#{Rails.application.routes.url_helpers.media_object_url(master_file.media_object_id).to_s}/manifest/canvas/#{master_file.id}"
+        @target = Rails.application.routes.url_helpers.transcripts_master_file_supplemental_file_url(master_file.id, transcript_id)
```
cjcolvar, line 58: I think it might be clearer and safer to make these be local variables and pass them as params instead of instance variables.

lib/avalon/transcript_search.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        text ||= cue
+
+        formatted_matches += [format_item(text, @target, time_cue: time_cue)]
+      end
+
+      formatted_matches
+    end
+
+    def format_item(result, target, time_cue: nil)
```
cjcolvar, line 84: I like how this method is just concerned about the structure of the json.

cjcolvar (commented):
Looks great!  This builds nicely on top of the work we previously did! :slightly_smiling_face:

cjcolvar (approved):

PR #5845, opened 2024-05-31
Index MODS recordChangeDate on parent media object

### LINKED ISSUE #5580: Create Timestamp Value in Solr for Object Updates
# Description
There is not currently a value in Solr that reflects a "last updated" time for object updates the record. For OAI-PMH harvesting or any data feed (like Avalon's Atom feed), it would be advantageous to have some type of date modified field in Solr that will cover new records, records with updated resource description metadata, or records that have become public (so changes to access control). Having a timestamp in Solr to cover the updates to any of the Fedora objects that make up the Avalon object would be a more elegant solution than relying on the current timestamp value, as every time the record gets re-indexed the Solr timestamp value is updated. Looking only to descriptive metadata updates and not updates on other parts of the record could be a simpler way to implement this that may meet most use cases.

Reported by Josh Westgard of UMD during their project to do OAI-PMH harvesting with their Avalon instance.

# Done Looks Like
- [ ] The Solr index contains a timestamp updated when an item's descriptive metadata is updated (possibly repurposing existing field instead of creating a separate one)
- [ ] Solr query or queries created for this purpose
- [ ] New timestamp value added to Atom feed (or Atom feed pointed to new value for update on object rather than update on index)

cjcolvar:
This field will act as a better indicator of when the object's metadata has changed instead of relying on media object's solr timestamp which gets updated any time the object gets reindexed. This new value is used in the ATOM feed and is available for use in OAI-PMH. 

Resolves #5580

### DIFF
--- a/app/models/mods_behaviors.rb
+++ b/app/models/mods_behaviors.rb
@@ -89,6 +89,8 @@ module ModsBehaviors
     # For full text, we stuff it into the mods_tesim field which is already configured for Mods doucments
     solr_doc['mods_tesim'] = self.ng_xml.xpath('//text()').collect { |t| t.text }
 
+    solr_doc['descMetadata_modified_dtsi'] = ActiveFedora::Indexing::DefaultDescriptors.iso8601_date(self.record_change_date.first)
+
     # TODO: Find a better way to handle super long fields other than simply dropping them from the solr doc.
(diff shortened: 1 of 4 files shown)


masaball (approved):
:+1:

PR #5846, opened 2024-05-31
Index avalon_resource_type the same as it is stored in fedora

### LINKED ISSUE #5836: Enforce Identical Strings for avalon_resource_type in Solr and Fedora
## Description
`avalon_resource_type` values are lower-case in Fedora but titlized in the solr document (https://github.com/avalonmediasystem/avalon/blob/c1a19e0a89fdea66ba40f7a1a93fd66c6dbe21aa/app/models/mods_behaviors.rb#L55)
This feels like a minor bug but one that could trip up services or processes which do exact string matching.

See https://github.com/avalonmediasystem/avalon/issues/5834 for original reported issue.

## Done Looks Like
- [x] Don't titlize avalon_resource_type values in Solr (make the index match Fedora)
- [x] Add helper method to Titlize in Blacklight display for resource type facet
- [x] `avalon_resource_type` in `media_objects/id.json` response is lower-case
- [x] Format facet on search results only has Titlized values (`Moving Image` and `Sound Recording`)

cjcolvar:
Shift titleization into front end through a helper method utilized by the Blacklight facet.

Resolves #5836

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -80,7 +80,7 @@ class CatalogController < ApplicationController
     #
     # :show may be set to false if you don't want the facet to be drawn in the
     # facet bar
-    config.add_facet_field 'avalon_resource_type_ssim', label: 'Format', limit: 5, collapse: false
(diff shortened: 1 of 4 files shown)


masaball (approved):
Looks good!

PR #5847, opened 2024-05-31
Playlist description fixes

### LINKED ISSUE #5839: Allow More Text for Playlist Description Field
## Description
Attempting to save a playlist description over 255 characters presents the user with this warning: `There are errors with your submission. Description is too long (maximum is 255 characters).` 

Playlist descriptions should be allowed to be greater length to better match the kind of textual data put into the field. For example,TEXT in MySQL can contain 65535 characters, which would be an appropriate available size.

## Done Looks Like
- [ ] Playlist description field is able to contain text greater than 255 characters

cjcolvar:
- Consistently label playlist description by using i18n translation
- Display new lines in playlist description by using `simple_format` helper
- Remove length restrictions on playlist description

Resolves #5839 and https://github.com/samvera-labs/ramp/issues/498

### DIFF
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -24,7 +24,6 @@ class Playlist < ActiveRecord::Base
 
   validates :user, presence: true
   validates :title, presence: true
-  validates :comment, length: { maximum: 255 }
   validates :visibility, presence: true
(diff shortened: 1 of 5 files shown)


masaball (approved):
:+1:

PR #5848, opened 2024-06-03
Make media object json response consistent between SpeedyAF proxy and ActiveFedora object

### LINKED ISSUE #5834: [Regression] MediaObject json changes in Avalon 7.7
Avalon 7.7 made all read-only requests be handled with data from Solr instead of Fedora including `/media_object/<id>.json`  While doing work on removing the ordered_aggregation relationship I found that https://github.com/avalonmediasystem/avalon/blob/c1a19e0a89fdea66ba40f7a1a93fd66c6dbe21aa/spec/controllers/media_objects_controller_spec.rb#L1191-L1211 was failing due to differences between the fedora objects and solr document.

- `record_identifier` isn't indexed so it is empty (`[]`) in Avalon 7.7 responses
  - Does `record_identifier` need to be in the media object json?
  - It gets passed in intercom so pushed records might be missing this, but ...
  - Whenever a mods document is saved it will add `record_identifier` if missing
  - Would it be better to not include in intercom push and instead always generate this for all MODS documents (new or pushed)?
- `avalon_resource_type` values are lower-case in Fedora but titlized in the solr document (https://github.com/avalonmediasystem/avalon/blob/c1a19e0a89fdea66ba40f7a1a93fd66c6dbe21aa/app/models/mods_behaviors.rb#L55)
  - This feels like a minor bug but one that could trip up services or processes which do exact string matching

## More Details
On MCO for the record Jon references above the difference in json when fetching from fedora vs. solr (speedy-af).  Unfortunately there are some more differences:
```
-   :avalon_resource_type=>["sound recording"],
+   :avalon_resource_type=>["Sound Recording"],
-   :identifier=>["AAY8650", "avalon:34123", "h53w827h39"],
+   :identifier=>["aay8650", "avalon:34123", "h53w827h39"],
-   :uniform_title=>["Bauernkantate"],
-   :statement_of_responsibility=>"J.S. Bach.",
+   :uniform_title=>[],
+   :statement_of_responsibility=>nil,
-   :format=>[],
-   :resource_type=>["sound recording"],
+   :format=>nil,
+   :resource_type=>["Sound Recording"],
-   :subject=>["Cantatas, Secular", "TBA"],
+   :subject=>["\n    Cantatas, Secular\n  ", "\n    TBA\n  "],
-   :record_identifier=>["http://127.0.0.1:8984/fedora4/rest/objects/avalon/nk/32/2d/62/nk322d627"],
+   :record_identifier=>[],
-   :bibliographic_id=>"AAY8650",
-   :bibliographic_id_label=>"other",
+   :bibliographic_id=>nil,
+   :bibliographic_id_label=>nil,
```

How critical are some of these? Which are surfaced in the UI vs which don't need to be represented?

## Done Looks Like
- [ ] Explore whether the fields above are the only ones that aren't matching
- [ ] Test against an example item in avalon-dev that has all available metadata fields filled
- [ ] Resolve missing or changed fields between Fedora and Solr

cjcolvar:
TODO: 
- [x] Figure out how to handle identifier_ssim since we previously downcased for indexing so we could downcase on query for matching (See comment on https://github.com/avalonmediasystem/avalon/issues/5834#issuecomment-2159057384)
- [x] Audit other MODS fields in JSON response to ensure they are getting indexed

Resolves #5834

### DIFF
--- a/app/models/concerns/media_object_mods.rb
+++ b/app/models/concerns/media_object_mods.rb
@@ -90,7 +90,7 @@ module MediaObjectMods
 
   # has_attributes :translated_title, datastream: :descMetadata, at: [:translated_title], multiple: true
   def translated_title
-    descMetadata.alternative_title
+    descMetadata.translated_title
   end
 
   def translated_title=(value)
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -318,8 +318,8 @@ class MediaObject < ActiveFedora::Base
       solr_doc["title_ssort"] = self.title
       solr_doc["creator_ssort"] = Array(self.creator).join(', ')
       solr_doc["date_ingested_ssim"] = self.create_date.strftime "%F" if self.create_date.present?
-      solr_doc['avalon_resource_type_ssim'] = self.avalon_resource_type.map(&:titleize)
-      solr_doc['identifier_ssim'] = self.identifier.map(&:downcase)
+      solr_doc['avalon_resource_type_ssim'] = self.avalon_resource_type
+      solr_doc['identifier_ssim'] = self.identifier
       solr_doc['note_ssm'] = self.note.collect { |n| n.to_json }
       solr_doc['other_identifier_ssm'] = self.other_identifier.collect { |oi| oi.to_json }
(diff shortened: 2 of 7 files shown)


masaball (approved):
:+1:

PR #5850, opened 2024-06-05
Overrides of ActiveFedora to avoid doing costly loads of list_source and master_files when saving

### LINKED ISSUE #5849: Saves of many-sectioned items still slow after migration from ordered_aggregation to JSON array of ids
The rendering of the edit page for an item is much faster but saving many-sectioned items is still slow after #5778.
100 section item (https://avalon-dev.dlib.indiana.edu/media_objects/707957642) still takes 21s for saving on Access Control page.
It appears that the save operation is still loading list_source (but not saving) twice which amounts to at least 9s of that 21s.
124 section item (https://avalon-dev.dlib.indiana.edu/media_objects/3f462541v) took 36s to save on Access Control page with 17s of that being loading list_source twice.
Adding another file though took 1.72 minutes of which only 17s was loading list_source twice.  It looks like all of the sections including associated objects were loaded as part of the save on the Manage Files page.  I'm thinking that this is probably what is taking the most time.  While adding a file is much worse than other save operations both need to be improved to bring save times down by removing loading and processing of list_source and any other time-consuming tasks.

cjcolvar:
Two overrides needed to realize the gains hoped for with  #5778 
- Skip attempting to update first/last properties of list_source on save
- Allow setting hasPart triples directly on media object to avoid `master_files=` which causes load of all master files and associated files (I was able to reuse overrides I had written as part of the Fedora 6 work. :slightly_smiling_face: )

The second change will skip writing to the proxy ActiveFedora::IndirectContainer object (`id/master_files`) which will become a snapshot of pre-migration state just like `list_source`.

Resolves #5849

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -249,10 +249,6 @@ class MediaObjectsController < ApplicationController
         if api_params[:replace_masterfiles]
           old_ordered_sections.each do |mf|
             p = MasterFile.find(mf)
-            # FIXME: Figure out why this next line is necessary
-            # without it the save below will fail with Ldp::Gone when attempting to set master_files in the MediaObject before_save callback
-            # This could be avoided by doing a reload after all of the destroys but I'm afraid that would mess up other changes already staged in-memory.
-            @media_object.master_files.delete(p)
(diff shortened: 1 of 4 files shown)


masaball (approved):
This looks good as far as I can tell

Issue #5842, opened 2024-05-24
Derivative Download Button Doesn't Appear for Valid Manager


joncameron:
## Description
Administrator users can see the Download button for derivative download on the Manage Files page, but Managers on that collection cannot. The Download button should display for Manager users and above on an item's Manage Files page.

## Example
On avalon-dev, 'joncameron' can't see the Download button at https://avalon-dev.dlib.indiana.edu/media_objects/wp988j841/edit?step=file-upload, but archivist1@example.com can. This was an item I created as the Manager-level 'joncameron' user.

## Done Looks Like
- [ ] The Download button for derivative download on the Manage Files page displays if the user is in the Manager role for the parent collection

joncameron:
👍

PR #5853, opened 2024-06-06
Fix derivative download for collection managers

masaball:
Related issue: #5842

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -325,7 +325,7 @@ class MasterFilesController < ApplicationController
   end
 
   def download_derivative
-    authorize! :manage, @master_file.media_object.collection
+    authorize! :download, @master_file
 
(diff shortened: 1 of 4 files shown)


cjcolvar (commented):
I think that using a custom ability for the download action makes more sense than relying on `:manage, Admin::Collection` which is only available to admin users before this PR.  Since Avalon hasn't granted `:manage, Admin::Collection` before I'm hesitant to do it for this feature.  Sorry that I missed this in the original PR.

What do you think about using a custom action instead?
 ```
can :download, [MasterFile, SpeedyAF::Proxy::MasterFile] do |master_file|
  @user.in?(master_file.media_object.collection.managers)
end
```
I think the check for `is_administrator?` in the view partial could also be dropped since administrators `can :manage, :all`

cjcolvar (approved):
Looks great!

PR #5860, opened 2024-06-11
Added more cypress tests, env files

charumitraravi:
**Change logs:**

1. Removed cypress_dev.config.js. 
2. Added cypress.dev.env.json, cypress.local.env.json, and cypress.staging.env.json inside the 'spec/cypress' directory. These files handle test data across different environments and are dynamically used by cypress.config.js based on the specified environment.
3. Updated cypress.config.js to fetch environment settings from process.env and merge them with config.env. This allows for dynamic configuration based on the specified environment during runtime.
4. Modified the docker-compose.yml to accept the CYPRESS_ENV variable during runtime. The command to run Cypress tests with a specified environment is now: **CYPRESS_ENV=dev docker-compose up cypress**.  Replace dev with staging or local to run tests using the respective env.json files. By default, the environment will be local if no environment is specified.
5. Cleaned up the playlist specification to remove false positives and failures. Added test cases in accordance with the manual test cases. 
6. Added additional cases to collection spec 
7. Introduced a new Browse spec and moved the "search by keyword" case from the navigation spec to align with the manual testing document.

**Note::**
 The user credentials inside the env.json files are placeholder values. These need to be replaced with actual values to run the tests. Another approach is to provide them as CYPRESS_ environment variables either through a .env file, as a run command, or as CI variables under secrets. I am currently working on the .env file approach so that we have a single source for loading credentials. This might involve updating the environment variables in the Docker Compose file and modifying the cypress.config.js file to load credentials from process.env.


cjcolvar (approved):
Looks great!  Really good job on figuring out a way to handle credentials via different env files!  I'm looking forward to how these tests keep growing and maturing.

Issue #5832, opened 2024-05-16
Poster Images Stretched In Media Player


joncameron:
## Description
Videos that do not match the ratio of the media player are stretched to fit.

## Example
https://avalon-dev.dlib.indiana.edu/master_files/4m90dv50r
- thumbnail (https://avalon-dev.dlib.indiana.edu/master_files/4m90dv50r/thumbnail) is 640x360
- original media is 720x1280 vertical orientation

### Before Playback
![Screenshot 2024-05-16 at 11 52 52 AM](https://github.com/avalonmediasystem/avalon/assets/1406618/4bc2eeb6-a5fa-43ac-b365-2a891addf783)

### During Playback
![Screenshot 2024-05-16 at 11 53 08 AM](https://github.com/avalonmediasystem/avalon/assets/1406618/67164ebe-655e-4660-808e-0eb48bb8b4aa)

## Done Looks Like
- [ ] Thumbnails are saved in the original aspect ratio with the same pixel width and height as the source

masaball:
@joncameron Can you retest this? I downloaded the media file from the linked item and have tried uploading it locally and into a new item on Avalon-dev and the thumbnail is created how I would expect:
<img width="607" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOU5JQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b26156aa0aa92af0910f023b05d0da58b4b593d9/Screenshot_20240607_133114.png" alt="Screenshot_20240607_133114.png" />

I also tested the results of creating a new thumbnail and that also generated the image with the proper aspect ratio. 

Maybe this example item was a weird one-off glitch?

joncameron:
After testing it looks like this is due to the media file and how Avalon is storing metadata. 

Two files to demonstrate: https://drive.google.com/drive/folders/1tlKzE2l-p61uSveianBbb6Db54G1YO2s
- pexels-roman-odintsov-7691954 (correct ratio stored in record)
- IMG_4190 (shows as the correct ratio in media player but not in thumbnail)

IMG_4190.MOV is hevc taken from my iPhone. It's a vertical video, but inside the MOV it's being stored as a horizontal ratio, then the rotation is marked in item metadata. Here's FFprobe output for this file:

```
[...]

PR #5866, opened 2024-06-12
Fix stretched posters

masaball:
Related Issue: #5832 

The root of the issue is that portrait video recorded on an iPhone (and possibly mobile devices in general) has metadata for an aspect ratio of "1280x720" with a flag for being rotated 90 degrees. The transcoded derivatives, however, have an aspect ratio that matches being a portrait video such as "606x1080". This disagreement in aspect ratio between the input and output causes the frame extraction to pull out an image of the wrong dimensions.

By setting the `display_aspect_ratio` off the derivatives, we will get a hopefully more accurate record. For existing items, we just compare the ratio on the master file against the derivative files to prevent extra/unrelated saves during the extraction.

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -312,8 +312,10 @@ class MasterFile < ActiveFedora::Base
     # with a rotation value. This can cause the aspect ratio on the master file to be incorrect. 
     # We can get the proper aspect ratio from the transcoded files, so we set the master file off the 
     # encode output.
-    high_output = Array(encode.output).select { |out| out.label.include?("high") }.first
-    self.display_aspect_ratio = (high_output.width.to_f / high_output.height.to_f).to_s
+    if is_video?
+      high_output = Array(encode.output).select { |out| out.label.include?("high") }.first
+      self.display_aspect_ratio = (high_output.width.to_f / high_output.height.to_f).to_s
+    end
 
     outputs = Array(encode.output).collect do |output|
       {
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -171,6 +171,7 @@ class MasterFile < ActiveFedora::Base
   after_transcoding :generate_waveform
   after_transcoding :update_ingest_batch
   # Generate and set the poster and thumbnail
+  after_transcoding :set_display_aspect_ratio
   after_transcoding :set_default_poster_offset
   after_transcoding :update_stills_from_offset!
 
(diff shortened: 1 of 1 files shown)


app/models/master_file.rb
```diff
@@ -605,9 +606,11 @@ def extract_frame(options={})
     end
 
     frame_size = (options[:size].nil? or options[:size] == 'auto') ? self.original_frame_size : options[:size]
+    # Handle existing files that may have inaccurate display_aspect_ratio saved.
+    aspect_ratio = actual_aspect_ratio != self.display_aspect_ratio.to_f ? actual_aspect_ratio : self.display_aspect_ratio
```
cjcolvar, line 610: Since this seems like an edge case I wonder if a backfill might be better than trying to handle the case dynamically.  After fixing `display_aspect_ratio` the backfill could regenerate the poster and thumbnail.

masaball: Are you thinking another migration rake task for this or would a different approach be better? And, regardless of approach, should we make an issue specifically for creating the backfill or just wrap it up in this work?

masaball: I think just having instructions in the upgrade notes is reasonable. You're right that this situation is, in general, pretty edge case so I'm fine with conserving the effort.

app/models/master_file.rb
```diff
@@ -308,6 +308,13 @@ def update_progress_on_success!(encode)
     # is stored as an integer string
     self.duration = encode.input.duration.to_i.to_s if encode.input.duration.present?
 
+    # Input videos that are in portrait orientation can have metadata showing landscape orientation
+    # with a rotation value. This can cause the aspect ratio on the master file to be incorrect. 
+    # We can get the proper aspect ratio from the transcoded files, so we set the master file off the 
+    # encode output.
+    high_output = Array(encode.output).select { |out| out.label.include?("high") }.first
+    self.display_aspect_ratio = (high_output.width.to_f / high_output.height.to_f).to_s
```
cjcolvar, line 316: This looks good.  I think it just needs to be in a check for video.
```suggestion
    if is_video?
      high_output = Array(encode.output).select { |out| out.label.include?("high") }.first
      self.display_aspect_ratio = (high_output.width.to_f / high_output.height.to_f).to_s
    end
```

masaball: Yep, good catch :100:

cjcolvar (approved):

PR #5870, opened 2024-06-14
Migrate remaining media objects to section_list array property

### LINKED ISSUE #5856: Create process for migrating remaining media objects's section lists
## Description

#5778 implements an in-place lazy migration of media object's `ordered_aggregation` modelling to a new property, `section_list`.  This migration is done in memory and persists when an object is saved.  This will migrate some of a repository's items but there will be many that won't get edited and thus won't get migrated within a set period of time.  At some point we'll want to ensure that the migration has been run on all media objects.  This will require a process to force the migration on all remaining un-migrated media objects.  The number of media objects that need to be migrated could be significant and may take a large amount of time to process (e.g. days) for large repositories.  Thus this process should be able to be run in a sidekiq background job or as a rake task run through `nohup`.  

Prior work: #5749, #5778, #5849, #5850

## Done Looks Like
- [ ] Task or job written for migration on all un-migrated media objects
- [ ] Task or job finds all media objects without a `section_list_ssim` field in solr then loads them through ActiveFedora and saves them

cjcolvar:
Resolves #5856 

I ran this locally and it correctly identified that only 1 of 2 items needed to be migrated and then did it successfully.  A subsequent run didn't find any to migrate as expected.

### DIFF
--- a/lib/tasks/avalon_migrations.rake
+++ b/lib/tasks/avalon_migrations.rake
@@ -93,5 +93,26 @@ namespace :avalon do
         puts("#{error.length} files failed to save. Refer to parent_id_backfill_errors.log for full list of SupplementalFile IDs.")
       end
     end
+    desc "Migrate MediaObjects from list_source linked list of sections to JSON array section_list property"
+    task media_object_section_list: :environment do
+      error_ids = []
+      mo_count = MediaObject.count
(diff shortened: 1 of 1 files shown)


masaball (approved):
Looks great!

PR #5873, opened 2024-06-14
Add max_upload_size to config/settings.yml and all for disabling limit

### LINKED ISSUE #5790: Change Default Web Upload File Size Limit
# Description
Avalon's file size limit for web upload is currently set to a default of 2 GB. This seems unnecessarily low and and an artificial restriction at this point, and this value can be removed unless specified in the system config.

# Done Looks Like
- [x] Limit is disabled if [max_upload_size](https://samvera.atlassian.net/wiki/spaces/AVALON/pages/1957954807/Configuration+Files) is not present in config (:none)
- [x] Avalon respects :none value for web upload files (no rejection of files based on size in server file handling)

cjcolvar:
Resolves #5790

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -121,3 +121,5 @@ recaptcha:
 derivative:
   # Choose whether collection managers and admins can download high quality derivatives
   allow_download: true
+# Maximum size for uploaded files in bytes (default is disabled)
+#max_upload_size: 2147483648 # Use :none or comment out to disable limit
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -176,12 +176,8 @@ class MasterFile < ActiveFedora::Base
 
   after_processing :post_processing_file_management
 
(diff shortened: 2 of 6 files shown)


config/settings.yml
```diff
@@ -121,3 +121,5 @@ recaptcha:
 derivative:
   # Choose whether collection managers and admins can download high quality derivatives
   allow_download: true
+# Maximum size for uploaded files in bytes (default is disabled)
+max_upload_size: 2147483648 # Use :none or comment out to disable limit
```
masaball, line 125: Should this be `:none` since the comment specifies the default is for the limit to be disabled?

cjcolvar: I was debating about this since it could be a surprising change for sysadmins but the ticket says disable so I'll make that change and will add a warning in the upgrade notes.

masaball (approved):
Looks good

PR #5875, opened 2024-06-17
Display Bibliographic ID distinctly from Other Identifiers and make searchable

### LINKED ISSUE #5813: Index Bibliographic Identifier Value
## Description
Avalon does not index the bibliographic id field or present it on the item page or in the manifest. This is a quirk of the way Avalon was designed, where the bibligraphic ID is always assumed to be entered for import from some external source of metadata. Upon import, the ID is added to the Other Identifiers listing and thereby added to the index for search and discovery.

This doesn't meet a use case where collection staff have an identifier for an item, which is not meant to be used for import from another system. This could be because it's an internal identifier to the department or unit or perhaps only available in some other non-machine accessible data source. Users should be able to put a value in that field and expect it to be then searchable in the repository, even if it is not an identifier used with the Import function.

The expectation for Bibliographic ID is that it is searchable and included in the descriptive metadata in some way.

The help text for Bibliographic ID in the Resource Description form (update this?)
> Bibliographic ID should be used for an external record identifier that can connect the Avalon item to a catalog record or other record for the original item. Type specifies the type of external record identifier. 

One option: upon Save for Resource Description, check Other Identifiers and add in the type and value of Bibliographic ID if it is not already present. This would get the value indexed 

Another option: make Bibliographic ID more prominent; give it its own dt/dd in the metadata display component. From one perspective it seems odd that this value would not be surfaced in descriptive metadata except for under the "Other Identifiers" listing.

## Done Looks Like
- [x] Bibliographic ID value added to solr index for the media object (existing or new separate field)
- [x] Bibliographic ID value added to all_text_timv
- [x] Add Bib ID to IIIF manifest as a discrete metadata field (placed before Other Identifiers listing)
- [x] De-dupe Other Identifiers listing to remove the bib id if it is present there as well

cjcolvar:
Resolves #5813

### DIFF
--- a/app/models/iiif_manifest_presenter.rb
+++ b/app/models/iiif_manifest_presenter.rb
@@ -107,9 +107,14 @@ class IiifManifestPresenter
 
   def display_other_identifiers(media_object)
     # bibliographic_id has form [:type,"value"], other_identifier has form [[:type,"value],[:type,"value"],...]
-    ids = media_object.bibliographic_id.present? ? [media_object.bibliographic_id] : []
-    ids += Array(media_object.other_identifier)
+    ids = Array(media_object.other_identifier) - [media_object.bibliographic_id]
+    return nil unless ids.present?
     ids.uniq.collect { |i| "#{ModsDocument::IDENTIFIER_TYPES[i[:source]]}: #{i[:id]}" }
+  end
+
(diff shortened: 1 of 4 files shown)


masaball (approved):
Looks good!

PR #5878, opened 2024-06-18
Bump VideoJS version dependency

masaball:
When we upgraded Ramp to video.js@^8.10.0, we neglected to bump the dependency in Avalon's package.json

Related issue: samvera-labs/ramp#526


Dananji (approved):
Good catch!

Issue #5869, opened 2024-06-13
Playlist paging buttons not working on iPhone (Safari)


elynema:
# Description

The buttons to page through playlists on avalon-dev are not working in Safari on iOS (note we test iOS 17 only). This includes both the buttons for individual pages and the prev/next buttons. 

These buttons are working fine on Android (Chrome) and in Chrome and Safari on the desktop.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFJpQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7cd5a3430cff06389892ecce077319602ef882c6/Screen%20Recording%202024-06-13%20at%204.31.07%20PM.mov)

# Done Looks Like

 - [x] prev/next buttons work to page through playlists for users with many playlists on iOS iPhone (Safari)
 - [x] buttons for individual pages of pages of playlists (page 1, 2) etc for on iOS iPhone (Safari)

masaball:
This was being caused by the width of the paging controls dropping them to the next line, where they are covered by the footer. I have a PR to set the footer to a negative z-index to ensure that it will be behind everything else, but should we also adjust the paging controls so that they don't get dropped to the next line? 

We would need to change the number of entries dropdown so it does not have explanatory text. And we could lessen the number of pages displayed in the controls. So something like

<img width="391" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFI4QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--260a363cf3bd1f1f6ba53da3264bc9448c9d420d/Screenshot_20240618_145115.png" alt="Screenshot_20240618_145115.png" />

or

<img width="390" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFY4QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6ad8936b346625b5c00a3549836f846227bc1bed/Screenshot_20240618_145240.png" alt="Screenshot_20240618_145240.png" />

elynema:
[...]

PR #5879, opened 2024-06-20
Set footer to negative z-index so paging works

masaball:
If controls such as the playlist paging ended up on the same level as the footer they were not clickable because the padding around the footer was covering them up. Setting the footer with a large negative z-index ensures that it will always be below the paging controls.

Related Issue: #5869


Dananji (approved):
Looks good!

PR #5880, opened 2024-06-20
Use player.src() to enable action buttons instead of player.readyState()

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5641#issuecomment-2166208478


cjcolvar (approved):
:+1:

Issue #5883, opened 2024-06-25
[BUG] Error when trying to edit a media object with no language on supplemental file


elynema:
**Describe the bug**
For at least one instance of a media item on avalon-dev, I get a 500 error (NoMethodError) when I go to Edit -> Manage Files a specific media object. The error seems to be related to language of supplemental file.

<img width="1402" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK0tQQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--77e46fc1a09cd684e231b7119a0dd3517b004625/image.png" alt="image.png" />

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Log in as archivist admin user
2. Go to https://avalon-dev.dlib.indiana.edu/media_objects/wm117n96b/edit
3. Click on Manage Files
4. You will see an error message.

**Expected behavior**
User should see the Manage File page, not the error. 

**Environment (please complete the following information):**
 - Device / platform: Mac laptop
 - Browser: Chrome
- [ x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Unknown

elynema:
Another example to test for this:

https://avalon-dev.dlib.indiana.edu/media_objects/3n203z084/edit?step=file-upload

joncameron:
👍

PR #5891, opened 2024-06-27
Display default language when supplemental file is missing language field

masaball:
Related issue: #5883


cjcolvar (approved):
Looks great!

Issue #5884, opened 2024-06-26
Long playlist description overflows when expanded


elynema:
# Description

In #5857, we added a 'show more' and 'show less' toggle for long playlist descriptions and only show 4 lines of the description by default. However, we still have formatting issues when a user expands the playlist description, especially if the font size is magnified for accessibility.

For longer descriptions (~120 words, the Playlist items get hidden, and only the title is visible)

<img width="200" alt="Screenshot 2024-06-26 at 10 36 49 AM" src="https://github.com/avalonmediasystem/avalon/assets/52535309/10334856-d765-4255-a1a1-264a70488b54">

For more than 150 words, the text starts to overflow beyond the page edge. 
<img width="300" alt="Screenshot 2024-06-26 at 10 40 02 AM" src="https://github.com/avalonmediasystem/avalon/assets/52535309/07bbce0f-5760-4ce3-8649-ddbb7ad96ddf">

elynema:
I think I vote for #1 if possible myself. We already have a lot of scrolling elements to deal with.

joncameron:
👍

PR #5894, opened 2024-07-01
Fix playlist description display issues

masaball:
Related issue: #5884 

This commit limits the height of the description section so that it does not overwhelm the playlist items section and does not extend past the bottom of the page. With limiting the height we also enable scrolling of the description when it overflows and mimic the structured nav "scroll to see more" message.


Dananji (approved):

Issue #5892, opened 2024-06-27
Remove transcripts from index when the supplemental files are deleted from MCO


elynema:
# Description

A user can delete a transcript or a caption that is marked to treat as a transcript from the Manage Files interface. When this happens, that text should also be removed immediately from the Solr index.

# Done Looks Like

- [x] Deleted transcripts are removed from index
- [x] Deleted captions that are marked as transcripts are removed from the index
- [x] Captions that have the 'Treat as transcript' box unchecked and then saved should also be removed from index
- [x] Deleted transcripts / captions should also cause the item to move to 'no' in the Has Captions and Has Transcripts facets (if no other transcripts / captions exist)
- [x] User should not have to click 'Continue' on the Manage Files page for these changes to be propagated to Solr.
~~- [ ] Investigate how to remove old deleted supplemental files from index on avalon-dev~~

elynema:
Has Caption / Has transcript is based off info in the media object in the index. This is not updated until you hit save/continue and the media object is updated in the index.

elynema:
This issue will not remove old previously deleted indexed supplemental files from the index. Dananji has noted deleted transcripts are still showing results for content search.

PR #5900, opened 2024-07-03
Remove transcripts from indexes

masaball:
Related issue: #5892

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -110,6 +110,7 @@ class SupplementalFilesController < ApplicationController
     @supplemental_file.attach_file(attachment) if attachment
 
     raise Avalon::SaveError, @supplemental_file.errors.full_messages unless @supplemental_file.save
+    raise Avalon::SaveError, @object.errors[:supplemental_files_json] unless @object.save
 
     flash[:success] = "Supplemental file successfully updated."
     respond_to do |format|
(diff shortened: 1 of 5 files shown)


app/controllers/supplemental_files_controller.rb
```diff
@@ -110,6 +110,7 @@ def update
     @supplemental_file.attach_file(attachment) if attachment
 
     raise Avalon::SaveError, @supplemental_file.errors.full_messages unless @supplemental_file.save
+    raise Avalon::SaveError, @object.errors[:supplemental_files_json] unless @object.save
```
cjcolvar, line 113: If the expected change is to the parent object's solr document then maybe it makes more sense to have the SupplementalFile model have an after_save callback which calls `update_index` on the parent object.  But maybe that would put too strong of a link between the two models (I think callbacks between models like this have caused us problems in the past and has the potential for unintentionally causing lots of saves/index updates) and doing the update here in the controller makes more sense.  If that's the case then I think the parent object doesn't actually need to be saved but only have `update_index` called on it.
```suggestion
    # Updates parent object's solr document
    @object.update_index
```

cjcolvar (commented):
Looks really good!  I had one question/suggestion around updates to a supplemental file but otherwise seems great!

app/controllers/supplemental_files_controller.rb
masaball: Yeah, I hesitate to use an after save callback. I started to try that when I was first working on this but it was causing a lot of issues with the testing suite and I imagine the finickiness could potentially translate to the live environs.

I will try `update_index` and see how that goes. I missed that that was a method that exists but is exactly what I was looking for.

cjcolvar (approved):
Nice!  The failed test is an unrelated flaky test so feel free to go ahead and merge.

Issue #5899, opened 2024-07-02
Link Back to Original Item Page in Embedded Player


joncameron:
# Description
In the embedded media player with video, include a link back to the section on the media object page as was present in the MediaElement.js embed.

## Done Looks Like
- [ ] Embedded player includes link back to item in MCO - button in toolbar

joncameron:
![Screenshot 2024-07-09 at 10 05 53 AM](https://github.com/avalonmediasystem/avalon/assets/1406618/3c0c15c9-2c0a-4fdc-b873-c4b6c9d6a3e1)
Link is appearing in the controls and functioning as expected.

PR #5906, opened 2024-07-08
Fix embedded player after Ramp rework

masaball:
Either the Video.js upgrade or other reworking of Ramp broke certain features of the embedded player. This work:

1. Fixes the "View in Repository" button so that the icon renders and button instances are not continuously added.
2. Fixes behavior so the quality selector is not present on audio items.
3. Removes the playback rate option from embedded audio player.

Related issues: #5899, #5732


cjcolvar (approved):
Looks good to me as far as I can tell!

Issue #5909, opened 2024-07-09
Don't Show Zero Counts for "Found In" Search Results


joncameron:
# Description
If there is a zero count for any of the 3 parts in the "Found in" displayed in Blacklight search (metadata, transcripts, sections), do not display that part.

# Current
<img width="762" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFBGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c39b503d8550219fcb8da8d5b24877bccfd98578/Screenshot%202024-07-09%20at%2010.40.56%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.40.56 AM.png" />

# Desired
<img width="753" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFhGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1923b4b9deb6428d6c535f1022356c907516bf2d/Screenshot%202024-07-09%20at%2010.45.50%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.45.50 AM.png" />

# Done Looks Like
- [x] Of the three fields that return match hits, only display those with non-zero counts

joncameron:
![Screenshot 2024-07-11 at 12 10 43 PM](https://github.com/avalonmediasystem/avalon/assets/1406618/5542565e-0d45-4990-b66e-ee078dde0a8a)

👍

PR #5913, opened 2024-07-09
Remove zero counts from "Found In" results

masaball:
Related issue: #5909


cjcolvar:
Test failure is unrelated so don't worry about it.

Issue #5908, opened 2024-07-09
Audio embedded player showing up with a scrollbar


joncameron:
# Description
Audio embeds are showing up with a scrollbar in the iframe element; it seems like there might be some extra padding or other styling that needs to be removed so that there is no scrolling that appears for the iframe embed.

~~The audio slider position circle is also out of position; it seems to be a little lower than expected and not centered on the slider/rail.~~

# Screenshots
<img width="603" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHpGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--28567641892565b80305e242ffe422f43e027498/Screenshot%202024-07-09%20at%2010.09.45%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.09.45 AM.png" />

<img width="606" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDNGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0cfaa7aa576f9dca7e2bc71ce32076b45b168a6e/Screenshot%202024-07-09%20at%2010.09.36%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.09.36 AM.png" />

# Done Looks Like
- [x] Audio embeds do not have hidden or scrollable elements within the given size iframe size (600 x 40 default) (2 points)
~~- [ ] Volume slider nub is centered on the volume slider rail (2 points in Ramp repo, to be split from this issue)~~ Moved to its own ticket: samvera-labs/ramp#559

joncameron:
Volume panel issue is coming from Ramp styling; can be fixed there. CSS needs to be scoped in Ramp so it doesn't get changed when the component is used in a different application.

masaball:
Note for testing: If you are using the W3Schools Tryit editor to test, you may need to clear your cache before testing in order for things to work properly. I ran into some issues with the player not loading at all or the current CSS not working because the editor page had cached conflicting copies of the embed.

joncameron:
![Screenshot 2024-07-12 at 10 57 33 AM](https://github.com/user-attachments/assets/fbcec15a-af82-4366-a350-ad7aaade8fa1)
👍

PR #5914, opened 2024-07-10
Remove margin and padding from embed body element

masaball:
Related issue: #5908

Certain html elements have built in margins or padding not reflected in the CSS readouts of inspect element. `<body>` is one of these elements. This built in margin was causing the embedded audio player to overflow the iframe it was embedded in, resulting in unintended scrolling behavior. Setting the embed layout's body to have no margin and no padding allows the embedded media to properly fill the iframe and avoid scrolling.


cjcolvar (approved):
Looks good to me.

Issue #5918, opened 2024-07-10
API Request for PUT Update Returns HTML


joncameron:
# Description
PUT API requests to update supplemental file binary files are returning an HTML response instead of the expected response, `{"id": :supplemental_file_id}`.

# Example
`curl -H "Avalon-Api-Key:1234abcd" -X PUT -F "file=@lunchroom.srt;type=text/plain" https://avalon-dev.dlib.indiana.edu/master_files/q237hs039/supplemental_files/360`

This command updated the supplemental file, but returned the following response:
`<html><body>You are being <a href="https://avalon-dev.dlib.indiana.edu/media_objects/0v838060s/edit?step=file-upload">redirected</a>.</body></html>%`

masaball:
I think this was a documentation error on my part. For API binary file updates we expect the user to include `-H "Accept: application/json"` in the curl request. If that is included in that PUT command, the response should be the expected `{ id: #sup_file_id }` response.

I think Chris and I changed the behavior to check the "Accept" header after I'd put together the example commands on the original ticket and I forgot to update the example. I will update the example on the other ticket, and add a failure case for if the "Accept" header is not included in that command.

joncameron:
Worked as expected! I'll make sure to note this when I update the API docs on the wiki.

PR #5919, opened 2024-07-10
Fail SupplementalFile Binary update if no Accept header

masaball:
Users updating supplemental files via CLI should use `-H "Accept:application/json"` in their requests to ensure a JSON response from an endpoint. For updating the binary file specifically, the request is through the HTML format endpoint and so having Accept headers needs to be required to force a JSON response in the CLI environment.

Related issue: #5918


cjcolvar (approved):
:+1: This seems like a good enough way to handle this.

PR #5921, opened 2024-07-10
Force vertical alignment of horizontal volume slider thumb to top

cjcolvar:
I couldn't find anything within Avalon's CSS that was overriding Ramp's CSS to cause this difference so I added a new override.

Before:
Screenshot from 2024-07-10 17-19-48

After (including https://github.com/samvera-labs/ramp/pull/562 which is not part of this PR):
Screenshot from 2024-07-10 17-19-14

Resolves https://github.com/samvera-labs/ramp/pull/559

### DIFF
--- a/app/javascript/components/Ramp.scss
+++ b/app/javascript/components/Ramp.scss
@@ -38,6 +38,9 @@
         min-height: 2.9em;
       }
     }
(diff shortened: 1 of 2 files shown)


Dananji (approved):
Looks good.

PR #5923, opened 2024-07-11
Add new fields to caption/transcript batch ingest

masaball:
Since implementing support for uploading supplemental files via batch ingest we have added additional editable fields to the various types of supplemental file. This commit adds support for setting these fields in batch ingest:

1. Transcript language
2. Marking captions to be treated as transcripts
3. Ensured that captions could be marked machine generated

### DIFF
--- a/lib/avalon/batch/entry.rb
+++ b/lib/avalon/batch/entry.rb
@@ -276,14 +276,14 @@ module Avalon
         filename.dup.insert(filename.rindex('.'), ".#{quality}")
       end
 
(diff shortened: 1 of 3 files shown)


cjcolvar (commented):
Looks good except for the boolean casting issue in my comment.  Do you think that should be a separate PR?

cjcolvar (approved):
Boolean casting will be handled in another PR so this one is good to go for merging. :tada:

Issue #5922, opened 2024-07-11
[BUG] Selected items page giving a 400 error


elynema:
**Describe the bug**
Select the Selected Items page and the page does not load. Provides a 400 error. This occurs over multiple users regardless of whether there are actually any selected items.

<img width="1402" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNHJSQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--02f12472d48a86b9f3352c5199b0864738eff6d4/image.png" alt="image.png" />

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Login to avalon-dev as any user
2. Go to the Selected Items page

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: Mac
 - Browser: chrome and safari
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Is this issue associated with a specific release? [e. g. 7.6]

**Additional notes**
Any additional information or notes that may be helpful in diagnosing or fixing the problem.

joncameron:
👍

PR #5924, opened 2024-07-11
Skip transcript search in bookmark context

masaball:
The "Selected Items" page was returning a 400 error because the search service that populates the page runs through our SearchBuilder model and generates a `q` solr_parameter. This triggered the transcript search logic. Something about the `q` generated by the search service broke the query building for `transcript_tsim`, possibly that the search service generates `q="{!lucene}(#object_ids)"`. Checking if we are in the bookmark controller context and skipping the transcript searching allows us to bypass the error.

Related issue: #5922


cjcolvar (approved):
:+1:

Issue #5920, opened 2024-07-10
Section-level Scope for Create Timeline Not Working


joncameron:
# Description
When creating new Timelines from an object page, the section-level scope option doesn't work, and instead the application always populates the Timeline with the range that populates the "Custom" field (current position to end of media).

# To Reproduce
- Go to https://avalon-dev.dlib.indiana.edu/media_objects/xw42n7934
- Move the playhead to a position in the media
- Select "Create Timeline"
- Selection the section as scope, "Over 1 Hour of Audio - Over 1 Hour of Audio"
- Select "Create Timeline"
- The Timeliner loads with the selection, but the range is always from the current position to the end, instead of the full range of the entire section

# Environment
- avalon-dev
- latest Chrome on macOS

# Done Looks Like
- [x] Selection the section as scope uses the full duration/range when populating Timeliner

joncameron:
Can no longer reproduce! Works as expected.

PR #5926, opened 2024-07-12
Set timelines without structure from full length of item

masaball:
Related issue: #5920 

Unstructured items were generating timelines based off of the current playhead for non-custom timelines when they should have been being set off the full length of the item. Adding a case for unstructured items to set the scope off the full length of the item takes care of it.


cjcolvar (approved):
Good catch!

PR #5928, opened 2024-07-12
Don't show "Found in:" if no found in hits exist

cjcolvar:
This is an edge case but putting it in just to be safe.  
I found an example:  Search "avalon-dev.dlib.indiana.edu" on avalon-dev.dlib.indiana.edu .  There are currently 4 results which all have no found in hits in either metadata, structure, or transcript.


masaball (dismissed):
Good catch!

masaball (approved):

Issue #5871, opened 2024-06-14
Delete does not cascade to child supplemental files


masaball:
# Description

When a media object or master file is deleted associated supplemental files are not removed, leaving orphan files in the system. Will leave orphaned supplemental files in database and on disk/in s3 bucket.

Delete callback may not work; may need to manually add a loop through supplemental files in the delete action in the controllers.

Old orphaned supplemental files could be cleaned up in the future by looking for files that aren't linked to any media files / master files.

# Done Looks Like

- [x] When a media object or master file is deleted, any associated supplemental files (caption, transcript, other supplemental file) should be deleted from ActiveStorage.

elynema:
Not sure how often we are deleting things that have supplemental files, and this does not seem to be a new problem, so this is probably not a 7.8 requirement.

joncameron:
* I added a file to an item
* confirmed the existence of the file in `/container-data/minio/supplementalfiles`
* deleted the item
* confirmed that the file was no longer in `supplementalfiles`

PR #5939, opened 2024-07-16
Delete supplemental files when parent object is deleted

masaball:
Related issue: #5871 

Do not merge until after 7.8 release.

### DIFF
--- a/app/jobs/bulk_action_jobs.rb
+++ b/app/jobs/bulk_action_jobs.rb
@@ -135,7 +135,7 @@ module BulkActionJobs
       documents.each do |id|
         media_object = MediaObject.find(id)
         supplemental_files = media_object.supplemental_files
-        DeleteChildFiles.perform_later(supplemental_files, nil)
+        DeleteChildFiles.perform_now(supplemental_files, nil)
 
         if media_object.destroy
           successes += [media_object]
--- a/app/jobs/bulk_action_jobs.rb
+++ b/app/jobs/bulk_action_jobs.rb
@@ -134,6 +134,9 @@ module BulkActionJobs
       successes = []
       documents.each do |id|
         media_object = MediaObject.find(id)
+        supplemental_files = media_object.supplemental_files
+        DeleteChildFiles.perform_later(supplemental_files, nil)
+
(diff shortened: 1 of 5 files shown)


app/jobs/bulk_action_jobs.rb
```diff
@@ -134,6 +134,9 @@ def perform(documents, _params)
       successes = []
       documents.each do |id|
         media_object = MediaObject.find(id)
+        supplemental_files = media_object.supplemental_files
+        DeleteChildFiles.perform_later(supplemental_files, nil)
```
cjcolvar, line 138: I think we can skip the overhead of enqueuing new jobs here and run this inline since we're already inside a background job.
```suggestion
        DeleteChildFiles.perform_now(supplemental_files, nil)
```

cjcolvar (commented):
This works for requests to delete MasterFiles but there are also requests to delete MediaObjects.  (See https://github.com/avalonmediasystem/avalon/blob/37cfbea682640796d9d33f4069c8a773537f581e/app/models/media_object.rb#L190)
Do you think it would work to move it into the MasterFile model?

spec/controllers/master_files_controller_spec.rb
```diff
@@ -228,6 +228,23 @@ class << file
       expect(controller.current_ability.can?(:destroy, master_file)).to be_falsey
       expect(post(:destroy, params: { id: master_file.id })).to render_template('errors/restricted_pid')
     end
+
+    context "master file with child supplemental files" do
+      let!(:master_file) { FactoryBot.create(:master_file, :with_media_object, :cancelled_processing, supplemental_files: [supplemental_file]) }
+      let(:supplemental_file) { FactoryBot.create(:supplemental_file) }
+
+      around(:example) do |example|
+        # In Rails 5.1+ this can be restricted to whitelist jobs allowed to be performed
```
cjcolvar, line 237: I guessing this was copied from other examples in the test suite, but I think the comment can be dropped as it isn't adding much (and removes the problematic "whitelist").
```suggestion
```

cjcolvar (approved):
Looks great!

Issue #5941, opened 2024-07-16
Batch Ingest Errors


joncameron:
**Describe the bug**
When using batch ingest, some of the rows are not processing and returning with errors on the test manifest.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Add the CSV below as a file to avalon_core-devel-8120/container-data/minio/masterfiles/dropbox/7_8_Test_Collection on muskrat
2. Receive an error via system email:
```
Initial processing of your batch ingest failed. Details are below. Please see the Batch Ingest Package Format for further help.

To replay this batch, please rename your spreadsheet to 7bb059c1-c6c5-4c0a-8d94-9949dc006d3b_manifest.csv and reupload it after making desired changes. Only unpublished items can be updated via this method. Retain all rows in the spreadsheet including those that have no changes.
The following rows of your spreadsheet failed:
Row	Error
3 	Permission denied @ dir_s_mkdir - /home/app/avalon/storage/w8 
```

**Expected behavior**
- The item is created and populated successfully with the values from the manifest

**Screenshots/Screen Recording**
If applicable, add screenshots or a screen recording to help explain your problem.

**Environment (please complete the following information):**
 - Firefox on macOS
 
- [x] avalon-dev
- [x] avalon-staging

**Release:**
7.8

Found while testing #5862.

joncameron:
At first I thought this might be because the files copied as owned by 'joncamer', but even with everything owned by app_avalon:avalon_dropbox it still is giving this error.

joncameron:
I tried again and removed the caption file-related columns to see if that was the issue. I still ran into the same error:

> Row	Error
> 3 	Permission denied @ dir_s_mkdir - /home/app/avalon/storage/ha/rt

joncameron:
I'm no longer receiving the errors. Batch is still broken for "skip transcoding" with multiple files (known issue: https://github.com/avalonmediasystem/avalon/issues/4970) but otherwise records are getting created as expected.

PR #5942, opened 2024-07-16
Make sure that ActiveStorage service is set

cjcolvar:
Avalon sets the ActiveStorage service based up the configuration in Settings in config/application.rb, but it appears that the Settings aren't always loaded by the time that config/application.rb runs.  Adding in this initializer ensures that the service in the Settings is configured in Rails.

Related to #5941


masaball (approved):
:+1:

Issue #5885, opened 2024-06-26
[BUG] Add to Playlist, Create Timeline, and Create Thumbnail buttons sometimes do not get enabled


elynema:
**Describe the bug**
Sometimes, the Add to Playlist, Create Timeline, and Create Thumbnail buttons never get enabled on page load. So far, this is seen on Android and iOS mobile devices, mostly after reloading the page numerous times.

Once this happens, reloading the page does not seem to re-enable the buttons. Starting playing the video and then reloading the page may re-enable the buttons.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. As an example where we have seen this happen, go to https://avalon-dev.dlib.indiana.edu/media_objects/fj236208t on iOS/Safari or Android Chrome.
2. Reload the page repeatedly until you see that these buttons remain gray.

**Expected behavior**
These buttons should become enabled once the player is ready. We could conceivably enable them immediately on page load in the mobile context, where we seem to be having issues, but if they are enabled too early before the player is ready, they will have bad data and not work properly.

**Questions**

- How many people are using these 3 buttons in the mobile context? Can we determine this via Google Analytics to help figure out how much of an issue this is?

**Environment (please complete the following information):**
 - Device / platform: Android / iOS iPhone
 - Browser: Chrome / Safari
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Is this issue associated with a specific release? [e. g. 7.6]

**Additional notes**
Any additional information or notes that may be helpful in diagnosing or fixing the problem.

elynema:
Dananji is not sure how we could fix this better than we already have. The buttons are supposed to be on an interval that checks every second to see if the player is ready and then enable themselves.

joncameron:
I tried on iOS Safari and Chrome on Android and could reproduce after many refreshes. I'll also check the analytics and add that information.

joncameron:
Can no longer reproduce 👍

PR #5947, opened 2024-07-18
Fix enabling of media object action buttons

masaball:
On page load, both initial and refreshed, in mobile browsers there was some sort of race condition that was disabling the Add to Playlist, Timeline, and Thumbnail buttons immediately after being enabled. Changing the event listener for disabling the buttons to be triggered on the player emptying instead of loading seems to fix that race condition, but uncovered a separate one when switching canvases. This second race condition behaved similarly to the first, where the enabling function would be run before the disabling function. Adding a fallback enabling function that triggers on loadstart seems to fix this second race condition. There may still be some flakiness, especially on apple devices, because events triggering properly sometimes just doesn't happen.

Related issue: #5885 

**Wait to merge until after 7.8**


Dananji (approved):
Looks great! 
This was a tricky one for sure 😌

cjcolvar (approved):

Issue #5949, opened 2024-07-22
[BUG] User cannot type title of a playlist item while adding to playlist when media is playing


elynema:
**Describe the bug**
If a media item is playing, the user cannot enter a title for the playlist item while adding to playlist. This does not happen if the media item is paused or not playing. This is only relevant for Current Track or Custom Timespan options in Add to Playlist where the user can enter a title for the playlist item.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to any item in avalon-staging or avalon-dev.
2. Click Play on one of the sections
3. Select Add to Playlist
4. Choose either Current Track or Custom Timespan
5. Try typing a title for the playlist item (the title will be auto-populated, but should be editable).
6. Every second or so the text entered by the user disappears and the default playlist item title gets re-populated.

**Expected behavior**
User-entered text in the Playlist item title box should be retained once the user starts typing and should not return to default.

**Environment (please complete the following information):**
 - Device / platform: Mac Laptop
 - Browser: Chrome
- [x ] avalon-dev
- [ x] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
This appears to be a regression bug in 7.8 release, as it is not happening in MCO.

joncameron:
Could no longer reproduce.

PR #5952, opened 2024-07-22
Retain user entered playlist item title during playback

Dananji:
Related issue: #5949 

Additional fix: reset default option selection in the add to playlist forn when alert is cleared after adding a playlist item.


cjcolvar (approved):

Issue #5948, opened 2024-07-19
Provide a setting to change where Rack tempfiles are stored


masaball:
An issue was encountered with large file uploads where the Rack::Multipart middleware can end up producing large (7+ GB) temp files. Having these large temp files written during upload filled up the root file space, which can cause major issues with other functionalities on the server. 

The [default factory used to produce the temp files](https://www.rubydoc.info/gems/rack/Rack/Multipart/Parser#TEMPFILE_FACTORY-constant) writes to `Dir.tmpdir` which returns the Operating System's temp filepath. For Linux systems this will usually be `/tmp`. If the root system is constrained on storage, `/tmp` may not be large enough to handle these temp files. Rack::Multipart can take an environment variable, `env[RACK_MULTIPART_TEMPFILE_FACTORY]`, to override the default factory. @cjcolvar suggested that we could add an entry to Settings where implementers could define a filepath that is not storage constrained, then have an initializer which checks that entry and sets the environment variable as necessary.

# Done Looks Like
 - [x] setting in settings.yml file for the path to upload temp files
 - [x] setting should be commented out by default, which would fallback to default used by Rack
 - [x] parser should check that path exists, if not, fall back to default and put out error message to log file

elynema:
Without this feature, we have to manually define the default temp file location in code for MCO.

elynema:
This is merged to avalon-dev and using default values at this point in time. Mason and Chris tested this on mco-staging with a different temp file location and ensured that it did work there.

masaball:
For testing: 
1. Check the logs for the POST request of the file upload: `Started POST "/master_files" for 140.182.72.228 at 2024-08-01 09:49:35 -0400`
2. A couple lines down from that will be an entry for the request parameters: 
```
Parameters: {
  "authenticity_token"=>"[FILTERED]", 
  "container_id"=>"br86bv39m", 
  "step"=>"file_upload", 
  "Filedata"=>[#<ActionDispatch::Http::UploadedFile:0x00007fcb8f2de260 @tempfile=#<Tempfile:/tmp/RackMultipart20240801-2491473-7yybef.mp4>, @content_type="video/mp4", @original_filename="lunchroom_manners.mp4", @headers="Content-Disposition: form-data; name=\"Filedata[]\"; filename=\"lunchroom_manners.mp4\"\r\nContent-Type: video/mp4\r\n">], 
  "id"=>"br86bv39m", 
  "original"=>""
}
```
[...]

PR #5958, opened 2024-07-26
Add setting for alternative rack tempfile directory

masaball:
Related issue: #5948 

The variable that needs to be set to override the default factory is part of the Rack `request.env` hash. Creating a custom middleware to handle injecting the lambda seemed to be the best approach. 

Having the error handling in the initialize method will log any issues once during app startup. The factory logic can be called multiple times during a single upload, so this seemed like the simplest way to get the error logged without triggering, potentially multiple times, on every upload in the application. It does run the risk of not being noticed that way though so I'm not sure which approach is actually better.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -123,3 +123,8 @@ derivative:
   allow_download: true
 # Maximum size for uploaded files in bytes (default is disabled)
 #max_upload_size: 2147483648 # Use :none or comment out to disable limit
+# Rack Multipart creates temporary files when processing multipart form data with a large payload.
+# If the default system /tmp directory is storage constrained, you can define an alternative here. 
+# Leave commented out to use the system default.
+# tempfile:
+#   location: ''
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -123,3 +123,8 @@ derivative:
   allow_download: true
 # Maximum size for uploaded files in bytes (default is disabled)
 #max_upload_size: 2147483648 # Use :none or comment out to disable limit
+# Rack Multipart creates temporary files when processing multipart form data with a large payload.
+# If the default system /tmp directory is storage constrained, you can define an alternative here. 
+# Leave commented out to use the system default.
+# tempfile:
+#   location: ''
```
cjcolvar, line 130: How about putting the default value here as a cue?
```suggestion
#   location: '/tmp'
```

lib/tempfile_factory.rb
```diff
@@ -0,0 +1,25 @@
+# Middleware to override the default Rack::Multipart tempfile factory: 
+# https://www.rubydoc.info/gems/rack/Rack/Multipart/Parser#TEMPFILE_FACTORY-constant
+class TempfileFactory
```
cjcolvar, line 3: I think it would be good if you could add some tests for this class if it isn't too hard.

cjcolvar (commented):
Nice work!

cjcolvar (approved):

PR #5959, opened 2024-07-26
Ensure that thumbnails show during migration/reindex

cjcolvar:
In Avalon 7.8 we changed the value of avalon_resource_type_ssim to be all lowercase from titlized.  This causes a matching problem until a full reindex of MediaObjects (done with the section list migration) can be completed.  This small change matches both cases so should keep showing thumbnails for both pre and post migrated objects.  This can be removed in a future Avalon release.


masaball (approved):
:+1:

PR #5960, opened 2024-07-29
Fix links in footer

masaball:
In #5879, we set the footer to a negative z-index in order to get playlist paging to work. We did not catch/realize that negative z-indexes disable links, because elements are actually placed behind their containers. This PR sets the footer z-index explicitly to 0 to reenable the "Contact Us" and other links in the footer. It also sets the playlist pagination z-index to 10, so that its actions will not be covered by the footer again.

### DIFF
--- a/app/assets/stylesheets/avalon/_footer.scss
+++ b/app/assets/stylesheets/avalon/_footer.scss
@@ -21,7 +21,7 @@
   width: 100%;
   min-height: 2.5rem;
   color: #999999;
-  z-index: -1000;
+  z-index: 0;
 }
 
 footer {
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
I tried out these changes in Firefox dev tools and the playlist pagination buttons work but the footer links do not.  I'm guessing that we need to try a different approach to avoid the overlapping divs.

masaball:
Weird. Worked when I tested it so I wonder what's different...anyway, I can look at it more deeply tomorrow.

cjcolvar (approved):
I retested and wasn't able to reproduce what I saw earlier.  I think this is good to go and the failing flaky test can be ignored.

PR #5963, opened 2024-07-30
Speed improvements for searching by avoiding solr queries

cjcolvar:
Included in this PR:
- Reuse existing ability model for testing if conditional facets should load because instantiating new ability objects requires a solr query to answer the ability check instead of using a cached value.
- Use `SpeedyAF::Base.for` to cast the SolrDocument into a proxy when constructing a search result's title link instead of using `#find` which calls out to solr.

With these changes searches with 10 results per page return in about 1s vs. 3s and 4s vs. 20s with 100 results per page.


masaball (approved):
Looks good to me

joncameron:
>  10 results per page return in about 1s vs. 3s and 4s vs. 20s with 100 results 

Nice!

PR #5966, opened 2024-07-31
Get rid of nil ids when migrating to section lists

cjcolvar:
This fixes an edge case where ordered_master_file_ids contain nils.


masaball (approved):
:+1:

Issue #5967, opened 2024-08-01
[BUG] ATOM feed requests are failing


masaball:
**Describe the bug**
I was looking at the logs on mco-staging and it looks like there is a bug with how ATOM requests interact with the transcript search. The ATOM feed is directly querying for `other_identifier_sim:query`, which goes through the search builder and becomes a query for `transcript_tsim:other_identifier_sim:query`  which solr can't parse. We had a similar issue with the Selected Items page which was fixed by skipping transcript search in the Bookmark context.  

**Expected behavior**
ATOM does not error when making requests

**Environment (please complete the following information):**
- [ ] avalon-dev
- [ ] avalon-staging
- [x] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
7.8

masaball:
@elynema @joncameron 
I ultimately tracked down that it was specifically the curly braces that were causing the search to fail and that they cause the failure in the front end search as well. Even though we are escaping special characters before sending the query to solr, they were still causing issue. I could not tell if it was an issue with the `RSolr.solr_escape` method or if it is an issue with solr not respecting the escaping. Or some other intermediate processing screwing things up. For now, I have a PR to skip the transcript search if a curly brace is in the query.

It is possible there are other edge cases with searches involving special characters, but it is probably not worth the effort to try to track them all down and resolve them right now.

elynema:
This would impact any Solr search. If we search the main site with a curly brace, we'll get an error from the "Found in" part of the main site search. Looks like there is also a console error if you do a search on the Transcript tab, but no error displays to the user. We can't exclude the curly braces from searches since they are needed in the search that RMD is submitting to Atom.

In the first pass at a solution, just not doing transcript search if there are curly braces in the search term.

We could merge this as a first pass fix for 7.8, and then create another issue to look at for next release to explore other alternatives.

PR #5968, opened 2024-08-01
Skip found in transcript if there are curly brackets in query

masaball:
On mco-staging, an Atom feed request was failing because the query was `other_identifier_sim:/GR[0-9]{8}/`. I eventually determined that it was not an issue with this being turned into `transcript_tsim:other_identifier_sim` and that it was not an issue with the Regex formatting of the query.

The behavior that I ultimately determined was the culprit was the curly braces. Any search conducted in Avalon that contains curly braces will fail, even though the search query is escaped for Solr. I do not know if this is a bug upstream in RSolr with the `solr_escape` method or if this is an issue in Solr itself where it is not respecting the escaping like it should in this context.

For now, skipping the transcript search if there are curly braces in the query seemed like the easiest approach.

Related issue: #5967


Dananji (approved):
Looks good 👍

Issue #5955, opened 2024-07-23
[BUG] Video items are showing the audio only thumbnail / poster


elynema:
**Describe the bug**
In both avalon-dev and avalon-staging, we are seeing examples where items that contain video files (or specific sections that are videos) are displaying with the audio only icon as thumbnails on the results list and as posters on the media object page.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. On avalon-staging, search for 'lunchroom manners'. 
2. Scroll to the third item. It shows an audio only thumbnail. 
3. Click on the item (https://avalon-staging.dlib.indiana.edu/media_objects/9c67wm82b).
4. The audio only icon is showing up as the poster image, as well.
5. Another example on avalon-dev: https://avalon-dev.dlib.indiana.edu/media_objects/8s45q876k

**Expected behavior**
The default thumbnail on the search results page (if there is no poster image) should be a video icon if the first section is a video and an audio icon if the first section is an audio only file. 
On the media object page, for any section that is a video, the default poster (if there is no poster image) should be a video icon.

**Screenshots/Screen Recording**
![image](https://github.com/user-attachments/assets/7c3bd69c-048e-4341-85fd-4ebd5f956334)

![image](https://github.com/user-attachments/assets/872df9ec-9695-4f90-ae0f-910d298fb244)

**Environment (please complete the following information):**
 - Device / platform: Mac desktop (probably everywhere)
 - Browser: Chrome
- [x] avalon-dev
- [x] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
This seems to be a 7.8 regression.

elynema:
Currently the browse / search results page pulls the poster from the first section of the item. We can stick with that behavior.

elynema:
We don't generally have many situations where we don't have a poster image for video. Maybe occasionally if a background job fails to generate poster image. So this is an edge case.

joncameron:
Video icon now displays instead of the audio icon as expected 👍

PR #5969, opened 2024-08-02
Display video icon when video lacks thumbnail

masaball:
Related issue: #5955 

Audio items flow through the `#get_frame` logic when displaying in the index, so we still need the redirect to the audio icon.

**Merge after 7.8 release**


cjcolvar (approved):
Looks good!

Issue #5983, opened 2024-08-13
Upgrade to Rail 7.2


joncameron:

joncameron:
Could be 8 points, depending on how painful the Rails upgrade is.

masaball:
Notes:
Active fedora: /lib/active_fedora/scoping.rb#ln69 ActiveSupport::PerThreadRegistry is deprecated/removed. Need to switch to thread_mattr_accessor
  - Has a PR for 7.1 upgrade that was never merged but tests are passing.
  - @cjcolvar will look at what's needed for 7.2

hydra-head needs to be updated. Most recent release is rails < 7.1 (@cjcolvar)

noid-rails needs to be updated for 7.2, main supports 7.1 but a release has not been cut with 7.1 support (@masaball)

samvera-persona needs to be updated. Most recent release is rails < 7.1  (@masaball)

browse-everything needs to be updated for 7.2, most recent release supports 7.1.
- We use our own fork of browse-everything. Should we commit that back upstream?
- Jonathan Rochkind is usually good about keeping this up to date. We could contact him about his plans for that before trying to upgrade it ourselves.
- Wait for upgrade of upstream, then rebase our fork. Potentially commit our fork back upstream later.

active_annotations needs to be updated. Most recent release is rails < 7.1. 
- Tried updating but circleci is not working. Github had deleted the public key it had on record. I tried reissuing the ssh key, but now it is erroring trying to communicate with dockerhub 
- Maybe we can bring this into avalon and stop having it as separate gem and just not worry about fixing circleci?

Non-samvera/non-local gems:
- selenium-webdriver gem does not depend on rails, but could be updated. This has potential to break stuff in the testing suite.
- If we can update selenium, there is a separate webdrivers gem which is no longer necessary with the most recent versions of selenium

Non-breaking messages from bundle update that may be relevant:
```
Post-install message from about_page:

  To use the git_log node type, add the following line to your Gemfile:
  gem 'rugged'
```
```
Post-install message from rdf-xsd:

  For best results, use nokogiri and equivalent-xml gems as well.
  These are not hard requirements to preserve pure-ruby dependencies.
```
```
Post-install message from devise:

[DEVISE] Please review the [changelog] and [upgrade guide] for more info on Hotwire / Turbo integration.

  [changelog] https://github.com/heartcombo/devise/blob/main/CHANGELOG.md
[...]

PR #5996, opened 2024-08-19
Upgrade to ruby 3.3 and enable jemalloc and YJIT

cjcolvar:
Install jemalloc, configure based upon https://gist.github.com/jjb/9ff0d3f622c8bbe904fe7a82e35152fc, and enable YJIT in Avalon containers.

Part of #5983


masaball (approved):
Looks good

Issue #5242, opened 2023-06-26
Error when requesting media_objects index json when user belongs to many groups/virtual groups


cjcolvar:
# Description
Seen in production (https://app.honeybadger.io/projects/54117/faults/97117493). Affects the Avalon Ingest API.

This error is due to the solr request uri being too big for a GET request.  In other places we've overridden ActiveFedora to allow switching solr requests to POST instead of GET.  In this case, ActiveFedora relation#to_a calls #find_each which ends up calling `ActiveFedora::SolrService.instance.conn.paginate` which bypasses our overrides.  A possible solution to this is to override `RSolr::Client#build_request` to fallback to `ActiveFedora::SolrService.default_http_method` instead of `:get`.  This would allow for configuring the default http method used by rsolr and would not require us to handle all of the cases in ActiveFedora or Avalon code where we reach into rsolr and execute methods directly instead of going through ActiveFedora::SolrService.

# Done Looks Like
- [ ] Override `RSolr::Client#build_request` to fallback to `ActiveFedora::SolrService.default_http_method` instead of `:get`

joncameron:
https://github.com/avalonmediasystem/avalon/pull/6020#pullrequestreview-2280796955
> If the rest of the test suite is passing then I think that should be sufficient to prove that this override is working.

PR #6020, opened 2024-09-03
Override RSolr::Client to fallback to ActiveFedora#default_http_method

masaball:
Related issue: #5242 

@cjcolvar I couldn't really figure out a good way to put in tests for this. If you have any ideas, maybe we can pair up and go over them. Manual testing seemed to be returning information as expected.


cjcolvar (approved):
Looks good!

I don't think a unit test is always necessary for overrides.  I think if the rest of the test suite is passing then I think that should be sufficient to prove that this override is working.

Issue #5762, opened 2024-04-02
Remove Mediainfo Dependency


joncameron:
# Description
Work related to incorrect duration values being written into Avalon metadata for files (see https://github.com/avalonmediasystem/avalon/pull/5754) has brought to attention Avalon's use of an older version of Mediainfo. Mediainfo is only being used for basic information gathering and could be removed as a dependency altogether in favor of the FFmpeg toolchain. This would free Avalon from using this older version of mediainfo, and reduce breaking changes related to mediainfo. 

This is separate from the Active Encode gem which still uses mediainfo. Active Encode uses the current version of mediainfo with the updated XML schema and does not have this same problem.

# Done Looks Like
- [ ] Mediainfo calls replaced with ffprobe

joncameron:
While testing this I passed in a video file (https://drive.google.com/file/d/110WAo6DaHKXAre5ksW7cLYsEOrBoDzyB/view?usp=drive_link) that went through as audio:
- https://avalon-dev.dlib.indiana.edu/media_objects/zg64tk959
- https://avalon-dev.dlib.indiana.edu/encode_records/1021

The raw object info in the dashboard reports the video stream as expected but the encode profile is for audio, and after file upload the flash message appears indicating that Avalon found it to be an audio item.

I tried an upload with lunchroom manners and it seemed fine and reported as a video: https://avalon-dev.dlib.indiana.edu/media_objects/8910jt59d. 

I tried another time with the Libya to Somalia file and could reproduce that it identifies as audio. This is all via web upload.

masaball:
Yes, that is a bug with this work. One of the checks we use for identifying video is to confirm that there is a display aspect ratio value. This check was because text files are treated as image/video files by ffprobe but don't generate that field, so we could use it to reject them. 

Running the example file through ffprobe, there is no display aspect ratio so it fails the video check. So looks like we will need to go back and reevaluate those checks to avoid false positives/negatives.

joncameron:
My proposal for this would be to use the existence of `duration` rather than aspect ratio and packet values.

[...]

PR #6023, opened 2024-09-05
Adjust non-media item filtering in ffprobe

masaball:
In QA it was found that a video was improperly processed as audio. This was a result of keying off display aspect ratio to filter out non-audio/video items, under the false assumption that all videos would have a display aspect ratio. 

Using [Marcel](https://github.com/rails/marcel), we can check the file mimetype before feeding it into ffprobe and filter out non-audio/video files.

Related issue: #5762

### DIFF
--- a/lib/avalon/ffprobe.rb
+++ b/lib/avalon/ffprobe.rb
@@ -14,10 +14,11 @@
 
 module Avalon
   class FFprobe
-    # media_path should be the output of `FileLocator.new(file_location).location`
+    attr_reader :json_output, :video_stream, :audio_stream
+    # @param [String] media_path the full path to the media file
     def initialize(media_path)
       ffprobe = Settings&.ffprobe&.path || 'ffprobe'
-      raw_output = `#{ffprobe} -i "#{media_path}" -v quiet -show_format -show_streams -count_packets -of json`
+      raw_output = `#{ffprobe} -i "#{media_path}" -v quiet -show_format -show_streams -count_frames -of json`
       # $? is a variable for the exit status of the last executed process. 
       # Success == 0, any other value means the command failed in some way.
       unless $?.exitstatus == 0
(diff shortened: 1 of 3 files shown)


lib/avalon/ffprobe.rb
```diff
@@ -14,10 +14,12 @@
 
 module Avalon
   class FFprobe
-    # media_path should be the output of `FileLocator.new(file_location).location`
-    def initialize(media_path)
+    attr_reader :json_output, :video_stream, :audio_stream
```
cjcolvar, line 17: A pattern I've seen before that I think would work well here is defining getter methods for these and using memoization.  I think that would also help cleanup the initializer method and make it lazily call ffprobe when one of these getters is called.

I'm thinking something like this:
```
  def initialize(media_file)
    @media_file = media_file
  end

  def json_output
    return @json_output unless @json_output.nil?
    return @json_output = {} unless valid_content_type?(@media_file)
    ffprobe = Settings&.ffprobe&.path || 'ffprobe'
    raw_output = `#{ffprobe} -i "#{@media_file.location}" -v quiet -show_format -show_streams -of json`
    # $? is a variable for the exit status of the last executed process. 
    # Success == 0, any other value means the command failed in some way.
    unless $?.exitstatus == 0
      Rails.logger.error "File processing failed. Please ensure that FFprobe is installed and that the correct path is configured."
      return @json_output = {}
    end
    @json_output = JSON.parse(raw_output).deep_symbolize_keys
    @json_output
  end

  def video_stream
    @video_stream ||= json_output[:streams].select { |stream| stream[:codec_type] == 'video' }.first
  end

  def audio_stream
    @audio_stream ||= json_output[:streams].select { |stream| stream[:codec_type] == 'audio' }.first
  end
```

lib/avalon/ffprobe.rb
```diff
@@ -26,35 +28,51 @@ def initialize(media_path)
...
 
     def video?
-      # ffprobe treats plain text files as ANSI or ASCII art. This sets the codec type to video
-      # but leaves display aspect ratio `nil`. If display_aspect_ratio is nil, return false.
-      # ffprobe treats image files as a single frame video. This sets the codec type to video
-      # but the packet/frame count will equal 1. If packet count equals 1, return false.
-      return true if @video_stream && @video_stream[:display_aspect_ratio] && @video_stream[:nb_read_packets].to_i > 1
+      return true if video_stream
 
       false
```
cjcolvar, line 39: I think this could be simplified to:
```suggestion
      video_stream.present?
```

lib/avalon/ffprobe.rb
```diff
@@ -26,35 +28,51 @@ def initialize(media_path)
...
     end
 
     def audio?
-      @json_output[:streams]&.any? { |stream| stream[:codec_type] == 'audio' }
+      return true if audio_stream
+
+      false
```
cjcolvar, line 45: I think this could be simplified similar to `video?` but should it also include a check that it isn't a video?
```suggestion
      audio? && !video?
```

lib/avalon/ffprobe.rb
```diff
@@ -26,35 +28,51 @@ def initialize(media_path)
...
+    private
+
+    def valid_content_type? media_file
+      # Remove S3 credentials or other params from extension output
+      extension = File.extname(media_file.location)&.gsub(/[\?#].*/, '')
+      # Fall back on file extension if magic bytes fail to identify file
+      content_type = Marcel::MimeType.for media_file.reader, extension: extension
+      return true if ['audio', 'video'].any? { |type| content_type.include?(type) }
+
+      false
```
cjcolvar, line 75: I think the conditional return is redundant.
```suggestion
      ['audio', 'video'].any? { |type| content_type.include?(type) }
```

lib/avalon/ffprobe.rb
```diff
@@ -26,35 +28,51 @@ def initialize(media_path)
...
     end
 
     def original_frame_size
-      "#{@video_stream[:width]}x#{@video_stream[:height]}" if video?
+      "#{video_stream[:width]}x#{video_stream[:height]}" if video?
+    end
+
+    private
+
+    def valid_content_type? media_file
```
cjcolvar, line 68: This method feels like a really good improvement!

lib/avalon/ffprobe.rb
```diff
@@ -14,10 +14,12 @@
 
 module Avalon
   class FFprobe
-    # media_path should be the output of `FileLocator.new(file_location).location`
-    def initialize(media_path)
+    attr_reader :json_output, :video_stream, :audio_stream
+    # @param [FileLocator] media_file a file locator instance for the media file
```
cjcolvar, line 18: Nice yarddoc making the param class explicit!

cjcolvar (commented):
Looks good overall.  I have some refactoring suggestions that I think would clean things up a big.

lib/avalon/ffprobe.rb
masaball: Good call. Was being explicit when working on it to make sure I had the logic right but should have gone back through and simplified.

PR #6025, opened 2024-09-06
Use official release of noid-rails

cjcolvar:
Part of follow on work of Rails 7.2 upgrade.


masaball (approved):
Looks good. I think the failing test is flaky, so this should be good to merge.

PR #6032, opened 2024-09-13
Improve thumbnail load times

masaball:
IndexedFiles must be reindexed for this work to have an effect. Without reindexing all thumbnails will display as the missing content icon.

Improvements needed:
- Populate with icon data instead of `asset_path` when item is missing thumbnail
- Because of being in the term frequency method, the work as it stands breaks the browse page. 
    - Move thumbnail logic out of term frequency method and into its own method
    - Or rework the checks and search logic so that the term frequency method runs without causing an error when `solr_parameters[:q]` is missing.
- This work probably needs an explicit fallback for when a video is missing a thumbnail or `content_ss` is otherwise unable to be pulled from solr.

Investigation needed:
- Does this affect overall search load times? Need to test with larger datasets.
- Ensure that the thumbnails are returned for both browse and search, and that this doesn't break other routes that run through the search builder/catalog controller.
- Is there a way to return the thumbnail data without needing a reindex?

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -66,7 +66,7 @@ module ApplicationHelper
 
     if master_file_id
       if video_count > 0
-        thumbnail_master_file_path(master_file_id)
+        document["sections"][:docs].first["thumbnail"][:docs].first["content_ss"]
       elsif audio_count > 0
         asset_path('audio_icon.png')
       else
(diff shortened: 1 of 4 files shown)


cjcolvar (commented):
This is great work, but I'm unsure if it is the best solution given the need to reindex and I'm wondering if data-uris might end up being more of a problem than an improvement.  I'd like to leave this PR in draft and talk over how it compares to alternative solutions.

masaball:
I'm going to close this PR because we decided not to pursue this particular approach further. Will create a new PR if/when we return to this issue.

joncameron:
Created #6879 for thumbnail migration from Solr to disk or another better location.

PR #6035, opened 2024-09-16
Optimize waveform generation

### LINKED ISSUE #5155: FFmpeg Process Hangs During Waveform Generation
# Description
FFmpeg processes used for generating waveforms are hanging in the IU Avalon instances, tying up system resources and sometimes requiring a manual server cull of zombie processes. A timeout value has been added to make sure these are cleaned on the server, but if they are not exiting as expected we should investigate why that's the case and resolve the problem.

Waveform jobs can take from 30-60 minutes for media with long durations.

Scaling up in Sidekiq could create issues with resource contention and CPU saturation. One waveform could complete fairly quickly on staging, but in production could run for hours. Running a lot of jobs concurrently on mco-staging could be a way to test this.

# Done Criteria
- [x]  Investigation into why processes don't return properly
- [x]  Code change to resolve (possibly need to refactor IO buffering or CPU saturation)

cjcolvar:
Changes propsed in this PR:
- FFmpeg downmix to mono wav instead of having waveform code check both channels for min/max peaks.
- Bring `gather_peaks` method into `get_normalized_peaks` so overall absolute value max can be computed while gathering peaks instead of needing a separate pass through the peaks after.
- Scale peaks in place instead of allocating another array
- Explicitly close both reader and IO instead of waiting for FFmpeg to terminate
- Avoid `flatten` since it is expensive when run so frequently

Note that the waveform fixture has slightly different values.  I found that running the waveform job on a real item on mco-staging yielded different length and values even when run with the same code.  But the small differences did not affect how it rendered in the SME so shouldn't be problem.

Resolves #5155


masaball (approved):
Looks good!

PR #6042, opened 2024-09-18
Allow manual loading of records with progressive download streams instead of HLS streams

### LINKED ISSUE #6003: Bring progress download code back into main branch
# Description

Some code changes were made to Avalon to get skip the IIIF manifest generation to skip HLS and just do progressive download (put MP3 URLs in the manifest). This is sitting in demo branch only, so needs to be brought into core and have tests added for it.

This would be for MSU implementation

cjcolvar:
This allows use of mp3s pushed into avalon via the ingest api (setting derivative `mime_type` and `hls_url`) to playback correctly in ramp.

Resolves #6003


masaball (approved):
Looks good to me!

PR #6046, opened 2024-09-20
Drop solr6 support

### LINKED ISSUE #5686: Retire Solr 6 Config
# Description
Avalon retains two sets of configuration files for Solr, based on the version. With the next release we should officially drop config for legacy Solr from the Avalon config and only provide a single config for the currently used version.

# Done Looks Like
- [ ] Remove legacy Solr config from source
- [ ] Update upgrade documentation to note the change and add to release announcement

cjcolvar:
As part of the upgrade to Avalon 7.7 we strongly encouraged upgrading to Solr 9.  With Avalon 8.0 we can drop support of Solr 6.

Resolves #5686


masaball (approved):
:+1:

PR #6047, opened 2024-09-20
Date issued is no longer a required field

### LINKED ISSUE #6021: Remove Date as a Required Field for Descriptive Metadata
# Description
Avalon has required Title and Date as fields that must exist for a record to be validated. This is a bad fit for certain types of metadata and is an artificial requirement that can be removed. Items may only have a creation date, for example, but no publication date as they were never published—but this specific value is the required field, leading people to put in inaccurate data to conform to expectations.

Shouldn't impact facets since null values don't impact index efforts.

Another issue to create: should Creation Date also be used for "Date" indexing as presented in the Blacklight facet search?

# Done Looks Like
- [x] Publication Date is no longer a required field
- [x] Validation for items does not required a Publication Date value
- [x] Documentation updated

cjcolvar:
Resolves #6021

### DIFF
--- a/app/models/iiif_manifest_presenter.rb
+++ b/app/models/iiif_manifest_presenter.rb
@@ -100,6 +100,7 @@ class IiifManifestPresenter
   end
 
   def combined_display_date(media_object)
+    #FIXME Does this need to change now that date_issued is not required and thus could be nil
     result = media_object.date_issued
     result += " (Creation date: #{media_object.date_created})" if media_object.date_created.present?
     result
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -157,7 +157,7 @@ class BookmarksController < CatalogController
       else
         case params['action']
         when 'publish'
-          if media_object.title.nil? || media_object.date_issued.nil?
+          if media_object.title.nil?
             errors += ["#{id}, Unable to Publish Item. Missing required fields."]
           else
             success_ids << id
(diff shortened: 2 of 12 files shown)


app/models/iiif_manifest_presenter.rb
```diff
@@ -100,6 +100,7 @@ def metadata_field(label, value, default = nil)
   end
 
   def combined_display_date(media_object)
+    #FIXME Does this need to change now that date_issued is not required and thus could be nil
     result = media_object.date_issued
     result += " (Creation date: #{media_object.date_created})" if media_object.date_created.present?
```
masaball, line 105: I think this will need to be changed. If someone provides a creation date but not an issued date then this method will raise an exception as written.

cjcolvar: @joncameron What do you think about un-combining the display date so it is two separate metadata values in the IIIF manifest?
```
"Publication date: #{media_object.date_issued}" if media_object.date_issued.present?
"Creation date: #{media_object.date_created}" if media_object.date_created.present?
```

joncameron: I would love it—that makes a lot more sense and especially with this change. I created #6050 for changing the manifest generation to separate the date fields.

masaball: Since a new issue has been created, I'll go ahead and approve this PR :+1:

masaball (approved):

PR #6049, opened 2024-09-23
Link to faceted search for subjects in IIIF manifest

cjcolvar:
Does this need more tests for the switch from text to links in the manifest metadata?


app/controllers/catalog_controller.rb
```diff
@@ -102,6 +102,8 @@ class CatalogController < ApplicationController
     config.add_facet_field 'has_captions_bsi', label: 'Has Captions', if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow", helper_method: :display_has_caption_or_transcript
     config.add_facet_field 'has_transcripts_bsi', label: 'Has Transcripts', if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow", helper_method: :display_has_caption_or_transcript
 
+    config.add_facet_field 'subject_ssim', label: 'Subject', if: false
```
masaball, line 105: Just making sure I understand. `if: false` will disable the subject field from showing in the list of facets on the browse page, but still allows the backend to perform the facet searches?

masaball:
I think it would probably be good to have a unit test that makes sure the iiif manifest subject metadata is being populated with links.

Dananji (approved):
Looks good 👍

PR #6053, opened 2024-10-01
Removed "Required" Text for Date Issued Description

joncameron:


cjcolvar (approved):
:tada: 
This reminds me that this will also need to be changed in documentation outside the codebase when we release 8.0.  Collection Manager's Guide is what I'm thinking of but there might be other places.

Issue #6050, opened 2024-09-24
Don't Merge Display Date in Manifest Metadata


joncameron:
# Description
Un-combining the display date so it is two separate metadata values in the IIIF manifest.

## Example
"Publication date: #{media_object.date_issued}" if media_object.date_issued.present?
"Creation date: #{media_object.date_created}" if media_object.date_created.present?

See https://github.com/avalonmediasystem/avalon/pull/6047/files/749215c386243e17037770743f8121f189657760#r1772058052

# Done Looks Like
- [x] Manifest generation changed to include both date values if present as separate fields

PR #6062, opened 2024-10-02
Separate display date into 'Creation date' and 'Publication date'

masaball:
Related issue: #6050

### DIFF
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -66,7 +66,3 @@ img.no-icon {
 .page-item span.page-link {
   color: $dark !important;
 }
-
-.blacklight-date_issued_ssi.col-sm-3 {
-  white-space: nowrap;
-}
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -66,3 +66,7 @@ img.no-icon {
 .page-item span.page-link {
   color: $dark !important;
 }
+
+.blacklight-date_issued_ssi.col-sm-3 {
+  white-space: nowrap;
+}
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -115,8 +115,8 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field 'title_tesi', label: 'Title', if: Proc.new {|context, _field_config, _document| context.request.format == :json }
-    config.add_index_field 'date_issued_ssi', label: 'Publication date'
-    config.add_index_field 'date_created_ssi', label: 'Creation date'
+    config.add_index_field 'date_issued_ssi', label: 'Date', if: Proc.new {|_context, _field_config, document| document[:date_issued_ssi].present? }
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/blacklight.scss
```diff
@@ -66,3 +66,7 @@ img.no-icon {
 .page-item span.page-link {
   color: $dark !important;
 }
+
+.blacklight-date_issued_ssi.col-sm-3 {
+  white-space: nowrap;
```
masaball, line 71: "Publication date" was too long to display on a single line on the search page with the default css, so override to keep it inline.

app/controllers/catalog_controller.rb
```diff
@@ -113,7 +113,8 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field 'title_tesi', label: 'Title', if: Proc.new {|context, _field_config, _document| context.request.format == :json }
-    config.add_index_field 'date_issued_ssi', label: 'Date', helper_method: :combined_display_date
+    config.add_index_field 'date_issued_ssi', label: 'Publication date'
+    config.add_index_field 'date_created_ssi', label: 'Creation date'
```
cjcolvar, line 117: Good catch!  I forgot about the combined dates on search results.  This probably also comes up in collection landing page search results.

@joncameron Which dates do you think should be displayed on search results?

app/views/playlist_items/_current_item.html.erb
```diff
@@ -39,7 +39,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 </div>
 
 <dl id="creation_metadata">
-  <%= display_metadata('Date', combined_display_date(@current_mediaobject), 'Not provided') %>
+  <%= display_metadata('Publication date', @current_mediaobject.date_issued, 'Not provided') %>
```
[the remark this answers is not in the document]
masaball: Right, good catch.

app/controllers/catalog_controller.rb
joncameron: Could it be done in a way that favors Publication Date, but shows Creation Date if it's present and Publication Date is not? 

Something like:
```ruby
if date_created.present? && !date_issued.present?
    config.add_index_field 'date_created_ssi', label: 'Creation date'
else
    config.add_index_field 'date_issued_ssi', label: 'Publication date'
end
```

cjcolvar: That might be possible.  We'd have to use a helper_method(s) since the config is static and conditionals can't be evaluated here.

joncameron: +1 to adding a helper method like the previous used combined_date if that's reasonable. The label could even continue to be generic "Date" if that makes things easier.

cjcolvar (commented):
Looks great and super close to merging.

app/views/playlist_items/_current_item.html.erb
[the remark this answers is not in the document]
masaball: Weird. It was not intentionally re-added. Maybe came back in from the rebase onto develop somehow...Will remove it again.

cjcolvar (approved):

PR #6067, opened 2024-10-07
Bump samvera gems to released versions

masaball:


Gemfile
```diff
@@ -29,9 +29,9 @@ gem 'terser'
 gem 'shakapacker'
 
 # Core Samvera
-gem 'active-fedora', git: 'https://github.com/samvera/active_fedora.git', branch: 'further_along_the_rails'
+gem 'active-fedora', '~> 15.0'
 gem 'active_fedora-datastreams', '~> 0.5'
-gem 'hydra-head', git: 'https://github.com/samvera/hydra-head.git', branch: 'af_version'
+gem 'hydra-head', '~> 13.0'
 gem 'ldp', '~> 1.1.0'
```
cjcolvar, line 35: I noticed that we're behind on `ldp` (current version 1.2.1).  What about changing this to be any 1.x version?
```suggestion
gem 'ldp', '~> 1.1'
```

cjcolvar (approved):

Gemfile
masaball: Setting to 1.x successfully updates to `1.2.0`. When I set the gemfile to force it to `1.2.1`, there are a lot of dependency conflicts. Is it still worth going ahead and updating to `1.2.0` here or should we do a separate PR later to try to untangle the dependency hell to get on the most recent version?

cjcolvar: Oh, odd.  Let's merge this as-is and can make the ldp upgrade a separate thing.

PR #6073, opened 2024-10-08
OneDrive integration

masaball:
This PR adds support for file upload from a user's OneDrive site. The work is primarily contained in our fork of browse-everything, but also includes an updated ActiveEncode to properly parse the download links provided by the Microsoft API.

To implement on Avalon, we will need to have the application registered in the Entra admin center. For initial testing we may be able to use the application information I registered for my local testing, but it may be worthwhile to go ahead and set up an application profile specifically for Avalon-dev. If you go to [entra.microsoft.com](https://entra.microsoft.com/) and login with your IU info, you should be able to go to "Applications -> App Registrations" and look up my test which is called "browseeverything-test" to see the current permissions, ids, redirects, etc.

Once the app is registered, the information can be plugged into Avalon-dev's `settings.yml` and hopefully everything will just work...

### DIFF
--- a/app/services/file_locator.rb
+++ b/app/services/file_locator.rb
@@ -118,8 +118,17 @@ class FileLocator
     when 'file'
       File.open(location,'r')
     else
-      URI.open(uri, 'r')
+      open_uri
     end
+  end
+
+  # Ruby 3.0 removed URI#open from being called by Kernel#open.
+  # Prioritize using URI#open, attempt to fallback to Kernel#open
+  # if URI fails.
+  def open_uri
+    URI.open(uri, 'r')
+  rescue
+    Kernel::open(uri.to_s, 'r')
   end
 
   def attachment
--- a/app/services/file_locator.rb
+++ b/app/services/file_locator.rb
@@ -118,7 +118,7 @@ class FileLocator
     when 'file'
       File.open(location,'r')
     else
-      Kernel::open(uri.to_s, 'r')
+      URI.open(uri, 'r')
     end
   end
 
--- a/config/browse_everything_providers.yml
+++ b/config/browse_everything_providers.yml
@@ -18,3 +18,9 @@ file_system:
 # sky_drive:
 #   :client_id: YOUR_MS_LIVE_CONNECT_CLIENT_ID
 #   :client_secret: YOUR_MS_LIVE_CONNECT_CLIENT_SECRET
+# sharepoint:
+#   client_id: YOUR_MICROSOFT_GRAPH_CLIENT_ID
+#   client_secret: YOUR_MICROSOFT_GRAPH_CLIENT_SECRET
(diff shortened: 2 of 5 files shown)


app/services/file_locator.rb
```diff
@@ -118,7 +118,7 @@ def reader
     when 'file'
       File.open(location,'r')
     else
-      Kernel::open(uri.to_s, 'r')
```
masaball, line 121: `Kernel::open` would fail to find the sharepoint files due to our Ruby version. Using `Kernel::open` to call `URI.open` was removed in Ruby 3.0, so we need to explictly use `URI.open` here.

config/initializers/browse_everything.rb
```diff
@@ -11,5 +11,14 @@
   else
     settings['file_system'] = { name: 'File Dropbox', home: Settings.dropbox.path }
   end
+  if Settings.dropbox.sharepoint
+    settings['sharepoint'] = { client_id: Settings.dropbox.sharepoint.client_id,
+                               client_secret: Settings.dropbox.sharepoint.client_secret,
+                               tenant_id: Settings.dropbox.sharepoint.tenant_id,
+                               grant_type: Settings.dropbox.sharepoint.grant_type,
```
masaball, line 18: For current implementation this line is unneeded. However, there are a couple different ways to grant an application access to Microsoft Graph. For our purposes at IU, we need to log in as specific users which requires the `authorization_code` grant type. This is currently hardcoded. The other type is for allowing the application itself to get access, so a specific user account is not needed, which uses the `client_credentials` grant type.

For the final version of this work it would be ideal to allow implementers to choose which auth flow works best for how they have sharepoint setup so I have left the `grant_type` param in place here as a placeholder/reminder.

config/browse_everything_providers.yml
```diff
@@ -18,3 +18,9 @@ file_system:
 # sky_drive:
 #   :client_id: YOUR_MS_LIVE_CONNECT_CLIENT_ID
 #   :client_secret: YOUR_MS_LIVE_CONNECT_CLIENT_SECRET
+# sharepoint:
+#   client_id: YOUR_MICROSOFT_GRAPH_CLIENT_ID
+#   client_secret: YOUR_MICROSOFT_GRAPH_CLIENT_SECRET
+#   tenant_id: YOUR_MICROSOFT_GRAPH_TENANT_ID
+#   scope: offline_access https://graph.microsoft.com/.default
+#   redirect_uri: https://example.com/browse/connect
```
masaball, line 26: The `redirect_uri` param should be the application's base url appended with `/browse/connect`. So Avalon-dev should be `https://avalon-dev.dlib.indiana.edu/browse/connect`. This same value should be what is entered in the `redirect_uri` setting for the application in the Entra admin center.

cjcolvar (approved):
Looks good so far!  If everything checks out on avalon-dev then the branches on B-E and A-E can get merged.

PR #6074, opened 2024-10-08
Fix updated element in atom feed entries when query string present

cjcolvar:
The two solr fields used to populate the updated element in entries were not being returned in solr responses when a query string is present.  This was because in Avalon 7.8 an explicit `fl` was added to the query for the 'Found in' functionality and these two fields were forgotten.  Adding them in resolves the issue.


masaball (approved):
Looks good!

PR #6082, opened 2024-10-22
Fedora6 7.8 rebase rebase

cjcolvar:
Rebase of #5990 on top of latest in develop before merging.  Supersedes #5990 and #5363.


cjcolvar:
Test failure is not rspec tests but an error communicating with code climate.

PR #6088, opened 2024-10-28
Adjust column size to avoid player overflow on browser resize

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/676

Fixes the player overflowing into the next column when browser is resized.


cjcolvar (approved):
:+1:

PR #6098, opened 2024-11-04
Don't remove file_location since it is used to render the file's filename

cjcolvar:
Thanks to John Merritt of University of Louisville for pointing out this issue in avalon slack.

### DIFF
--- a/app/jobs/master_file_management_jobs.rb
+++ b/app/jobs/master_file_management_jobs.rb
@@ -81,14 +81,8 @@ module MasterFileManagementJobs
         when 'file' then File.delete(oldpath)
         when 's3'   then FileLocator::S3File.new(locator.source).object.delete
         end
-      	masterfile.file_location = ""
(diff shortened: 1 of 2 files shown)


masaball (approved):
This change looks good. It does make me think though should we consider uncoupling the section name from the file location? It feels a little weird to continue storing a file location when the file has actually been removed.

cjcolvar:
I think we're still storing a file location when we move the file and the file might then get moved to tape off disk.  So I'm thinking retaining it here is somewhat analogous.  If a section title is manually entered then we use that for section name but fallback to file location.  I'm wondering if we should store original filename in a separate field to get that decoupling that you're suggesting.

masaball:
That's a good point about the files being moved to tape storage being analogous...

If we were to decouple, I think storing original filename as a separate field would probably be the easiest/most sensible approach. 

I am fine with either approach.

Issue #6001, opened 2024-08-21
Provide default required metadata in IIIF manifest for newly uploaded media


elynema:
# Description

When a file is uploaded into Avalon from the web, and the user tries to preview the page before adding any descriptive metadata, Ramp gives an error:

<img width="1330" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkYyQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--59bb9a3af4faced10a31f19a4c18b6149e4fd474/image.png" alt="image.png" />

This error is due to Ramp was missing manifest.label which I'm guessing is Title in the manifest's metadata which wasn't in the manifest.  The labels in the structure were also null.

Prior to 7.8, when a media file had been uploaded, but no metadata added, the preview page would have a default title and date displaying. These default values did not populate into the Resource Description section when editing the item.

<img width="1330" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMTkyQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--021fd820543b031b88f80c4188c3ce2f26f5d451/image.png" alt="image.png" />

# Questions
- [ ] Should users be able to upload multiple files before creating a title/date? Even pre 7.8 release on Archivo does not seem to allow this.

# Done Looks Like
- [ ] the IIIF manifest generated by Avalon for a media item without a title and/or date should provide a default auto-generated value
- [ ] For title, the value should be the media item identifier [does this work if there are multiple sections?]
- [ ] For date, the default value should be "Date not provided"
- [ ] These default values should not be populated into the metadata for the item in Avalon
- [ ] Should default metadata be passed for the structure?

masaball:
As long as a user has not moved past the file upload screen they can upload multiple files. The validations on the media object are only activated after the file upload step has been "completed". 

We could potentially expand that so users can always attach more files but that would take some thought on how to do the rework and should be its own ticket if we want to look into it.

elynema:
@masaball I see one potential tweak here. After uploading a single item, there is a section with a blank title. After uploading 2 items, the file names populate the sections.

[...]

PR #6105, opened 2024-11-07
Use structure_title for populating structure list

masaball:
~~We have completely reworked how media object, master file, and playlist item titles are generated/displayed so limiting the creation of a display title from filename to only when there are multiple sections is no longer necessary.~~

Actually, we created a method called `structure_title` specifically for populating the structure list but apparently never switched to it. Reverted the changes to `display_title` and tests, switched the canvas presenter to use `structure_title` instead.

Related issue: #6001

### DIFF
--- a/app/models/concerns/master_file_behavior.rb
+++ b/app/models/concerns/master_file_behavior.rb
@@ -108,7 +108,7 @@ module MasterFileBehavior
                  structuralMetadata.section_title
                elsif title.present?
                  title
-               elsif file_location.present?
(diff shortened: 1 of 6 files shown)


cjcolvar (commented):
Can you please add a test which shows this difference between these two methods? (A test which would fail with `display_title` but passes with `structure_title`.)

cjcolvar (approved):
Looks good.  Thanks for adding the test!

PR #6106, opened 2024-11-08
Set fedora url env variable to right value in test container

masaball:
This PR also removes the version number from the top of the compose file. There is a deprecation message that it is no longer used.


cjcolvar (approved):
:+1:

PR #6125, opened 2024-11-18
Make override explicit by doing it in the original module in an initializer

### LINKED ISSUE #6124: [BUG] Incorrect links to items in search results
**Describe the bug**
Search results list items with the title or id of the item as a link into the detailed view of the item.  Blacklight by default makes this link `/catalog/:id`  but we have overridden it to use our controller `/media_objects/:id`.  The issue appears to be that Avalon's overrides to blacklight helpers are not getting loaded correctly.  This appears to only be an issue in the production environment and not development which leads me to believe it might be related to zeitwerk autoloading.

**To Reproduce**
1. Go to avalon-staging.dlib.indiana.edu
2. Click on Browse
3. Hover over search result links and inspect the link url

cjcolvar:
Resolves #6124 

Before this change the overridden helper method was not loading correctly and Blacklight's original implementation was executing causing the return value to be a SolrDocument which led to incorrect urls in the search results.  This is probably due to a change in the way zeitwerk autoloading works in Rails 7.2 and only manifests in production environment.  It doesn't appear in the development environment because of a reloading workaround that targets the development environment (see https://github.com/avalonmediasystem/avalon/pull/1747).  This workaround also fixes the issue in production but it the change in this PR feels like a better solution for this specific problem as well as being more limited in scope.

### DIFF
--- a/app/helpers/blacklight/local_blacklight_helper.rb
+++ b/app/helpers/blacklight/local_blacklight_helper.rb
@@ -25,10 +25,6 @@ module Blacklight::LocalBlacklightHelper
     blacklight_config.facet_fields.map {|facet,opts| opts[:group]}.uniq
   end
 
-  def url_for_document doc, options = {}
-    SpeedyAF::Base.for(doc.to_h.with_indifferent_access)
-  end
-
   def contributor_index_display args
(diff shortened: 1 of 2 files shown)


Dananji (approved):

PR #6128, opened 2024-11-18
Fix processing of google drive files

masaball:
When we made the switch from mediainfo to ffprobe, we created a new class for processing the file. Because Google Drive utilizes headers/params for its authorization flow, this change resulted in the file information becoming divorced from the auth token. By explicitly passing the auth token through to the classes downloading and processing the file, we are able to fix this oversight.

### DIFF
--- a/lib/avalon/ffprobe.rb
+++ b/lib/avalon/ffprobe.rb
@@ -23,7 +23,9 @@ module Avalon
       return @json_output unless @json_output.nil?
       return @json_output = {} unless valid_content_type?(@media_file)
       ffprobe = Settings&.ffprobe&.path || 'ffprobe'
-      raw_output = `#{ffprobe} -i "#{@media_file.location}" -v quiet -show_format -show_streams -of json`
+      # Include authorization headers for cases like Google Drive
+      header = "-headers 'Authorization: #{@media_file.auth_header.fetch('Authorization')}'" if @media_file.auth_header.present?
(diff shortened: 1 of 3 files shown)


lib/avalon/ffprobe.rb
```diff
@@ -23,7 +23,9 @@ def json_output
       return @json_output unless @json_output.nil?
       return @json_output = {} unless valid_content_type?(@media_file)
       ffprobe = Settings&.ffprobe&.path || 'ffprobe'
-      raw_output = `#{ffprobe} -i "#{@media_file.location}" -v quiet -show_format -show_streams -of json`
+      # Include authorization headers for cases like Google Drive
+      header = "-headers 'Authorization: #{@media_file.auth_header.fetch('Authorization')}'" if @media_file.auth_header.present?
+      raw_output = `#{ffprobe} #{header} -i "#{@media_file.location}" -v quiet -show_format -show_streams -of json`
```
masaball, line 28: This is downloading the file for processing, but we have already downloaded the file as part of the [`valid_content_type?` check](https://github.com/avalonmediasystem/avalon/blob/ecb7995e42339a9e35cf5a8c9d78336c03814c48/lib/avalon/ffprobe.rb#L79). 

It would probably be best as a separate ticket, but should we rework this so that when we are trying to access an internet hosted file we download to a tempfile and use the tempfile for all the processing instead of downloading it multiple times?

masaball:
Yeah, I'm not sure that any tests we tried to add would actually be looking at anything substantial.

PR #6130, opened 2024-11-19
Pin to specific revisions in github until releases of gems are made.

cjcolvar:
Turns out that `active_fedora-datastreams` didn't need a release or any code changes so I reverted to `~> 0.5.0`.  `speedy_af` got a release because it didn't actually depend on an `active_fedora` release.  This leaves `active_fedora` as the only dependency without a tag or release so I pinned it to a specific git commit.  I did this because I think it might take some time for community review of the Fedora 6 support in ActiveFedora and a release to be cut.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,9 +30,8 @@ gem 'shakapacker'
 
 # Core Samvera
 #gem 'active-fedora', '~> 15.0'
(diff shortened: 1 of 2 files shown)


Dananji (approved):

Issue #5702, opened 2024-02-29
Add Caption Files to Section During Processing


joncameron:
# Description
Media files added to Avalon may contain caption data within. Right now if a user has a container file with captions, they must process the file to pull the caption data into a separate file, then attach the captions file to the record in Avalon.

See https://github.com/avalonmediasystem/avalon/issues/2964 for a similar issue relating to pulling data/metadata from files upon ingest.

## Caption Types
- There are many caption formats; to begin we can only support the existing supported formats. Having ffmpeg convert the existing embedded text into a VTT could also be a way to handle this.
 
## Example File
This file contains Lunchroom Manners with the existing VTT for testing embedded into the container:
- https://drive.google.com/file/d/13KTK8LOrhp5c5n95wmqwHI0pJc8AZoRg/view?usp=drive_link
 
Active Encode will need to be updated to be able to pull out these supplemental files, and Avalon should be able to see that the data is there and create an Active Storage record for the data. One caveat is that this would be used for the ffmpeg adapter, but it might require extra effort for the AWS adapter. This could possibly be a feature that is restricted to the ffmpeg adapter, at least for the first pass.

# Done Looks Like
- [ ] Active Encode looks for caption/subtitle data in in media files during processing
- [ ] Valid caption data is written to a caption file and attached to the media file in Avalon if present

elynema:
Option for 8.0 release; depends on effort

joncameron:
Poster and thumbnail would be a nice thing to have in addition; others would be spritemaps for trick mode, waveforms, etc.

elynema:
Josh Bennett over at Media School is using Adobe Premier to create transcripts / captions for his media before uploading into MCO, so he is a potential collection owner who could test this out.

joncameron:
@masaball On avalon-dev I uploaded the Lunchroom Manners with embedded caption file example file (via web upload), and the video came through but no caption files showed up on the section afterward.

masaball:
The caption files do not show up until the file is fully processed, and the page does not currently automatically refresh so it will not reflect any attached captions until you manually reload the page. I meant to make a follow up ticket to look into that, but had forgotten. Checking the `encode_records` page, I found your test items and it looks like they have captions attached and working:

https://avalon-dev.dlib.indiana.edu/media_objects/9s161623j

https://avalon-dev.dlib.indiana.edu/media_objects/t722h885b

joncameron:
[...]

PR #6145, opened 2024-12-11
Create caption files during transcoding

masaball:
Related issue: #5702


cjcolvar (commented):
Waiting to review this once ActiveEncode PR gets merged.

app/models/watched_encode.rb
```diff
@@ -32,23 +32,37 @@ class WatchedEncode < ActiveEncode::Base
...
-          File.delete file.location
+      encode.output.collect! do |output|
+        if output.url.include?('supplemental_files')
+          new_file = SupplementalFile.new(tags: ['caption'], parent_id: record.master_file_id)
+          new_file.attach_file(FileLocator.new(output.url).location, io: true)
+          new_file.save
+          output.url = if Settings.active_storage.bucket.present?
+                      "s3://#{Settings.active_storage.bucket}/#{new_file.file.blob.key}"
+                    else
+                      new_file.file.blob.url
```
cjcolvar, line 44: This works but I thought there was a method in ActiveStorage that could give us the url in a backend independent way.  `new_file.file.url` appears to be that but for files stored on disk we would need to set `ActiveStorage::Current.url_options` and for s3 files we may want to strip the query params.  Given all that maybe this is a future improvement and not worth doing now.

app/models/supplemental_file.rb
```diff
@@ -33,9 +33,14 @@ class SupplementalFile < ApplicationRecord
   after_update_commit :update_index, prepend: true
   after_destroy_commit :remove_from_index
 
-  def attach_file(new_file)
-    file.attach(new_file)
-    extension = File.extname(new_file.original_filename)
+  def attach_file(new_file, io: false)
```
cjcolvar, line 36: Nice!

app/models/master_file.rb
```diff
@@ -314,11 +314,11 @@ def update_progress_on_success!(encode)
     # We can get the proper aspect ratio from the transcoded files, so we set the master file off the 
     # encode output.
     if is_video?
-      high_output = Array(encode.output).select { |out| out.label.include?("high") }.first
+      high_output = Array(encode.output).select { |out| out.label&.include?("high") }.first
       self.display_aspect_ratio = (high_output.width.to_f / high_output.height.to_f).to_s
     end
 
-    outputs = Array(encode.output).collect do |output|
+    outputs = Array(encode.output).reject { |output| output.subtitles.nil? }.collect do |output|
```
cjcolvar, line 321: I think it would be good to have a clearer test that only captures the named outputs requested when the encode is created.  Maybe something like this is sufficient?
```suggestion
    outputs = Array(encode.output).select { |output| ["high", "medium", "low"].include?(output.label) }.collect do |output|
```

app/models/watched_encode.rb
```diff
@@ -32,23 +32,37 @@ class WatchedEncode < ActiveEncode::Base
...
+                    end
+          output.id = new_file.id
```
cjcolvar, line 46: How about making this the global id?  Then it can be verified that the output is actually a SupplementalFile when adding it to the MasterFile.
```suggestion
          output.id = new_file.to_global_id.to_s
```

app/models/master_file.rb
```diff
@@ -352,6 +357,12 @@ def update_derivatives(outputs, managed = true)
     save
   end
 
+  def add_supplemental_files(ids)
+    self.supplemental_files += SupplementalFile.find(ids)
+
+    save
```
cjcolvar, line 363: Could the save be moved up into `update_progress_on_success!` To avoid saving both here and in `update_derivatives`?

cjcolvar (commented):
This is looking good overall!  Exciting to see these changes in avalon to handle additional non-Derivative outputs from ActiveEncode.  I'm hopeful the approach in this PR can be easy to extend if/when we add additional outputs other than captions.

spec/models/watched_encode_spec.rb
```diff
@@ -101,6 +101,42 @@
...
+
+        it 'creates a SupplementalFile' do
+          expect { encode.create! }.to change { SupplementalFile.all.count }.by(1)
+          expect(master_file).to have_received(:update_progress_on_success!)
+        end
+
+        it 'uploads to Minio' do
+          encode.create!
+          expect(master_file).to have_received(:update_progress_on_success!)
+        end
```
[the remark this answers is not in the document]
masaball: I'm not sure. The Active Storage service is set at application boot and so the test suite is set to DiskService and I have been struggling to find a way to override that or mock it to test the Supplemental File. 

Maybe we just remove this test and trust that Active Storage will properly do its thing?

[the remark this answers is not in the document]
cjcolvar: I think it would be good to keep this test as a check that the supplemental file gets attached in the `after_completed` callback.  So maybe ActiveStorage service independent would be best.  Do you think this tests enough?
```
supplemental_file = master_file.supplemental_files.last
expect(supplemental_file.file.present?).to eq true
expect(supplemental_file.file.byte_size > 0).to eq true
```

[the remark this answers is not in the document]
masaball: That set of tests is probably sufficient, though Active Storage has an `attached?` method which may be better/more appropriate to check than `present?`. Will make the changes on Monday and see how it goes.

app/models/master_file.rb
masaball: As I've been working on getting other stuff to work with the spec test, I don't think that this test will work consistently enough. 

I'm not sure that outputs are guaranteed to be named "high", "low", etc. In `config/ffmpeg_presets.yml`, the label for outputs is a configurable option. I doubt that people would change that but they could. Also, do we know if files processed through the S3 adapters are always labelled "high", etc? Can we guarantee that is consistent?

Maybe a better test for now would be to reject `output.format == 'vtt'` like we filter off of in watched_encode? Or would there be an easy way to get the actual values the encode was created with and filter off those?

cjcolvar: I think getting the output labels from the original encode request would be ideal but rejecting `output.format == 'vtt'` seems good enough for now.

cjcolvar (approved):
Looks great!

Issue #6094, opened 2024-10-30
solrconfig.xml provided in solr/conf compatibility with Solr 9


sgurnick:
I provisioned a fresh Solr 9.7 server in conjunction with an upgrade of Avalon from `7.6` to `7.7.2`. It is my understanding that Avalon `7.7.2` does support the use of Solr 9. 

After I confirmed my standalone Solr 9 instance was up and running, I created a new core and used the files provided in [solr/conf](https://github.com/avalonmediasystem/avalon/tree/v7.7.2/solr/conf) of the Avalon repo for the `7.7.2` tag. 

After restarting Solr, the core could not load. These warnings and errors appeared in the `solr.log` file:
```
2024-10-30 17:09:32.929 WARN  (coreLoadExecutor-13-thread-1) [c: s: r: x:avalon t:] o.a.s.c.SolrConfig Couldn't add files from /opt/solr-9.7.0/contrib/analysis-extras/lib filtered by null to classpath: java.nio.file.NoSuchFileException: /opt/solr-9.7.0/contrib/analysis-extras/lib
2024-10-30 17:09:32.929 WARN  (coreLoadExecutor-13-thread-1) [c: s: r: x:avalon t:] o.a.s.c.SolrConfig Couldn't add files from /opt/solr-9.7.0/contrib/analysis-extras/lucene-libs filtered by null to classpath: java.nio.file.NoSuchFileException: /opt/solr-9.7.0/contrib/analysis-extras/lucene-libs
2024-10-30 17:09:32.931 WARN  (coreLoadExecutor-13-thread-1) [c: s: r: x:avalon t:] o.a.s.c.SolrConfig Couldn't add files from /opt/solr-9.7.0/contrib/extraction/lib filtered by .*\.jar to classpath: java.nio.file.NoSuchFileException: /opt/solr-9.7.0/contrib/extraction/lib
2024-10-30 17:09:32.931 WARN  (coreLoadExecutor-13-thread-1) [c: s: r: x:avalon t:] o.a.s.c.SolrConfig Couldn't add files from /opt/solr-9.7.0/dist filtered by solr-cell-\d.*\.jar to classpath: java.nio.file.NoSuchFileException: /opt/solr-9.7.0/dist

...

[...]

PR #6162, opened 2025-01-08
Update config to reference correct solr 9 paths for contrib modules

cjcolvar:
Fix provided by @sgurnick in #6094.


masaball (approved):
Looks good

PR #6173, opened 2025-01-22
Use safe navigator in case comments or tags are nil

cjcolvar:
Should resolve https://app.honeybadger.io/projects/54117/faults/112541833

Most playlists in our production site have empty strings for comments but a few had `nil` which caused them not to open.  I changed all of those playlists to empty string in the database but this code change should also handle this edge case (and tags too for good measure).


masaball (approved):
:+1:

PR #6175, opened 2025-01-22
Fixes for single downloads from Google Drive

masaball:
This commit includes three fixes:
1. The way that we were generating and referencing the temporary file marked it as ready for Garbage Collection before we were finished making use of it and so we would get intermittent "file does not exist" errors if the garbage collector ran at the wrong time. Switching to `Tempfile.create` fixes this issue, but requires us to manually remove the tempfile after use.
2. Even though we would download to a local file, we were still passing the google auth_headers around to the various modules and methods under the assumption it was safe. This was an incorrect assumption. FFProbe has been adjusted to only use the auth headers if the object passed in is http/https.
3. The final issue is the same as in #2 but related to active encode's processing. By setting the auth header to nil after it is no longer needed, we can prevent it getting passed further and causing issues.

### DIFF
--- a/app/models/master_file.rb
+++ b/app/models/master_file.rb
@@ -210,6 +210,7 @@ class MasterFile < ActiveFedora::Base
       else
         local_file = FileLocator.new(file, filename: file_name, auth_header: auth_header).local_location
         saveOriginal(File.open(local_file), file_name, dropbox_dir)
+        auth_header = nil
       end
     else #Batch
       saveOriginal(file, File.basename(file.path), dropbox_dir)
--- a/app/services/file_locator.rb
+++ b/app/services/file_locator.rb
@@ -104,8 +104,13 @@ class FileLocator
     end
   end
 
+  # Tempfile.create generates a regular File object instead of a Tempfile object:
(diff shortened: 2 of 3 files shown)


masaball:
That is actually handled by the `FileMover` methods, which are called within `saveOriginal`. All of those methods remove the source file as part of their operation.

As currently programmed the tempfile is created in the system default location which looks at `Dir.tmpdir`. This will usually be `/tmp`. So if the working path is configured to use /tmp, the job would clean it up if the file mover didn't. But if the working path is set to something else, then the job would miss it.

I can adjust the tempfile generation so that it writes to the working path if present and we can guarantee that there is the fallback if the FileMover fails to remove it.

cjcolvar:
Thanks for clarifying.  Creating the tempfile in `Dir.tmpdir` is probably best because `saveOriginal` will create the working file after `FileMove` moves the temp file into the collection dropbox.  I think there is the off chance that an exception occurs but then the file would just stay in the temp directory which should get cleaned up eventually or at least be clearer that it is a temp file.  Also I'm guessing that the CleanupWorkingFileJob probably wouldn't get enqueued and run if an exception did happen.

Issue #6134, opened 2024-11-25
[BUG] Error Displayed When Clicking 'Display Returned Items' on Checkouts Page


abankhele:
**Describe the bug**
The "Display Returned Items" button on the Checkouts page triggers an error when clicked, instead of displaying the list of returned items.

**To Reproduce**

Steps to reproduce the behavior, including the results:

1. Go to 'https://avalon-staging.dlib.indiana.edu/' and log in
2. Go to the Checkouts page, ensuring there are returned items available.
3. Click on the Display Returned Items button.
4. Observe the error popup displayed instead of the expected list.

**Expected behavior**
Clicking the Display Returned Items button should display a list of items that were returned, without any errors.

**Screenshots/Screen Recording**
<img width="1358" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOHpqQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f59f328102d3feeab7fb998709a4cff49137fb13/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform : Laptop
 - OS: macos

- [ ] avalon-dev
- [x] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

abankhele:
Clicking the Display Returned Items button now displays a list of items that were returned. The issue is fixed on avalon-dev.

PR #6176, opened 2025-01-23
Skip checkouts belonging to deleted items

masaball:
Related issue: #6134 

Checkouts created for items that were subsequently deleted are not cleared from the database. This caused a datatables error when looking at the Checkouts page. This PR sets these errored checkouts to nil and then clears them from the results JSON so that datatables populated correctly.

This is out of scope for this issue, but should we have some sort of handling to delete checkout records when the referenced media object is deleted? Or a rake task/cron job for clearing these orphaned records semi-regularly?

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -29,7 +29,7 @@ class CheckoutsController < ApplicationController
             else
               user_array(checkout)
             end
-          end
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
This seems fine for now but I agree that it would be nice to clear out checkouts of deleted objects.

PR #6177, opened 2025-01-24
Support solr 9.8 by setting SOLR_MODULES env variable

cjcolvar:
Based upon https://github.com/projectblacklight/blacklight/pull/3497


cjcolvar:
The build had one unrelated flaky test so I think this is actually working fine.

masaball (approved):

Issue #6109, opened 2024-11-11
Store Original Filename for Section in a Separate Field


joncameron:
# Description
If a section title is manually entered then we use that for section name on the item page; otherwise the file location is used. file_location is stored as the full path or URI to the location of the source media file. To get a better decoupling between these and avoid having to do extra string parsing, we should store original filename in a separate field.

# Done Looks Like
- [x] original_filename added to masterfile properties
- [x] original_filename contains just the filename of the source media
- [x] set original_filename as the section label default when generating the IIIF manifest for the item
- [x] backfill as needed (as rake task or script to populate the field)

joncameron:
Should this default label value populated by the original_filename field be indexed into the record for ease of searching?

joncameron:
See #5521 as well.

joncameron:
👍

PR #6180, opened 2025-01-30
Add `original_filename` field to MasterFile

masaball:
Related issue: #6109

### DIFF
--- a/lib/tasks/avalon_migrations.rake
+++ b/lib/tasks/avalon_migrations.rake
@@ -114,5 +114,18 @@ namespace :avalon do
         puts "Migration completed successfully."
       end
     end
+    desc "Backfill original filename property on MasterFile records"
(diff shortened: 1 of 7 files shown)


lib/tasks/avalon_migrations.rake
```diff
@@ -114,5 +114,18 @@ namespace :avalon do
         puts "Migration completed successfully."
       end
     end
+    desc "Backfill original filename property on MasterFile records"
+    task master_file_original_filename: :environment do
+      count = 0
+      missing = MasterFile.where(original_filename: nil)
+      missing.each do |mf|
```
cjcolvar, line 121: I'm fine with merging this now, but as part of testing we should test that this works and is performant when running on a large repository.

cjcolvar (approved):
Looks good!

Issue #6160, opened 2025-01-07
Read only mode


elynema:
# Description

For large production sites (like IU) to migrate to 8.0 requires a Fedora 4 to Fedora 6 data migration. We anticipate that will take several days at IU, and we do not currently have a straightforward process to perform a delta update of records updated/created/deleted in the production system during the migration process.

It would be useful to have a "read only" mode we could enable in Avalon where users could access materials, but not make changes.

Actions that should be unavailable include:
- edit item (including structure, supplemental files, access/discovery, etc)
- publish / unpublish item
- create item
- delete item
- create collection
- delete collection
- edit collection metadata
- edit item access at the collection level
- create thumbnail/poster
- selected items actions: merge, move, update access control, publish / unpublish, delete
- push item from Dark Avalon to MCO
- batch ingest
- API uploads of structure / supplemental files
- API item creation

This feature can be scoped to apply to Fedora data and actions, and not things that rely on the database. 

The custom "Read only administrator" configuration in the MDPI Internal Content Access Avalon instance can be used to provide a base for this functionality.

# Done Looks Like

- [x] flesh out list of actions that should be unavailable
- [x] investigate difficulty in setting this up as a feature in the codebase
- [x] read-only mode should hide action/edit buttons and show a relevant splash message if feasible

elynema:
What about actions that don't impact Fedora, just the Avalon database?
- create / edit timeline
- create / edit playlist
- check out / return CDL items
- select / unselect items

elynema:
repository_read_only_mode in settings - defaults to false; can be set to true. Would need to add documentation about that settings.

Do we want a banner that displays to collection staff when the system is in read only mode? This could also be useful to display messages to collection staff at other times (ex: upcoming outages).

elynema:
Tested on avalon-dev as admin user. Great job making buttons not available. The only oddity I saw was that on the Manage Collection page, the buttons are disabled rather than not appearing. This is not a big deal as it still accomplishes our goals. The Change Poster button looks a little funky since it is overlaid on top of the upload button.

<img width="1223" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVhoQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9fbb76b9702d47a56bbceb353e673e725a2bfde4/image.png" alt="image.png" />

[...]

PR #6183, opened 2025-02-03
Repository read only mode

cjcolvar:
Implements #6160

This PR attempts to block write actions in two contexts, UI and rails console, when `Settings.repository_read_only_mode` is set to true.

For UI actions this is handled in `Ability` by tacking `cannot` statements at the end of the Ability processing so they should take precedence.  UI write elements like buttons shouldn't appear or be enabled because the ability for the actions is revoked.

For rails console actions, I overrode `ActiveFedora::Base.readonly?` which should raise a `ReadOnlyRecord` error if a write operation is attempted.

Operations that work at a lower level than ActiveFedora or outside of rails could still modify fedora or solr but they should only be available to system administrators who would be the ones setting Avalon into read-only mode so it is up to them to not mess things up. :slightly_smiling_face:

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -22,7 +22,8 @@ class Ability
                          :marker_permissions,
                          :encode_dashboard_permissions,
                          :timeline_permissions,
-                         :checkout_permissions]
+                         :checkout_permissions,
+                         :repository_read_only_permissions]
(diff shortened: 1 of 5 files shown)


masaball (approved):

PR #6188, opened 2025-02-05
Adding data-testids for cypress tests

abankhele:
I have added data-testid attribute in html tags. Collection and Playlist now use this atrribute for testing.


db/schema.rb
```diff
@@ -196,7 +196,7 @@
   create_table "playlists", force: :cascade do |t|
     t.string "title"
     t.bigint "user_id", null: false
-    t.text "comment"
+    t.string "comment"
```
cjcolvar, line 199: This is a bit odd but we don't commit this change even though it was automatically made by rails.  Text vs. string can be different with different DBs (mysql vs. postgresql) and I've always been afraid that switching might cause the field to get truncated.  Please undo this change and I'll look into if it is safe to make or not.
```suggestion
    t.text "comment"
```

spec/cypress/integration/collection_spec.js
```diff
@@ -0,0 +1,235 @@
...
+		Cypress.on('uncaught:exception', (err, runnable) => {
+
+			return false;
+		});
+
+        // Wait for the intercepted request and validating the response. Intercepting the async behaviour
+        cy.wait('@createCollection').then((interception) => {
+            expect(interception.response.statusCode).to.eq(302);
+            expect(interception.response.headers.location).to.include('/admin/collections/');
+        });
```
cjcolvar, line 39: This is neat!  I haven't seen `cy.intercept` before.

spec/cypress/integration/item_spec.js
```diff
@@ -80,22 +80,22 @@ context('Item', () => {
...
+    ).click()
+
+
     // Validate the item title, collection, and publication date
     cy.get('.page-title-wrapper h2').should('contain.text', item_title);
 
     cy.get('div.ramp--tabs-panel').within(() => {
       cy.get('div.tab-content dt')
-        .contains('Date')
+        .contains('Publication date') //changed from Date
```
cjcolvar, line 98: Nice catch!  We changed this in the UI but forgot to change it here.

app/javascript/components/collections/CollectionCardBody.js
```diff
@@ -17,7 +17,7 @@
 import React from 'react';
 
 const CollectionCardBody = ({ children }) => (
-  <div className="card-body">
+  <div className="card-body" data-testid="collectionCardBody">
```
cjcolvar, line 20: It looks like this div might get rendered on the page multiple times.  Will having the same testid repeated on the page cause an issue?

app/views/modules/_access_object.html.erb
```diff
@@ -71,7 +71,7 @@ Unless required by applicable law or agreed to in writing, software distributed
         <% if members.present? %>
         <% members.each do |member_object| %>
         <tr>
-          <td class='access_list_label'>
+          <td  class='access_list_label' data-testid='accessListLabel'>
```
cjcolvar, line 74: It looks like this one will be repeated too.

spec/cypress/integration/playlist_spec.js
```diff
@@ -61,75 +60,86 @@ context('Playlists', () => {
     // Validate the presence of the video.js element
     cy.get('video[data-testid="videojs-audio-element"]')
       .should('exist')
-      .and('have.class', 'video-js')
-      .and('have.class', 'vjs-big-play-centered');
-
+      
+      // .and('have.class', 'video-js')
+      // .and('have.class', 'vjs-big-play-centered');
+      //why were they added
```
cjcolvar, line 66: I agree.  I don't think these checks added any value.

cjcolvar (commented):
This looks really good overall!  Nice job making the tests more resilient and fixing up the failing tests!

I am seeing kebab case (e.g. `delete-table-view`) as the most commonly used for test ids in other applications (like Ramp).  Do you think it would be worth it to switch?  Sorry I didn't catch this earlier when you presented your work.

abankhele:
Thank you!
I was following camel-case for data-testid but yes it makes more sense to switch to kebab-case like ramp, making it consistent throughout.

Dananji (commented):
This looks great! 🎉 

I had one suggestion, when creating the `data-testid` attributes, does it makes more sense to prefix these ids with the page name? 

For example: for playlist index page table use `data-testid="playlist-index-table"`

Since this test suite could potentially grow to a larger set of tests, I think it would be helpful to have a format like this with future implementations as well readability of the code.

abankhele:
Thanks for the suggestion! I’ll update the data-testids to include the page name.

app/javascript/components/collections/CollectionCardBody.js
cjcolvar: If there are multiple collections being rendered in the collection listing page then I think this testid won't be unique which could cause problems.  But if this isn't causing the tests to fail then I think it is okay for now.  

For unique ids I think they could incorporate an index or the id of the collection.  But this would require some more significant refactoring of the javascript components to put in place.  In the future if you do run into a problem with this testid we can fix the components then.

cjcolvar (approved):
Looks great!  Thanks for your hard work on this!

app/javascript/components/collections/CollectionCardBody.js
abankhele: Okay, I can check if we can get unique ids for each card. For now I am getting them for checking if the collection that we searched for is visible.
`cy.get("[data-testid='collection-card-body']").contains('a', search_collection); `.

abankhele:
Thank you so much!

Issue #6147, opened 2024-12-13
Pull label and language from embedded captions to populate in Avalon caption


elynema:
# Description

When pulling caption tracks from the .mp4 container (see #5702 for initial work), the caption label/name is currently be auto-generated based on the filename.

It would be great to pull any label and language from the text track (if they exist) to populate those fields in Avalon.

# Done Looks Like
- [ ] Investigate location in file container metadata for label and language values associated with timed text tracks
- [ ] find an example of a media file with these qualities for testing
- [ ] implement retrieval of label and language metadata for adding to supplemental file metadata fields

elynema:
Metadata track has both title and handler name. If both are present, mediainfo concatenates them with " / " between, so could see labels coming through like "English / English." This is a mediainfo quirk; create a new issue if we want to change this behavior.

joncameron:
I created a Lunchroom Manners with the embedded caption set to German and it came through as expected!

<img width="792" alt="Image" src="https://github.com/user-attachments/assets/6b9caa93-6498-49c8-9c3d-14e2e57fc527" />

PR #6189, opened 2025-02-06
Handle embedded subtitle metadata (label, language)

masaball:
Related issue: #6147 

Wait to review until samvera-labs/active_encode/pull/143 has been merged

### DIFF
--- a/app/models/language_term.rb
+++ b/app/models/language_term.rb
@@ -15,6 +15,79 @@
 class LanguageTerm
   class LookupError < Exception; end
 
+  class Iso6391
+    Terms = File.join(Rails.root, 'config/iso639-1.yml')
+
+    class << self
+      def convert_to_6392(term)
+        raise LookupError, "Incorrect number of characters. Term must be a string of 2 alphabetic characters." unless /[a-zA-Z]{2}/.match?(term)
+        lang_text = self.find(term).text
+        # ISO 639-1 can have multiple languages defined for a single code:
+        # 'es': "Spanish | Castilian".
+        # ISO 639-2 does not follow the same convention, so we iterate through
+        # any multi languages until we get a match from the ISO 639-2 standard.
+        lang_text = lang_text.split('|').map(&:strip) if lang_text.include?('|')
+        Array(lang_text).each do |text|
+          begin
+            @alpha3 = LanguageTerm.find(text).code
+            break
+          rescue LookupError
+            next
+          end
+        end
+
+        raise LookupError, "Unknown language: `'#{value}" if @alpha3.nil?
+
+        return @alpha3
+      end
+
+      def map
+        @@map ||= self.load!
+      end
+
+      def find(value)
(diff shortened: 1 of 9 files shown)


app/models/language_term.rb
```diff
@@ -15,6 +15,79 @@
 class LanguageTerm
   class LookupError < Exception; end
 
+  class Iso6391
```
cjcolvar, line 18: It seems like it might be possible to DRY this up by making this a subclass of `LanguageTerm` and making some constants that get overridden (e.g. controlled vocabulary URL and file path).

If you do go this route, you should be careful that the class instance variable `@@map` isn't shared between the two implementations.

app/models/watched_encode.rb
```diff
@@ -36,6 +36,25 @@ class WatchedEncode < ActiveEncode::Base
...
+            when output.language.length == 3
+              # Ensure that output.language is in the controlled vocabulary because
+              # the language field is validated on the SupplementalFile model.
+              new_file.language = LanguageTerm.find(output.language).code
+            else
+              new_file.language = nil
+            end
+          rescue LanguageTerm::LookupError
+            new_file.language = nil
+          end
```
cjcolvar, line 57: I'm wondering if this could be pulled out to a static method on `LanguageTerm`.  Which makes me wonder if `LanguageTerm` as it currently exists should turn into a subclass of itself focused on Iso6392 and `LanguageTerm.find` would turn into something like:
```
def self.find(term)
  case term.length
  when 3
    Iso6391.convert_to_6392(term)
  when 2
    Iso6392.find(term).code
  else
    nil
rescue LookupError
  nil
end
```

cjcolvar (commented):
I like where this is going.  Expanding on LanguageTerm feels like it could be useful in the future beyond this PR.

cjcolvar (approved):
Looks great!

app/models/watched_encode.rb
```diff
@@ -36,6 +36,12 @@ class WatchedEncode < ActiveEncode::Base
       encode.output.collect! do |output|
         if output.format == "vtt"
           new_file = SupplementalFile.new(tags: ['caption'], parent_id: record.master_file_id)
+          new_file.label = output.label.presence
+          begin
+            new_file.language = LanguageTerm.find(output.language).code if output.language.present?
+          rescue LanguageTerm::LookupError
+            new_file.language = nil
+          end
```
cjcolvar, line 44: Not a big deal but I've read that begin/rescue has an overhead cost so it would be better here to skip that block by moving the guard clause outside of it.
```suggestion
          if output.language.present?
            begin
              new_file.language = LanguageTerm.find(output.language).code
            rescue LanguageTerm::LookupError
              new_file.language = nil
            end
          end
```

masaball: That makes sense. Good catch!

cjcolvar (approved):

Issue #6190, opened 2025-02-07
[BUG] docker-compose run createbuckets fails


jmcgranahan:
**Describe the bug**
On an Ubuntu 24.04.1 LTS server within an AWS environment with no other applications running, we are attempting to install the newest version of Avalon since it is compatible with Fedora 6. However, at the step to createbuckets, it always fails:

mc: <ERROR> Unable to initialize new alias from the provided credentials. Get "http://minio:9000/probe-bsign-mk1dfiympem9dhqqjp2mmsftlegtvj/?location=": dial tcp 172.18.0.2:9000: connect: connection refused.

**To Reproduce**
Steps to reproduce the behavior, including the results:
Launch an Ubuntu 24.04.1 LTS server in AWS using their Graviton processor. 
Install Docker - apt install docker
Install Docker-compose - curl -SL "https://github.com/docker/compose/releases/download/v2.32.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose

check & verify Docker/Docker-compose -
Docker version 27.5.1, build 9f9e405
Docker Compose version v2.32.4

Then:
1. Go to https://github.com/avalonmediasystem/avalon and follow the "Quickstart development with Docker"
2. git clone https://github.com/avalonmediasystem/avalon
3. cd avalon
4. cp config/controlled_vocabulary.yml.example config/controlled_vocabulary.yml
5. docker-compose pull
6. docker-compose run createbuckets

docker-compose run createbuckets
[+] Creating 1/1
 ✔ Container avalon-minio-1  Created                                                                                                                                                                         0.0s
[+] Running 1/1
 ✔ Container avalon-minio-1  Started                                                                                                                                                                         0.5s
mc: Configuration written to `/root/.mc/config.json`. Please update your access credentials.
mc: Successfully created `/root/.mc/share`.
mc: Initialized share uploads `/root/.mc/share/uploads.json` file.
mc: Initialized share downloads `/root/.mc/share/downloads.json` file.
**mc: <ERROR> Unable to initialize new alias from the provided credentials. Get "http://minio:9000/probe-bsign-mk1dfiympem9dhqqjp2mmsftlegtvj/?location=": dial tcp 172.18.0.2:9000: connect: connection refused.**
Bucket created successfully `myminio/fcrepo`.
Bucket created successfully `myminio/masterfiles`.
Bucket created successfully `myminio/derivatives`.
Bucket created successfully `myminio/supplementalfiles`.
[...]

PR #6191, opened 2025-02-10
Wait on minio container before attempting to create buckets

cjcolvar:
As reported in #6190 creation of buckets sometimes fails.  This appears to be a timing issue where the minio container is not available when the createbuckets container attempts to connect.  This is resolved by changing the `depends_on` in the createbuckets container to check that the minio container is healthy instead of just created which is the default.  (See https://docs.docker.com/compose/how-tos/startup-order/#control-startup for more details.)

Also included in this PR: use minio latest tag

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,7 +173,7 @@ services:
     ports: []
 
   minio:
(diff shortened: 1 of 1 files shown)


masaball (approved):
:+1:

PR #6193, opened 2025-02-11
Additional fixes for read-only mode

cjcolvar:
Part of #6160

- Administrative facets should be visible
- Manage Content should list collections
- Delete button should not appear next to collections on manage content page
- Push button should not appear on selected items page or item view page
- Apply to All Existing Items button on collection edit page should be disabled

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -91,18 +91,18 @@ class CatalogController < ApplicationController
     config.add_facet_field 'rights_statement_ssi', label: 'Rights Statement', limit: 5, helper_method: :rights_statement_facet_display, solr_params: { "facet.contains" => "http"}
 
     # Hide these facets if not a Collection Manager
-    config.add_facet_field 'workflow_published_sim', label: 'Published', limit: 5, if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow"
-    config.add_facet_field 'avalon_uploader_ssi', label: 'Created by', limit: 5, if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow"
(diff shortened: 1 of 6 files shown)


app/models/collection_search_builder.rb
```diff
@@ -64,7 +64,7 @@ def apply_gated_discovery(solr_parameters, permission_types = discovery_permissi
 
     # Copied from SearchBuilder
     def only_published_items(solr_parameters, ability = current_ability)
-      if ability.cannot? :create, MediaObject
```
cjcolvar, line 67: `can? :create, MediaObject` is equivalent to `is_member_of_any_collection?`
See https://github.com/avalonmediasystem/avalon/blob/1ab561e38327afa172fcb0f62fe8d6a2f33cf283/app/models/ability.rb#L71-L73

app/controllers/catalog_controller.rb
```diff
@@ -91,18 +91,18 @@ class CatalogController < ApplicationController
     config.add_facet_field 'rights_statement_ssi', label: 'Rights Statement', limit: 5, helper_method: :rights_statement_facet_display, solr_params: { "facet.contains" => "http"}
 
     # Hide these facets if not a Collection Manager
-    config.add_facet_field 'workflow_published_sim', label: 'Published', limit: 5, if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow"
-    config.add_facet_field 'avalon_uploader_ssi', label: 'Created by', limit: 5, if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow"
-    config.add_facet_field 'read_access_group_ssim', label: 'Item access', if: Proc.new {|context, config, opts| context.current_ability.can? :create, MediaObject}, group: "workflow", query: {
+    config.add_facet_field 'workflow_published_sim', label: 'Published', limit: 5, if: Proc.new {|context, config, opts| context.current_ability.can?(:read, :administrative_facets)}, group: "workflow"
```
masaball, line 94: Where is `administrative_facets` coming from? Is that a blacklight built in?

masaball (approved):

Issue #6192, opened 2025-02-10
[BUG] Files tab is showing up when there are only transcript files


elynema:
**Describe the bug**
Due to transcript files being downloadable from the transcript tab, we stopped displaying them on the Files tab (to prevent redundancy). However, the Files tab now shows up with the message "No Supplemental file(s) in Manifest" when there is only a transcript file.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to https://media.dlib.indiana.edu/media_objects/1v53kc546
2. Click on Files tab
3. See error

**Expected behavior**
The Files tab should not be rendered if there are no files to list for download. If there are only transcript files(s) available as supplemental files, the File tab should not be rendered.

**Screenshots/Screen Recording**
<img width="997" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelhPQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b2b6d3f1848cc9f1da5ffdeae0e051d4590bbe5f/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: MacOS laptop
 - OS: Sequoia
 - Browser: Chrome
- [ ] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [x] mco-production
- [ ] demo

**Release:**
I am guessing this was introduced in Release 7.8, but have not tested to be sure?

**Additional notes**
I anticipate this issue will show up on avalon-dev and mco-staging, as well.

elynema:
Pass a variable to Ramp based on whether a media item has supplemental files present. Current code is not set up to distinguish transcripts from other supplemental files. Need to adjust the scoping on the hasFiles check.

elynema:
@elynema Find and link to issue where we decided not to display transcripts on the Files tab.

abankhele:
This issue is fixed and works as expected on avalon-dev.

PR #6200, opened 2025-02-19
Change has_files in item-view to include only caption and supplemental files in sections

Dananji:
Related issue: #6192

### DIFF
--- a/app/presenters/speedy_af/proxy/media_object.rb
+++ b/app/presenters/speedy_af/proxy/media_object.rb
@@ -144,6 +144,10 @@ class SpeedyAF::Proxy::MediaObject < SpeedyAF::Base
     sections.select { |master_file| master_file.supplemental_files(tag: tag).present? }.map(&:id)
   end
 
+  def sections_with_rendering_files(tags)
+    tags.select { |t| sections_with_files(tag: t).present? }.present?
+  end
+
   def permalink_with_query(query_vars = {})
     val = permalink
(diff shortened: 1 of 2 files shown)


app/presenters/speedy_af/proxy/media_object.rb
```diff
@@ -144,6 +144,10 @@ def sections_with_files(tag: '*')
     sections.select { |master_file| master_file.supplemental_files(tag: tag).present? }.map(&:id)
   end
 
+  def sections_with_rendering_files(tags)
+    tags.select { |t| sections_with_files(tag: t).present? }.present?
+  end
```
masaball, line 149: The convention in Ruby is for methods returning boolean values to end with `?`. And there is an `any?` method which we could use instead of `select {}.present?` which should be more efficient.

```suggestion
  def sections_with_rendering_files?(tags)
    tags.any? { |t| sections_with_files(tag: t).present? }
  end
```

To keep parity between the proxy and the model, could you add this method to the MediaObject model as well? We should probably have modules shared between each Model and its Proxy (or at least MediaObject and MasterFile since we touch those most often) so that we only need to add methods in one place in cases like this. Something for @cjcolvar and I to talk over maybe.

Dananji: Thanks for the feedback! I will make the suggested changes.

masaball (approved):

PR #6208, opened 2025-02-26
Tell bots not to request m3u8 files

### LINKED ISSUE #4405: Change robots.txt to Exclude M3U8
# Description
Robots.txt should be updated to exclude m3u8 manifest routes so these don't get scraped (no value in letting bots hit those endpoints).

# Done Looks Like
- [ ] robots.txt file for Avalon excludes m3u8 routes

cjcolvar:
Resolves #4405

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,4 @@
 # See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+
+user-agent: *
+disallow: /*.m3u8


public/robots.txt
```diff
@@ -1 +1,4 @@
 # See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+
+user-agent: *
+disallow: /*.m3u8
```
masaball, line 4: After looking over the documentation, I don't think this will work. 

> Note also that globbing and regular expression are not supported in either the User-agent or Disallow lines. The '\*' in the User-agent field is a special value meaning "any robot". Specifically, you cannot have lines like "User-agent: *bot*", "Disallow: /tmp/*" or "Disallow: *.gif".

With that, I think we would need to disallow each possible m3u8 endpoint so something like:
```suggestion
disallow: /index.m3u8
disallow: /auto.m3u8
disallow: /high.m3u8
disallow: /medium.m3u8
disallow: /low.m3u8
```

cjcolvar: I was looking at Google's docs and it allows wildcards: https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt

From reading that page it looks like specifying each route probably won't work for us because the disallow isn't relative so we would need to know the masterfile ids which leads back to the wildcard.

masaball (approved):

Issue #6209, opened 2025-02-26
[BUG] "Add to Playlist" Button Adds Entire Media Object Instead of Selected Section


abankhele:
**Describe the bug**
When clicking the "Add to Playlist" button and selecting a specific section to add, the entire media object is added instead of just the selected section.

**To Reproduce**

1. Navigate to Media Object.
2. Click on Add to Playlist.
3. Select Current Section (e.g., section_name).
4. Ensure that "Create playlist items for each track/subsection" is checked.
5. Click the Add button.
6. Open the playlist and observe that playlist items are created for all sections of the media object, not just the selected section.

**Expected behavior**
Only the selected section should be added to the playlist, rather than the entire media object.

**Screenshots/Screen Recording**

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjhJQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e28da451eeb3b03a4c1794f020691d44ed729a96/Screen%20Recording%202025-02-26%20at%201.54.01%E2%80%AFPM.mov)

**Environment (please complete the following information):**
 - Device :Desktop
 - OS:macos
 - Browser : chrome
- [ ] avalon-dev
- [ ] avalon-staging
- [x] mco-staging
- [ ] mco-production
- [ ] demo

joncameron:
Reproducible in avalon-dev as well, or restricted to mco-staging?

elynema:
@joncameron This seems like it needs to be cherrypicked out to MCO with the 8.0 release b/c we don't want to introduce this bug?

elynema:
@abankhele Looks like this is available to QA on avalon-dev. Can you take a look?

abankhele:
Only the selected section is added to the playlist, not the entire media object. I tested it, and it works as expected.

PR #6215, opened 2025-03-03
Add current section to playlist

masaball:
Related issue: #6209


masaball:
Looking back at the ticket, perhaps I may have had too simple a structure on my test item as I'm not sure I had multiple divs. 

Regardless, I was interpreting the current section option to be for adding an entire masterfile and then for adding just a subset of `div`s or `span`s you would use the current track option. If I have that behavior right, seeing all four spans in the playlist would be the expected behavior. So it would be that the modal just has the `current-section` name wrong...

cjcolvar (approved):

Issue #6214, opened 2025-03-03
[BUG] Streaming issues across browsers 


elynema:
# Description

Seeing inconsistent streaming behaviors across browsers on mco-staging. In Safari, content will stream ok, but in Chrome it gives error in the media player: "media could not be loaded. Network error or media format is not supported" and content does not load.

In the Chrome network inspector, I see a 505 for the request for the auto.m3u8 file and then all the request for streaming chunks get 404 (Not found)

# To reproduce

1. Go to this item https://mco-staging.dlib.indiana.edu/media_objects/08612s75g or https://mco-staging.dlib.indiana.edu/media_objects/2r36v5177 in Chrome
2. Open up dev console
3. switch between sections to watch for m3u8 requests

# Questions

- [ ] Chris believes there is a bug in the mp3 streaming code new in 8.0 combined with older existing Fedora data. Do we need to create a bugfix release for this to solve problems in 8.0 release?

elynema:
@abankhele Chris believes he has fixed the issues with streaming on mco-staging that we were seeing across browsers. Can you test this on mco-staging and comment here?

abankhele:
I have tested the issue, and it now works properly on Chrome and Safari. I can successfully stream the video.

elynema:
I am not able to replicate the message I was seeing: ""That stream was not recognized. Defaulting to the first available stream for the resource" today on avalon-dev or mco-staging when editing existing items or uploading new sections. 

I'm going to close this issue. We ca re-open a new issue if that problem pops back up.

PR #6216, opened 2025-03-03
Force mimetype to application/x-mpegurl when the hls_url is an m3u8

cjcolvar:
Fix for #6214 

As part of enabling MP3 streaming a check was added to IIIFCanvasPresenter for the m3u8 mimetype.  That mimetype is the default when a derivative doesn't have one set which I believe is true for all files transcoded using ffmpeg.  Older transcodes (possibly dating back to matterhorn) do have a mimetype and that causes issues in the front end when it comes through.  This PR forces the m3u8 mimetype when the HLS url being passed to the presenter has an m3u8 extension.


masaball (approved):
:+1:

PR #6229, opened 2025-03-11
Make finders case insensistive to match validators

cjcolvar:
I was investigating this honeybadger report (https://app.honeybadger.io/projects/54117/faults/103019048) and noticed that the username/email returned by the auth service was uppercase but a prior user account in Avalon had been made with the username/email lowercase.  This led to an error when using `find_or_create_by_username_or_email` because the user wasn't found but creating failed due to the case insensitive validations on username and email.  Making the special finders used by `find_or_create_by_username_or_email` to do case insensitive queries should fix this problem.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,8 +43,8 @@ class User < ActiveRecord::Base
   before_destroy :remove_bookmarks
 
   def username_email_uniqueness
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good

PR #6234, opened 2025-03-17
Speed up item show view

cjcolvar:
Part of #6214 

Changes proposed in this PR:
- Memoize `@media_object` to avoid multiple calls to `load_resource` causing additional calls to solr
- Pass `@media_object` to calls to `secure_streams` to avoid having to call out to solr for CDL check in `SecurityHelper`
- Use Solr instead of Fedora for fetching section in `set_active_file`


masaball (approved):
Looks good

PR #6236, opened 2025-03-19
Move clean up of expired StreamTokens into background job run by sidekiq cron

cjcolvar:
This work was originally done in an attempt to deal with a production slowdown due to the database being overloaded and table write locks being hard to acquire.  By moving the deletion out into a background job we reduce the number of database write calls to one every few minutes instead of one every request which could potentially be hundreds in the same time frame.


masaball (approved):
Looks good

Issue #5543, opened 2023-12-21
Avoid fetching section data when rendering item show view


cjcolvar:
With ramp handling most of the item view page and it's source of information is the IIIF manifest Avalon shouldn't need to fetch section level data in order to render the show view.  Currently it does which slows down page render a lot.  If we can get rid of all of these references then Avalon should only have to fetch the MediaObject and page render should be speedy.

- Ramp component uses lots of information in props: master file count, has_structuralMetadata, transcripts, supplemental files, etc.
- `_item_view.html.erb` needs current stream for oembed links in `<head>` 
  (I'm confused how this is supposed to work since the page head doesn't dynamically get updated when sections change and the oembed links are only for the active section.  Maybe this is correct since the oembed is machine read only?)
- Some other partials need `ordered_master_file_ids`, `@masterFiles`, and `@media_object.master_files`
- The Share panel is currently built from stream endpoint info
- Stream details for active section are provided on page render instead of fetched via the `/stream` endpoint like subsequent sections on section change (related to #5502)

# Done Looks Like
- [ ] Investigation into items and reducing extraneous requests
- [ ] New issues created as needed

elynema:
Ramp props are needed to know whether to render certain components, ex: whether to render the Transcripts tab. Not sure how we could re-work the conditional loading of these components pulling that info from the manifest. We would need to parse the manifest on Avalon side, not just on Ramp side because the components in Ramp are initiated by Avalon. Is there a way for Ramp to handle conditional display of components based on what is in the manifest? This is challenging as the instantiating app decides which components from the library to render and where.

elynema:
@elynema write up last point about stream request as a separate issue

joncameron:
Revisiting oEmbed would be good as a separate ticket, and a good starting point for that work (method calls in partials). Also: is oEmbed working the way we want it to?

elynema:
/stream endpoint is no longer called, but stream details are still being returned as a hash for page render.

Issue #5502, opened 2023-12-05
Share URLs in IIIF Manifest


joncameron:
# Description
Share URLs are currently being built within Avalon to populate the Share form fields on the item page. Could these instead come via the IIIF manifest for the item? Where would an appropriate location to put this be if so, and would it add too much extra data to the manifest?

joncameron:
- Manifest URL is already available in the manifest itself
- If adding a separate component for the Share actions and links, the manifest URL could live there as well

masaball:
Having looked around a bit at other IIIF implementations:
* UV does not have share information directly/explicitly in the manifest. As far as I can tell, UV dynamically generates the share links based on the host name and the ID field for each canvas. Because our manifest IDs are not actually resolvable URIs outside the manifest context, this will not work for us
* Clover does not seem to have an explicit share section, but includes `homepage` and `see also` in their "About" sidebar
* Mirador uses the `homepage` property to populate the share section

For purposes of greatest compatibility with other clients, I do not think it is the right choice to create a service. I conceptually like the cleanness of it but no other client would be able to parse it to provide the share information so it would only be usable by Ramp.

With everything I've looked at, it seems like the `homepage` property is our best bet to move forward with. [From the spec](https://iiif.io/api/presentation/3.0/#homepage) the homepage property is an array JSON objects, so we can have all the different share links we provide in this single place. These would be: manifest (this does not need to be in the homepage array, but might be good to have it there for completeness?), section url, item url, LTI share urls if that is enabled, and embed url. I think for the embed, we would need to generate the embed iframe dynamically on the Ramp side. Or we could maybe have the share links in `homepage` but have the embed iframe in a `metadata` block since that allows more arbitrary information?

As for security concerns, I don't think there would really be any with having these links in the manifest? These links are public anyways. If someone were to somehow get a manifest from an item that is locked down the links would be navigating to endpoints with our standard access checks in place so people should not be able to access things they shouldn't.

masaball:
@joncameron What are next steps on this ticket? Do you think another discussion with the findings from the investigation would be beneficial or do you have a decision on the implementation?

joncameron:
Could add homepage to a canvas, and one at the root, so we can have share links for objects and sections in appropriate scope.

See https://iiif.io/api/presentation/3.0/#331-external-links; we may need some extra handling for embeds. But just for share links, the `homepage` property should fit the bill.

[...]

PR #6240, opened 2025-03-24
Move share links into JS so it doesn't need the /stream endpoint

cjcolvar:
Part of #5543 and #5502

These changes remove the slow calls to `/stream` that happen on item view page load and section switches.  The share information is embedded in the page in the JS on page render which should be _much faster_ than calling `/stream`.


masaball (approved):
Looks good to me

Issue #6223, opened 2025-03-07
Re-request m3u8 file to refresh stream token


masaball:
## Description

In the discussion for #6076 it was decided that it would likely be easiest to refresh the stream tokens by re-requesting the m3u8 via calling `player.src(player.currentSources*())`. Once the video stream has been reset, we should be able to return the user to their current time by leveraging information from the state in Ramp.

Should this re-request be on an interval (re-request every ~5 minutes) or would it be better to try to detect that the player is failing to receive the next segment?

## Done looks like
- [ ] Investigate player events to determine if we can detect when the hls segments are failing from stream token expiration
- [ ] Implement an error catch or interval to reset the m3u8
- [ ] Ensure that the player is returned to its previous time/state after token refresh

joncameron:
I tested this in Chrome, Firefox and iOS Safari. I could no longer reproduce the issue with timeouts happening during long periods of pause.

PR #6242, opened 2025-03-24
Refresh stream token every 5 minutes when media is paused

masaball:
Related issue: #6223

### DIFF
--- a/app/assets/javascripts/player_listeners.js
+++ b/app/assets/javascripts/player_listeners.js
@@ -542,7 +542,7 @@ function handleCreateTimelineModalShow() {
 /**
  * Handler to refresh stream tokens via reloading the m3u8 file
  */
-function initM3U8Reload(player) {
+function initM3U8Reload(player, mediaObjectId, sectionIds, sectionShareInfos) {
   if (player && player.player != undefined) {
     if (firstLoad === true) {
       player.player.on('pause', () => {
@@ -561,6 +561,7 @@ function initM3U8Reload(player) {
       // Set player to user's active time after reloading m3u8
       player.player.on('loadeddata', () => {
         player.player.currentTime(currentTime);
+        buildActionButtons(player, mediaObjectId, sectionIds, sectionShareInfos);
       });
     }
   }
--- a/app/assets/javascripts/player_listeners.js
+++ b/app/assets/javascripts/player_listeners.js
@@ -22,6 +22,8 @@ let firstLoad = true;
 let streamId = '';
 let isMobile = false;
 let isPlaying = false;
+let reloadInterval = false;
+let currentTime;
 
 /**
  * Bind action buttons on the page with player events and re-populate details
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/ramp_utils.js
```diff
@@ -360,3 +360,9 @@ function closeAlert() {
     $('#playlistitem_scope_track').prop('checked', true);
   }
 }
+
+/** Refresh stream token by reloading active m3u8 */
+function m3u8Reload() {
+  player = document.getElementById('iiif-media-player');
+  player.player.src(player.player.currentSources());
+};
```
cjcolvar, line 368: For this scenario (user is still online), I think we could avoid resetting the player and instead just fetch the auto.m3u8:
```
fetch(player.player.currentSources()[0]["src"]);
```
For the other scenario (token already expired) I think we would need to set the player sources like you're doing.

cjcolvar (dismissed):
I like how this is looking!  This seems to cover the scenario where a user is still active on a page and the `m3u8Reload` method is isolated.  I'm hoping that it could possibly be reused to handle the scenario when a user still has a player page up but has been offline for a long time.  Did you find a player event that can be caught when a user attempts to play after a token has expired?

masaball:
I was not able to find a player event that would reliably and uniquely indicate that the hls segments were failing. That is why I went with an interval to refresh the token. 

I'm unsure what the differentiation between online/offline is that you are making? In my testing the reset works whether the token has already expired or not.

app/assets/javascripts/ramp_utils.js
masaball: In testing the fetch this morning, it looks like it does actually work for both cases. That would probably be better than fully resetting the player and looks like it removes the need for the `on(loadeddata, ...)` listener.

app/assets/javascripts/player_listeners.js
```diff
@@ -536,3 +538,30 @@ function handleCreateTimelineModalShow() {
...
+
+      player.player.on('play', () => {
+        if (reloadInterval !== false) {
+          clearInterval(reloadInterval);
+          reloadInterval = false;
+        }
+      });
+
+      player.player.on('waiting', () => {
+        m3u8Reload();
```
cjcolvar, line 563: Does this work with just fetch or does the player need the sources set again in this case?

cjcolvar (commented):
This looks okay to me.  I'd like @Dananji to take look over it too since she probably knows video.js the best of us.

app/assets/javascripts/player_listeners.js
masaball: It was working with just the fetch when I was messing with things this morning.

Dananji (approved):
This looks great!! 

I do have one question, does the player show a loading spinner or something while `m3u8Reload()` is fetching the source files?

If not and the fetch request takes a bit of time; we could maybe add the loading spinner manually on the `waiting` event and remove it on successful fetching of the sources? I think this would improve the user experience.

I don't think this is a blocker for this PR. We could probably re-visit this after this is merged and tested in `avalon-dev` to see if it's needed.

masaball:
In messing with everything locally there was a loading spinner when playback hit the end of the buffered content until the fetch finished processing. But we can see if the behavior is different or inconsistent on avalon-dev and adjust if needed :+1:

Issue #6181, opened 2025-01-31
Ensure proper siloing of docker environments


masaball:
## Description
A [recent change](https://github.com/avalonmediasystem/avalon/commit/66e8a69e373a24adfb5e0e77d088e8efa47e868f) to move the OCFL store into Minio broke the siloing of Fedora between the different environments. This results in the OCFL of the non-test environment being deleted whenever a spec test is run.

## Done Looks Like
- [ ] Review all of the components (S3 buckets, fedora, DB, Solr, etc.) that are being built by docker
- [ ] Ensure environments are properly de-coupled from each other.

joncameron:
Theoretically only a problem for devs: test suite is cleaning out the DB anytime tests are run.

PR #6243, opened 2025-03-24
Test docker environment shouldn't overwrite development environment when running tests

cjcolvar:
Completes #6181 

Make fedora-test service usable by test independent of the fedora service (networking and storage)
Test environment doesn't use minio or streaming so do not depend on them and separate test services do not need to be created for them.

Also included in this PR is a bump to the fedora image to handle a tomcat security upgrade.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,23 +25,19 @@ services:
     volumes: []
 
   fedora: &fedora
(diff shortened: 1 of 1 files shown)


masaball (approved):
:+1:

Issue #6226, opened 2025-03-10
Re-test and fix enabling media object action buttons in mobile browsers


masaball:
Creating this issue to reexamine the fix(es) put in place for #5885.

In smoke testing on mco-staging it was found that sometimes the share link is being set to the previous canvas, or flicking back and forth between the old and new share link, when switching sections. This is a consequence of [this block of code](https://github.com/avalonmediasystem/avalon/blob/1ab561e38327afa172fcb0f62fe8d6a2f33cf283/app/assets/javascripts/player_listeners.js#L74).

That code block was meant to ensure that the action buttons always get enabled and are populated with the correct information. However, at least when building the share button, it is encountering a race condition that sets the share links improperly. The code flow in those listeners needs to be re-examined so that all of the media object buttons are populated with the information for the current canvas. 

Additionally, the `if` statement is too restrictive. Currently only users with editor or higher permissions will have this code trigger because we are checking the status of all 3 action buttons (playlist, timeline, thumbnail). "Create Thumbnail" does not render for regular users so this code never triggers for them.

## Done Looks Like
- [x] Code to enable and populate buttons triggers for all users that the buttons render for
- [x] Media object page buttons are consistently enabled on all devices
- [x] Media object page button sub menus are properly populated on all devices

elynema:
Chris worked on some changes to eliminate need to reach out to /stream endpoint. Will this issue resolve with that PR?

elynema:
See #336 on local board for more details about this problem.

elynema:
Estimate this at a 3 as changes may impact all buttons on media object page, so needs testing.

cjcolvar:
I opened a new PR with just the removal of calls to `/stream`: #6240.  This can be retested once that is merged.

masaball:
From my testing this afternoon, #6240 appears to have fixed the issue with the share link flickering and reverting to the previously active session.

elynema:
@abankhele Can you verify that the playlist, timeline, and thumbnail buttons seem to be activated on iOS and Android mobile devices?

abankhele:
Tested on iOS and Android mobile devices, I can see the playlist, timelines, and thumbnail buttons are enabled.

joncameron:
👍

PR #6245, opened 2025-03-26
Attempt to re-enable buttons if any is disabled

masaball:
Related issue: #6226


cjcolvar (approved):
:+1:  Makes sense.  Good catch!

Issue #6246, opened 2025-03-26
Loading many-sectioned items is slow


elynema:
# Description

Testing in mco-staging with Fedora 6 (Avalon 8.0), loading of many-sectioned items seems particularly slow, especially after removing the linked list implementation that was part of 7.8 release.

- Example 100 section item: https://mco-staging.dlib.indiana.edu/media_objects/5138kd134 - load time 20-40 seconds to player ready for playback
- Example 194 section item: https://mco-staging.dlib.indiana.edu/media_objects/tt44qc078 - load time 55 seconds to player ready for playback
- Example 100 section item on MCO (for comparison): https://media.dlib.indiana.edu/media_objects/gq67kq31s - load time more like 10 seconds to player ready for playback
- 
Probably unsurprisingly, the majority of the time seems to be spent creating the manifest.json.

This should be tested as well between a standard user account and an admin account.

Performance testing data: https://docs.google.com/spreadsheets/d/1mi88YbutsW7dIHMRpNrpS5BnZOZuYjluHAjS7hdA2b8/edit?usp=sharing

## Done Looks Like
- [ ] Investigate why excess Solr calls are happening
- [ ] Investigation of other potential causes in the slow response times

joncameron:
Related: #6000

joncameron:
I'll test this on staging after https://github.com/avalonmediasystem/avalon/pull/6251 is merged.

joncameron:
I've created 100- and 200-section items on avalon-staging for testing:
- https://avalon-staging.dlib.indiana.edu/media_objects/1c18df77c
- https://avalon-staging.dlib.indiana.edu/media_objects/tm70mv20j

joncameron:
After the boolean clause setting got changed this is tested and working as expected, with better performance than seen previously.

PR #6249, opened 2025-03-28
Do not include legacy captions in IIIF manifests

cjcolvar:
Part of #6246 

Including any "legacy" captions in manifests was causing a separate solr call for each section whether it had legacy captions or not.  Removing them from manifests allows us to skip those solr calls keeping the number of solr calls around 6-7 no matter the number of sections in an item.  Legacy captions should have been migrated to supplemental file captions as part of upgrading to 7.7 and they are not shown in the manage files page nor anywhere else in the UI.

### DIFF
--- a/app/presenters/speedy_af/proxy/master_file.rb
+++ b/app/presenters/speedy_af/proxy/master_file.rb
@@ -67,6 +67,7 @@ class SpeedyAF::Proxy::MasterFile < SpeedyAF::Base
   end
 
   def captions
+    return nil unless has_captions?
     load_subresource_content(:captions) rescue nil
   end
 
--- a/spec/models/iiif_playlist_canvas_presenter_spec.rb
+++ b/spec/models/iiif_playlist_canvas_presenter_spec.rb
@@ -187,7 +187,7 @@ describe IiifPlaylistCanvasPresenter do
 
       it "includes paths to supplemental but NOT legacy caption files" do
         expect(subject.any? { |content| content.body_id =~ /supplemental_files\/#{caption_file.id}\/captions/ }).to eq true
-        expect(subject.any? { |content| content.body_id =~ /master_files\/#{master_file.id}\/captions/ }).not_to eq true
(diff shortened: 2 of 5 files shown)


app/presenters/speedy_af/proxy/master_file.rb
```diff
@@ -67,6 +67,7 @@ def supplemental_files(tag: '*')
   end
 
   def captions
+    return nil unless has_captions?
```
cjcolvar, line 70: `has_captions?` is true if the section has legacy captions OR supplemental file captions.  This guard clause will avoid solr calls when the section does not have any captions at all, but if the section has supplemental file captions only it will still attempt to find legacy captions in solr.  This is a compromise to avoid calling fedora but still avoid some solr calls.  Also, I don't believe this method is actually called anywhere.

spec/models/iiif_playlist_canvas_presenter_spec.rb
```diff
@@ -182,12 +182,12 @@
...
         expect(subject).to all be_a(IIIFManifest::V3::AnnotationContent)
-        expect(subject.length).to eq 2
+        expect(subject.length).to eq 1
       end
 
-      it "includes paths to supplemental and legacy caption files" do
+      it "includes paths to supplemental but NOT legacy caption files" do
         expect(subject.any? { |content| content.body_id =~ /supplemental_files\/#{caption_file.id}\/captions/ }).to eq true
-        expect(subject.any? { |content| content.body_id =~ /master_files\/#{master_file.id}\/captions/ }).to eq true
+        expect(subject.any? { |content| content.body_id =~ /master_files\/#{master_file.id}\/captions/ }).not_to eq true
```
masaball, line 190:
```suggestion
        expect(subject.any? { |content| content.body_id =~ /master_files\/#{master_file.id}\/captions/ }).to eq false
```

I don't really feel strongly one way or the other, but would it be better to switch the check to match on false instead of negating true? Whatever the choice, ln 219 should match for consistency.

masaball (approved):
Looks good. Just one non-blocking suggestion.

masaball (approved):

PR #6252, opened 2025-03-31
Data-testids added to mediaobject and item

abankhele:
Added data-testids for media-objects and item testcase.


spec/cypress/integration/media_object_spec.js
```diff
@@ -15,105 +15,118 @@
...
-  const caption = Cypress.env('MEDIA_OBJECT_CAPTION_2');
+  const media_object_id = Cypress.env('MEDIA_OBJECT_ID');
+  const media_object_title = Cypress.env('MEDIA_OBJECT_TITLE');
+  const caption = Cypress.env('MEDIA_OBJECT_CAPTION');
 
   beforeEach(() => {
     cy.login("administrator")
+    cy.intercept({
+      method: 'GET',
+      url: /\/media_objects\/.*\/section\/(?!undefined).*\/stream/
```
cjcolvar, line 26: We got rid of javascript calls to `/media_objects/<media_object_id>/section/<section_id>/stream` in #6240.  Maybe watching for the IIIF manifest call would be a good replacement?
```suggestion
      url: /\/media_objects\/.*\/section\/(?!undefined).*\/manifest.json/
```

spec/cypress/integration/playlist_spec.js
```diff
@@ -134,6 +154,10 @@ context('Playlists', () => {
     cy.login('administrator');
     cy.visit('/');
     cy.get('#playlists_nav').contains("Playlists").click();
+    //if in case the created playlist goes on next page
+    cy.get('#Playlists_filter').within(()=>{
+      cy.get('input[type="search"]').type(playlist_title_public);
+    });
```
cjcolvar, line 160: Nice!  This is more resilient this way.

spec/cypress/integration/item_spec.js
```diff
@@ -224,40 +309,54 @@ context('Item', () => {
...
+    cy.wait('@updateResourceDescription').then((interception) => {
+      expect(interception.response.statusCode).to.eq(302);
+    });
 
-    // Navigate to the preview page
-    cy.get('li.nav-item.nav-success').contains('a.nav-link', 'Preview').click();
+    
+    //Navigate to the item
+    
+    cy.intercept('GET', '**/stream').as('getmediaobject');
```
cjcolvar, line 338: Like the other pattern intercepting `/stream` this one should maybe be changed to `manifest.json` as well.
```suggestion
    cy.intercept('GET', '**/manifest.json').as('getmediaobject');
```

cjcolvar (commented):
This looks really good!  I really like how you keep improving the cypress tests to be cleaner and more resilient as you go!

spec/cypress/integration/media_object_spec.js
cjcolvar: The sequence of calls that happen before the player is ready to stream and the buttons enabled should be:

1. `/media_objects/<media_object_id>`
2. `/media_objects/<media_object_id>/manifest.json`
3. `/master_files/<section_id>/auto.m3u8`
4. Request for m3u8 from streaming server
5. Request for ts chunk from streaming server

I saw a 500ms to 1s gap between #3 completing and #5 completing but it could potentially be longer.  The url pattern for 4 and 5 will be different depending on the streaming server (avalon-dev is different from mco-staging) but it will always end with `<filename>.ts?token=<streaming_token>`.

```suggestion
      url: /\.ts\?token=\w+$/
```

cjcolvar (commented):
Looking good!  I had one suggestion and the only other thing left is resolving the merge conflicts  (let me know if you need any help with that).

docker-compose.yml
```diff
@@ -220,3 +220,4 @@ services:
       - npms:/e2e/node_modules
     environment:
       - CYPRESS_ENV=${CYPRESS_ENV:-local}
+      - CYPRESS_grep=${CYPRESS_grep:-@critical}
```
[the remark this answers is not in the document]
abankhele: If we make it all capital, I think grep plugin won't be detected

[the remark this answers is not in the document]
cjcolvar: I found an [old issue](https://github.com/cypress-io/cypress/issues/26880) about this which says `@cypress/grep` isn't getting updated anymore and look at using https://github.com/bahmutov/cy-grep.  For this PR I think what you have is fine for now but it would probably be good to switch to `cy-grep` in a later PR.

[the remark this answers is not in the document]
abankhele: Okay, with new pr I'll make the change to cy-grep.

cjcolvar (approved):
Looks great!

Issue #6221, opened 2025-03-05
Return checkouts when CDL is disabled for a Collection


elynema:
## Description

Currently, when a collection has CDL disabled, any items in that collection that were checked out remain on the list of Checkouts for individual users. If the user clicks on the item from the Checkouts page, they will be taken to the item page, but there will be no CDL info. Presumably, the item will be removed from the Checkouts page when it expires.

While this behavior may be acceptable since it still provides users access to the resource, it is confusing.

Would a better solution be to automatically return anything checked out in a collection when CDL is disabled? And perhaps the same across all collections if CDL is turned off at the application level?

## Done Looks Like
- [x] Add a background job to run after CDL is turned off for a collection to return all items

joncameron:
This would be a different thing for the case of CDL being disabled at the application level, rather than collection level. More discussion should happen for an application level change.

joncameron:
Worked as expected on avalon-dev.

PR #6261, opened 2025-04-11
Return checkouts when CDL is disabled for collection

masaball:
Related issue: #6221

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -347,13 +347,7 @@ class Admin::CollectionsController < ApplicationController
       collection.default_hidden = params[:hidden] == "1"
     end
     if params[:save_field] == "cdl"
-      case params[:cdl]
-      when "1"
-        collection.cdl_enabled = true
-      else
-        BulkActionJobs::ReturnCheckouts.perform_later(collection.id, nil)
-        collection.cdl_enabled = false
-      end
+      collection.cdl_enabled = params[:cdl] == "1"
     end
   end
 
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -347,7 +347,13 @@ class Admin::CollectionsController < ApplicationController
       collection.default_hidden = params[:hidden] == "1"
     end
     if params[:save_field] == "cdl"
-      collection.cdl_enabled = params[:cdl] == "1"
+      case params[:cdl]
+      when "1"
+        collection.cdl_enabled = true
+      else
+        BulkActionJobs::ReturnCheckouts.perform_later(collection.id, nil)
+        collection.cdl_enabled = false
+      end
     end
   end
 
--- a/app/jobs/bulk_action_jobs.rb
+++ b/app/jobs/bulk_action_jobs.rb
@@ -272,4 +272,14 @@ module BulkActionJobs
       [successes, errors]
     end
   end
+
+  class ReturnCheckouts < ActiveJob::Base
+    queue_as :bulk_return_checkouts
+    def perform(collection_id, _params)
+      collection = Admin::Collection.find(collection_id)
+      collection.media_object_ids.each do |id|
+        Checkout.active_for_media_object(id)&.update(return_time: DateTime.current)
+      end
+    end
+  end
 end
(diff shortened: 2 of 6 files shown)


app/jobs/bulk_action_jobs.rb
```diff
@@ -272,4 +272,14 @@ def perform(collection_id, overwrite = true, save_field = nil)
...
     end
   end
+
+  class ReturnCheckouts < ActiveJob::Base
+    queue_as :bulk_return_checkouts
+    def perform(collection_id, _params)
+      collection = Admin::Collection.find(collection_id)
+      collection.media_object_ids.each do |id|
+        Checkout.active_for_media_object(id)&.update(return_time: DateTime.current)
+      end
```
cjcolvar, line 282: I was thinking about how to make this more efficient and found a couple things:

`active_for_media_object` can take an array which will do one query like `WHERE checkout.media_object_id IN ['id1', 'id2', 'id3']`.  I'm not sure if we'll run into issues with collections with tons of media objects though.

`update_all` will do a single SQL update using the WHERE clauses setup in `active_for_media_object` instead of one update (and transaction) per checkout.  Things to note are that  `update_all` doesn't run validations and callbacks (which seems fine in this case) and it doesn't update the `updated_at` timestamp.  We can manually update `updated_at` as part of the `update_all` though and it will be close enough to the timestamp Rails would have given it.

```suggestion
      Checkout.active_for_media_object(collection.media_object_ids).update_all(return_time: DateTime.current, updated_at: DateTime.current)
```

app/controllers/admin/collections_controller.rb
```diff
@@ -347,7 +347,13 @@ def update_access_settings(collection, params)
       collection.default_hidden = params[:hidden] == "1"
     end
     if params[:save_field] == "cdl"
-      collection.cdl_enabled = params[:cdl] == "1"
+      case params[:cdl]
+      when "1"
+        collection.cdl_enabled = true
+      else
+        BulkActionJobs::ReturnCheckouts.perform_later(collection.id, nil)
```
cjcolvar, line 354: This is probably fine for now, but I'm wondering about hooking this in to a Admin::Collection callback (like how `around_save :reindex_members` works) instead of enqueuing it here.  This might have the benefit of working whether the collection gets CDL disabled via the UI, script, or rails console.  It could also move it later so that it only gets enqueued if the collection save is successful.  What do you think?

app/jobs/bulk_action_jobs.rb
```diff
@@ -272,4 +272,14 @@ def perform(collection_id, overwrite = true, save_field = nil)
       [successes, errors]
     end
   end
+
+  class ReturnCheckouts < ActiveJob::Base
+    queue_as :bulk_return_checkouts
```
cjcolvar, line 277: Do you think this needs a new queue?  I'm guessing it won't happen very often and thus would probably be fine on the default queue.  If you think it would be better being on its own then you need to add it to `config/sidekiq.yml` to let sidekiq know to run them.

cjcolvar (commented):
This seems very functional as is but I had a few suggestions/questions.

app/jobs/bulk_action_jobs.rb
masaball: I think it is fine as part of the default queue for now. I would expect it will be a fairly rare thing to trigger in real world use and we can always add a queue later if it becomes problematic for an implementer.

app/jobs/bulk_action_jobs.rb
```diff
@@ -272,4 +272,12 @@ def perform(collection_id, overwrite = true, save_field = nil)
...
```
cjcolvar, line 277: Removing this will have the job use the default queue.
```suggestion
```

app/jobs/bulk_action_jobs.rb
```diff
@@ -272,4 +272,12 @@ def perform(collection_id, overwrite = true, save_field = nil)
...
+    def perform(collection_id, _params)
```
cjcolvar, line 278: I don't think the `_params` parameter is needed here.  Other jobs in BulkActionJobs follow the pattern `perform(documents, params)` since they are getting queued from the BookmarksController and having them all have the same method signature made things simpler.  But this job is only ever going to be called on one collection and won't be used in the BookmarksController so I think removing the unused param makes sense.
```suggestion
    def perform(collection_id)
```

cjcolvar (commented):
This is looking good.  I like how the callback turned out!

cjcolvar (approved):

Issue #5989, opened 2024-08-14
Optimize # of Solr requests when adding item(s) to a playlist


elynema:
# Description

When adding multiple sections / tracks to a playlist at one time, it requires about a second per section / track. If adding 9 sections, that means 9 seconds to add all to a playlist. Looks like there are about 4-6 Solr requests per playlist item, so expense grows with number of sections.

Could we optimize this?

Example item: https://media.dlib.indiana.edu/media_objects/qn59qp839

This item has 6 tracks and takes 6-7 seconds to add all to playlist.

joncameron:
Using this 100-section example item: https://avalon-dev.dlib.indiana.edu/media_objects/12579s30p, adding all sections to a playlist takes less than a second. A huge improvement!

PR #6268, opened 2025-04-17
Use `load_reflections` when instantiating Speedy::AF records in `MediaObject#add_to_playlist`

masaball:
Related issue: #5989 

This commit also moves the `playlist.items +=` call out of the `each` block so that it is only triggered once.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -109,22 +109,23 @@ class MediaObjectsController < ApplicationController
 
   # POST /media_objects/avalon:1/add_to_playlist
   def add_to_playlist
-    @media_object = SpeedyAF::Proxy::MediaObject.find(params[:id])
+    @media_object = params[:post][:masterfile_id].present? ? SpeedyAF::Proxy::MediaObject.find(params[:id]) : SpeedyAF::Proxy::MediaObject.find(params[:id], load_reflections: true)
     authorize! :read, @media_object
     masterfile_id = params[:post][:masterfile_id]
     playlist_id = params[:post][:playlist_id]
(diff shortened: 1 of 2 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -109,22 +109,25 @@ def new
 
   # POST /media_objects/avalon:1/add_to_playlist
   def add_to_playlist
-    @media_object = SpeedyAF::Proxy::MediaObject.find(params[:id])
+    if params[:post].present?
+      @media_object = params[:post][:masterfile_id].present? ? SpeedyAF::Proxy::MediaObject.find(params[:id]) : SpeedyAF::Proxy::MediaObject.find(params[:id], load_reflections: true)
+    end
```
cjcolvar, line 114: This guard seems a little odd since the rest of the action expects params[:post] to be present.  This might be a good place to use strong parameters to require the params are present.  The failing test(s) in the controller specs would need to have the required params added to the call.
```suggestion
    add_to_playlist_params = params.require(:post).permit(:masterfile_id, :playlist_id, :playlistitem_scope)
    @media_object = add_to_playlist_params[:masterfile_id].present? ? SpeedyAF::Proxy::MediaObject.find(params[:id]) : SpeedyAF::Proxy::MediaObject.find(params[:id], load_reflections: true)
```
Adding strong parameters might be out of scope for this PR, but it feels like the test(s) should have the parameters added either way so they reflect expected usage.

cjcolvar (approved):
I have a suggestion but not sure if it is really needed now so feel free to merge as you see fit.

cjcolvar (approved):

cjcolvar (approved):

PR #6269, opened 2025-04-21
Refactor to reduce solr requests needed to render playlist edit page

cjcolvar:
While looking at #6268 I noticed the large number of solr requests being made to render the playlist edit page when there are many playlist items.  This PR shouldn't break anything and reduce the solr calls to 1 when rendering the form.

### DIFF
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -99,6 +99,14 @@ class Playlist < ActiveRecord::Base
     access_token == token && visibility == Playlist::PRIVATE_WITH_TOKEN
   end
 
+  def clips_with_section_proxies
+    cached_clips = clips.to_a
+    section_ids = cached_clips.collect(&:master_file_id).uniq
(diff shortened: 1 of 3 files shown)


app/models/playlist.rb
```diff
@@ -99,6 +99,14 @@ def valid_token?(token)
     access_token == token && visibility == Playlist::PRIVATE_WITH_TOKEN
   end
 
+  def clips_with_section_proxies
+    cached_clips = clips.to_a
+    section_ids = cached_clips.collect(&:master_file_id).uniq
+    sections = SpeedyAF::Proxy::MasterFile.where("id:#{section_ids.join(' id:')}")
+    cached_clips.map {|c| c.master_file = sections.find { |mf| mf.id == c.master_file_id }}
```
masaball, line 106: Should this be `map!` instead? Or does stuff get persisted anyway even with not doing the change in place?

masaball (approved):
Looks good as far as I can tell. Just have a question about the choice of method to be using.

Issue #6279, opened 2025-04-29
[BUG] v8.0 - Unable to download a MP3 from the Media Object file upload page


KyleCRat:
**Describe the bug**
mp3 downloads do not work. MediaObject.find(mo).download_path returns invalid path from the `https://github.com/avalonmediasystem/avalon/blob/main/lib/avalon/configuration.rb:49-58` file. I believe due to the regex not accounting for mp3 files, only m3u8

**To Reproduce**
- Upload MP3 to use the Direct MP3 media streaming feature available in 8.0
- navigate https://dml.test.lib.msu.edu/media_objects/MEDIA_OBJECT_ID/edit?step=file-upload
- Try to click "Download" on any MP3 media file

**Expected behavior**
The MP3 download would initiate

**Screenshots/Screen Recording**
<img width="1150" alt="Image" src="https://github.com/user-attachments/assets/070fa1bb-78d6-4aa1-a7a1-aadb52348b26" />

**Environment (please complete the following information):**
 - Device / platform Macbook Pro
 - OS: OSX
 - Browser [e.g. chrome version 90.0.4430.93]
 - Michigan State University Avalon Development

**Release:**
8.0

**Additional notes**
Lines 44-59 that have Regex's that need be able to work with MP3 urls
```
        # `Settings.streaming.server` returns symbols in testing environment
        # but strings in dev/prod. Explicitly call to_sym for consistency.
        location = case Settings.streaming.server.to_sym
                   # HLS Url templates can be found in config/url_handlers.yml
                   when :generic, :adobe
                     url.gsub(/(?:#{Regexp.escape(http_base)}\/)(?:audio-only\/)?(.*)(?:\.m3u8)/, '\1')
                   when :nginx
                     url.gsub(/(?:#{Regexp.escape(Settings.streaming.http_base)}\/)(.*)(?:\/index\.m3u8|)/, '\1')
                   when :wowza
                     # Wowza HLS urls include the extension between the base and relative path.
                     # "http_base/extension:path/filename.extension/playlist.m3u8"
                     # (?:.*?:) is a non-capturing group that will non-greedily match
                     # any character until the first colon. This removes the extension from
                     # the middle of the path.
                     url.gsub(/(?:#{Regexp.escape(Settings.streaming.http_base)}\/)(?:.*?:)(.*)(?:\/playlist.m3u8)/, '\1')
                   end
```

joncameron:
[...]

PR #6283, opened 2025-04-29
Provide download path for side-loaded mp3 derivatives with nginx streaming

cjcolvar:
Related to #6279 

This was overlooked when implementing the mp3 derivative streaming support.  There may be other places in the code that still assume HLS streaming and which might break when tried with side-loaded mp3s.


masaball (approved):
Looks good to me

joncameron (approved):

Issue #6259, opened 2025-04-09
Add "Alternative Title" Field to Item Metadata


joncameron:
## Description
Alternative title is a field in MODS that is being populated by the MARC 246 field in our default transformation for bibliographic import. Adding this field would allow this metadata to be retained and presented as part of the Avalon record. Currently this field is discarded from an item's MODS XML document on save, since there's no field in Avalon available for it. 

Ex in MCO: “no mas bebes” - https://media.dlib.indiana.edu/media_objects/zw12zp470
Catkey: 19270348

```
<mods xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-4.xsd">
<titleInfo usage="primary">
<title>No más bebés = No more babies</title>
</titleInfo>
<titleInfo type="alternative">
<title>No more babies</title>
</titleInfo>
```

## Done Looks Like
- [x] A new field added to Avalon's metadata for Alternative Title, with the following details
- [x] Appears in the Resource Description form

### Fields

- Name: Alternative Title
- MODS mapping: titleInfo@type="alternative"/title
- Repeatable
- Resource Description Form: “Alternative Title"
- Batch spreadsheet: “Alternative Title”

Alternative Title is used for display in search results and single item views. Only the first 32 characters of a title are included in search results listings. Recommended use is to reflect the content captured in digitized media files (such as the title of the piece performed or a short description of the content of a home movie).

joncameron:
Search will be something to look at—how does the inclusion of titles in different languages affect search function?

joncameron:
New ticket to be created for the question of the 880 field: Should one of the 880 fields be mapped to this field by default? What's the behavior in the marc2mods xslt?

joncameron:
👍

joncameron:
This was tested and is working in both.

PR #6302, opened 2025-05-12
Make alternative_title field user editable and viewable

masaball:
Related issue: #6259

### DIFF
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -344,6 +344,7 @@ class MediaObject < ActiveFedora::Base
       #Add all searchable fields to the all_text_timv field
       all_text_values = []
       all_text_values << solr_doc["title_tesi"]
+      all_text_values << solr_doc["alternative_title_ssim"]
       all_text_values << solr_doc["creator_ssim"]
       all_text_values << solr_doc["contributor_ssim"]
       all_text_values << solr_doc["unit_ssim"]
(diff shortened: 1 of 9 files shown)


masaball:
The failing test is passing when I run it locally, so I'm not sure what's going on.

app/models/media_object.rb
```diff
@@ -344,6 +344,7 @@ def to_solr(include_child_fields: false)
       #Add all searchable fields to the all_text_timv field
       all_text_values = []
       all_text_values << solr_doc["title_tesi"]
+      all_text_values << solr_doc["alternative_title_ssim"]
```
cjcolvar, line 347: This looks good.

Leaving a note here for later that this will require a reindexing in order to make the field available for searching.  This could be a limited reindexing...something like:
```
MediaObject.where('alternative_title_ssim:[* TO *]').each(&:update_index)
```

cjcolvar (commented):
When testing I noticed that alternative title didn't appear in search results if there were search terms (`q`).  Since `alternative_title_ssim` is a new blacklight index field it needs to be added to the search builder found in functionality otherwise it isn't included to be returned in search results.  I think changing the search builder to construct the `fl` from the list of blacklight index fields would resolve this issue.
```
fl = blacklight_config.index_fields.keys + ['id', 'has_model_ssim', 'avalon_resource_type_ssim', 'descMetadata_modified_dtsi', 'timestamp']
```

masaball:
Good catch on the search results. I thought I had checked that, but it must have gotten lost in the shuffle when I got distracted looking into what was happening with the helper methods.

cjcolvar (approved):
:+1: Looks good!

Issue #5056, opened 2023-02-07
Upgrade to Blacklight 8


joncameron:
# Description
Blacklight 8.4 has been released: https://github.com/projectblacklight/blacklight/wiki/Upgrading-to-Blacklight-8. Should Avalon upgrade Blacklight versions as well, and what would this entail?

This will be longer-term, and we'll need to ensure the Samvera components we use are compatible. If we want to do a Bootstrap upgrade, there will be lots of JavaScript configuration decisions to make. Now that Avalon is on Rails 7, is it time to make this move? Things might continue working as they are, but rethinking the way we do JS bundling might be needed.

There are deprecation warnings that pop currently when running Avalon; these will need to be resolved as part of this process. We'll need to either change things in Avalon or our dependencies so we're not using deprecated methods.

# Done Looks Like
- [x] Upgrade to Blacklight 8.10
- [x] Refactor Blacklight customizations to the new way of encapsulating components within the Ruby code

# QA 
Compare avalon-dev (Blacklight 8) with avalon-staging to verify functionality and identify major changes to UI / user experience
- [x] Search and browse functionality  
- [x] display of metadata
- [x] Display and functionality of breadcrumbs
- [x] sorting search results
- [x] paging search results
- [x] select items to add to selected items list / select all
- [x] selected items page - sorting
- [x] selected items page - paging
- [x] selected items page - clear selected items button
- [x] selected items page - select individual items / select all to take actions
- [x] selected items page -- all action buttons for selected items

masaball:
Note: Blacklight 8 does not have a release for Rails 7.2 yet. They are working on it, a beta version has been merged to main but it may be a minute.

masaball:
[Rails 7.2.0 has been added to Blacklight's test suite](https://github.com/projectblacklight/blacklight/pull/3167) and merged, so we should be able to start this upgrade with the next 8.x release.

masaball:
Blacklight 8.4.0 has been released which includes Rails 7.2 coverage, so we can move forward on this when we have cycles.

elynema:
Could break this up into a basic Blacklight upgrade + separate work to get our customizations refactored.

Blacklight 9 beta is out. Could consider doing more refactor when we move to 9.

masaball:
[...]

PR #6316, opened 2025-05-20
Upgrade to Blacklight 8

masaball:
Related issue: #5056


masaball:
Since we build the Media Objects page from IIIF rather than Blacklight results, should we remove the section from `catalog_controller.rb` that sets show fields? (`config.add_show_field ...`)

app/components/formless_document_action_component.rb
masaball, line 1: This file is for the publish/unpublish actions on the bookmarks page. They do not require generating a modal or other confirmation so we add `method: post` to the link creation so the action is automatically performed when the button is clicked.

app/components/index_header_media_object_component.html.erb
masaball, line 1: This adds our expected styling classes to the media object title field on the index page. It is possible that this is logic that could be passed in from the config or by overriding the partial that renders the component, but this seemed like the easier/more obvious approach.

app/components/index_metadata_media_object_component.html.erb
masaball, line 1: This partial adds the `found in: ` fields to the metadata display

app/components/index_metadata_media_object_component.rb
masaball, line 1: In order to have the `found in: ` information available to the partial, we need to make sure that the SolrDocument is passed through to the partial. This also adds our specific styling classes to the initialization of the component.

app/components/thumbnail_media_object_component.html.erb
masaball, line 1: This component adds the placeholder image(s) if a thumbnail is not found.

app/views/bookmarks/_tools.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render Blacklight::Document::ActionsComponent.new(document: nil,
+                                                      tag: 'div',
+                                                      classes: "#{controller_name}Tools",
+                                                      link_classes: 'btn btn-outline',
```
masaball, line 4: This partial override is specifically to change the classes passed into the initialization of the actions component to match prior Avalon behavior/rendering.

joncameron (dismissed):
👍

app/views/catalog/_document_media_object.atom.builder
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 xml.entry do
-  xml.title index_presenter(document).label(document_show_link_field(document))
+  xml.title document_presenter(document).heading
```
cjcolvar, line 4: I noticed that upstream they DRY this out by storing the presenter in a local variable:
`presenter = document_presenter(document)`

app/views/catalog/_document_media_object.atom.builder
```diff
@@ -19,9 +19,8 @@ xml.entry do
 
   with_format(:html) do
     xml.summary "type" => "html" do
-      xml.text! render_document_partial(document,
-      :index,
-      document_counter: document_counter)
+      document_component = blacklight_config.view_config(:atom).summary_component
+      render document_component.new(document: document_presenter(document), component: :div, show: true)
```
cjcolvar, line 23: Does this need `xml.text!` to write out the rendered content?

cjcolvar (commented):
This is looking so good!!  I wasn't expecting the view_components to come together so quickly.  I'm loving how clean the overrides are now!  I wonder if it'll be easier to write unit tests for the custom components now (which could be a future PR.)

app/views/catalog/_document_media_object.atom.builder
masaball: Yep. I assumed the html in the output was a improper data being returned because they'd changed things and so removed `xml.text!` because that had better formatted output. But looking at the upstream builder and output from MCO it is supposed to have the html in the output.

Having put the `xml.text!` back into place though, I'm seeing that there is a difference in the html being rendered out. So need to dig into that.

Issue #6298, opened 2025-05-08
Change Location Field Key/Label


joncameron:
## Description
The "Location" in Avalon actually pulling from 6xx$z, which is really Geographic Subject (see https://samvera.atlassian.net/wiki/spaces/AVALON/pages/1957954564/Metadata+Fields). In Avalon this is both "Location" and  "Geographic Subject". Changing the label would reduce semantic confusion and be more clear about the use of the field; for example, MODS also already has a "location" field, which is used for the location of the electronic resource, not geographic information. 

## Done Looks Like
- [ ] "Geographic Subject" used for this field's label instead of "Location" or "Location(s)"

PR #6317, opened 2025-05-20
Use "Geographic Subject" label instead of location

masaball:
Related issue: #6298


app/models/mods_behaviors.rb
```diff
@@ -55,7 +55,7 @@ def to_solr(solr_doc = Hash.new, opts = {})
     solr_doc['series_ssim'] = gather_terms(self.find_by_terms(:series))
     #filter formats based upon whitelist
     solr_doc['resource_type_ssim'] = (gather_terms(self.find_by_terms(:resource_type)) & ['moving image', 'sound recording' ])
-    solr_doc['location_ssim'] = gather_terms(self.find_by_terms(:geographic_subject))
+    # solr_doc['location_ssim'] = gather_terms(self.find_by_terms(:geographic_subject))
```
cjcolvar, line 58: :+1: I like how this is left commented out.  In code it can historically document that some solr documents may have that field.

cjcolvar (approved):
Looks good!

PR #6329, opened 2025-05-28
Added Cypress tests

abankhele:
Added more test cases
updated cypress grep
made it compatible with mco env


cjcolvar:
Superseded by #6340

Issue #6321, opened 2025-05-21
Update application to Rails 8 and test


joncameron:

masaball:
When working on updating our dependencies, I ran into redis-rails and active_elastic_job seem to not be compatible with Rails 8. The functionality of active_elastic_job is now covered by the official AWS SDK gem. Unsure how much code will need to be changed to switch over.

 What are we using redis-rails for? It doesn't seem like it is super maintained anymore. Has its functionality been brought into core rails or another gem we can switch to?

PR #6360, opened 2025-06-10
Upgrade to Rails 8

masaball:
Related issue: #6321


Gemfile
```diff
@@ -167,7 +167,7 @@ end
 
 # Install the bundle --with aws when running on Amazon Elastic Beanstalk
 group :aws, optional: true do
-  gem 'active_elastic_job'
+  # gem 'active_elastic_job'
```
masaball: Support for this protocol has been added to the official `aws-sdk-rails` gem and so `active_elastic_job` seems to have effectively been deprecated: https://github.com/active-elastic-job/active-elastic-job/pull/162. From the conversation on that PR, it seems like a seamless switch? Do we need to test anything specific to make sure that things are working as expected? @cjcolvar

Gemfile
```diff
@@ -93,7 +93,7 @@ gem 'marc'
 # Jobs
 gem 'activejob-traffic_control'
 gem 'activejob-uniqueness'
-gem 'redis-rails'
+# gem 'redis-rails'
```
masaball: This gem hasn't had a new release since 2017. According to the README most, if not all, of the gem's functionality became included in base Rails in version 5.2 and maintenance on this gem was sunset at that point. I see that this gem was primarily a cache store but what were we specifically using this gem for? What do we need to test to verify that it is still happening? @cjcolvar

app/presenters/collection_presenter.rb
```diff
@@ -56,7 +56,7 @@ def website_link
     view_context.link_to label, document["website_url_ssi"] if document["website_url_ssi"].present?
   end
 
-  def as_json(_)
+  def as_json(*)
```
masaball, line 59: Rails 8 was not happy with the underscore argument when the function was called without any arguments. Switching to the optional asterisk allows it to process whether arguments are passed to it or not.

bin/dev
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env ruby
+exec "./bin/rails", "server", *ARGV
```
masaball, line 2: This `bin/dev` script seems to be the new preferred way to boot an application. This was auto-generated by the update script but I have not really experimented with using it yet.

config/environments/production.rb
```diff
@@ -32,6 +31,9 @@
   # Do not fall back to assets pipeline if a precompiled asset is missed.
   config.assets.compile = true
 
+  # Cache assets for far-future expiry since they are all digest stamped.
+  # config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
```
masaball, line 35: This will probably be good to turn on once we have updated the asset pipeline, if we fully switch over to propshaft.

config/environments/production.rb
```diff
@@ -55,23 +57,28 @@
...
-  # config.ssl_options = { redirect: { exclude: ->(request) { request.path == "up" } } }
+  # config.ssl_options = { redirect: { exclude: ->(request) { request.path == "/up" } } }
 
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
-  # Prepend all log lines with the following tags.
+  # Log to STDOUT with the current request id as a default log tag.
   config.log_tags = [ :request_id ]
+  # config.logger   = ActiveSupport::TaggedLogging.logger(STDOUT)
```
masaball, line 67: We define the logger in a block elsewhere in this file, so we can comment out the default.

config/routes.rb
```diff
@@ -6,7 +6,7 @@
   concern :searchable, Blacklight::Routes::Searchable.new
   concern :exportable, Blacklight::Routes::Exportable.new
 
-  resource :catalog, only: [:index], as: 'catalog', path: '/catalog', controller: 'catalog' do
+  resource :catalog, only: [], as: 'catalog', path: '/catalog', controller: 'catalog' do
```
masaball, line 9: Rails 8 added an error when `only` and `except` have disallowed arguments, which was triggered here and for the persona users. Leaving `only` empty maintains prior behavior.

Gemfile
cjcolvar: Sounds like using https://github.com/aws/aws-activejob-sqs-ruby should work.  I think the things to make the change would be:
- Add `aws-activejob-sqs` to gemfile
- Set the queue_adapter to :sqs in settings.yml
- Remove `active_elastic_job` gem
- Remove `lib/active_job/queue_adapters/better_active_elastic_job_adapter.rb` and the line in `config/environments/production.rb` loading that file.

Gemfile
cjcolvar: Looks like we could get rid of the gem and set the `cache_store` to `:redis_cache_store` in `config/initializers/cache_store.rb` and possibly switch to a `url` parameter instead of the seperate `host`, `port`, `db`, `namespace`.
https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

Issue #6306, opened 2025-05-14
Add display helper method for Alternative Title


masaball:
## Description
The text tip for Alternative Title says that the field will be truncated to 32 characters when displayed on the index page. We will need to add a helper method so Blacklight renders that properly.

Additionally, since this is a multi-entry field how should entries be concatenated in the display? Separated by comma, semi-colon, new-line character, etc.?

## Done Looks Like
- [ ] Add helper method for Alternative Title display
- [ ] Limit to display of 3 Alternative Title field entries on the search results page (similar to other field)
- [ ] Separate entries by semicolon

PR #6363, opened 2025-06-11
Add display helper method for alternative_title

masaball:
Related issue: #6306

### DIFF
--- a/app/helpers/blacklight/local_blacklight_helper.rb
+++ b/app/helpers/blacklight/local_blacklight_helper.rb
@@ -17,6 +17,11 @@ module Blacklight::LocalBlacklightHelper
     ModsDocument::RIGHTS_STATEMENTS[arg]
   end
 
+  def alternative_title_index_display args
+    field = args[:document][args[:field]]
+    field.first(3).map { |f| truncate(f, length: 32) }.join(", ")
+  end
+
   def contributor_index_display args
     args[:document][args[:field]].first(3).join("; ")
   end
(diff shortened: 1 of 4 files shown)


app/helpers/blacklight/local_blacklight_helper.rb
```diff
@@ -17,6 +17,11 @@ def rights_statement_facet_display arg
     ModsDocument::RIGHTS_STATEMENTS[arg]
   end
 
+  def alternative_title_index_display args
+    field = args[:document][args[:field]]
+    field.first(3).map { |f| truncate(f, length: 32) }.join(", ")
```
cjcolvar: FWIW Multiple values for contributors are joined with a semicolon.

masaball: I did see that, but the ticket said to use a comma. @joncameron Should we switch to using a semi-colon here to match contributors? Or should we switch contributors to use a comma? Or we could just leave things as they are?

FWIW, I think the blacklight default for multi-value fields is to use commas.

joncameron: @masaball Let's switch to using a semicolon here to match contributors. We have contributor values with commas in the string and most likely we'll see the same for titles, so using a semicolon instead of the Blacklight default comma makes sense.

I updated to done criteria in the ticket as well.

cjcolvar:
I think the failing test here is due to the upgrade to ruby 3.4 and the formatting changes.

> Hash#inspect rendering have been changed. [[Bug #20433](https://bugs.ruby-lang.org/issues/20433)]
> Symbol keys are displayed using the modern symbol key syntax: "{user: 1}"
> Other keys now have spaces around =>: '{"user" => 1}', while previously they didn’t: '{"user"=>1}'

PR #6366, opened 2025-06-12
Search result titles include duration and fallback to object id when title isn't present

### LINKED ISSUE #6337: Blacklight 8: media objects without title do not display any title on the search results page
# Description

Media Objects without title do not display anything in the header. Prior displayed the id. Could be related to removal of document_show_link_field method with no replacement, but I would have expected the labelling behavior to have been maintained?

Will require some investigation to determine if there is a new configuration option for handling this case. If not, we may need to manually add the fallback title in one of our customizations.

Examples are available on the last page of search results in avalon-dev (when sorted by title).

# Done Looks Like

- [x] Media objects without title show the media object identifier as the title in Blacklight search results.

cjcolvar:
The commit restores the pre-BL8 title behavior by moving the search_result_label helper into the component and splitting some time formatting helpers into their own helper file to make a more minimal include into the component.  Also setup testing for components.

Resolves #6337

### DIFF
--- a/app/helpers/time_formatting_helper.rb
+++ b/app/helpers/time_formatting_helper.rb
@@ -1,0 +1,50 @@
+# Copyright 2011-2025, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+# 
+# You may obtain a copy of the License at
+#   
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+module TimeFormattingHelper
+  # We are converting FFprobe's duration output to milliseconds for
+  # uniformity with existing metadata and consequently leaving these
+  # conversion methods in place for now.
+  def milliseconds_to_formatted_time(milliseconds, include_fractions = true)
+    total_seconds = milliseconds / 1000
+    hours = total_seconds / (60 * 60)
+    minutes = (total_seconds / 60) % 60
+    seconds = total_seconds % 60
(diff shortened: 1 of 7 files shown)


app/helpers/time_formatting_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+module TimeFormattingHelper
+  # We are converting FFprobe's duration output to milliseconds for
+  # uniformity with existing metadata and consequently leaving these
+  # conversion methods in place for now.
+  def milliseconds_to_formatted_time(milliseconds, include_fractions = true)
```
masaball, line 19: I saw that this method is used a few times in the `media_objects_helper`. Do we need to throw an include in there or do helpers see each other without explicit inclusion?

spec/components/index_header_media_object_component_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+require 'rails_helper'
+
+RSpec.describe IndexHeaderMediaObjectComponent, type: :component do
```
masaball, line 17: This looks great. Good base to start building out other customization testing from.

app/helpers/time_formatting_helper.rb
cjcolvar: By default all helpers are included in all controllers so they should be available at the same time.
https://guides.rubyonrails.org/configuring.html#config-action-controller-include-all-helpers

masaball: Cool. I had a feeling that was the case but couldn't remember.

cjcolvar: It was good to ask.  I had to look it up and discovered the rails config setting.  So I learned something!

PR #6368, opened 2025-06-12
Fix Blacklight JSON responses

### LINKED ISSUE #6359: [BUG] Collection show page is not loading media object list
**Describe the bug**
When navigating to the show page for a specific collection the list of media objects is not loading.

**To Reproduce**
1. Go to the homepage
2. Click on "Collections" in the navigation bar
3. Click on any random collection
4. See the show page with perpetual loading spinner below the "Search within collection" field
5. Open web console
6. See "Error retrieving results" message

**Expected behavior**
The page should load with cards for each of the collection's child media objects

 - Laptop
 - Ubuntu
 - Firefox
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Additional notes**
Only present on Dev so likely something stemming from the Blacklight 8 upgrade. Error message in Rails logs: `NoMethodError (undefined method 'show_solr_document_url' for an instance of #<Class:0x00007ba70f172f00>)`. Code trace shows it failing on generating the document_url in the [index.json.jsbuilder partial](https://github.com/projectblacklight/blacklight/blob/de5ddbb4852abc1ab145e7e55afa15de903ddabe/app/views/catalog/index.json.jbuilder#L17)

cjcolvar:
Blacklight 8 changed where `url_for_document` is implemented so the call to `polymorphic_url` was returning a url helper name (`show_solr_document_url`) that didn't exist.  Fixing Avalon's override of `url_for_document` causes the correct url helper to be returned (`media_object_url`).

Resolves #6359

### DIFF
--- a/config/initializers/blacklight.rb
+++ b/config/initializers/blacklight.rb
@@ -1,7 +1,7 @@
 require 'blacklight'
 
 Rails.application.config.to_prepare do
-  module Blacklight::UrlHelperBehavior
+  Blacklight::SearchState.class_eval do
     def url_for_document doc, options = {}
       SpeedyAF::Base.for(doc.to_h.with_indifferent_access)
     end


masaball (approved):
Looks good!

PR #6369, opened 2025-06-12
Use new combined redis url setting

cjcolvar:
The `docker-compose.yml` file switched to use `SETTINGS__REDIS__URL` from the component ENV vars in #6360.  This PR continues the approach taken there and applies it to the other two initializers that set redis connections.  This resolves sidekiq being unable to connect to redis.  (I had to rebuild my worker container for this change to work but that may have been peculiar to my environment.)


masaball (approved):
Good catch. Looks good!

Issue #6307, opened 2025-05-15
Optimize use of new Blacklight 8 design patterns for customizations


elynema:
# Description

In #5056, Mason is doing the initial work to migrate Avalon's Blacklight customizations into new design patterns. This mostly consists of moving override code to a new location. However, there could be additional optimizations to simplify custom HTML files by pulling code out into separate files to improve readability and maintainability.

# Done Looks Like
- [ ] Investigate additional optimizations, removing unneeded code and refactoring as needed

joncameron:
Blacklight 8's move from partials to a component architecture is a change we should leverage and make sure we're doing efficiently.

PR #6371, opened 2025-06-12
Update catalog partial overrides to modern blacklight code

masaball:
Related issue: #6307


app/components/facet_field_pagination_component.rb
```diff
@@ -0,0 +1,6 @@
+class FacetFieldPaginationComponent < Blacklight::FacetFieldPaginationComponent
+  def initialize(**kwargs)
+    super
+    @span_button_classes = 'btn btn-primary'
```
cjcolvar: It looks like the only difference in the component markup is this variable used for css classes.  What do you think of overriding the default kwarg value for `button_classes` in the method signature and removing this line and the view partial?

app/components/index_header_media_object_component.rb
```diff
@@ -1,7 +1,7 @@
 class IndexHeaderMediaObjectComponent < Blacklight::DocumentTitleComponent
   include TimeFormattingHelper
 
-  def initialize(title = nil, document: nil, presenter: nil, as: :h3, counter: nil, classes: 'index_title document-title-heading col', link_to_document: true, document_component: nil, actions: true)
+  def initialize(**kwargs)
```
cjcolvar, line 4: Nice!

app/components/facet_field_pagination_component.rb
masaball: In order to match the original customization, a couple of the `helper.link` lines in the markup use the default `button_classes` so this felt like a clearer/cleaner approach than overriding `button_classes` to have all the necessary values and filtering in the markup.

Maybe it would be worth looking at the effect of removing the customization entirely? Or we could skip overriding the initialization and just hardcode in the markup override?

cjcolvar: How about we skip the customization entirely and deal with this in CSS?
It seems like the class changes are to deal with the style differences when a button is active or not.  There is the `active` class so I'm thinking that we could add a rule to style `.facet-pagination .btn-outline-secondary.active` and that should allow us to skip the custom component.
FWIW it looks like there are also [properties specifically for bootstrap](https://getbootstrap.com/docs/5.2/components/buttons/#variables) which could be set for `.facet-pagination .btn-outline-secondary` (or possibly `.btn-outline-secondary` as a whole) that are new in Bootstrap 5.2 which Blacklight appears to be using.

app/assets/stylesheets/blacklight.scss
```diff
@@ -33,6 +33,10 @@ img.no-icon {
   @extend .btn-outline;
 }
 
+.facet-pagination .btn-outline-secondary.active {
+  @extend .btn-primary;
```
cjcolvar, line 37: Nice!  I forgot about `@extend`.  This seems like a much cleaner way to preserve the old styling in the simplest way possible.

cjcolvar (approved):
Looks great!  It feels good to pay down technical debt and it seems like keeping up to date with Blacklight should be easier now.

Issue #6338, opened 2025-06-02
Blacklight 8: Select all checkbox errors


elynema:
# Description

- Select all checkbox on index page stays disabled when clicked after initial selection
- Select all checkbox on bookmarks page does not toggle selections as expected for first several clicks. Select all checkbox loses the check on first click, second click re-checks the box but unchecks all the selected items, third click loses the check but does not re-check the items, 4th click and up toggles everything as expected.

If reload the index page after clicking on Select All, it shows similar behavior to the select all checkbox on the bookmarks page.

Could be related to a change in how Blacklight core Javascript is supposed to be imported. The new import method didn't work when doing upgrade, so old import was put in place. 

# Done Looks Like

- [ ] Select All checkboxes work as expected on index page and bookmarks page  and can be used to select or unselect everything on the page with a single click.

elynema:
@abankhele Since this is likely a JS thing, it would be good to test on both Chrome and Safari, as well as on an iOS and Android device.

abankhele:
While testing I have observed this bug

On the Index Page(after a fresh reload):

- Click Select All → all checkboxes on page 1 become checked.
- Navigate to page 2, then return to page 1 → the “Select All” box remains checked.
- Uncheck any two individual items → those two items become unchecked, but the “Select All” box stays checked.
- Uncheck the “Select All” box → nothing happens (the remaining items stay checked).
- Check “Select All” again → only the two previously unchecked items become checked; the rest of the items toggle off, even though “Select All” is checked.

On the Bookmarks Page (after a fresh reload):

- 1. Click Select All → all bookmark checkboxes become checked.
- 2. Uncheck any one item → that item becomes unchecked, but the “Select All” box stays checked.
- 3. Uncheck the “Select All” box → no change occurs (the other items remain checked).
- 4. Check “Select All” again → the previously unchecked item becomes checked, and all the others toggle off, despite the box being checked.

elynema:
@abankhele Did you observe these in Chrome?

@Dananji It looks like there are some bugs to take a look at with this one still?

abankhele:
Yes this was on Chrome - macos

elynema:
@abankhele Dananji has made some updates - can you re-test this one?

elynema:
@abankhele Can you give this one a test on Android and iOS as well? I feel like sometimes JS is wonky in those environments.

abankhele:
Tested on Android and ios as well, working as expected.

PR #6374, opened 2025-06-16
Select all

Dananji:
Related issue: #6338


cjcolvar (approved):
Looks great to me!  So good to get rid of that coffeescript and really clean things up!

PR #6376, opened 2025-06-16
MediaObject#destroy should not fail when it can't find all sections

cjcolvar:
Sometimes sections get deleted but the parent media object doesn't get updated to remove the section's id from section_ids.  In this scenario avalon will currently fail to delete the remaining sections and the media object itself.  This PR adds a safe_load mode to #sections which only loads sections that are in both solr and fedora.  This mode is used to ensure that MediaObject#destroy will not fail when it is in an inconsistent state and some of its sections have already been deleted.

This PR comes from an issue raised in Avalon slack: https://samvera.slack.com/archives/C1C3C4F5L/p1747259886294209


masaball (approved):
Looks good to me.

Issue #6351, opened 2025-06-09
[A11y] Playlist index page 


Dananji:
### Description

From the accessibility checks performed on playlist related pages in https://github.com/avalonmediasystem/avalon/issues/6312, the following were identified for playlist index page.
| Success criterion | WCAG 2.1 Level | Accessibility checker | Error | Relevant HTML element |
| --- | --- | --- | --- | --- |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE| Form field missing a label | Tag filter drop-down menu |
| 1.4.3 Contrast (Minimum) | Level AA | SiteImprove & WAVE | Color contrast (2.85:1) doesn't meet min requirement | Disabled `Previous` button on the table pagination |
| 1.4.3 Contrast (Minimum) | Level AA | SiteImprove & WAVE | Color contrast (2.54:1) doesn't min requirement | Highlighted page number button on the table pagination |

### Done Looks Like
 - [ ] For the first issue: Since there is a label for this form element, we can probably use it to fix these issues by making the form label part of the drop-down making it accessible (Details of implementations: https://dequeuniversity.com/rules/axe/4.10/label?application=AxeFirefox)
 - [ ] Fix color contrast for table pagination elements

joncameron:
These are resolved and no longer flagged in Siteimprove.

PR #6378, opened 2025-06-16
Playlist index page color contrast and form element a11y issues

Dananji:
Related issue: #6351

### DIFF
--- a/app/views/playlists/index.html.erb
+++ b/app/views/playlists/index.html.erb
@@ -42,15 +42,13 @@ Unless required by applicable law or agreed to in writing, software distributed
 
     <% unless playlists.empty? %>
       <div class="tag_filter_container">
-        <label>
-          Filter:
-          <select name="tag_filter" class="tag_filter form-control input-sm" style="width:auto">
-            <option value="" selected="selected"></option>
-            <% current_user.playlist_tags.each do |tag| %>
-            <option value="<%= tag %>"><%= tag %></option>
-            <% end %>
-          </select>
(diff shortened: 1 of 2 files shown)


app/views/playlists/index.html.erb
```diff
@@ -42,12 +42,15 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-        </select>
+        <label>
+          Filter:
+          <select name="tag_filter" class="tag_filter form-control input-sm" style="width:auto">
+            <option value="" selected="selected"></option>
+            <% current_user.playlist_tags.each do |tag| %>
+            <option value="<%= tag %>"><%= tag %></option>
+            <% end %>
+          </select>
+        </label>
```
cjcolvar: Why does the label wrap the select element?  Would it be better for it to just wrap `Filter:`?

Dananji: Yes, `<label />` can only wrap `Filter:` text. Then, we would need to add an `id` attribute to `<select />` field to build a relationship between them.

And this syntax can be used to associate a label with a form field without introducing extra HTML attributes like `id`. I opted for this syntax, because it doesn't add more attributes to the `<select />` field.

Looking at the other forms in the Avalon code, I see that we are following the syntax with separated `<label />` and form fields. I will change this to follow the same syntax 👍

cjcolvar: :+1: That makes sense.  Thanks for explaining!

PR #6381, opened 2025-06-17
Fix unsupported ARIA by adding role attr. to share-tabs

Dananji:
Accessibility fixes around unsupported ARIA attributes in the share tabs in item view page.
Related: https://github.com/avalonmediasystem/avalon/pull/6380


cjcolvar (approved):
Looks good!

PR #6389, opened 2025-06-24
Work towards not fetching sections data on item show view

cjcolvar:
In this PR:
- Change ramp component prop to boolean and switch to using `section_ids`
- DRY up `supplemental_files(tags)` method and optimize it by using `locate_many` and memoization
- Use branch of speedy-af that adds allowlists for loading reflections
- Fix stream token bulk creation code (possibly difference in upsert return values between mysql and postgres)

These last two changes speed up many sectioned items considerably on mco-staging.  For a 244 section item the page request takes ~2s and the manifest ~8s (down from ~14s and ~30s on MCO).

I'm happy to split this out into separate PRs if any of the commits need more work/review.

Part of #5543

### DIFF
--- a/app/models/concerns/supplemental_file_read_behavior.rb
+++ b/app/models/concerns/supplemental_file_read_behavior.rb
@@ -29,11 +29,7 @@ module SupplementalFileReadBehavior
     # encountered in a live environment but came up in automated
     # testing. Adding a rescue on fail to locate allows us to skip
     # these out of sync files.
-    @supplemental_files ||= begin
-                              GlobalID::Locator.locate_many(JSON.parse(supplemental_files_json))
-                            rescue ActiveRecord::RecordNotFound
-                              nil
-                            end.compact
+    @supplemental_files ||= GlobalID::Locator.locate_many(JSON.parse(supplemental_files_json), ignore_missing: true)
     return [] if @supplemental_files.blank?
 
     case tag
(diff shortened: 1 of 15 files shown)


app/models/concerns/supplemental_file_read_behavior.rb
```diff
@@ -0,0 +1,48 @@
...
+    # If the supplemental_files_json becomes out of sync with the
+    # database after a delete, this check could fail. Have not
+    # encountered in a live environment but came up in automated
+    # testing. Adding a rescue on fail to locate allows us to skip
+    # these out of sync files.
+    @supplemental_files ||= begin
+                              GlobalID::Locator.locate_many(JSON.parse(supplemental_files_json))
+                            rescue ActiveRecord::RecordNotFound
+                              nil
+                            end.compact
```
masaball:
```suggestion
    @supplemental_files ||= GlobalID::Locator.locate_many(JSON.parse(supplemental_files_json), ignore_missing: true)
```

Looks like `GlobalID::Locator.locate_many` has an [`ignore_missing` option](https://github.com/rails/globalid/blob/e9548a3531a89abfaf036f88d4dac732355a5b80/lib/global_id/locator.rb#L56). Have you seen if that would work for us here?

cjcolvar: Neat!  I'll update this PR.

PR #6401, opened 2025-06-26
Fix select-all logic after page reload and individual selection changes

Dananji:
Related issue: #6338 

Changes:
- Override Blacklight's `bookmarked?` helper method to use `current_or_guest_user.bookmarks` to mark selected items on page reload
- Change the jQuery selector to check for `<input/>`'s `checked` attribute instead of `.checked` class in the `<label />` when filtering bookmarks for select-all operation

### DIFF
--- a/app/helpers/blacklight/local_blacklight_helper.rb
+++ b/app/helpers/blacklight/local_blacklight_helper.rb
@@ -59,6 +59,12 @@ module Blacklight::LocalBlacklightHelper
     result
   end
 
+  # Override of blacklight helper identify if the document is in the user's bookmarks
+  # https://github.com/projectblacklight/blacklight/blob/fe4b92fa4866f192c50f717cfa24bbf528d3c8ee/app/helpers/blacklight/document_helper_behavior.rb#L45
+  def bookmarked? document
+    current_or_guest_user.bookmarks.any? { |b| b.document_id == document.id && b.document_type.to_s == document.class.to_s }
+  end
+
   # Override of blacklight helper to add row class
   def render_document_class(document)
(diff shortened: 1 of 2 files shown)


app/helpers/blacklight/local_blacklight_helper.rb
```diff
@@ -59,6 +59,12 @@ def constraints_string(localized_params = params)
     result
   end
 
+  # Override of blacklight helper identify if the document is in the user's bookmarks
+  # https://github.com/projectblacklight/blacklight/blob/fe4b92fa4866f192c50f717cfa24bbf528d3c8ee/app/helpers/blacklight/document_helper_behavior.rb#L45
+  def bookmarked? document
+    current_or_guest_user.bookmarks.any? { |b| b.document_id == document.id && b.document_type.to_s == document.class.to_s }
```
cjcolvar: I don't fully understand why this change was necessary.  In the upstream Blacklight implementation it looks like the `current_bookmarks` method uses the `@document` or `@response.documents` instance variables and memoizes the result instead of doing a DB query to fetch all bookmarks of the user.  Can you please explain what was failing behavior you were seeing with the Blacklight implementation?

Dananji: When there are selected documents (bookmarked), the browse page was not displaying them as selected items on page reload. This was I what I was seeing in my local dev environment, but I don't see this in `avalon-dev`.
Maybe it's due a data inconsistency in my local dev environment? I changed `FEDORA_BASE_PATH` in `docker-compose.yml` after I had ingested some items with previous configuration. 
I just tested this after restoring the `docker-compose.yml` settings and it seem to be working without this code. I'll remove this and push the changes.

Issue #6356, opened 2025-06-09
[Keyboard a11y] Edit/Cancel playlist details buttons


Dananji:
### Description

From accessibility checks performed in https://github.com/avalonmediasystem/avalon/issues/6312, it was identified that the edit and cancel playlist details buttons are not reachable with `Tab` navigation in the edit playlist page.
To observe this behavior;
1. When edit form is closed: use `Tab` navigation to reach the edit playlist details button (next to `Playlist Details` text)
2. When edit form is open: use `Tab` navigation to reach the `Cancel` button at the end of the playlist details form to close the form

In Firefox, go to `Accessibility` tab in the dev-tools and check `Show Tabbing Order` to see the tab order of the elements on the page. Observe that both edit playlist details and cancel buttons are not in the tab order

### Done Looks Like
- [ ] A keyboard user can reach and activate the edit playlist details button
- [ ] A keyboard user can reach and activate the cancel changes made to the playlist details form
- [ ] These buttons are included in the page's tab order when visible

joncameron:
Working great! 👍

Issue #6354, opened 2025-06-09
[A11y] Playlist edit page


Dananji:
### Description

From the accessibility checks performed on playlist related pages in https://github.com/avalonmediasystem/avalon/issues/6312, the following were identified specifically for playlist edit page.
| Success criterion | WCAG 2.1 Level | Accessibility checker | Error | Relevant HTML element |
| --- | --- | --- | --- | --- |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE| Button missing a text alternative | Playlist item edit button with SVG |
| 1.4.3 Contrast | Level AA | SiteImprove & WAVE | Color contrast (3.95:1) doesn't meet min requirement | `No description` and `No tags` text in in playlist details |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE | Form field missing a label | `<input>` form field for playlist item position | 
| 3.3.2 Labels or Instructions | Level A | WAVE & axeDevtools | Ensure an element's role supports its ARIA attributes | Edit playlist details button and cancel edits button |

The input form related issues are handled in #6353

For the ARIA attributes issue, more information was provided by `axeDevtools`: https://dequeuniversity.com/rules/axe/4.10/aria-allowed-attr?application=AxeFirefox

### Done Looks Like
 - [ ] Add a screen reader accessible label to the playlist details edit button
 - [ ] Fix color contrast for greyed out text in the details panel
 - [ ] Add labels for the input form fields for playlist item positions
 - [ ] Both playlist item details edit button and cancel buttons have ARIA attributes supported by the element's role

PR #6372, opened 2025-06-13
fix(a11y): improve accessibility for playlist edit page elements

priyankc:
Addresses the following accessibility issues:

### 4.1.2 Name, Role, Value (Level A):
- Added aria-label to playlist item edit buttons.
- Updated Cancel span to a semantic button element with appropriate attributes.
- Labeled input fields for playlist item positions using aria-label.

### 1.4.3 Contrast (Level AA):
- Updated greyed-out text color to meet WCAG minimum contrast ratio (now #6c757d).
- Labels or Instructions (Level A):
- Replaced <a> tag with <button> for playlist details edit button and added screen reader label.

These changes enhance compatibility with screen readers and automated accessibility testing tools like WAVE, axeDevtools, and SiteImprove.

Fixes #6354

PR #6402, opened 2025-06-26
Fix a11y issues in playlist edit page

Dananji:
Related issue: #6354 and #6356


Dananji:
This PR incorporates changes from #6372, which addresses the same issue. I previously requested changes on #6372, but the author has not yet responded. During my subsequent review, I observed that more UI elements, specifically individual playlist item edit forms, still exhibit issues (which I had missed in the initial investigation). This PR includes additional fixes beyond those in #6372 to resolve these outstanding concerns and addresses the previously requested modifications.

@cjcolvar, what's the recommended approach for integrating these changes? Given the unresponsiveness on #6372, would it be preferable to merge this PR directly, or should we still aim to merge #6372 first and then apply the supplementary changes from this PR?

cjcolvar:
I think it would be best to merge their PR then open a PR with any additional changes /  improvements.

Issue #6403, opened 2025-06-26
[BUG] Exception Occurs When Parsing SRU query


aguilarm-umd:
**Describe the bug**
This issue is related to the workflow when retrieving metadata from Alma. 
When doing batch imports containing catalog numbers in the metadata csv there is a parsing issue with calling the escape method within sru.rb

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Setup an avalon instance locally using the docker-compose file
2. Update the settings.yml with the following settings within bib_retreiver
```
bib_retriever:
  default:
    protocol: sru
    url: https://usmai-umcp.alma.exlibrisgroup.com/view/sru/01USMAI_UMCP
    query: rec.id=%{bib_id} # I also found the same issue using rec.id='%s'
    retriever_class: 'Avalon::BibRetriever::SRU'
    retriever_class_require: 'avalon/bib_retriever/sru'
```
3. Edit the commands for the avalon and worker containers, to run `bash` and `dumb-init -- bash` respectively
4. Set a breakpoint at the start of the url_for method in sru.rb 
5. Exec into each of the containers, and run their respective commands
6. Create a collection in Avalon, and then copy into the directory a folder containing this csv file, with arbitrary video files that match the file names in the csv (All that matters is that the sru.rb file gets called)
7. Verify with the ruby debugger when stepping through the program that this exception appears
`#<Addressable::URI::InvalidURIError: Invalid scheme format: 'rec.id='{'>`

**Screenshots/Screen Recording**
I’ll include  a screenshot of the error I got when I was debugging

The batch import would succeed when I tried, but the items would
not have the metadata filled in for them as I was expecting

**Environment (please complete the following information):**
 - Device / platform: Apple M1 Pro
 - OS: macOS Sonoma 14.7.3 
 - Browser: chrome version 137.0.7151.120 

**Release:**
Was seeing this issue with 7.5.1, but the issue within sru.rb is still there within the latest versions of Avalon, so the issue probably still remains in later versions

**Additional notes**
I have a fix for this, so I can go add it to this issue, I wasn’t sure if I needed to create this issue this first or not

<img width="1800" alt="Image" src="https://github.com/user-attachments/assets/1b1facd1-f5bc-4fb2-9e77-818f9e79c90e" />

[wapava-test-metadata.csv](https://github.com/user-attachments/files/20934865/wapava-test-metadata.csv)

aguilarm-umd:
[...]

PR #6404, opened 2025-06-26
Fix Addressable::URI::InvalidURIError in sru.rb

aguilarm-umd:
This fixed the issue I had with not being able to
import metadata from Alma, the batch import would
run this, fail, but the import would still suceed, just without the metadata pulled from Alma.


aguilarm-umd:
Here's the related issue I created for this: https://github.com/avalonmediasystem/avalon/issues/6403

PR #6411, opened 2025-07-02
Allow configuring streaming auth callback referer header

cjcolvar:
When setting up Wowza in a local dev environment I needed to pass a referer other than rails root in order for wowza to contact avalon.  This could also be useful for other infrastructure setups where the streaming server could make auth calls to avalon on an internal address instead of the public address.

### DIFF
--- a/lib/avalon/m3u8_reader.rb
+++ b/lib/avalon/m3u8_reader.rb
@@ -24,7 +24,7 @@ module Avalon
         new(io.read, recursive: recursive)
       elsif io.is_a?(String)
         if io =~ /^https?:/
-          URI.open(io, "Referer" => Rails.application.routes.url_helpers.root_url) { |resp| new(resp, Addressable::URI.parse(io), recursive: recursive) }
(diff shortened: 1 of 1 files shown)


lib/avalon/m3u8_reader.rb
```diff
@@ -24,7 +24,7 @@ def self.read(io, recursive: true)
         new(io.read, recursive: recursive)
       elsif io.is_a?(String)
         if io =~ /^https?:/
-          URI.open(io, "Referer" => Rails.application.routes.url_helpers.root_url) { |resp| new(resp, Addressable::URI.parse(io), recursive: recursive) }
+          URI.open(io, "Referer" => Settings.streaming.auth_referer || Rails.application.routes.url_helpers.root_url) { |resp| new(resp, Addressable::URI.parse(io), recursive: recursive) }
```
masaball, line 27: Should this param be added (maybe commented out?) to the `settings.yml` file so that implementers know it is available without having to view the external docs?

masaball (approved):
Looks good. Small question on documenting the setting, but feel free to merge if you think it is already clear enough.

lib/avalon/m3u8_reader.rb
cjcolvar: @masaball I added this.  Could you please re-review?

Issue #6430, opened 2025-07-14
[BUG] Sorting by date on results page leads to error


elynema:
**Describe the bug**
Attempting to sort by date produces a Blacklight error.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to avalon-dev.
2. Click on Browse.
3. Select date from sort drop-down field.

**Expected behavior**
Results should be sorted by date.

**Screenshots/Screen Recording**
<img width="1364" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemkxQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--64e4eba0147a0c9401f3dafc27eac5b0e8a4ceac/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: both Android and desktop
 - OS: Android / macOS
 - Browser: Chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
This is working on avalon-staging, so seems to be associated with the 8.1 release.

elynema:
Does the Date sort by Creation date or Published Date?

masaball:
Date is sorted by publication date (`date_issued_ssi` in solr). This looks to be a conflict with the view helper I made for the human readable date display...unsure what the exact cause might be.

PR #6437, opened 2025-07-15
Fix date helper to handle "unknown/unknown"

masaball:
Related issue: #6430

In the original implementation of the view helper for human readable dates an assumption was made about "unknown/unknown" being edtf compliant. This was not the case, so the edtf-humanize gem errored in an environment containing that value. Adding a guard clause to handle that case takes care of the issue.


cjcolvar (commented):
This change seems fine but I think it would be good to have a fallback catch for if there are any other values that edtf-humanize might fail to parse.  (Maybe this should result in a `nil`?)  Adding a few quick tests for this method would probably be good to ensure we're getting the values we're hoping for.

Issue #6436, opened 2025-07-14
[BUG] Routing error when attempting to add or remove supplemental files


abankhele:
**Describe the bug**
Observing a routing error when attempting to add or remove supplemental files from the Manage Files section of a media object.

**To Reproduce**
Steps to reproduce the behavior, including the results:

1. Navigate to: https://avalon-dev.dlib.indiana.edu/media_objects/6d56zw687/edit?step=file-upload
2. Attempt to upload a new supplemental file and observe a routing error.
3. For an existing item with a supplemental file, attempt to delete the file observe that the routing error occurs.

**Expected behavior**
Uploading and removing supplemental files should function correctly.

**Screenshots/Screen Recording**
<img width="1459" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelcyQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e199b802be99a523d67b86b3c32f8d674e791631/image.png" alt="image.png" />
<img width="1462" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemUyQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d125ccc0b2d0e3c74d57b785a6581c48af5995d2/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser: Chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

elynema:
This is not happening on avalon-staging, so it is an 8.1 regression.

masaball:
@elynema @abankhele Doing a quick test locally this seems to only be affecting Supplemental Files put on the media object itself. Captions, transcripts, and generic files at the section level are uploading fine. Does that match the behavior you are seeing, or are you seeing issues with section supplemental files as well?

abankhele:
Observed only on item supplemental file. Uploading and deleting on section files is working fine.

joncameron:
Tested in dev and this error no longer occurs adding or removing media object-level supplemental files.

PR #6438, opened 2025-07-15
Fix supplemental file path helpers

masaball:
Related issue: #6436

A lack of parantheses in the conditional statements determining object type for supplemental file path generation caused valid SpeedyAF::Proxy::MediaObject objects to fail the check and fallback to nil. Added in the missing parantheses and added additional tests to hopefully catch any future issues.


cjcolvar (approved):
Good catch!  Nice job updating tests too!

Issue #6348, opened 2025-06-06
[BUG] Searching for “:” in the global search bar causes a Blacklight::Exceptions::InvalidRequest error


abankhele:
**Describe the bug**
When you enter a colon (“:”) into the global search box on the Avalon Development site, the application raises a Blacklight::Exceptions::InvalidRequest error. On Avalon Staging, the same input simply returns “No results found” as expected.

**To Reproduce**
Steps to reproduce the behavior, including the results:

1. Navigate to: https://avalon-dev.dlib.indiana.edu/
2. In the global search field, type : (a single colon) and press Enter or click Search.
3. Observe that the page displays a Blacklight::Exceptions::InvalidRequest in CatalogController#index error instead of showing “No results found.”

**Expected behavior**
Entering a colon (“:”) in the search box should not trigger an exception. Instead, the page should display a message such as “No results found for your search,” matching the behavior on Avalon Staging.

**Screenshots/Screen Recording**
Avalon-dev
<img width="1329" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3BMQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4d41a3ed1437c0550579f1c15527960557a8eb87/image.png" alt="image.png" />
Avalon-staging
<img width="1335" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzlMQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--33ca49a46ab1deefdd203ed664b50800590c9610/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform : laptop
 - OS: macos
 - Browser : chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

masaball:
This issue is tied to our "found in" queries. It did not manifest locally until I added a transcript to a document.

It looks like it may be possible to sub `:` for `\:` in the search builder, escaping the character for Solr. However, I've encountered some weird behavior with the "found in" search and so am unable to tell whether or not adding the escape character breaks other parts of the search.

joncameron:
Resolved when making a catalog search request for `:`.

PR #6447, opened 2025-07-17
Escape solr special characters when performing section and transcript searches

masaball:
Related issue: #6348

This PR also includes some formatting fixes for the affected files.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -50,9 +50,12 @@ class SearchBuilder < Blacklight::SearchBuilder
   def search_section_transcripts(solr_parameters)
     return unless solr_parameters[:q].present? && SupplementalFile.with_tag('transcript').any? && !(blacklight_params[:controller] == 'bookmarks')
 
-    return if solr_parameters[:q].match?(/[\{\}]/)
-    # In order for the multi-word query to work we need to NOT escape the query AND replace the spaces as +; this is also true for quoted phrase searches
-    transcript_subquery = "transcript_tsim:#{solr_parameters[:q].gsub(/ /, '+')}"
+    # In order for the multi-word query to work we need to NOT RSolr.solr_escape the query AND replace the spaces as +; this is also true for quoted phrase searches
+    # We can manually escape solr special characters that cause issues.
+    query = solr_parameters[:q].gsub(/([(){}\[\]\^\*?:])/, "\\\\\1").tr(' ', '+')
+    # Quotes need special handling to allow phrase searches.
+    query = query.gsub(/"/, "\\\\\1") if query.count('"').odd?
+    transcript_subquery = "transcript_tsim:#{query}"
     solr_parameters[:defType] = "lucene"
     solr_parameters[:q] = "({!edismax v=\"#{RSolr.solr_escape(solr_parameters[:q])}\"}) {!join to=id from=isPartOf_ssim}{!join to=id from=isPartOf_ssim}#{transcript_subquery}"
   end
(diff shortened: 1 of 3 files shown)


app/models/search_builder.rb
```diff
@@ -75,9 +78,9 @@ def term_frequency_counts(solr_parameters)
     solr_parameters["sections.rows"] = 1_000_000
     sections_fl = ['id']
     transcripts_fl = ['id'] if transcripts_present
-   
-    # Add fields for each term in the query 
-    terms = solr_parameters[:q].split
+
+    # Add fields for each term in the query, explictly escape closing parenthesis to prevent error
+    terms = solr_parameters[:q].gsub(/(\))/, "\\\\\1").split
```
cjcolvar, line 81: Weird!  I wonder what makes closing parenthesis special that it needs this extra handling.  I tried out some additional cases and they all seemed to fail on closing parenthesis whether they were matched or unmatched.  Oddly open parenthesis don't seem to have a problem.
```
(example)
( Example )
Example ) captions
((Example))
(() Lisp is (great)())
```

app/models/search_builder.rb
```diff
@@ -32,27 +32,30 @@ def only_published_items(_permission_types = discovery_permissions, _ability = c
...
     return unless solr_parameters[:q].present? && SupplementalFile.with_tag('transcript').any? && !(blacklight_params[:controller] == 'bookmarks')
 
-    return if solr_parameters[:q].match?(/[\{\}]/)
-    # In order for the multi-word query to work we need to NOT escape the query AND replace the spaces as +; this is also true for quoted phrase searches
-    transcript_subquery = "transcript_tsim:#{solr_parameters[:q].gsub(/ /, '+')}"
+    # In order for the multi-word query to work we need to NOT RSolr.solr_escape the query AND replace the spaces as +; this is also true for quoted phrase searches
+    # We can manually escape solr special characters that cause issues.
+    query = solr_parameters[:q].gsub(/([(){}\[\]\^\*?:])/, "\\\\\1").tr(' ', '+')
+    # Quotes need special handling to allow phrase searches.
+    query = query.gsub(/"/, "\\\\\1") if query.count('"').odd?
```
cjcolvar: This appears to escape all quotes if there is an odd number of them.  So a stray unmatched quote at the end invalidates all quotes from making phrase searches?  Would it be better to escape just the last one or it that assuming too much about the query?

cjcolvar: I did some manual testing and I'm running into issues with certain queries all of which have matched quotes so shouldn't be getting escaped in this line. (Last example is from MCO and works there.)
```
test "item"
"test item" test
"test item
" 1997 TV AD Campaign", 30 spots, January 
```

cjcolvar (commented):
Looks like this is an improvement and close!

app/models/search_builder.rb
masaball: Just to clarify, is there additional work that needs to be done on the escaping here or were you just testing the original behavior?

And yeah, it is very weird that it is just specifically closing parenthesis.

app/models/search_builder.rb
masaball: Hmmm...yeah, that is a good point about invalidating the phrase searches.

For the case of odd number of quotation marks, what do you think about not trying to escape them at all? We can catch the error and display a message about "All quotes in search must be matched".

As for your examples of stuff still failing, the first 2 and the last one look like they are giving a different error. The character issues return a "Cannot parse" error, whereas those examples are returning a "no field name specified in query and no default specified via 'df' param" error. I'm guessing that something about the combo of phrase search and multiword search is confusing solr. Unsure on how best to handle that...

The third one that starts with a single quote looks like it is returning a parsing error as well, so would probably be covered by us catching the error, if we go that route.

cjcolvar (approved):
Looks great!

PR #6448, opened 2025-07-17
Supplemental file links in ramp render inline and not force download

### LINKED ISSUE #6399: Support opening supplemental files in-line instead of downloading
## Description

The supplemental files that are linked on the Files tab are always downloaded and never open in a new browser tab. This is also currently also happening when we put links in a .vtt being display in the Transcripts tab.

Is there a change to the configuration of the endpoint provided in the IIIF manifest for supplemental files that would allow them to be opened in a new tab instead of downloading?

Also being done in AVAnnotate; supplemental files are available in MCO; external users can access this document. It always goes to download; how can we prevent this from happening? We set the content-disposition in the controller when handling requests for supplemental files, and this is where the logic will likely go to change this behavior.

## Done Looks Like
- [ ]  Requests for supplemental files are handled by the browser rather than content-disposition or other application-governed behavior

## QA
See this example record for QA: https://avalon-dev.dlib.indiana.edu/media_objects/12579s29x. All these links should open the HTML in a new tab, instead of forcing download.
- [ ] Full collection details link as Related URL in resource description
- [ ] Full collection details link in .vtt file after each cue
- [ ] Full collection link on Files tab

cjcolvar:
Resolves #6399 

Change content disposition header from `attachment` (which forces download) to `inline` (which allows browser to display or download if filetype isn't supported).  This required a change in ramp to skip special handling of links because it was creating an anchor tag and setting the `download` attribute which if present browsers will download instead of respecting the inline content disposition.

### Testing guidance

1. Upload supplemental files of different types to both item and section
1. Test that downloads work correctly with correct filename on item view page
1. Upload captions and transcripts (both machine generated and not)
1. Test that captions and transcripts work correctly in ramp
1. Test that transcript download works correctly and renames file if machine generated

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -85,11 +85,11 @@ class SupplementalFilesController < ApplicationController
       format.html { 
         # Redirect or proxy the content
         if Settings.supplemental_files.proxy
-          send_data @supplemental_file.file.download, filename: @supplemental_file.file.filename.to_s, type: @supplemental_file.file.content_type, disposition: 'attachment'
+          send_data @supplemental_file.file.download, filename: @supplemental_file.download_filename, type: @supplemental_file.file.content_type, disposition: "inline; filename=#{@supplemental_file.download_filename}"
         else
           # Rails 7.0 adds a config option to protect against "open redirects". We override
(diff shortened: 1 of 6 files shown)


masaball (approved):
Looks good to me

PR #6449, opened 2025-07-17
Add display_title to media object section_labels so it will be searchable if only filename

### LINKED ISSUE #6440: [BUG] Searches are not matching on some section titles
**Describe the bug**
When performing a search with a term that should match a section title the media object is not being returned. If the term matches in the metadata or transcripts, the search page does not list any matches in "sections".

**To Reproduce**
1. Go to Avalon-dev
2. Click on the search bar
3. Enter "android-chrome.mp4"
4. Click the search button
5. See that no results are returned
6. Go to https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p
7. See that section 6 is titled "android-chrome.mp4"

**Expected behavior**
Media objects with section titles that match the search query are returned on the search page and have "found in: sections" count.

 - Laptop
 - Ubuntu 24.04
 - Firefox
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Additional notes**
This appears to be from [adding `section.title`](https://github.com/avalonmediasystem/avalon/blob/ba9668892ec7da05cad58aeb42074ed92a10d389/app/models/media_object.rb#L278-L281) when generating the section labels index field. For user facing pages, we use `section.display_title` which falls back to the section filename when `section.title` is not present. This discrepancy is confusing when a user sees an apparent section label and tries to search it later but gets no results.

cjcolvar:
Resolves #6440 

This change would fix newly created items but a partial reindex would be needed to fix items that already exist with this issue.

### DIFF
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -287,7 +287,9 @@ class MediaObject < ActiveFedora::Base
   end
 
   def section_labels
-    all_labels = sections.collect{ |section| section.structural_metadata_labels << section.title}
+    # Need to add both title and display_title to avoid a regression because it is possible to have a title different than
+    # the section's structrualMetadata.section_title which is preferred in display_title
+    all_labels = sections.collect { |section| (section.structural_metadata_labels << section.title << section.display_title).uniq }
     all_labels.flatten.uniq.compact
   end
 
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good!

Issue #6413, opened 2025-07-07
[BUG] Tooltips in update access control modal on selected items page are not working


masaball:
**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to "Browse" page and check one or more of the "Select" boxes
2. Click on "Selected Items" in the Navbar
3. Click on "Update Access Control" button
4. Click on the question mark next to any label under the "Special Access" heading
5. Observe that nothing happens

**Expected behavior**
When the question mark icon is clicked, tooltip text should render next to/below the label explaining the purpose of the field.

**Environment (please complete the following information):**
 - Laptop
 - OS: Ubuntu 24.04
 - Firefox 139.0
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Additional Notes**
If you check the page html in dev tools, the tooltip text is present on the page. It is just not being unhidden when attempting to toggle it.

PR #6453, opened 2025-07-18
Use asset file for Blacklight modal script

masaball:
Related issue: #6413

In Blacklight 8 the code used to populate a modal was changed to reinitialize script tags: https://github.com/projectblacklight/blacklight/blob/a4ae88392ff6c0aba5bae473f9bffb21b203450e/app/javascript/blacklight-frontend/modal.js#L125

This change assumes that the script tag will be using a src attribute rather than having the scripts inline. Since our script tag was using inline javascript this caused the modal to be generated with an empty script tag, breaking the typeahead and tooltip functionality on the modal. Providing the javascript via an asset file referenced in the src attribute fixes the issue.


app/views/bookmarks/update_access_control.html.erb
```diff
@@ -139,18 +139,6 @@ Unless required by applicable law or agreed to in writing, software distributed
...
-  $('.form-text .close').on("click", function(event){
-    event.preventDefault()
-    $(this).parent().collapse('toggle')
-  });
-  $('.typeahead.from-model').each(function() {
-    initialize_typeahead($(this));
-  });
-  </script>
+  <script src="<%= asset_path('pop_help.js') %>"></script>
+  <script src="<%= asset_path('autocomplete.js') %>"></script>
```
cjcolvar, line 143: It `autocomplete.js here the coffeescript or a different one?

masaball: It is the autocomplete coffeescript file. We have to reference `autocomplete.js` here to make sure we are pulling in the compiled version of the file, but it is just the coffeescript.

Issue #6463, opened 2025-07-22
[BUG] Fail to fetch manifest error for any title with unknown date


elynema:
**Describe the bug**
When clicking to view media object from results, I'm getting a Failed to Load Manifest error. This appears to be happening for any item that has a date of "unknown"

For example: https://avalon-dev.dlib.indiana.edu/media_objects/jm214p131/

The manifest.json request returns with this error:

<img width="1370" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNWpPQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8b950fb68a88871a314aaf9320614bcb68f599cd/image.png" alt="image.png" />

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to avalon-dev.dlib.indiana.edu
2. Click Browse, then sort by date. This will sort the unknown date items to the top
3. Click on any title with an unknown date. For example: https://avalon-dev.dlib.indiana.edu/media_objects/jm214p131/
4. See the Failed to Load Manifest Error from Ramp.

**Expected behavior**
Media object should load properly in Ramp.

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macOS
 - Browser: Chrome/Safari
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
This is an 8.1 regression.

abankhele:
Clicking on media objects with unknown date is working as expected.

PR #6465, opened 2025-07-22
Make edtf humanization configurable

masaball:
Related issue: #6463

An issue was run into where we fixed date handling for the blacklight helper but failed to apply it to the iiif manifest generation as well. This PR moves the humanization into a lambda function in the avalon configuration. This allows a one stop place to adjust the logic and will allow other implementers to more easily adjust the humanization to their specifications.

This PR also includes some minor fixes to appease rubocop.


cjcolvar (approved):
Looks great!  Love how you moved this into a configurable lambda!

PR #6472, opened 2025-07-23
Updated ReadMe for cypress

abankhele:
Updated Readme for cypress test.
We have to now create 3 users and no testing media.


cjcolvar (approved):
Thanks for updating this!

PR #6478, opened 2025-07-24
Quote transcript search to allow for mixed phrase and non-phrase searching

cjcolvar:
TODO: Rebase on develop after #6447 is merged and write tests including showing that this doesn't force all transcript searches into phrase searches


app/controllers/application_controller.rb
```diff
@@ -40,7 +40,6 @@ class ApplicationController < ActionController::Base
   rescue_from RSolr::Error::Timeout, :with => :handle_solr_connection_error
   rescue_from Blacklight::Exceptions::ECONNREFUSED, :with => :handle_solr_connection_error
   rescue_from Faraday::ConnectionFailed, :with => :handle_fedora_connection_error
-  rescue_from Blacklight::Exceptions::InvalidRequest, :with => :handle_invalid_request_error
```
cjcolvar, line 43: This is no longer needed as the reworked search doesn't raise an error when there are unmatched quotes.

app/models/search_builder.rb
```diff
@@ -50,12 +50,17 @@ def add_access_controls_to_solr_params(solr_parameters)
...
     return unless solr_parameters[:q].present? && SupplementalFile.with_tag('transcript').any? && !(blacklight_params[:controller] == 'bookmarks')
 
-    # In order for the multi-word query to work we need to NOT RSolr.solr_escape the query AND replace the spaces as +; this is also true for quoted phrase searches
+    # In order for the multi-word query to work we need to NOT RSolr.solr_escape the query; this is also true for quoted phrase searches
     # We can manually escape solr special characters that cause issues.
-    query = solr_parameters[:q].gsub(/([(){}\[\]\^\*?:])/, "\\\\\1").tr(' ', '+')
-    transcript_subquery = "transcript_tsim:#{query}"
-    solr_parameters[:defType] = "lucene"
-    solr_parameters[:q] = "({!edismax v=\"#{RSolr.solr_escape(solr_parameters[:q])}\"}) {!join to=id from=isPartOf_ssim}{!join to=id from=isPartOf_ssim}#{transcript_subquery}"
+    query = solr_parameters[:q].gsub(/([\(\)\{\}\[\]\^\*\?\:\$\+\-\"\/])/, '\\\\\\\\\1')
```
cjcolvar: For some reason the switch from `"\\\\1"` to `\\\\1'` changed the result from `\\\u0001` to the expected `\\+` (in the case of `+`).  For certain characters they needed to be double-escaped in the subquery so I changed this to the unwieldy ` '\\\\\\\\\1'`.

I escaped every character in the regex to be safe. I also think it improves readability slightly, but I'm open to reverting this.

cjcolvar: I escaped every character in the regex to be safe.  I also think it improves readability slightly, but I'm open to reverting this.

Issue #6417, opened 2025-07-08
Remove Jasny bootstrap


masaball:
## Description
> This extension is for Bootstrap 4. Almost all of the components of this extension have been added to Bootstrap 5, making this library redundant. Because Bootstrap 4 has reached its end of life, Jasny Bootstrap will no longer be updated.

Jasny bootstrap has effectively been deprecated and is supposed to be unnecessary in Bootstrap 5. 

Do we know what components we were using Jasny for so we can try to implement the BS5 way?

# Jasny Components
- File upload button (styled with Jasny)
- More (to be investigated and compiled)

## Done Looks Like
- [ ] Investigation to determine all components currently using Jasny
- [ ] Implement alternatives to retain functionality without Jasny dependency

masaball:
Vanilla BS5 upload form options:
<img width="562" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3ZnQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6753334a8d72e28c6812ff367ceb35ba31c18f04/Screenshot_20250729_101621.png" alt="Screenshot_20250729_101621.png" />

<img width="556" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzNnQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--00070759fcac39190ed694a0313fc830d39a490e/Screenshot_20250729_101921.png" alt="Screenshot_20250729_101921.png" />

Bootstrap has more or less completely gutted the customization options for file input forms. As far as I have been able to determine there is no built-in way to change the text of the "Browse..." label. There are various discussions on StackOverflow and the like where you can use a different form input type and override the default behavior to recreate the file input workflow via javascript. That feels hacky and potentially brittle.

If we are deadset on a different button label or on the behavior/appearance of our current implementation we could just continue using Jasny Bootstrap. It is no longer being maintained so any security issues that may eventually be discovered we would likely be on the hook for fixing ourselves and we would need to keep it updated to properly interface with any classes BS changes in future releases. So this could become a bit of a maintenance nightmare.

joncameron:
[...]

PR #6487, opened 2025-07-29
Replace Jasny Bootstrap with vanilla Bootstrap file input

masaball:
Related issue: #6417

This PR leaves some of the Jasny classes in place because we have styling or javascript behavior that is triggered from those classes and, since the classes do not conflict with vanilla BS, it seemed most straightforward to leave that code alone.


Dananji (approved):

PR #6488, opened 2025-07-29
Fix tab focus and tab order in playlist details edit page

Dananji:
Related issue: #6356

Changes:
- switched `data-bs-target` to `href` in `<a>` tag for `Cancel` button in playlist edit details form. Adding `href` adds the button to the tab order of the page
- added `:focus-visible` pseudo-class to `.btn-outline` Bootstrap class to keep the outline visible when using keyboard focus. - **This change is applied to all the buttons in Avalon using `.btn-outline` class**
- add `:focus-visible` pseudo-class to the edit-icon button for both playlist and playlist-item details edit forms to make it visibly focused upon keyboard `Tab` focus

### DIFF
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -1240,6 +1240,12 @@ td {
   &:hover {
     background-color: $lightgray;
   }
+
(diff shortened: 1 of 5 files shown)


masaball (approved):
Looks good

Issue #6422, opened 2025-07-08
[A11y] Timeline edit details page


Dananji:
### Description

From the accessibility checks performed on timeline related pages in https://github.com/avalonmediasystem/avalon/issues/6314, the following issue was identified for the timeline edit page.

| Success criterion | WCAG 2.1 Level | Accessibility checker | Error | Relevant HTML element |
| --- | --- | --- | --- | --- |
| 1.4.3 Contrast (Minimum) | Level AA | SiteImprove & WAVE | Color contrast (1.84 : 1) doesn't meet min requirement | Help texts for visibility, tags and copy LTI link |
| 1.4.3 Contrast (Minimum) | Level AA | SiteImprove & WAVE | Color contrast (3.76 : 1) doesn't meet min requirement | Copy button for LTI link |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE| Form field missing a text alternative | `<input>` textbox for  timeline `Name` |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE | Form field missing a label | `<span>` combobox for `Tags` |
| 4.1.2 Name, Role, Value | Level A | SiteImprove & WAVE | Form field missing a label | `<input>` textbox for `Tags` |

### Done Looks Like
- [ ] SiteImprove/WAVE/ANDI bookmarklet/axeDevtool don't complain about any accessibility issues for the UI components on the page

joncameron:
Looks good! All of these are resolved in the accessibility checkers.

PR #6498, opened 2025-07-30
Fix timeline edit page a11y issues

Dananji:
Related issue: #6422


cjcolvar (approved):
I didn't test this out in a browser but looks good from a code perspective.

PR #6506, opened 2025-08-05
Fix timeliner container height for wider viewports

Dananji:
A bug was introduced with the [previous fix](https://github.com/avalonmediasystem/avalon/pull/6504) made to keep the timeline from overflowing into the footer where, the navigation bar was getting hidden in larger viewports. This PR fixes this bug.


masaball (approved):
:+1:

PR #6509, opened 2025-08-06
Update dependencies to releases or tags

masaball:
Also includes a bump of ruby-saml to take care of a security notice.


cjcolvar (approved):
👍

PR #6512, opened 2025-08-07
Add setting to change transcript tab title for EVIADA items

Dananji:
Related issue: https://github.iu.edu/AvalonIU/Avalon/issues/341


cjcolvar (commented):
This looks good but I think it would be best to leave the EVIADA part in our IU fork.

cjcolvar (approved):
Looks great!

Issue #6395, opened 2025-06-25
Move away from twitter-typeahead


masaball:
## Description
For our typeahead form fields we have been using the gem `twitter-typeahead-rails v0.11.1.pre.corejavascript`. This gem is unmaintained and Blacklight switched away from it in BL8: https://github.com/projectblacklight/blacklight/pull/2576 which I think this was what caused #6347.

https://github.com/github/auto-complete-element seems to be actively maintained and was Blacklight's choice for switching. We should work to follow suit or find another alternative.

## Done Looks Like
- [ ] Update typeahead code to use auto-complete-element gem
- [ ] Remove twitter-typeahead-rails gem from Avalon dependencies

joncameron:
Looks good! I only noticed one small bug: #6542.

PR #6515, opened 2025-08-12
Replace twitter-typeahead with @github/auto-complete-element

masaball:
Related issue: #6395


app/views/layouts/avalon.html.erb
```diff
@@ -32,6 +32,9 @@ Unless required by applicable law or agreed to in writing, software distributed
   <%= stylesheet_link_tag "application", media: "all" %>
   <%= stylesheet_pack_tag 'application' %>
   <%= javascript_pack_tag 'application' %>
+  <script type="module">
+    import githubAutoCompleteElement from 'https://cdn.skypack.dev/@github/auto-complete-element';
```
cjcolvar: What do you think about moving this to a npm/yarn dependency?  I'm thinking that we could control the version better that way and not have a dependency on the CDN being up.

Gemfile
```diff
@@ -26,7 +26,6 @@ gem 'jquery-ui-rails'
 gem 'react_on_rails'
 gem 'sassc-rails'
 gem 'sprockets-es6'
-gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
```
cjcolvar, line 29: 🎉

app/controllers/objects_controller.rb
```diff
@@ -28,17 +30,17 @@ def show
   def autocomplete
     expires_now
     model = Module.const_get(params[:t].to_s.classify)
-    render json: model.send(:autocomplete, params[:q].strip, params[:id])
+    @results = model.send(:autocomplete, params[:q].strip, params[:id])
```
cjcolvar, line 33: Is it worth retaining the older JSON response as an option since it might possibly be useful as an API endpoint?
```
@results = model.send(:autocomplete, params[:q].strip, params[:id])
respond_to do |format|
  format.html
  format.json { render json: @results }
end

cjcolvar (commented):
Looks great!  Super nice to be on a supported dependency!

app/views/layouts/avalon.html.erb
masaball: I initially was trying for an npm/yarn dependency for exactly the reasons you've laid out but was having trouble getting it working. But I had some unrelated issues at the start of working on this and maybe I didn't circle back when I got those resolved...I'll give it another shot because it would be preferable to not rely on a CDN :+1:

app/controllers/objects_controller.rb
masaball: I think that's a good idea. Doesn't hurt to keep around and may be something someone is using or could use.

cjcolvar:
This potentially fixes #6517 and #6530

cjcolvar (approved):
Looks great!

PR #6521, opened 2025-08-18
Backend and edit form for audio description support

masaball:


app/models/iiif_canvas_presenter.rb
```diff
@@ -96,225 +96,232 @@ def service
...
-    def section_processing?(master_file)
-      !master_file.succeeded?
-    end
+  def section_processing?(master_file)
+    !master_file.succeeded?
+  end
 
-    def supplemental_captions_transcripts
-      master_file.supplemental_files(tag: 'caption') + master_file.supplemental_files(tag: 'transcript')
+  def supplemental_accessibility_files
```
cjcolvar, line 159: I wonder if we could remove the need for this method by extending `MasterFiles#supplemental_files` to take an array for the `tag` kwarg.

spec/models/supplemental_file_spec.rb
```diff
@@ -298,21 +318,11 @@
...
-        "00:00:57.900 --> 00:01:05.700 Then, instead of getting into line at the lunchroom, Mr. Bungle pushed everyone aside and went right to the front.",
-        "00:01:06.000 --> 00:01:11.800 Even though this made the children laugh, no one thought that was a fair thing to do."
-      ] }
-      
       it 'splits the text by time cue' do
-        expect(subject).to match_array parsed_text
+        expect(subject).to be_a Array
+        expect(subject.length).to eq 10
+        expect(subject.all? { |s| s.is_a?(String) }).to eq true
+        expect(subject[0]).to eq "00:00:01.200 --> 00:00:21.000 [music]"
```
cjcolvar, line 325: Nice cleanup!

cjcolvar (approved):
This looks great!  I had one suggestion but it's not a blocker so could be handled in a follow up PR if at all.

PR #6522, opened 2025-08-18
Move blacklight helper overrides into CatalogController 

cjcolvar:
This allows them to take precedence over the ones supplied by Blacklight which fixes the layout for search results.  

This PR also includes new `staging` and `worker-staging` services in the `docker-compose.yml` that use the same dependency services (fedora, solr, redis, db, HLS streaming, minio) as the `avalon` service but with `RAILS_ENV=production`.  This allows for testing issues which only come up in the production environment.


masaball (approved):
Looks good to me

Issue #6529, opened 2025-08-18
[BUG] Assign Special Access: Begin Date & “Add” buttons not disabled


abankhele:
**Describe the bug**
In Edit Media Object > Assign special access, the Begin date fields for Avalon Group and IP Address or Range remain enabled, and all Add buttons are also enabled, while the rest of the fields are correctly disabled. This is inconsistent with the Collection page, where the Add buttons are disabled in the same state.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to https://avalon-staging.dlib.indiana.edu/media_objects/6t053f984/edit?step=access-control
2. Observe that the begin date are not disabled.
3. The Add buttons under Assign Special Access are not disabled

**Expected behavior**
The begin date field and add button behaviour should be consistent across the page

**Screenshots/Screen Recording**
<img width="1267" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL0VXQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--54a7de06d3dd081082a9d9f29b17d417e212f651/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser: chrome
- [ ] avalon-dev
- [x] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

Dananji:
The `Begin Date` form fields should not have been disabled, the code was picking up the wrong element to disable in the typeahead forms. Instead the `Add` button at the end of the form row should be disabled.

joncameron:
Working as expected on staging; the form for https://avalon-staging.dlib.indiana.edu/media_objects/6t053f984/edit?step=access-control is adding new values and date ranges as expected.

PR #6535, opened 2025-08-19
Correctly filter the 'Add' in the form row

Dananji:
Related issue: #6529 

Get the 'Add' button in the form row with typeahead form fields by filtering elements using the text on the button element instead of getting the element next to the typeahead form field. This makes it work with any form where there are additional form fields in the row.


cjcolvar (approved):
Great job making this specific and no longer brittle!

PR #6543, opened 2025-08-22
Delete autocomplete.js.coffee

masaball:
When rebasing the auto-complete-element work onto the most recent develop before merging it changed the handling of autocomplete.js.coffee from removal to simply being an empty file. This commit is to clean that up.


cjcolvar (approved):
🎉

Issue #6244, opened 2025-03-26
Publish / unpublish action is slow


elynema:
# Description

Testing version 8.0 in mco-staging shows that publish and unpublish actions appear slower than anticipated or desired, in comparison to version 7.8 running in production. 

To publish a single section item, we are seeing about 6 seconds each for the publish / unpublish API call. In MCO, that is more like 1-2 seconds.

joncameron:
Data center locations for Fedora/Solr in MCO and MCO-staging shouldn't be enough of an impact to show this level of increase in execution times.

joncameron:
On dev it's much improved and I haven't run into any slowness like experienced before. I've tested it in a number of different situations and I think this is in a good place now.

PR #6546, opened 2025-08-25
Refactor MediaObjectsController#update_status method

masaball:
Related issue: #6244

This commit makes three minor improvements to the `update_status` method:

1. Removes the extra save action from the publish code flow. This was originally put in place because of issues with permalink generation, but I did not run into any issues with a single save in testing.

2. Moves the initial ability check into the publish code flow, meaning that each code path is only performing one ability check.

3. Moves `MediaObject.find` out of the loop to grab all of the relevant media objects in a single query instead of making a query for each object.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -400,53 +400,50 @@ class MediaObjectsController < ApplicationController
         success_ids << id
         success_count += 1
       else
-        errors += [ "#{media_object.title} (#{params[:id]}) permission denied" ]
+        errors += ["#{media_object.title} (#{params[:id]}) permission denied"]
       end
     end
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
I think this looks great but I'm a little hesitant to move the `cannot? :update` ability check since it isn't exactly the same as `can? :unpublish` and is actually checking a little bit more than it.  It might be possible to go ahead with this change but I think we'd have to think it through and probably write some more tests.  Do you think the time savings is worth that effort?

masaball:
That's fair. I expect time savings from that particular change would be pretty minimal so probably not worth the extra effort. I'll update this to leave the ability checks in place.

cjcolvar (dismissed):
Looks good!

Issue #6533, opened 2025-08-18
[BUG] Publish/Unpublish success message shows wrong count for multi-object items


abankhele:
**Describe the bug**
When publishing or unpublishing an item that contains multiple media objects, the success banner always reads “1 media object successfully published.” (or “unpublished.”) instead of reflecting the actual number of media objects in the item.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to https://avalon-staging.dlib.indiana.edu/media_objects/6t053f984
2. Click on Publish/ Unpublish button.
3. Observe the error says as '1 media object successfully published.' where as there are 2 media objects present in the item.

**Expected behavior**
The success message should not show the number of media objects published if that number is 1, or disentangle this message from the batch message which requires a numeral

**Screenshots/Screen Recording**
<img width="1424" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeElYQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cddb6ed4f63cd4a094c71e5bef8f87e8552f1001/image.png" alt="image.png" />

<img width="1466" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFFYQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5bd914fb18f2a6d2fe75f5516d5a3c16c81b04c1/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser : chrome
- [ ] avalon-dev
- [x] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

PR #6548, opened 2025-08-26
Do not display item count when publishing a single item

masaball:
Related issue: #6533

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -449,7 +449,7 @@ class MediaObjectsController < ApplicationController
               elsif success_count == 1
                 "Media object successfully #{status}ed."
               elsif success_count > 1
-                "#{success_count} #{'media object'.pluralize(success_count)} successfully #{status}ed."
+                "#{success_count} media objects successfully #{status}ed."
               end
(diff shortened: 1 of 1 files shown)


app/controllers/media_objects_controller.rb
```diff
@@ -444,8 +444,13 @@ def update_status
...
       end
     end
-    message = "#{success_count} #{'media object'.pluralize(success_count)} successfully #{status}ed." if success_count.positive?
-    message = "Unable to #{status} #{'item'.pluralize(errors.count)}: #{ errors.join('<br/> ') }" if errors.count > 0
+    message = if errors.count.positive?
+                "Unable to #{status} #{'item'.pluralize(errors.count)}: #{errors.join('<br/> ')}"
+              elsif success_count == 1
+                "Media object successfully #{status}ed."
+              elsif success_count > 1
+                "#{success_count} #{'media object'.pluralize(success_count)} successfully #{status}ed."
```
cjcolvar: If we're checking that there is more than 1 then we can hard code the pluralization. 🙂 
```suggestion
                "#{success_count} media objects successfully #{status}ed."
```

cjcolvar (dismissed):
Looks good!

PR #6557, opened 2025-09-03
Add file upload form for audio descriptions

### LINKED ISSUE #6556: [BUG] Caption/Transcript language is not being persisted
**Describe the bug**
When trying to save metadata for a caption or transcript, the selected language is not being persisted to the database.

**To Reproduce**
1. Go to https://avalon-dev.dlib.indiana.edu/media_objects/5138jd902/edit?step=file-upload (or any item with captions/transcripts)
2. Click on the edit button for the section
3. Click on the edit button for the caption
4. Select the language input field
5. Enter a new language
6. Hit the save button for the caption
7. Shift-Reload the page to make sure you aren't loading cached data
8. Reopen the edit forms
9. See that the caption language is still "English"

**Expected behavior**
Users can select a language besides English and it will be persisted after saving.

 - Device: Laptop
 - OS: Ubuntu 24.04
 - Firefox
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
8.1.1

masaball:
Resolve #6556 and rebase the audio description branch before review and merging


app/assets/stylesheets/avalon.scss
```diff
@@ -946,6 +946,24 @@ h5.card-title {
           display: none;
         }
         display: flex;
+
+        .mb-3 {
+          margin-bottom: 0.5rem !important;
+        }
```
cjcolvar: Why does this bootstrap class need to be changed?  From the docs it appears this would be equivalent to mb-2. (https://getbootstrap.com/docs/5.2/utilities/spacing/#notation)
It this to avoid having to change the CSS classes back and forth in JS when the edit form is expanded?  I guess it looks like this predates this PR so probably doesn't need to be dealt with in this PR.

It does look like `mb-3` is handled on lin 942, could it be changed there instead of here?

masaball: The `.mb-3` elements are the input fields in the edit form. Bootstrap changed the default CSS for this class to `margin-bottom: 1rem !important` in BS5, which was overriding our custom CSS so that the checkboxes below the input forms were rendering too low (right on the line separating sections instead of in between the inputs and the line). 

I had missed that we had that custom CSS in place and had assumed we were using the BS4 default previously, which is why I overrode it at line 950 instead of 942. And the choice to override was entirely a "when you have a hammer, everything is a nail" approach. I need to work on not necessarily defaulting to overriding when working with CSS. Looking over things, it would probably be better to just use `.mb-2` or even `.mb-1` assuming that doesn't break other aspects of the styling.

PR #6558, opened 2025-09-03
Fix language persistence on supplemental file update

masaball:
Related issue: #6556

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -166,6 +166,13 @@ class SupplementalFilesController < ApplicationController
     end
 
     def supplemental_file_params
+      # Typeahead field requires a unique name for proper association between input and dropdown.
+      # This param is part of the regular params hash, not the child supplemental_file hash.
+      # Requests should still only contain one language param, filter for it to include in sup_file_params.
+      if params[:supplemental_file].present? && params[:supplemental_file][:language].blank?
+        language_param = params.keys.select { |key| key.match?(/^language_\d+$/) }
+        params[:supplemental_file][:language] = params[language_param.first]
+      end
       # TODO: Add parameters for minio and s3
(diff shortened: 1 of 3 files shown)


app/controllers/supplemental_files_controller.rb
```diff
@@ -166,6 +166,13 @@ def validate_params
     end
 
     def supplemental_file_params
+      # Typeahead field requires a unique name for proper association between input and dropdown.
+      # This param is part of the regular params hash, not the child supplemental_file hash.
+      # Requests should still only contain one language param, filter for it to include in sup_file_params.
+      if params[:supplemental_file].present? && params[:supplemental_file][:language].blank?
+        language_param = params.keys.select { |key| key.match?(/^language_\d+$/) }
```
cjcolvar: I wonder if there is a different way to handle this.  I think we have some places where we use JS to mung the form data before posting.  I think I'd prefer that over this since the backend shouldn't really have to deal with the idiosyncrasies of the frontend.  On the other hand this works and preserves the prior cleaner API.  I guess this is okay for now.

cjcolvar (approved):
Looks good.

app/controllers/supplemental_files_controller.rb
masaball: Thanks for bringing that up. Prompted my memory and I had forgotten rails conventions and my own previous work for a minute there. We don't even need to use javascript. 

 When providing arguments to the new autocomplete module, if we pass `supplemental_file[language]` as the `field_name` parameter, that generates the `{ supplemental_file: { language: 'English' } }` child within the parameters object. Then we can pass in a unique `for_field` parameter for each section so the typeahead doesn't mis-associate the dropdowns.

cjcolvar (approved):
Looks great!  Glad the standard approach worked out!

Issue #6384, opened 2025-06-23
Cache IIIF Presentation Manifests


joncameron:
## Description
Keeping a cached or pre-generated manifest available for media objects would speed item view up considerably. Manifests for items in Avalon are currently dynamically generated in all cases.

## Done Looks Like
- [ ] Investigate methods to cache or pre-generate Presentation 3 manifests for media objects

PR #6563, opened 2025-09-09
Cache Media Object IIIF manifests

masaball:
Related issue: #6384 

This PR leverages Rails low-level caching to cache our media object IIIF manifests. From what I was seeing, this seemed like the most straightforward/robust path. If we move forward with this, should we have a different expiration time on the cached items?

### DIFF
--- a/spec/controllers/media_objects_controller_spec.rb
+++ b/spec/controllers/media_objects_controller_spec.rb
@@ -16,7 +16,6 @@ require 'rails_helper'
 
 describe MediaObjectsController, type: :controller do
   include ActiveJob::TestHelper
-  include ActiveSupport::Testing::TimeHelpers
 
   render_views
 


app/presenters/speedy_af/proxy/media_object.rb
```diff
@@ -154,6 +154,12 @@ def workflow
     WorkflowDatastream.find("#{id}/workflow")
   end
 
+  def cache_key_with_version
+    # Proxy has two potential values for last updated, use the latest one for version number
+    version = [@attrs[:system_modified], @attrs[:descMetadata_modified]].max
+    "media_object/#{id}-#{version.to_fs(:number)}"
```
cjcolvar: FWIW Each solr doc has a `_version_` field which gets updated anytime a document gets indexed.  Since it appears that this cache_key is relying solely on the media object solr doc and not any other associated doc then I think we could use this field here.

```suggestion
    "media_object/#{id}-#{@attrs[:_version_]}"
```

cjcolvar (approved):
Looks great!

I was a little concerned about `secure_stream_infos` being called inside the cache block but it looks like the manifest doesn't actually use any of the user/session specific data  and the HLS endpoints should be able to create tokens if they are hit without tokens being generated here first.

cjcolvar (approved):

spec/controllers/media_objects_controller_spec.rb
```diff
@@ -16,6 +16,7 @@
 
 describe MediaObjectsController, type: :controller do
   include ActiveJob::TestHelper
+  include ActiveSupport::Testing::TimeHelpers
```
cjcolvar: I think this isn't needed anymore with your last commit.
```suggestion
```

cjcolvar (dismissed):

spec/controllers/media_objects_controller_spec.rb
masaball: Good catch

cjcolvar (approved):

Issue #6542, opened 2025-08-21
Typeahead Dropdown on Resource Description Appears Until Value is Selected


joncameron:
## Description
In the Resource Description form, entering a value in a typeahead field such as Series do not disappear as expected unless the user selects on of the list items. This means that the dropdown list stays on the page even if the user clicks outside the dropdown, or focus is moved to another form element. 

It appears that the only way to get it to go away is the `Esc` key or selecting on of the presented list items in the dropdown. The dropdown automatically opens on focus so someone tabbing through the page can end up with this dropdown blocking form fields or other content.

### Current Behavior

![Image](https://github.com/user-attachments/assets/8cf4d925-ff08-4371-9331-a6940e099ea8)

## Done Looks Like
- [ ] Dropdown closes once it loses focus

joncameron:
Timeboxed to 2 points.

PR #6574, opened 2025-09-12
Close typeahead dropdown when input loses focus

masaball:
Related issue: #6542

### DIFF
--- a/app/javascript/auto-complete-open.js
+++ b/app/javascript/auto-complete-open.js
@@ -16,11 +16,16 @@
 
 import Autocomplete from '@github/auto-complete-element/dist/autocomplete.js'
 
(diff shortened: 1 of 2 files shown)


masaball:
Oh, good catch on that being firefox only. Interesting that it has been present in FF since version 1.5 and no other browser has bothered to implement it.

I just connected some dots and think I found a way that doesn't require a timer or listening to events at all. If we monkeypatch `Autocomplete.prototype.open` so that we set `interactingWithList = false` instead of true like the base library does, we should be good. There is an `onResultsMouseDown` function in the class that sets that bool to true when the options list is moused down on, which fires before the blur event. In quick testing just now the behavior was working as I would expect.

PR #6589, opened 2025-09-23
Transition Avalon to jsbundling-rails

masaball:
One potential pain point moving forward is that the react-on-rails gem is developed by the developers of shakapacker and is expected to be used in conjunction with it. Everything seems to be loading and working as I would expect but there may eventually be updates to react-on-rails or advanced features we eventually want to use which could break the current integration. It shouldn't be terribly difficult to revert back to Shakapacker if that becomes necessary or if jsbundling-rails causes too many other issues/headaches. Or we could switch to loading our React components manually instead of relying on the helpers from react-on-rails.

The other biggest potential headache that I see for our use of jsbundling is SME. I'm not sure what exactly the root cause is but something in SME breaks with the default minimizer settings. I was able to get it to work by enabling a couple options for Terser's mangle ability but it may end up being another pain point.

In addition to the asset pipeline update, this PR also moves us over to the modern `bin/dev` initialization instead of directly calling `rails server`

Note: There are a handful of deleted files such as `app/javascripts/collection_index.html` that appeared to be old 'Hello World'-type implementations so I removed them without replacement.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,6 @@ gem 'csv'
 
 # Assets
 gem 'bootstrap', '~> 5.0'
-gem 'coffee-rails', '~> 5.0'
 gem "font-awesome-rails"
 gem 'jsbundling-rails'
 gem 'jquery-rails'
@@ -50,8 +49,6 @@ gem 'activerecord-session_store', '>= 2.0.0'
 gem 'acts_as_list'
 gem 'api-pagination'
 gem 'avalon-about', git: 'https://github.com/avalonmediasystem/avalon-about.git', branch: 'main'
-# gem 'bootstrap-sass', '< 3.4.1' # Pin to less than 3.4.1 due to change in behavior with popovers
-# gem 'bootstrap-toggle-rails'
 gem 'bootstrap_form'
 gem 'iiif_manifest', '~> 1.6'
 gem 'rack-cors', require: 'rack/cors'
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,4 +1,4 @@
 //= link_tree ../images
-//= link application.css
 //= link application.js
 //= link_tree ../builds .js
+//= link application.css
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -9,10 +9,10 @@ module.exports = (env, argv) => {
     mode: argv.mode,
     devtool: isProd ? false : 'eval-source-map',
     entry: {
-      application: '/home/app/avalon/app/javascript/application.js',
-      embed: '/home/app/avalon/app/javascript/embed.js',
-      'iiif-timeliner': '/home/app/avalon/app/javascript/iiif-timeliner.js',
-      'server-bundle': '/home/app/avalon/app/javascript/server-bundle.js',
+      application: path.resolve(__dirname, '..', '..', 'app/javascript/application.js'),
+      embed: path.resolve(__dirname, '..', '..', 'app/javascript/embed.js'),
+      'iiif-timeliner': path.resolve(__dirname, '..', '..', 'app/javascript/iiif-timeliner.js'),
+      'server-bundle': path.resolve(__dirname, '..', '..', 'app/javascript/server-bundle.js'),
     },
     module: {
       rules: [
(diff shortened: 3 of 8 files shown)


app/assets/config/manifest.js
```diff
@@ -0,0 +1,4 @@
+//= link_tree ../images
```
cjcolvar, line 1: I don't see this directory.  Does it need to be added with a `.keep` like `builds`?

app/views/layouts/embed.html.erb
```diff
@@ -22,9 +22,8 @@ Unless required by applicable law or agreed to in writing, software distributed
 
    <%= csrf_meta_tags %>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <%= favicon_link_tag %>     
-    <%= stylesheet_pack_tag 'embed' %>
-    <%= javascript_pack_tag 'embed' %>
+    <%= favicon_link_tag %>
+    <%= javascript_include_tag "embed", "data-turbo-track": "reload", type: "module" %>
```
cjcolvar, line 26: Is the stylesheet tag not needed here?

spec/rails_helper.rb
```diff
@@ -146,6 +146,8 @@
 
   config.before :each do
     DatabaseCleaner.strategy = :truncation
+    # Clear out faker unique generator between tests
+    Faker::UniqueGenerator.clear
```
cjcolvar, line 150: Nice catch!

Gemfile
```diff
@@ -53,7 +52,7 @@ gem 'acts_as_list'
 gem 'api-pagination'
 gem 'avalon-about', git: 'https://github.com/avalonmediasystem/avalon-about.git', tag: 'avalon-r8.0'
 # gem 'bootstrap-sass', '< 3.4.1' # Pin to less than 3.4.1 due to change in behavior with popovers
-gem 'bootstrap-toggle-rails'
+# gem 'bootstrap-toggle-rails'
```
cjcolvar: Can this line be removed?

config/webpack/webpack.config.js
```diff
@@ -1,25 +1,105 @@
...
-  ],
-};
+  const config = {
+    mode: argv.mode,
+    devtool: isProd ? false : 'eval-source-map',
+    entry: {
+      application: '/home/app/avalon/app/javascript/application.js',
+      embed: '/home/app/avalon/app/javascript/embed.js',
+      'iiif-timeliner': '/home/app/avalon/app/javascript/iiif-timeliner.js',
+      'server-bundle': '/home/app/avalon/app/javascript/server-bundle.js',
```
cjcolvar: Is it a problem for these to be absolute paths?

package.json
```diff
@@ -1,21 +1,27 @@
...
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
+    "@babel/plugin-transform-class-properties": "^7.27.1",
+    "@babel/plugin-transform-nullish-coalescing-operator": "^7.27.1",
     "@babel/preset-react": "^7.0.0",
     "@babel/runtime": "^7.21.1",
     "@github/auto-complete-element": "^3.8.0",
-    "@rails/webpacker": "^5.4.4",
+    "@popperjs/core": "^2.11.8",
     "@samvera/ramp": "https://github.com/samvera-labs/ramp.git#a0b3ebf",
+    "babel": "^6.23.0",
```
cjcolvar: It looks like this package has been deprecated in favor of `babel-cli`.  Would it work to use that instead?

app/assets/config/manifest.js
masaball: That line is pointing at `app/assets/images` which is a directory that already existed and I don't think I had to touch so isn't part of the commit. If you check the branch itself it should be present.

app/views/layouts/embed.html.erb
masaball: It is not, as far as I could tell. The embedded player component imports its CSS/SCSS in the JSX file, so webpack is setup to transpile that CSS and then it gets delivered as part of the javascript bundle.

config/webpack/webpack.config.js
masaball: Good catch, it probably will be a problem. Any standard docker deploy would be fine, but MCO for example would fail because we set that up with different parent directory naming. I will switch all the absolute paths in the config to be relative.

package.json
masaball: Good catch. I will make the switch and see what happens.

cjcolvar (approved):
Looks good.

Issue #6565, opened 2025-09-10
Set up cssbundling-rails to transpile and bundle SCSS


joncameron:
## Description
Setting up cssbundling-rails to transpile and bundle our SCSS. Shakapacker can be set up in advance of this work to serve only JS so we can focus on just the CSS bundling. Our CSS bundling should be separate from JS in our asset pipeline. 

## Done Looks Like
- [ ] migrate Avalon to `dartsass-rails` gem
- [ ] cssbundling-rails added to Avalon
- [ ] cssbundling-rails used for CSS transpilation and bundling for Avalon's asset pipeline

PR #6603, opened 2025-10-01
Transition Avalon to cssbundling-rails

masaball:
Related issue: #6565 

Some notes:
* There may be minor issues with fontawesome icons. The Rails gem that we had been using was pinned to 4.7.0 which came out 9 years ago. I upgraded to the free version of 6.7.2. I have not noticed anything broken yet, but there are possibly changes to icons or icons we were using that are no longer in the free tier so we will want to keep an eye out.
* External gems that have sass stylesheets are not handled by default in cssbundling-rails. A workaround is to manually add the gems as load-path arguments in the yarn command then import as normal in the application.css. See https://github.com/rails/cssbundling-rails/issues/81


cjcolvar (commented):
This looks great as far as I can tell!

I'm guessing that you'll need to add `yarn build:css` to `docker_init.sh` like we did in #6604 so the CSS bundle gets built for avalon-dev.  If you agree can you please add it to this PR?

PR #6615, opened 2025-10-13
Make Derivative#location_url and Derivative#hls_url be dynamically derived from Derivative#derivativeFile

cjcolvar:
I'm exploring if streaming urls can be derived dynamically so switching streaming server base urls won't involve changing metadata, saving, and indexing all of the Derivative objects.

### DIFF
--- a/app/models/derivative.rb
+++ b/app/models/derivative.rb
@@ -71,22 +71,36 @@ class Derivative < ActiveFedora::Base
 
   def initialize(*args)
     super(*args)
-    self.managed = true
+    self.managed = true if self.managed.nil?
   end
 
-  def set_streaming_locations!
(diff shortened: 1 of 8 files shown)


cjcolvar:
Save this for after Avalon 8.2 gets cut.

masaball (approved):
Looks good to me. 

Should there be a follow-up ticket on this that removes the `location_url` property from the model and index once everything is fully tested? Or is there still a need for the field to be in the metadata?

cjcolvar:
Good question.  It doesn't seem like we really need both `location_url` and `hls_url` anymore (and probably haven't for years).  I'll write up a new ticket.

cjcolvar:
Created https://github.com/avalonmediasystem/avalon/issues/6870

Issue #6054, opened 2024-10-02
Add Unit Objects


joncameron:
# Description
Adding Units as objects in Avalon.

## Unit Metadata
- Title
- Description
- Poster Image
- Contact Email
- Website URL
- Website Label

# Done Looks Like
- [ ]

PR #6616, opened 2025-10-13
Add new Units model, controller, views

masaball:
Related issue: #6054 

By changing Units from a controlled vocab list to a model, this PR changes Collections to have a foreign key relationship to the unit. Is this how it should actually be implemented? There is definitely the `has_many -> belongs_to` database relationship, but remediating the data on application update is likely significant. All existing Unit values would need to have a corresponding instance created and then each collection would need to be changed to have the appropriate foreign key. This is probably doable with a rake task, but we need to make sure it is ironclad. 

Edit: created potential rake task. It uses `update_all` which is more efficient since it is a single database injection, but does skip validations and callbacks. Since this is a rake task that should only need to be used once, maybe it is better to cycle through and individually update each record so all the callbacks and everything are properly set off...

TODO:
* Do we need to implement a public `Units` tab like we have with collections? Or is the Unit meant as an administrative aspect that is only surfaced for organizing the collection index? If we need the tab, the views need to be added and controller/model code needs to be tweaked for functionality. If we are not adding the tab we can likely remove some of the files added in this PR.
* Verify that permissions are accurate to the levels of control we want each staff role to have at the unit level.
* Is it possible to initialize a default unit when the application is first created?


app/models/admin/collection.rb
```diff
@@ -193,7 +188,7 @@ def as_json(options={})
     {
       id: id,
       name: name,
-      unit: unit,
+      unit: unit.name,
```
cjcolvar: Should this use the safe navigator?
```suggestion
      unit: unit&.name,
```

app/models/admin/collection.rb
```diff
@@ -77,16 +75,12 @@ class Admin::Collection < ActiveFedora::Base
 
   has_subresource 'poster', class_name: 'IndexedFile'
 
-  around_save :reindex_members, if: Proc.new{ |c| c.name_changed? or c.unit_changed? }
-  around_save :return_checkouts, if: Proc.new{ |c| c.cdl_enabled_changed? && c.cdl_enabled == false }
+  around_save :reindex_members, if: Proc.new { |c| c.name_changed? or c.unit_changed? }
```
cjcolvar, line 79: We'll need to think about how this trigger needs to change since the collection no longer knows when the unit's name changes.  We could change this to be just `c.name_changed?` and have a similar `around_save` in Unit which triggers `reindex_members` on all of its collections.  This could be a _lot_ of reindexing so we may want a warning.

This doesn't need to be handled in this PR.

app/models/iiif_manifest_presenter.rb
```diff
@@ -134,7 +134,10 @@ def display_collection(media_object)
   end
 
   def display_unit(media_object)
-    "<a href='#{Rails.application.routes.url_helpers.collections_url(filter: media_object.collection.unit)}'>#{media_object.collection.unit}</a>"
+    # The proxy object returns the string of the unit name, the actual MediaObject
+    # returns the Unit itself so we have to specifically request the name in that case.
+    unit = media_object.is_a?(SpeedyAF::Proxy::MediaObject) ? media_object.collection.unit : media_object.collection.unit.name
```
cjcolvar: We could make this `media_object.collection.unit.name` for both cases or if we want to keep the logic for different cases then I think it is actually possible to make it shorter for the proxy: `media_object.unit.first`.

app/views/admin/collections/_form.html.erb
```diff
@@ -28,7 +28,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <div class="modal-body">
         <%= f.text_field :name, 'data-testid': 'collection-update-name' %>
         <% if @collection.new_record? || can?(:update_unit, @collection)%>
-          <%= f.select(:unit, Admin::Collection.units, {}, {'data-testid': 'collection-update-unit'} ) %>
+          <%= f.select(:unit_id, Admin::Unit.all.map { |u| [u.name, u.id] }, {}, {'data-testid': 'collection-update-unit'} ) %>
```
cjcolvar: I think it would be good to figure out a convenience method for this that fetches the names from solr and maybe memoizes it.  As is I think this will fetch all unit objects from fedora.

config/initializers/active_fedora_general.rb
```diff
@@ -215,3 +215,54 @@ def readonly?
     @readonly || Settings.repository_read_only_mode
   end
 end
+
+# Attempt to patch ActiveFedora to register when foreign key changes
+# See https://github.com/rails/rails/pull/42751 for ActiveRecord implementation
```
cjcolvar, line 220: Are these methods to support the unit_changed? check in the around_save callback in Admin::Collection?  If so, I wonder if we really need them.  I suspect that collections changing units will be extremely rare and we don't have any mechanism in the UI to do that.  These methods are conceptually cool I'm just afraid of things not working exactly as expected.  (Like I think that ActiveFedora may go ahead and persist the change on setting so dirty tracking might be moot.)

app/models/unit_search_builder.rb
```diff
@@ -0,0 +1,62 @@
...
+    ability = Ability.new(user) if user.present?
+    ability ||= current_ability
+    apply_gated_discovery(solr_parameters, discovery_permissions, ability) if ability.cannot? :discover_everything, Admin::Unit
+  end
+
+  def gated_discovery_join(solr_parameters)
+    temp_solr_parameters = {}
+    add_access_controls_to_solr_params_if_not_admin(temp_solr_parameters)
+
+    query =  "{!join from=heldBy_ssim to=id}"
```
cjcolvar, line 42: I think we may need to doubly join here to get from the unit down to the media object.  I forget if it works to chain them:
`{!join from=heldBy_ssim to=id}{!join from=isMemberOfCollection_ssim to=id}`

lib/tasks/avalon_migrations.rake
```diff
@@ -127,5 +127,33 @@ namespace :avalon do
...
+    task admin_units: :environment do
+      if ENV['unit_admin_username'].nil?
+        abort "You must specify a username. This user must be part of the Unit Administrator group. Example: rake avalon:migration:admin_units unit_admin_username=user@example.edu"
+      end
+
+      units = {}
+      units_cv = Avalon::ControlledVocabulary.find_by_name('units', sort: true)
+      units_cv.each do |unit|
+        unit_model = Admin::Unit.new(name: unit, unit_admins: [ENV['unit_admin_username']])
+        unit_model.save!
```
cjcolvar, line 147: Maybe this could be changed into a something akin to a find_or_create_by_name.  If this tries finding any that exist before creating then it would be possible to run this rake task again (in case of errors the first time) without creating duplicate unit objects.

lib/tasks/avalon_migrations.rake
```diff
@@ -127,5 +127,33 @@ namespace :avalon do
...
     end
+    # TODO: Test
+    desc "Migrate existing units to new Admin::Unit model and associate them with the appropriate Admin::Collections"
+    task admin_units: :environment do
+      if ENV['unit_admin_username'].nil?
+        abort "You must specify a username. This user must be part of the Unit Administrator group. Example: rake avalon:migration:admin_units unit_admin_username=user@example.edu"
+      end
+
+      units = {}
+      units_cv = Avalon::ControlledVocabulary.find_by_name('units', sort: true)
```
cjcolvar, line 138: Should this list be reduced to only the ones currently being used by collections?  Or is it helpful to create unit objects that are empty?

lib/tasks/avalon_migrations.rake
```diff
@@ -127,5 +127,33 @@ namespace :avalon do
...
+      rescue ActiveFedora::RecordInvalid => e
+        puts "Creation failed for #{unit}: #{e.message}"
+      end
+
+      units.each do |k, v|
+        collections = Admin::Collection.where(unit: k)
+        collection_ids = collections.map(&:id)
+        # update_all inserts straight into the database, skipping validations
+        # and callbacks. Manually call ReindexJob to ensure that solr is updated.
+        collections.update_all(unit_id: v)
```
cjcolvar: I'm not sure if `update_all` exists (or works as expected) for ActiveFedora::Base objects.  I wonder if this loop should go over all collections one by one instead of units.  Then any collections that might somehow not have a valid unit can be flagged and not skipped over.

config/initializers/active_fedora_general.rb
masaball: Yes, they are intended to support `unit_changed?`. There is the ability to change unit in the Collection edit modal, but you're right that it should be an extremely rare thing to have to do. I'm fine with locking that down and removing the option for users to change unit if @joncameron doesn't see a pressing need for it to stay.

app/models/unit_search_builder.rb
masaball: Do we need to drill down to the media object? In the current implementation, this is primarily (solely?) used by the Units index where my assumption was we would be only concerned that child collections were being returned.

app/views/admin/collections/_form.html.erb
masaball: Good catch. Most applications it probably wouldn't be too big a hit, but I expect MCO would get pretty slow.

Issue #6564, opened 2025-09-10
Remove jquery dependency


joncameron:

Dananji:
It seems the `ready` defined in the `gridjs` library is not supported anymore, which is needed to make adjustments to the view once the table is rendered on the page. I was not able to get tag filter working with the `gridjs` implementation, because there was no supported event handlers in `gridjs` that I could use to attach the tag filter change event to refresh the data in the table.

Therefore, I'm experimenting with a ReactJS approach. Starting off with a vanilla ReactJS implementation and if it seems to be complex, I will explore `@tanstack/react-table` library.

PR #6617, opened 2025-10-13
Replace jQuery datepikcer with js-datepicker

Dananji:
Related issue: #6564

Changes in this PR:
- replace jQuery datepicker with `js-datepicker` library, which is a dependency free JS library
- some overrides to the default styling coming from `js-datepicker`
- conversion of jQuert code in the changed files to vanilla JS

### DIFF
--- a/app/assets/javascripts/access_control_step.js
+++ b/app/assets/javascripts/access_control_step.js
@@ -30,20 +30,24 @@ window.initializeDatepickers = function (container) {
     };
 
     let setupRange = (id, isBeginDate) => {
+      if (!id) return;
       let pairedInput = null;
       const pairedDateId = isBeginDate ? id.replace('_begin', '_end') : id.replace('_end', '_begin');
       pairedInput = root.querySelector(`#${pairedDateId}, [name="${pairedDateId}"]`);
 
       if (pairedInput) {
-        isBeginDate ? pickerOptions.onSelect = function (instance, date) {
-          if (pairedInput.datepicker) {
-            pairedInput.datepicker.setMin(date);
-          }
-        } : pickerOptions.onSelect = function (instance, date) {
-          if (pairedInput.datepicker) {
-            pairedInput.datepicker.setMax(date);
-          }
-        };
+        switch (isBeginDate) {
+          case true:
+            pickerOptions.onSelect = function (instance, date) {
+              pairedInput.datepicker?.setMin(date);
+            };
+            break;
+          case false:
+            pickerOptions.onSelect = function (instance, date) {
+              pairedInput.datepicker?.setMax(date);
+            };
+            break;
+        }
       }
     };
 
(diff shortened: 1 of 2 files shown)


app/views/bookmarks/update_access_control.html.erb
```diff
@@ -105,8 +104,8 @@ Unless required by applicable law or agreed to in writing, software distributed
                 <div class="d-flex flex-column flex-sm-row mb-1">
                   <%= render partial: 'modules/autocomplete_input', locals: { field_name: 'class',
                                                                               model: 'externalGroup' } %>
-                  <%= text_field_tag "add_class_begin", nil, placeholder: 'Begin Date (yyyy-mm-dd)', class: 'form-control date-input access_date', autocomplete: 'off', type: 'date' %>
-                  <%= text_field_tag "add_class_end", nil, placeholder: 'End Date (yyyy-mm-dd)', class: 'form-control date-input access_date', autocomplete: 'off', type: 'date' %>
+                  <%= text_field_tag "add_class_begin", nil, placeholder: 'Begin Date (yyyy-mm-dd)', class: 'form-control date-input access_date', autocomplete: 'off' %>
+                  <%= text_field_tag "add_class_end", nil, placeholder: 'Begin Date (yyyy-mm-dd)', class: 'form-control date-input access_date', autocomplete: 'off' %>
```
masaball:
```suggestion
                  <%= text_field_tag "add_class_end", nil, placeholder: 'End Date (yyyy-mm-dd)', class: 'form-control date-input access_date', autocomplete: 'off' %>
```

app/assets/javascripts/access_control_step.js
```diff
@@ -12,5 +12,81 @@
...
+        isBeginDate ? pickerOptions.onSelect = function (instance, date) {
+          if (pairedInput.datepicker) {
+            pairedInput.datepicker.setMin(date);
+          }
+        } : pickerOptions.onSelect = function (instance, date) {
+          if (pairedInput.datepicker) {
+            pairedInput.datepicker.setMax(date);
+          }
+        };
+      }
```
masaball, line 51: This feels kind of unreadable to me. The mix of the nested `if` statements and multiline ternary is hard to parse at a glance. Would using a switch statement and safe navigators work here? Something like:
```javascript
if (pairedInput) {
  switch(isBeginDate) {
    case true:
      pickerOptions.onSelect = function (instance, date) {
        pairedInput.datepicker?.setMin(date);
      }
      break;
    case false:
      pickerOptions.onSelect = function (instance, date) { 
        pairedInput.datepicker?.setMax(date);
      }
      break;
  }
}
```

app/assets/javascripts/access_control_step.js
```diff
@@ -12,5 +12,81 @@
...
+    input.dataset.datepickerInitialized = 'true';
+  });
+};
+
+document.addEventListener('DOMContentLoaded', function () {
+  // Initialize on page load
+  window.initializeDatepickers();
+
+  // Watch for dynamically added content using MutationObserver for Blacklight modals
+  const observer = new MutationObserver(function (mutations) {
```
masaball, line 71: Nice! I got the blacklight modals to display bootstrap popovers via explicitly adding the JS to the partial in a `<script/>` tag which was fine because it was just one partial. But using a MutationObserver would be good to keep in mind if we ever add any more modals that need popovers.

app/assets/javascripts/access_control_step.js
Dananji: Yes, this way seems clearer and easier to understand at a first glance.

PR #6619, opened 2025-10-14
Remove select2

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/6564

Changes in this PR:
- replace `select2` with a JS library called `tom-select` which provides similar functionality and can be maintained via yarn/npm
- clear `select2` vendor assets for JS and CSS

### DIFF
--- a/app/assets/javascripts/add-playlist-option.js
+++ b/app/assets/javascripts/add-playlist-option.js
@@ -12,10 +12,10 @@
 //   specific language governing permissions and limitations under the License.
 // ---  END LICENSE_HEADER BLOCK  ---
 
-// This script will enable support for the html5 form attribute
-// This should only be needed for IE but is currently applied wholesale
-// to all disjointed submit elements as is needed in some of the workflow steps
-
+/**
+ * This script will enable support adding 'Add new playlist' and search functionality in
+ * add to playlist dropdown menu via tom-select JS library.
+ */
 document.addEventListener('DOMContentLoaded', () => add_new_playlist_option());
 
 this.add_new_playlist_option = function () {
@@ -36,7 +36,7 @@ this.add_new_playlist_option = function () {
     return iconHtml + '<b>' + addnew + termText + '</b>';
   };
 
-  // Helper function to update "Add new playlist" option in dropdown
+  // Helper function to update "Add new playlist" option with typed search term
   const updateAddNewPlaylistOption = function (dropdown_content, searchTerm) {
     const addNewOption = dropdown_content.querySelector('[data-value="' + addnew + '"]');
     if (addNewOption) {
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -40,12 +40,12 @@
 //= require url-search-params-polyfill/index.js
 //= require sortablejs/Sortable.min.js
 //= require js-datepicker/dist/datepicker.min.js
+//= require tom-select/dist/js/tom-select.complete.min.js
 
 //= require moment/min/moment-with-locales.min.js
 //= require hls.js/dist/hls.min.js
 
 // include all of our vendored js
 //= require_tree ../../../vendor/assets/javascripts/.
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/application.js
```diff
@@ -46,5 +46,6 @@
 
 // include all of our vendored js
 //= require_tree ../../../vendor/assets/javascripts/.
+//= require tom-select/dist/js/tom-select.complete.min.js
```
cjcolvar: Does this need to be loaded after vendored js?  If not then I think it more logically fits above with other npm/yarn managed requires.  I think the include all vendored js and then require_tree is boilerplate so would be more readable to not put something else in it.

app/views/bookmarks/merge.html.erb
```diff
@@ -14,13 +14,7 @@ Unless required by applicable law or agreed to in writing, software distributed
...
 %>
 
-<div data-blacklight-modal="container">
-  <script type="text/javascript">
-    $('#target_collection_id').select2({
-      dropdownParent: $('#blacklight-modal')
-    });
-  </script>
-
+<div data-blacklight-modal="container">7
```
cjcolvar: Looks like a small typo here.
```suggestion
<div data-blacklight-modal="container">
```

app/assets/javascripts/add-playlist-option.js
```diff
@@ -16,140 +16,339 @@
 // This should only be needed for IE but is currently applied wholesale
 // to all disjointed submit elements as is needed in some of the workflow steps
```
cjcolvar: This comment looks incorrect.  I'm thinking it should probably be removed or replaced with something that explains the function of this JS file.

app/views/playlists/_tag_form.html.erb
```diff
@@ -24,28 +24,51 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+      if (submitButton) {
+        submitButton.addEventListener('click', function(e) {
+          e.preventDefault();
+          const initialTags = tagSelect.getValue();
+          const tagsArray = Array.isArray(initialTags) ? initialTags : [initialTags];
+          document.getElementById('taglist').value = JSON.stringify(tagsArray);
+          document.getElementById('playlist_form').submit();
+        });
+      }
     });
```
cjcolvar: Is this form submission handling needed?  I see that the `onChange` is setting the hidden input to be ready for form submission.  The timeline tag form doesn't seem to have this submit click handling.  That makes me think that either it needs to be added there or it is safe to remove it here.

app/views/timelines/_tag_form.html.erb
```diff
@@ -24,28 +24,44 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+          this.setTextboxValue('');
+        },
+        render: {
+          no_results: function(data, escape) {
+            return null;
+          },
+          option_create: function(data, escape) {
+            // Customize the HTML and text for the create new tag message
+            return '<div class="create">Create new tag: <strong>' + escape(data.input) + '</strong></div>';
+        }
```
cjcolvar: Looks like this closing brace is out of alignment.
```suggestion
          }
```

cjcolvar (commented):
Great work switching out these libraries and maintaining the behavior!  Not an easy task!

app/views/playlists/_tag_form.html.erb
Dananji: Yes 👍 Good catch!

Dananji: I checked both files, and it seems both files didn't need the extra tag parsing and setting. The `onChange` function sets these values as expected. So I removed the extra handling for both files.

Dananji:
Additionally I added `style="display: none;"` to the `<select />` element in the `timelines/_tag_form.html.erb` to stop it from flashing the un-styled `<select />` element with its pre-populated options before `tom-select` initializes.

cjcolvar (approved):

PR #6621, opened 2025-10-15
Use a solr query tailored for speed instead of Model.where.first when validating uniqueness

cjcolvar:
I'm not certain but I think Model.where.first could possibly pull back all records from Fedora for the given model in certain cases.  At best it still fetches a record from Fedora after making the solr query which isn't necessary for this validator.


masaball (approved):
Looks good to me

Issue #6057, opened 2024-10-02
Collections Inherit Unit Staff Roles


joncameron:
# Description

PR #6626, opened 2025-11-05
Inherit roles from units

cjcolvar:
#6057

### DIFF
--- a/spec/models/unit_spec.rb
+++ b/spec/models/unit_spec.rb
@@ -525,4 +525,36 @@ describe Admin::Unit do
       expect(ReindexJob).to have_been_enqueued.with(@unit.collection_ids)
     end
   end
+
+  describe ".autocomplete" do
+    let!(:unit) { FactoryBot.create(:unit, name: 'test unit') }
+
+    it "returns an array of hashes" do
+      resp = Admin::Unit.autocomplete("test")
+      expect(resp).to be_a Array
+      expect(resp.first).to match(a_hash_including(id: unit.id))
+      expect(resp.first).to match(a_hash_including(display: 'test unit'))
+    end
+
+    it "returns partial matches" do
+      resp = Admin::Unit.autocomplete("te")
+      expect(resp).to be_a Array
+      expect(resp.first).to match(a_hash_including(id: unit.id))
+      expect(resp.first).to match(a_hash_including(display: 'test unit'))
+    end
+
+    it "is not case sensitive" do
+      resp = Admin::Unit.autocomplete("Te")
+      expect(resp).to be_a Array
+      expect(resp.first).to match(a_hash_including(id: unit.id))
+      expect(resp.first).to match(a_hash_including(display: 'test unit'))
+    end
+
+    it 'successfully returns when there is whitespace' do
+      resp = Admin::Unit.autocomplete("Test un")
+      expect(resp).to be_a Array
+      expect(resp.first).to match(a_hash_including(id: unit.id))
+      expect(resp.first).to match(a_hash_including(display: 'test unit'))
+    end
+  end
 end
--- a/app/assets/stylesheets/avalon.scss
+++ b/app/assets/stylesheets/avalon.scss
@@ -782,7 +782,14 @@ h5.card-title {
 
 .autocomplete_popup {
   z-index: 100;
-  width: 100%;
+
+  &#admin_collection\[unit_name\]-popup {
+    white-space: nowrap;
+  }
+
+  &:not(#admin_collection\[unit_name\]-popup) {
+    width: 100%;
+  }
 }
 
 #blacklight-modal {
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -103,7 +103,8 @@ class Admin::CollectionsController < ApplicationController
 
   # POST /collections
   def create
-    @collection = Admin::Collection.create(collection_params.merge(managers: [current_user.user_key], governing_policy_id: collection_params[:unit_id]))
+    unit_id = collection_params['unit_id'].presence || convert_unit(collection_params["unit_name"])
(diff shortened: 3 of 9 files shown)


app/models/concerns/admin_unit_behavior.rb
```diff
@@ -22,11 +22,15 @@ def managers
     edit_users & collection_managers.to_a
   end
 
+  def unit_admins_and_managers
+    unit_admins + managers
+  end
+
   def editors
     edit_users - unit_admins - managers
```
masaball, line 30:
```suggestion
    edit_users - unit_admins_and_managers
```

Since we made the convenience method, should we go ahead and use it here?

app/models/concerns/admin_collection_behavior.rb
```diff
@@ -18,15 +18,27 @@ def managers
     edit_users & collection_managers.to_a
   end
 
+  def inherited_managers
+    unit.unit_admins_and_managers
+  end
```
masaball, line 23: We could maybe test these convenience methods, though that's probably redundant since I think the base methods are tested in the Admin::Unit suite...

lib/hydra/multiple_policy_aware_ability.rb
```diff
@@ -24,9 +24,9 @@ module Hydra::MultiplePolicyAwareAbility
   def policy_ids_for(object_id)
     policy_ids = policy_id_cache[object_id]
     return policy_ids if policy_ids
-    solr_results = ActiveFedora::Base.search_with_conditions({id: object_id}, fl: governed_by_solr_field)
+    solr_results = ActiveFedora::SolrService.query("id:#{object_id} OR _query_:\"{!join to=id from=isGovernedBy_ssim}id:#{object_id}\"", fl: [governed_by_solr_field])
```
masaball: If possible, maybe we add tests for this change and the change in `multiple_policy_aware_access_controls_enforcement`. Since the queries are being made more complex, it would be good to have something in place to verify the behavior.

cjcolvar: I think the ability tests I added in unit, collection, and media object test this change sufficiently albeit indirectly.

spec/models/collection_spec.rb
```diff
@@ -16,103 +16,159 @@
...
-      it{ is_expected.to be_able_to(:update_managers, collection) }
-      it{ is_expected.to be_able_to(:update_editors, collection) }
-      it{ is_expected.to be_able_to(:update_depositors, collection) }
-      it{ is_expected.to be_able_to(:destroy, collection) }
-      it{ is_expected.to be_able_to(:update_access_control, collection) }
+      let(:ability) { Ability.new(user) }
+      let(:user) { User.where(Devise.authentication_keys.first => collection.managers.first).first }
+
+      it "can perform all actions on an existing collection" do
+        expect(ability).not_to be_able_to(:create, Admin::Collection)
```
masaball, line 38: Shouldn't collection manager be able to create Admin::Collection? Or is that an admin only thing?

If they should be able to, this test passing would indicate that either our Ability set up is wrong or the be_able_to check is somehow broken/insufficient.

If it is an admin only thing, maybe the `it` statement should be adjusted or the creation check should be its own separate test block? At a glance I see "can perform all actions" and then the first test is checking that the manager can't perform an action. Only after a moment to think through "existing collection excludes creation" does it make sense.

spec/models/collection_spec.rb
```diff
@@ -188,21 +244,21 @@
   end
 
   describe "managers" do
-    let!(:user) {FactoryBot.create(:manager)}
-    let!(:collection) {Admin::Collection.new}
+    let!(:user) { FactoryBot.create(:manager) }
+    let!(:collection) { Admin::Collection.new(unit: FactoryBot.create(:unit)) }
```
masaball, line 248: Is the `unit: FactoryBot.create(:unit)` arg necessary? The Admin::Collection factory should be providing that already and if it isn't, we should fix that factory. 

This applies for the editor and depositor test blocks as well.

spec/models/unit_spec.rb
masaball, line 1: I'm not seeing a test block for managers. Is that something we need to add?

spec/presenters/admin_collection_presenter_spec.rb
```diff
@@ -18,73 +18,83 @@
...
 
   describe 'abilities' do
-    subject(:ability) { Ability.new(user, {}) }
+    subject(:ability) { Ability.new(User.where(Devise.authentication_keys.first => user).first) }
 
     context 'when manager' do
       let(:user) { manager }
 
-      it { is_expected.to be_able_to(:destroy, presenter) }
+      it { is_expected.not_to be_able_to(:destroy, presenter) }
```
masaball:
```suggestion
      it { is_expected.to be_able_to(:destroy, presenter) }
```

I think this got changed accidentally. Though the fact it is passing while the inherited from unit test with the opposite check is also passing indicates there is something wrong with the ability definitions or that the `be_able_to` check or negation of that check is not functioning as expected.

masaball (changes requested):

spec/models/collection_spec.rb
masaball: Wait, I missed this isn't a FactoryBot call for the Admin::Collection creation. Ignore.

spec/models/unit_spec.rb
cjcolvar: Good catch!  I've added it and made a change to the ability.  I wonder if all of these are correct but I think we can revisit this in a future PR.

spec/models/collection_spec.rb
cjcolvar: Based on the conversation with Jon C. I think that only unit admins can create collections but maybe I'm wrong about that.

I'll add a little more to the test description to help clarify.

spec/presenters/admin_collection_presenter_spec.rb
cjcolvar: Good catch!  I had an issue with the test setup and shouldn't have changed this test expectation.  This should be resolved now.

PR #6631, opened 2025-11-10
Use autocomplete for unit selection on collection new/edit

masaball:
This commit makes use of the auto-complete-element 'data-autocomplete-value' feature which allows a separate input value to be assigned to a given display value. The issue is that when a selection is made, the input is changed to the input value. So a user inputting 'Defaul' then selecting 'Default Unit' will see the input field change to the unit id. This could be confusing for users, but might be alright?

### DIFF
--- a/app/models/admin/unit.rb
+++ b/app/models/admin/unit.rb
@@ -247,7 +247,7 @@ class Admin::Unit < ActiveFedora::Base
     # We can get around this by using format.
     filter = format('has_model_ssim: "%s"', "Admin::Unit")
 
-    search_array = ActiveFedora::SolrService.query(solr_query, rows: 1000, fq: filter)
+    search_array = ActiveFedora::SolrService.query(solr_query, rows: 1000, fq: filter, fl: [:id, :name_ssi])
 
     search_array.map { |value| { id: value[:id], display: value[:name_ssi] } }
   end


app/models/admin/unit.rb
```diff
@@ -238,6 +238,20 @@ def as_json(_options = {})
...
+  def self.autocomplete(query, _id = nil)
+    # Search name_uniq_si for case insensitivity since that field is downcased
+    # name_uniq_si has no whitespace, so remove whitespace from the query
+    solr_query = "name_uniq_si: (#{query.downcase.gsub(/\s+/, '')}*)"
+    # To take advantage of solr automagically escaping characters the query has to be in single quotes.
+    # This runs counter to ruby's string interpolation which requires the string to be in double quotes.
+    # We can get around this by using format.
+    filter = format('has_model_ssim: "%s"', "Admin::Unit")
+
+    search_array = ActiveFedora::SolrService.query(solr_query, rows: 1000, fq: filter)
```
cjcolvar: You can scope this to just the two fields needed for a tiny bit of saved bandwidth and processing.
```suggestion
    search_array = ActiveFedora::SolrService.query(solr_query, rows: 1000, fq: filter, fl: [:id, :name_ssi])
```

cjcolvar (approved):

PR #6637, opened 2025-11-12
Initialize tags in playlist/timelines edit forms

Dananji:
This reverts a buggy code change made in https://github.com/avalonmediasystem/avalon/pull/6619.

The absence of this code breaks the playlist/timeline details show page whenever a user edits relevant details and saves the edit form without interacting with the tag-select form field. Because with this particular workflow the hidden input that passes current tags to the back-end remains empty at the time of the form submission. And this default value is an empty string, which breaks the array-related methods associated with `playlist.tags` values.

Related issue: #6564


masaball (approved):
👍

PR #6644, opened 2025-11-20
Collections inherit special access from parent Unit

masaball:


cjcolvar (approved):
This looks good as far as I can tell from a quick read through.  I think it's fine to merge this in so we can get closer to testing it out in its entirety on avalon-dev.

Issue #6649, opened 2025-12-03
MasterFilesController#create takes linearly more time the larger the file


cjcolvar:
**Describe the bug**
Ingesting a large file from the dropbox takes a really long time to start processing.  It appears that files might take linearly more time the larger they are.  A couple quick tests had a 300MB audio file take 90 seconds to start processing in MediaConvert after returning a 504 response from timing out.  A 2.5G file took 5 minutes and a user reported that a 5.6G file took 12 minutes in their AWS instance.  From the logs it appears that some part of the controller logic around the first or second save of the MediaObject is the source of the slowdown.  This might be common to dropbox and direct upload files as well as AWS and non-AWS installs.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to Manage Files edit step for an item (e.g. https://demo.avalonmediasystem.org/media_objects/g158bh506/edit?step=file-upload)
2. Click on 'Open Dropbox'
3. Select a large file (1G+)
4. Click 'Submit'
5. Notice 504 after waiting 60 seconds
6. Notice encoding job takes minutes before appearing the MediaConvert dashboard

**Expected behavior**
Request should complete within a few seconds and transcoding should start shortly after.

**Environment (please complete the following information):**
- [ ] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [X] demo

PR #6661, opened 2025-12-09
Only request magic bytes for mimetype verification

masaball:
Related issue: #6649

We analyze input files on upload to determine if they are audio or video and to grab some of the metadata to fill out the record. This process involves checking the file mimetype, which was using our `FileLocator.reader` method. In S3 environments this would download the entire file, resulting in linearly increasing upload times based on file size. By requesting just the first ~16 bytes of the file for this case, we should be able to reduce upload time of multi-gig files from several minutes down to less than 10 seconds.

This also memoizes the content type method so that the download should hopefully only happen once per instance.

### DIFF
--- a/app/services/file_locator.rb
+++ b/app/services/file_locator.rb
@@ -141,13 +141,15 @@ class FileLocator
   end
 
   def magic_bytes
+    # Magic bytes for relevant file types should max out at 16, so only request
+    # that range when checking mimetype.
     case uri.scheme
     when 's3'
-      S3File.new(uri).object.get(range: '0-16').body
+      S3File.new(uri).object.get(range: 'bytes=0-15').body
     when 'file'
-      File.open(location, 'r')
+      File.read(location, 16)
     else
-      open_uri
+      open_uri.read(16)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/services/file_locator.rb
```diff
@@ -134,7 +134,18 @@ def reader
...
+  end
+
+  def magic_bytes
+    case uri.scheme
+    when 's3'
+      S3File.new(uri).object.get(range: '0-16').body
+    when 'file'
+      File.open(location, 'r')
     else
       open_uri
```
cjcolvar: I think it would be good if these other cases also returned the first 16 bytes too.  

```suggestion
      File.read(location, 16)
    else
      open_uri.read(16)
```

FWIW some quick testing locally shows a very slight speedup for `'file'` when working with large files and fetching just the first 16 bytes (~1ms with a 2.5GB file).  I didn't see any consistent difference with `open_uri.read(16)` and https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/1080/Big_Buck_Bunny_1080_10s_30MB.mp4.

cjcolvar (commented):
Great find!  I love the `#magic_bytes` method!  Could you please add some unit tests for it?

I'm really looking forward to getting this in place in production and seeing how it helps reduce those super long `MasterFilesController#create` requests!

cjcolvar (approved):
Looks great!

PR #6662, opened 2025-12-10
Media objects inherit from collection and unit

masaball:


config/locales/en.yml
```diff
@@ -200,7 +200,8 @@ en:
       Lending Period is the length of time that an item may be checked out for.
       Day and Hour values must be positive integers.
     apply_to_all_warning: "This action will affect %s items. Do you wish to proceed?"
-    inherited_role: 'This role is inherited from Unit "%{unit_name}" and cannot be removed.'
+    collection_inherited_role: 'This role is inherited from Unit "%{unit_name}" and cannot be removed.'
+    media_object_inherited_role: 'This role is inherited from Collection "%{collection_name}"'
```
cjcolvar: What do you think about renaming these to reflect where they are inheriting from rather than where they are inherited into?
```suggestion
    unit_inherited_role: 'This role is inherited from Unit "%{unit_name}" and cannot be removed.'
    collection_inherited_role: 'This role is inherited from Collection "%{collection_name}"'
```

Issue #6286, opened 2025-04-30
Add Units to 'Manage Content' Admin Page


joncameron:
## Description
With Units as objects, CRUD actitivies should be possible within Avalon much as Collections are now. To this end the "Manage Content" page can become a more generic page where users can view Units and Collections on which they have a staff role. 

## Done Looks Like
- Manage content route changed from https://avalon-dev.dlib.indiana.edu/admin/collections to https://avalon-dev.dlib.indiana.edu/[placeholder]
- Manage page includes Units and Collections
- Units layout copied from Collections (tabular listing, hyperlinks to administrative pages for each, and buttons for create and delete.
- A new route name to be decided on and added instead of "admin": "/manage", etc.

## Quick Mockup 
```https://avalon-dev.dlib.indiana.edu/admin```
<img alt="Image" src="https://github.com/user-attachments/assets/0b998c97-afe8-4764-96d4-7897f50fb736" />

PR #6669, opened 2025-12-15
Add an /admin/dashboard page

masaball:
Related issue: #6286

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -22,14 +22,7 @@ class Admin::DashboardController < ApplicationController
     # Allow the number of collections to be greater than 100
     blacklight_config.max_per_page = 100_000
     builder = ::CollectionSearchBuilder.new([:add_access_controls_to_solr_params_if_not_admin, :only_wanted_models, :add_paging_to_solr], self).rows(100_000)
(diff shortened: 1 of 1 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  before_action :load_and_authorize_units, only: [:index]
+
+  def load_and_authorize_collections
+    authorize!(params[:action].to_sym, Admin::Collection)
+    repository = CatalogController.new.blacklight_config.repository
+    # Allow the number of collections to be greater than 100
+    blacklight_config.max_per_page = 100_000
+    builder = ::CollectionSearchBuilder.new([:add_access_controls_to_solr_params_if_not_admin, :only_wanted_models, :add_paging_to_solr], self).rows(100_000)
+    if params[:user].present? && can?(:manage, Admin::Collection)
+      user = User.find_by_username_or_email(params[:user])
```
cjcolvar: I forget what the use case of `params[:user]` was.  It might be only useful in API requests and if so then it probably isn't needed here (and only in the original Admin::UnitsController and Admin::CollectionsController).

cjcolvar (commented):
This looks great!

cjcolvar (approved):

PR #6670, opened 2025-12-16
Override ActiveStorage to bypass forcing attachment disposition

cjcolvar:
#6448 changed the content disposition for downloads from ActiveStorage from the Avalon side but ActiveStorage has an internal mime type check which forces file types not in an allowlist to "attachment" disposition.  ~This might have started happening when we upgraded rails or wasn't enforced in development mode.~  This seems to only be enforced in production and why we missed this before.  This PR overrides ActiveStorage to bypass the internal check and allow inline for all mime types.

### DIFF
--- a/config/initializers/active_storage.rb
+++ b/config/initializers/active_storage.rb
@@ -1,2 +1,19 @@
 # This is set in config/application.rb but it seems like the settings aren't loaded at that point so doing it again here
 Rails.application.config.active_storage.service = (Settings&.active_storage&.service.presence || "local").to_sym
+
+Rails.application.config.to_prepare do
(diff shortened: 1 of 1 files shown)


masaball (approved):
The code looks good. For whatever reason, I'm a little leery of _every_ mimetype being allowed for inline but looking at [the defaults](https://guides.rubyonrails.org/configuring.html#config-active-storage-content-types-to-serve-as-binary) for `content_types_to_serve_as_binary` most of the entries are for various XML formats which we would want to serve inline. So seems like changing the configs around effectively amounts to the same end result as this override. Lets go ahead with this approach :+1:

Issue #6663, opened 2025-12-10
API requests should respond 422 when id param doesn't match type of endpoint


cjcolvar:
Sometimes users make requests using an id for an object of a type that doesn't match the route.  Currently Avalon will attempt to fetch the object with an ActiveFedora finder method specific to the model class matching the route which will raise an error when the requested type and found type do not match. This error is not caught leading to a generic 500 response.  A more specific status code of 422 would help differentiate this situation from a case when the server has an unexpected error.

For example, if there is a media object with id abc1234 then `/admin/collections/<id>.json` will respond with status code 500 and response body:
```
{"status":500,"error":"Internal Server Error"}
```

For non-API requests, should we redirect the user to the appropriate route for the id or the home page and display an error flash message like we do for other errors?

This came up in local Time Based Media Systems workflow discussions.

## Done Looks Like
- [x] API requests receive a 422 response for wrong types in request URL
- [x] Non-API requests display the flash message and sends the user to the homepage, rather than a 500

joncameron:
Created #6726 for the cases Chris mentioned in this comment.

PR #6671, opened 2025-12-16
Return more specific error for model mismatch

masaball:
Related Issue: #6663


app/controllers/application_controller.rb
```diff
@@ -190,6 +190,14 @@ def current_ability
     end
   end
 
+  rescue_from ActiveFedora::ModelMismatch do |exception|
+    if request.format == :json
+      render json: { errors: ["Requested resource type does not match type of #{params[:id]}"] }, status: 422
+    else
+      render '/errors/model_mismatch', status: 422
```
cjcolvar: I think we decided that it would be better to redirect to the homepage with a flash message instead of a custom error page.  Sorry that that didn't make it into the ticket description.

cjcolvar (approved):
Looks great!

PR #6673, opened 2025-12-18
Use delete_all because it is more performant

cjcolvar:
We don't need to worry about running callbacks and anyways Blacklight uses delete_all in it's implementation.

This is a follow up to #6660


masaball (approved):
:+1:

PR #6679, opened 2026-01-09
Clean out environment before running asset compilation

cjcolvar:
Avalon docker image building may include local environment variables which has the potential of exposing them by 3rd party JS libraries due to an issue in shakapacker <9.5.0.  This change clears out the environment before asset compilation as a precaution.  This workaround will be unnecessary once we upgrade to shakapacker 9.5.0.  See https://github.iu.edu/advisories/GHSA-96qw-h329-v5rg


Dananji (approved):
🎉

Issue #6612, opened 2025-10-09
Make MediaConvert the default transcoder for AWS in Avalon 8.2


cjcolvar:
## Description
When Avalon 8.2 gets released ElasticTranscoder will be EOL. The changes in avalon, avalon-docker, and avalon-terraform were made to be backwards compatible and thus left ElasticTranscoder as the default.  We should make MediaConvert the default and remove ElasticTranscoder from the options and code.

See #6117 

## Done Looks Like
- [x] Remove references to ElasticTranscoder
- [x] Set MediaConvert as default for avalon, avalon-docker, and avalon-terraform

PR #6681, opened 2026-01-13
Remove Elastic Transcoder related settings and code

masaball:
Related Issue: #6612

AWS has sunset Elastic Transcoder so we need to be fully on Media Convert in our next release. For this PR, we are removing all elastic transcoder specific code changing references over to media convert where it makes sense.

Potential follow-up PR: Do we need to establish any rake tasks related to media convert?

### DIFF
--- a/spec/helpers/upload_form_helper_spec.rb
+++ b/spec/helpers/upload_form_helper_spec.rb
@@ -23,7 +23,7 @@ describe UploadFormHelper, type: :helper do
 
     context 'with media convert' do
       let(:transcoder) { :media_convert }
-      it 'returns true if using Elastic Transcoder' do
+      it 'returns true if using MediaConvert' do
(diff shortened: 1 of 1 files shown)


spec/helpers/upload_form_helper_spec.rb
```diff
@@ -21,8 +21,8 @@
       allow(Settings.encoding).to receive(:engine_adapter).and_return(transcoder)
     end
 
-    context 'with elastic transcoder' do
-      let(:transcoder) { :elastic_transcoder }
+    context 'with media convert' do
+      let(:transcoder) { :media_convert }
       it 'returns true if using Elastic Transcoder' do
```
cjcolvar:
```suggestion
      it 'returns true if using MediaConvert' do
```

Issue #6675, opened 2026-01-07
Issues with Collection permissions


masaball:
## Known Issues
* Unit Admin and Inherited Collection Manager cannot create collection due to missing unit on new collection
    * Can we setup Avalon to automatically create a "Default Unit" on application creation to initialize collections with?
* Unit Admin & Inherited Collection Manager do not have permissions to move items
* Editors can edit collection information. Is that broken? Permissions table says that shouldn't be happening.
* Manage Content page does not display the parent unit when a user is assigned directly to a collection
* Create Collection button not appearing on Manage Content page for manager assigned directly at collection level
* Editor and Depositor assigned at collection level can move items. Permissions table says that shouldn't be allowed.

joncameron:
Ability model might be finicky; will need to make sure tests align with the new permissions table.

PR #6694, opened 2026-01-21
Update permissions for new unit hierarchy

masaball:
Related issue: #6675

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -233,7 +233,7 @@ class Ability
 
       cannot :update, [Admin::Collection, SpeedyAF::Proxy::Admin::Collection] do |collection|
         # Editors and Depositors should not be able to edit collection level information
-        !is_manager_of?(collection)
+        (not (full_login? || is_api_request?)) || !is_manager_of?(collection)
       end
 
       cannot :destroy, [Admin::Collection, SpeedyAF::Proxy::Admin::Collection] do |collection, other_user_collections=[]|
@@ -241,11 +241,11 @@ class Ability
       end
 
       cannot :update, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
-        !is_admin_of?(unit)
+        (not (full_login? || is_api_request?)) || !is_admin_of?(unit)
       end
 
       cannot :destroy, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
-        (not (full_login? || is_api_request?)) || !@user.in?(unit.unit_admins)
+        (not (full_login? || is_api_request?)) || !is_admin_of?(unit)
       end
 
       can :intercom_push, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|


app/controllers/application_controller.rb
```diff
@@ -139,10 +139,14 @@ def get_user_collections(user = nil)
       if user.blank?
         SpeedyAF::Proxy::Admin::Collection.where("has_model_ssim:Admin\\:\\:Collection").to_a
       else
-        SpeedyAF::Proxy::Admin::Collection.where("has_model_ssim:Admin\\:\\:Collection AND inheritable_edit_access_person_ssim:#{user}").to_a
+        unit_collections = SpeedyAF::Proxy::Admin::Unit.where("has_model_ssim:Admin\\:\\:Unit AND inheritable_edit_access_person_ssim:#{user}").to_a.map(&:collections).flatten
+        collections = unit_collections + SpeedyAF::Proxy::Admin::Collection.where("has_model_ssim:Admin\\:\\:Collection AND inheritable_edit_access_person_ssim:#{user}").to_a
+        collections.uniq
```
cjcolvar: Is there a more efficient way to do this by combining it into one solr query (might need a join)?  I'm thinking that for a user that has edit access to many units could spawn many solr queries.  It looks like before this was 2 solr queries but now will be n + 2 for n units returned by the first query given `unit_proxy.collections` does a solr query.

app/models/ability.rb
```diff
@@ -348,7 +360,7 @@ def is_admin_of?(unit)
   def is_member_of?(collection)
      is_administrator? ||
        @user.in?(collection.managers, collection.editors, collection.depositors) ||
-       @user.in?(collection.unit.unit_admins, collection.unit.managers, collection.unit.editors, collection.unit.depositors)
+       @user.in?(collection.inherited_managers, collection.inherited_editors, collection.inherited_depositors)
```
cjcolvar, line 367: Nice!  This is much better!

app/models/ability.rb
```diff
@@ -155,19 +163,19 @@ def custom_permissions(user=nil, session=nil)
         end
 
         can :update, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
-          is_editor_of_unit?(unit)
+          is_admin_of?(unit)
```
cjcolvar, line 166: I'm surprised tests don't need to be updated with this change and the others on unit.  Maybe we're missing some ability tests for units?

app/models/ability.rb
```diff
@@ -219,15 +227,23 @@ def custom_permissions(user=nil, session=nil)
...
+        !is_manager_of?(collection)
       end
 
       cannot :destroy, [Admin::Collection, SpeedyAF::Proxy::Admin::Collection] do |collection, other_user_collections=[]|
         (not (full_login? || is_api_request?)) || !is_manager_of?(collection)
       end
 
+      cannot :update, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
+        !is_admin_of?(unit)
+      end
```
cjcolvar, line 245: Is this `cannot` necessary?  I don't see a more permissive `can :update Admin::Unit` above that needs restricting here.  Maybe this `cannot` should have the full login / non-api restrictions like destroy?

masaball: It oddly does seem to be necessary. Even though the `can :update, Admin::Unit` block should be limiting things to just users `in_admin_of?(unit)`, I could still perform unit admin only actions as an editor. I'm not sure where the issue could be arising from, but I saw there was a similar `cannot :update, Admin::Collection` around line 234, so copied it for the units. Perhaps both entries should have the full login checks?

cjcolvar: Looking at this again, I think the `can :update` might be coming from `edit_permissions` (upstream and the override on line 30) because managers and editors are included in `edit_users` and thus get `:update` by default.  I think the `cannot :update` and `:destroy` `Admin::Unit`/`Admin::Collection` make sense now but that the additional full login check should be added to them.

app/models/ability.rb
```diff
@@ -233,19 +233,19 @@ def custom_permissions(user=nil, session=nil)
...
       end
 
       cannot :update, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
-        !is_admin_of?(unit)
+        (not (full_login? || is_api_request?)) || !is_admin_of?(unit)
       end
 
       cannot :destroy, [Admin::Unit, SpeedyAF::Proxy::Admin::Unit] do |unit|
-        (not (full_login? || is_api_request?)) || !@user.in?(unit.unit_admins)
+        (not (full_login? || is_api_request?)) || !is_admin_of?(unit)
```
cjcolvar, line 248: Thanks for cleaning this up too!

cjcolvar (approved):
Looks great! 🎉

Issue #6654, opened 2025-12-04
Setup background cron job to cleanup searches table


cjcolvar:
## Description
The searches DB table can fill up a DB with lots of rows of data that doesn't get used anywhere in avalon.  Blacklight provides a rake task to clean up this table.  Instead of creating a cron job to call the rake task Avalon could provide a background job that runs regularly with sidekiq-cron like similar jobs for cleaning up sessions or stream tokens.

## Done Looks Like
- [ ] Rake task added for routine cleanup of search DB table

PR #6695, opened 2026-01-22
Avoid saving searches entirely; remove unneeded searches clean up job

cjcolvar:
Related to #6654 

~This uses a Blacklight configuration to mark all searches as crawler agents and thus none should be saved.~
~FIXME: Unit test is passing but I'm still seeing Search objects get persisted to the DB in local manual testing.~

2nd pass: override Blacklight method entirely because configuration was only applying to cases of non-logged in searches but we don't need searches saved for any requests.


masaball (approved):
👍

Issue #6349, opened 2025-06-06
[BUG] “Clear Selected Items” button removes all bookmarks instead of only the selected ones


abankhele:
**Describe the bug**
On the Bookmarks page, clicking Clear Selected Items should remove only the items that are currently checked. Instead, it unchecks all items, then clears every bookmark, even those that were not selected.

**To Reproduce**
Steps to reproduce the behavior, including the results:
Go to:

1. https://avalon-dev.dlib.indiana.edu/catalog?q=&search_field=all_fields&utf8=%E2%9C%93
2. In the catalog results, select the checkboxes for any five items.
3. Click Selected Items in the navigation bar to view your bookmarks. You should see those five items listed.
4. On the Bookmarks page, click Select All—this action unchecks all five bookmarks.
5. Check only two of the bookmarks.
6. Click Clear Selected Items.
7. Observe that the Bookmarks page is now empty: all five bookmarks have been removed instead of just the two that were checked.

**Expected behavior**
When Clear Selected Items is clicked, only the items with checkboxes checked should be removed from Bookmarks. Unchecked items should remain bookmarked and still appear after the action.

**Screenshots/Screen Recording**

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFZNQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6e7168e73524bb9a702fd9c329771ce90969458e/Screen%20Recording%202025-06-06%20at%2012.54.40%E2%80%AFPM.mov)

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser : Chrome
- [ ] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**
Is this issue associated with a specific release? [e. g. 7.6]

**Additional notes**
Any additional information or notes that may be helpful in diagnosing or fixing the problem.

Dananji:
Examples from other implementers.

IUCAT
<img width="939" height="1084" alt="Image" src="https://github.com/user-attachments/assets/6a50ad27-322d-478d-ae6c-d8bfe72ff026" />

Cornell
<img width="1140" height="1020" alt="Image" src="https://github.com/user-attachments/assets/88095123-01b7-48d8-a33e-fe99e9853b9f" />
[...]

PR #6702, opened 2026-01-30
Clear selected items

Dananji:
Related issue: #6349 

This PR adds new checkboxes to the selected items page with the ability to clear some of the selected items on the page as well as the functionality to clear all items and hides the in-built select checkboxes.

Screenshot 2026-01-30 at 3 13 11 PM

How it works;
- On page load all the items are selected (checked) and the user has the ability to select/deselect all using the 'Select All' checkbox.
- Once all items are unchecked, the bulk action buttons and the new 'Clear selected items' button are disabled
- If the user selects a subset of the items, the bulk action buttons are disabled to block a bulk action on a subset of the items on the page

Screenshot 2026-02-03 at 9 48 20 AM

### DIFF
--- a/app/assets/javascripts/bookmark_selection.js
+++ b/app/assets/javascripts/bookmark_selection.js
@@ -20,8 +20,8 @@
 class BookmarkSelectionManager {
   constructor() {
     this.selectedIds = new Set();
-    this.initializeEventListeners();
     this.selectAllOnLoad();
+    this.initializeEventListeners();
     this.updateUI();
   }
 
@@ -54,11 +54,6 @@ class BookmarkSelectionManager {
     // Form submission for selected items
     document.querySelectorAll('form[data-requires-selection]').forEach(form => {
       form.addEventListener('submit', (e) => this.handleFormSubmit(e));
-    });
-
-    // Check for if all items on page are selected
-    document.querySelectorAll('a[data-requires-selection], button[data-requires-selection], .bulk-actions[data-requires-selection] a').forEach(el => {
-      el.addEventListener('click', (e) => this.handleActionClick(e));
     });
   }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/bookmark_selection.js
```diff
@@ -0,0 +1,184 @@
...
+*/
+
+/**
+ * Manages selection state for bookmarks page
+ */
+class BookmarkSelectionManager {
+  constructor() {
+    this.selectedIds = new Set();
+    this.initializeEventListeners();
+    this.selectAllOnLoad();
```
cjcolvar: Would it make sense to switch these two?  Ff setting `cb.checked = true` in `selectAllOnLoad()` triggers the change event listener then I think it would be more efficient to initialize the state before adding the listeners.

app/assets/javascripts/bookmark_selection.js
```diff
@@ -0,0 +1,184 @@
...
+  handleActionClick(e) {
+    // Allow 'Clear selected items' button to work on partial selection
+    if (e.target.dataset.testid === 'remove-selected-btn') {
+      return;
+    }
+
+    const totalItemsCount = document.querySelectorAll('.bookmark-selection').length;
+    // If not all items are selected, show alert and prevent action
+    if (totalItemsCount > this.selectedIds.size) {
+      window.alert("Please use 'Clear selected items' to remove deselected items before performing actions on a subset.");
```
cjcolvar: Instead of an alert what do you think about enabling/disabling the bulk action buttons in `updateUI()`?

app/views/bookmarks/_document_list.html.erb
cjcolvar, line 1: What are the changes in this file from upstream?  I'm wondering if it is possible to override a view component instead.  If that isn't possible then can you please add a comment at the top of the file listing the changes?

cjcolvar (commented):
When I ran this locally it didn't seem to be working.  I'm still seeing the old selected checkbox on the right, the select all isn't working, and the clear selected items button doesn't seem to be enabling.  It could be that I have old cached JS or the JS isn't fully loading.

app/assets/javascripts/bookmark_selection.js
Dananji: Yes, that's much cleaner. I will change this to block buttons.

app/views/bookmarks/_document_list.html.erb
Dananji: This adds a checkbox to the left of each selected item, which has its own event listener in the JS module. The difference between this checkbox and the native Blacklight checkbox is that the custom one doesn't automatically update the bookmark count in the nav-bar upon check/uncheck event, but waits for the user to submit the change via `Clear selected items` button to update the count.
Yes, will do 👍

Dananji: I wasn't clear from looking at upstream if there's a document list view component. So, I added a comment to this partial override.

Issue #6692, opened 2026-01-21
Managers can move items into the collection they are already in


masaball:
## Description
The "Move" modal on the Selected Items page has a dropdown list of collections that an item can be moved into. This lists all collections that the manager is assigned, including the collection the item is already in. Trying to perform the move action to the item's original collection does not seem to have any negative effects, but it is a potentially confusing user experience particularly in the case of a user only being manager of a single collection.

Should we disable the move button if a user is manager of a single collection? Should we exclude an item's current collection from the list of possible move options?

## Done Looks Like
- [ ] An item's current collection is removed from the set of possible move options presented

joncameron:
👍

PR #6703, opened 2026-02-03
Limit collection move dropdown list

masaball:
Related issue: #6692

Behavior changes:
* Managers of a single collection receive warning they do not have permissions to move items to another collection
* When a single item is selected, the collection it belongs to is removed from the dropdown list.

### DIFF
--- a/app/views/bookmarks/move.html.erb
+++ b/app/views/bookmarks/move.html.erb
@@ -24,7 +24,7 @@ Unless required by applicable law or agreed to in writing, software distributed
   <%= form_tag url_for(:controller => "bookmarks", :action => "move"), :id => 'move_form', :method => :post do %>
     <div class="modal-body">
       <div class="row-fluid">
-        <% if  @candidates.count <= 1 || @documents.empty? %>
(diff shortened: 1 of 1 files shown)


app/views/bookmarks/move.html.erb
```diff
@@ -16,12 +16,15 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+  <% if @documents.length == 1 %>
+    <% @candidates = @candidates.reject { |c| c.id == @documents.first["isMemberOfCollection_ssim"].first } %>
+  <% end %>
   <%= render partial: 'modal_header', locals: { label: 'move' } %>
 
   <%= form_tag url_for(:controller => "bookmarks", :action => "move"), :id => 'move_form', :method => :post do %>
     <div class="modal-body">
       <div class="row-fluid">
-        <% if  @candidates.count == 0 || @documents.empty? %>
+        <% if  @candidates.count <= 1 || @documents.empty? %>
```
cjcolvar: I think this test needs to be a little more nuanced given the candidate removal when there is only one document.

```suggestion
        <% if  (@documents.length == 1 && @candidates.empty?) || (@documents.length > 1 && @candidates.count <= 1) || @documents.empty? %>
```

Issue #6691, opened 2026-01-16
Support Forced Subtitles


joncameron:
## Description
Avalon doesn't currently support forced subtitles, which are sometimes provided as external files. The ability to designate a file as "forced" in the listing of captions on a masterfile should be available to meet this use case.

## Forced Subtitles
> Forced subtitles are common on movies and only provide subtitles when the characters speak a foreign or alien language, or a sign, flag, or other text in a scene is not translated in the localization and dubbing process.

_https://en.wikipedia.org/wiki/Subtitles#Categories_

> On our service, Forced Narrative subtitles are only displayed if Subtitles and CC are set to "off" in the user's playback settings. When the user activates a full Subtitle or SDH/CC file, the FN subtitle does not display and for this reason, we require that all Forced Narrative events are also included in each full Subtitle and SDH/CC file.

_https://partnerhelp.netflixstudios.com/hc/en-us/articles/217558918-Understanding-Forced-Narrative-Subtitles_

## Done Looks Like
- [x] "Forced" binary option (checkbox?)
- [x] Only one caption file can be set to "Forced" at a time
- [ ] When timed text display is disabled in the media player and a "Forced" subtitle track exists for the masterfile, the forced subtitle track is displayed

masaball:
The behavior attribute has a list of accepted values and none of them seem quite right for this case. However, metadata is able to be used at any level of a manifest and can contain arbitrary data. @joncameron What do you think of adding a metadata entry to the caption annotation to denote a forced subtitle?

PR #6711, opened 2026-02-06
Allow captions to be marked "forced"

masaball:
Related issue: #6691

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -18,7 +18,6 @@ class SupplementalFilesController < ApplicationController
 
   before_action :set_object
   before_action :authorize_object
-  after_action :deduplicate_forced, only: [:create, :update]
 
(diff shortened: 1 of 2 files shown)


app/controllers/supplemental_files_controller.rb
```diff
@@ -255,6 +258,17 @@ def update_tags
       end
     end
 
+    def deduplicate_forced
```
cjcolvar: Instead of removing the forced flag from any older files on create/update what do you think about adding a validation check (like `validate_params`) and returning an `Avalon::BadRequest` if the request attempts to set forced when there is already a forced file on the object?  (I would be tempted to use `validate_params` but I think this should have a specific error message different than `missing required parameters`.)

masaball: That would make sense to do. I was going for ease of use, but it would probably be better for it to be a "hey, are you sure about this?" type of check. I'll take a look at reworking it.

Issue #6289, opened 2025-05-02
Encrypt API tokens at rest in DB


elynema:
## Description
It would be nice to have encrypted API keys in the database. Will need to make sure that API keys in headers are passed in in a way that it could be compared to encrypted DB values.

## Done Looks Like
- [ ] Encrypt API tokens at rest in DB

joncameron:
When I tried to run some rake tasks for tokens on avalon-dev they errored out (snippet below). Both of these failed:
`bundle exec rake avalon:token:list`
`bundle exec rake avalon:token:generate username=joncameron email=joncamer@iu.edu`

@masaball is this related to this change or a separate thing?
```
rake aborted!
ActiveRecord::Encryption::Errors::Configuration: Missing Active Record encryption credential: active_record_encryption.deterministic_key (ActiveRecord::Encryption::Errors::Configuration)
/usr/local/bundle/gems/activerecord-8.0.2.1/lib/active_record/encryption/config.rb:44:in 'block (2 levels) in <class:Config>'
/usr/local/bundle/gems/activerecord-8.0.2.1/lib/active_record/encryption/scheme.rb:98:in 'ActiveRecord::Encryption::Scheme#deterministic_key_provider'
/usr/local/bundle/gems/activerecord-8.0.2.1/lib/active_record/encryption/scheme.rb:58:in 'ActiveRecord::Encryption::Scheme#key_provider'
...
```

masaball:
It is related to this change. Seems that the deploy to dev didn't set the ENV variables for the encryption credentials. I'll take a look and make sure that the configuration is all in line on dev.

masaball:
@joncameron I've added the credentials to avalon-dev so it should hopefully work now. Some actions may still error because of unencrypted data present in the database, but I would expect generating a token and using that new token to access things should work. We can investigate clearing the old tokens or changing settings to allow use of the unencrypted data if it gets troublesome.

joncameron:
Worked as expected with a newly issued token 👍

PR #6714, opened 2026-02-10
Encrypt user API tokens

masaball:
Related issue: #6289

See https://guides.rubyonrails.org/active_record_encryption.html. Implementers will need to run `bin/rails db:encryption:init` to generate the encryption keys and then they will need to add those keys to the ENV variables defined in `config/application.rb`.

This change breaks existing API tokens unless unencrpyed data support is enabled in the config. Existing API tokens should be removed and regenerated, if possible.


cjcolvar (approved):
This looks great!  Good job finding a standard rails way of doing this!

Issue #6121, opened 2024-11-15
Improve Process for Updating Ramp Builds in Avalon


joncameron:
# Description
Currently a new Ramp build is required to be pulled into Avalon via a PR. Is there a way we could automate this process, so that new code committed to Ramp automatically builds and is added to Avalon, or that code deploy to the Avalon repo could automatically build Ramp based on a git hash?

Dananji:
Using GitHub workflows created https://github.com/avalonmediasystem/avalon/pull/6725

Use `workflow_dispatch` to trigger a workflow in a non-default branch (`develop` branch) in Avalon from Ramp.
Using `workflow_dispatch` along with `ref` flag, allows us to create the PR off of a non-default branch.

The Git workflow in Ramp is as follows;

1. Build the latest code from `main`
2. If there's new code use `gh-pages` library to create/update the `ramp-build` branch with latest build files and `package.json` file. Using `gh-pages` we are able to only keep track of the build files.
3. Check if the build files have changed to proceed to the next steps. This prevents from invoking the rest of the workflow on non-code changes (README updates, docs updates, ... ) that don't need to be updated in Avalon.
4. If there's changes notify Avalon to create a PR to update the Ramp dependency via `workflow_dispatch` created with,
   - `ref` flag set to `develop` to target `develop` branch in Avalon
   -  commit hash variable for the Ramp commit as `github.event.inputs.ramp_commit`
   -  a PAT created with `repo` and `workflow` scopes in Ramp

A corresponding Git workflow is then created in Avalon's `develop` (target) branch. And a a dummy workflow is created in the `main` (default) branch to help the workflow API to register the workflow in the Git repo. Once this is registered by the `main` branch, the workflow from `develop` is able to use it to invoke the steps.

The Git workflow in Avalon's `develop` branch is as follows;

1. Check for existing open PR to update ramp dependency from `update-ramp-build` branch
2. If there's an open PR, then collect the commit hashes in the PR body
3. Update the existing PR's body by attaching the new commit to the end of the collected commit information from before and force-push the latest build
4. If there are no open PRs to update Ramp, create a new PR with the commit hash information in the PR body

PR #6723, opened 2026-02-19
Add dummy workflow file to trigger workflow API

Dananji:
Related issue: #6121 

This PR adds a placeholder for registering the `update-ramp.yml` Git workflow for the Git API to update Ramp dependency in Avalon when a new build is created in Ramp repository.


cjcolvar:
@Dananji I temporarily allowed bypassing the required status checks and then put the requirement back in place after rebase merging.  I'm going to see how hard it is to rebase develop on main.

Issue #6285, opened 2025-04-30
Disable Special Permissions Inheritance for Media Objects


joncameron:
## Description
With permission inheritance applied to items from membership at the collection and unit levels, there may be some restricted items in a general collection that collection staff do not want to be broadly accessible to members of that group. Toggling permissions inheritance (for special access users and groups, not staff) will allow specific items to be restricted even with broad higher-level permissions.

For example, a researcher has requested access to a collection. Collection staff want to add this researcher for access to the contents of the collection, but there may be restrictions on specific items within the broader collection that should be honored. This might be seen in an ATM collection, where a donor has requested a specific piece of media be restricted for legal or ethical reasons, but endorses the rest of the collection being open and available. The donor might still want to be able to stream the item—their guest account added to special access—but want to prevent viewing for any additional researchers added for view access across the entire collection.

joncameron:
Listing usernames and group names who have inherited access would also be helpful in some cases. That or a link back directory to the collection or unit admin page so item editors could see the listing directly.

joncameron:
Gated discovery, Solr queries, other checks will need to be accounted for.

abankhele:
Tested in avalon-dev. The “Disable parent permissions” checkbox works as expected. When the checkbox is selected, access is revoked for users listed under “Assign special access,” and when it is unchecked, their access is restored. I verified this by impersonating the affected users as well.

This observation is related to collections: a group added at the collection level in Avalon Groups is inherited by items in both Avalon Groups and External Groups.

joncameron:
Created https://github.com/avalonmediasystem/avalon/issues/6747 for the behavior Akanksha mentioned.

PR #6729, opened 2026-02-26
Allow individual media objects to have inherited permissions disabled

masaball:
Related issue: #6285

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -216,6 +216,10 @@ class Ability
       #   can :manage, Avalon::ControlledVocabulary
       # end
 
+      cannot :read, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
+        media_object.disable_inheritance? && is_inherited_from_parent?(media_object)
+      end
+
       cannot :update, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
         (not (full_login? || is_api_request?)) || (!is_member_of?(media_object.collection)) ||
           ( media_object.published? && !is_manager_of?(media_object.collection) )
@@ -388,6 +392,10 @@ class Ability
   def is_editor_of_unit?(unit)
     is_administrator? ||
       @user.in?(unit.editors_managers_and_unit_admins)
+  end
+
+  def is_inherited_from_parent?(media_object)
+    @user.in?(media_object.inherited_read_users) || !(@user_groups & media_object.inherited_read_groups).empty?
   end
 
   def is_member_of_any_collection?
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -217,7 +217,7 @@ class Ability
       # end
 
       cannot :read, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
-        media_object.disable_inheritance? && is_inherited_from_parent?(media_object)
+        media_object.disable_inheritance? && is_exclusively_inherited_from_parent?(media_object)
       end
 
       cannot :update, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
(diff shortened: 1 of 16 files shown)


app/views/modules/_special_access.html.erb
```diff
@@ -38,3 +46,9 @@ Unless required by applicable law or agreed to in writing, software distributed
           input_disabled: !can_update %>
   </div>
 </div>
+
+<script>
+  document.querySelector('#disable_inheritance').addEventListener('change', function() {
+    this.closest('form').submit();
```
cjcolvar: @joncameron 
Do we want the disable inheritance checkbox autosave or would an explicit save be better?

app/models/ability.rb
```diff
@@ -216,6 +216,10 @@ def custom_permissions(user=nil, session=nil)
       #   can :manage, Avalon::ControlledVocabulary
       # end
 
+      cannot :read, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
+        media_object.disable_inheritance? && is_inherited_from_parent?(media_object)
```
cjcolvar: I think the logic here might need to be more complex.  I'm thinking of the scenario where a user is in the `inherited_read_users`, inheritance is disabled, and then the user is added to `read_users` to narrow the special access from the parent collection/unit.  It looks like this case will block the user from reading the item when I think they probably should be able to.  Maybe we need a method like `is_exclusively_inherited_from_parent?` which does this check: `!@user.in?(media_object.read_users) && @user.in?(media_object.inherited_read_users)` (and the same for groups).  Does that make sense?

app/models/search_builder.rb
```diff
@@ -28,13 +28,17 @@ def only_wanted_models(solr_parameters)
...
-    [policy_clauses, 'workflow_published_sim:"Published"'].compact.join(" OR ")
+    [policy_clauses(permission_types: [:edit]), 'workflow_published_sim:"Published"'].compact.join(" OR ")
   end
 
   def limit_to_non_hidden_items(_permission_types = discovery_permissions, _ability = current_ability)
     [policy_clauses, "(*:* NOT hidden_bsi:true)"].compact.join(" OR ")
   end
 
+  def limit_to_inheritance_enabled_items(_permission_types = discovery_permissions, _ability = current_ability)
+    [policy_clauses(permission_types: [:edit]), '(*:* NOT disable_inheritance_bsi:true)'].compact.join(" OR ")
```
cjcolvar: It seems like this needs the same kind of nuance I'm suggesting in ability.  E.g. A user could not have edit access, the item could have inheritance disabled, AND have read access granted on the item.  As is it seems like this clause would block all non-edit users from items with inheritance disabled.

cjcolvar (changes requested):
This is looking really good but when reviewing it again I think I found that things need to be more complex than originally thought.

app/models/ability.rb
masaball: That makes sense and is a good catch. A lot of potential for edge cases in all this, so good to keep an eye out for potential ones and preempt them :+1:

app/views/modules/_special_access.html.erb
cjcolvar: @masaball When testing this out locally this is working but when I save the form the disable inheritance checkbox gets reverted to unchecked.  I know there is a test for when an update happens changing user/group/ip access but this seems to happen when visibility or hidden (or nothing) is changed.  Probably need a change to the access control step model.

cjcolvar (commented):
I did some manual testing and this mostly appears to be working as expected.  You've added a good number of tests to the media objects controller.  Do you think we need to add any to the catalog controller or do you think all of the search changes are covered?

I found some other issues but they aren't directly related to this PR so I'll put them in a separate issue.

masaball:
I think the issue with the save there is that the disable inheritance checkbox is in a separate form than the regular save button effects. What do we think about pulling the disable inheritance into its own card within the main permissions form? Something like: 
Screenshot_20260305_100743

And that would remove the autosave, it would be using the standard save workflow of the rest of the options.

cjcolvar:
@masaball That works for me.  
@joncameron what do you think?

joncameron:
Works for me too, let's go for that and tweak styling and placement later if needed.

cjcolvar (approved):
Thanks for working through all this!

masaball:
Thanks for your patience reviewing all of this!

Issue #5998, opened 2024-08-20
MasterFiles#show should handle deleted objects gracefully


cjcolvar:
# Description
When a user tries to navigate directly to a section that has been deleted via /master_files a generic error (500) page is shown.  SpeedyAF should handle the case of deleted objects (tombstones) by raising a not found/deleted error which Avalon can catch the same way it does for `ActiveFedora::NotFound` / `Ldp::Gone` ("Item Has Been Deleted" page). Persistent URLs and embeds might encounter this case.

Does this mean that we need to store tombstones in Solr or have a fallback check to Fedora when not found in Solr?

See https://app.honeybadger.io/projects/54117/faults/110470099 and https://app.honeybadger.io/projects/54117/faults/106930302

# Done Looks Like
- [x] Add fallback check to Fedora when not present in Solr
- [ ] Add tombstones to Solr if the above strategy turns out to not be as viable (performance reasons etc.)

joncameron:
👍

PR #6730, opened 2026-03-02
Redirect to fedora on missing MasterFile proxy

masaball:
Related issue: #5998

Can reexamine approach if this is found to be too non-performant in more real world environments.

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -377,9 +377,9 @@ protected
 
   def set_masterfile_proxy
     @master_file = SpeedyAF::Proxy::MasterFile.find(params[:id], load_reflections: true)
-    if params[:id].blank? || @master_file.nil?
-      flash[:notice] = "MasterFile #{params[:id]} does not exist"
-    end
+    flash[:notice] = "MasterFile #{params[:id]} does not exist" if params[:id].blank?
+    set_masterfile if @master_file.nil?
+    @master_file
   end
 
   # return deflated waveform content. deflate only if necessary
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -369,15 +369,13 @@ class MasterFilesController < ApplicationController
 
 protected
   def set_masterfile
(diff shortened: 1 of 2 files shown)


app/controllers/master_files_controller.rb
```diff
@@ -377,9 +377,9 @@ def set_masterfile
 
   def set_masterfile_proxy
     @master_file = SpeedyAF::Proxy::MasterFile.find(params[:id], load_reflections: true)
-    if params[:id].blank? || @master_file.nil?
-      flash[:notice] = "MasterFile #{params[:id]} does not exist"
-    end
+    flash[:notice] = "MasterFile #{params[:id]} does not exist" if params[:id].blank?
```
cjcolvar: This check is done in `set_masterfile` so I don't think it needs to be done here now that `set_masterfile` is called.

app/controllers/master_files_controller.rb
cjcolvar, line 375: From what I can tell `MasterFile.exists?` and `MasterFile.find` both make a HEAD request to fedora then a GET request if the HEAD was successful.  Based on that I don't think both are needed.  I know that error handling can be expensive but I think it is probably much cheaper than the two calls to fedora in this case.

```suggestion
def set_masterfile
  @master_file = MasterFile.find(params[:id])
rescue ArgumentError, ActiveFedora::ObjectNotFoundError
  flash[:notice] = "MasterFile #{params[:id]} does not exist"
end
```

And we don't want to rescue `Ldp::Gone` as well because we want it to bubble and get handled in `ApplicationController`, right?

cjcolvar (commented):
Cool, this does seem easier than adding a tombstone to solr.  We can reexamine this if fedora is getting too many requests.  I added a couple suggestions that I think would tighten this up.

app/controllers/master_files_controller.rb
masaball: We have an application level catch for `ActiveFedora::ObjectNotFoundError` as well so maybe we just rescue `ArgumentError`? But otherwise, that makes sense.

cjcolvar (approved):
Looks great!

PR #6731, opened 2026-03-03
Update dependencies

cjcolvar:
Note that the bump of some JS dependencies requires node 20.9.0+ which should be okay since we've recommended 20.x since at least avalon 7.7 and we probably want to upgrade node since 20.x is EOL in about a month anyways.


Dananji (approved):
Looks great! 👍

Issue #6719, opened 2026-02-16
[BUG] Selected items are cleared after Move action for Manager role but not for Admin


abankhele:
**Describe the bug**
When logged in as a Manager (manager of two collections), performing a Move action from the Selected Items page clears/unselects the selected items after the move completes. This does not happen when logged in as an Admin.

**To Reproduce**
Steps to reproduce the behavior, including the results:

1. Go to https://avalon-dev.dlib.indiana.edu/
2. Log in as a user with the Manager role who is manager of one or more collections.
3. Add item/s to Selected Items.
4. Go to the Selected Items page.
5. Select the item/s and perform the Move action to a destination collection.
6. Observe that after the move completes, the item/s are no longer selected (selection is cleared).
7. Repeat the same steps as an Admin user.
8. Observe that the selected items remain selected after the move.

**Expected behavior**
After completing a Move action, the selection state should behave consistently across roles. The selected items should selected for both Manager and Admin.

**Screenshots/Screen Recording**

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGZ6REE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ff6d9c89a6ae49fc5c9a2584e64ed956ad3c89ad/Screen%20Recording%202026-02-16%20at%202.00.42%E2%80%AFPM.mov)

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser : chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

joncameron:
👍

PR #6732, opened 2026-03-03
Use `-=` when changing MO collection governing policy

masaball:
Related issue: #6719

The loss of bookmark was caused by the `governing_policies.delete` clause. The delete action triggers a save, which would trigger the after save callback to the `remove_bookmarks` method. The media object has no governing policy in that moment, so the bookmark would get removed because the manager did not have read permissions for the media object. Changing from `delete` to `-=` prevents this callback from being triggered at that moment and improves the efficiency of the background job by removing the superfluous save within the `collection=` method and relying on the `if media_object.save` within the job itself.

### DIFF
--- a/app/models/media_object.rb
+++ b/app/models/media_object.rb
@@ -212,7 +212,7 @@ class MediaObject < ActiveFedora::Base
   def collection= co
     old_collection = self.collection
     self._collection= co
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Good catch!  Great job tracking this down!

PR #6733, opened 2026-03-03
Build docker image with node 24

cjcolvar:
Also correct warnings in Dockerfile
Also deal with some deprecation warnings: `punycode` in JS and dart sass 2 deprecation warnings.
Note that there are still deprecation warnings about `url.parse()` in JS and `@import` for dart sass 3.


masaball (approved):
Looks good to me

Issue #6737, opened 2026-03-04
Inherited permissions issues identified from testing


cjcolvar:
- [x] Inherit messages for permissions inherited from unit can be misleading: says collection when it really comes from unit
- [x] Visibility is getting inherited and appearing in groups (e.g. `public` inherited from collection)
- [x] Inherited group showing up in both group and external group in work
- [x] Inherited IPs showing up in group, external group, and IP in work

PR #6741, opened 2026-03-09
Fix issues with inherited permissions

cjcolvar:
- Inherit visibility and hidden from collection
- Remove visibility and hidden from unit
- Fix issue with groups appearing in multiple sections of media object special access
- Standardize and simplify language of inherited special access permissions
- Remove inherited permissions from special access panel when inheritance disabled
- Merge VirtualGroups into MediaObjectBehavior concern since the changes require knowledge of MediaObjects and it was only being used in MediaObject and its proxy.  This made testing easier.

#6737


app/models/search_builder.rb
```diff
@@ -32,14 +34,20 @@ def only_published_items(_permission_types = discovery_permissions, _ability = c
   end
 
   def limit_to_non_hidden_items(_permission_types = discovery_permissions, _ability = current_ability)
-    [policy_clauses, "(*:* NOT hidden_bsi:true)"].compact.join(" OR ")
+    media_object_hidden_clause = "hidden_bsi:true"
+    collection_hidden_clause = "{!join from=id to=isGovernedBy_ssim}default_hidden_bsi:true"
```
masaball, line 38: Were these added as reference notes while you were working on this? Doesn't look like we actually use these variables in the search clause.

masaball (changes requested):
Overall, this looks good to me. Just a couple variables created that should maybe be removed.

app/models/search_builder.rb
cjcolvar: I was planning on splitting these out for readability.  Do you think it is easier to read with the variables or as one long line?  I'm open to either approach.

masaball: I think having them pulled out as variables is more readable. Those solr queries can get hard to parse, so I feel having them split into their constituent parts makes it easier to piece it all together.

masaball (approved):
Looks good to me

PR #6744, opened 2026-03-13
Fixes in spec files and tagging of the test cases

abankhele:
1. Fixed test cases of spec files.
2. Tagged with Testomat ID and updated the title to match with Testomat
3. Made CDL spec configurable - test cases would be conditional based on enabling/disabling of the CDL


cjcolvar (approved):
Only did a quick skim through but looks good.  After this is merged we will have to go through and fix up anything that has been broken by changes to avalon since this PR was created.

cjcolvar:
Thanks @abankhele for all of your efforts on getting the cypress test suite to where it is now!!

PR #6748, opened 2026-03-18
Refactor thumbnail retrieval

masaball:
This commit makes some minor changes to how thumbnails are retrieved and formatted for display. This removes an unnecessary nested link structure from the html and simplifies some of the underlying code structure. It also includes a variety of small rubocop linting changes.


cjcolvar (approved):
👍 Looks good as far as I can tell.

Issue #6646, opened 2025-11-20
Add Donor Field to Descriptive Metadata


joncameron:
## Description
A collection manager requested that a field for donor information be added to Avalon's metadata profile. That seems like a reasonable request, but we'll need to consult with a metadataist to figure out what the right values and mapping are.

> We have many films that are all "educational" films. It makes sense for users to see these grouped together in and MCO collection; however, it would also be great to distinguish where the film came from/from whom the film came.
> 
> As a concrete example, we have a collection "Agency for Instructional Television" which has 1026 items. AIT was an educational TV program. Then we have the John Huston (8 items) and John Allman(24 items) collections. Everything in the John Huston and John Allman collections is part of the AIT TV program. We'd love to have all of this in one collection, but would like to be able to retain that items came from different sources. 

### Note from Music cataloging
DONOR COLLECTIONS: MARC fields 590 and 690
The music library uses 590 to name the donor regularly, and uses the 690 for an authorized version of the collection name. The 590 should go under the Notes heading in MCO, with a subheading LOCAL NOTE. They are named in a 690 as a local subject heading--and if the 690 comes over to MCO and lives under the Subjects heading, that is fine; it could have a subheading of LOCAL SUBJECT, but it is ok if it doesn’t. The 690 is also used for true subjects--we’ve commandeered it for music donor collections because it is hyperlinked and indexed in IUCAT, so it should remain under the Subject heading in MCO...not be changed to something different. I think the ATM uses the 690 the same way (local subjects AND local collection names), but I don’t know about any other collections.
Here’s an example of how the two are consistently used now: https://iucat.iu.edu/catalog/4008362 The example is a book, but it uses the 590 and 690 the way we do now. 

## Done Looks Like
- [ ] "Donor" field added to descriptive metadata
- [ ] "Donor" is a free text field
- [ ] "Donor" is indexed—in the descriptive metadata display it will show as a hyperlink, linked to the catalog search filtering for items with a matching "Donor" field value (like subject is handled currently)
- [ ] Consult with metadata expert on the best way to represent this in MODS

joncameron:
This will not be based on an authority lookup or controlled vocabulary.

cjcolvar:
I talked with @jlhardes and she was suggesting that maybe the existing Topical Subject field could be used.  This field appears in the description form and item view page as "Subject" where it is linked to a filtered search.

[...]

PR #6754, opened 2026-03-24
Add a donor note to descriptive metadata

masaball:
Related issue: #6646

### DIFF
--- a/app/assets/javascripts/dynamic_fields.js
+++ b/app/assets/javascripts/dynamic_fields.js
@@ -60,7 +60,7 @@ document.addEventListener('DOMContentLoaded', function () {
         el.setAttribute('data-bs-target', target.join('_'));
       });
 
-      if (query('.typeahead'), current_input_group) {
+      if (query('.typeahead', current_input_group)) {
         // Add new input group for typeahead field, e.g. Language(s)
         query('.typeahead', current_input_group).setAttribute('open', false);
         const new_autocomplete = query('.typeahead', new_input_group);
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/dynamic_fields.js
```diff
@@ -60,7 +60,7 @@ document.addEventListener('DOMContentLoaded', function () {
         el.setAttribute('data-bs-target', target.join('_'));
       });
 
-      if (query('.typeahead'), current_input_group) {
+      if (query('.typeahead', current_input_group)) {
```
cjcolvar, line 63: Good catch!

app/models/iiif_manifest_presenter.rb
```diff
@@ -126,7 +126,9 @@ def note_fields(media_object)
   end
 
   def gather_notes_of_type(media_object, type)
-    media_object.note.present? ? media_object.note.select { |n| n[:type] == type }.collect { |n| n[:note] } : []
+    notes = media_object.note.present? ? media_object.note.select { |n| n[:type] == type }.collect { |n| n[:note] } : []
+    notes.map! { |note| "<a href='#{Rails.application.routes.url_helpers.search_catalog_url(utf8: true, search_field: 'all_fields', q: note)}'>#{note}</a>" } if notes.present? && type == 'acquisition'
```
cjcolvar: I think it would be good to treat this like a facet (like subject) instead of a general search.  In order to do that I think it would be easiest to create a new index field in MediaObject's to_solr specifically for donors.  Take a look at https://github.com/avalonmediasystem/avalon/pull/6049 as a reference since there is a change needed in CatalogController.

cjcolvar (approved):
Looks great!

PR #6755, opened 2026-03-24
Bump rails version to latest 8.0.x patch

cjcolvar:
I had to add `application/x-subrip` as an alternative to `text/srt` because marcel 1.1.0 added handling of `srt`.


masaball (approved):
Looks good

Issue #6751, opened 2026-03-24
 [BUG] Tooltips are not working on resource description page


masaball:
**Describe the bug**
The resource description section of the item edit page has tooltips for each descriptive metadata field. These are supposed to be activated by clicking the question mark icon next to a field name. Currently clicking on the question marks does nothing.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to any item edit page
2. Click on 'Resource Description'
3. Click on any question mark icon next to a fieldname
4. Nothing happens

**Expected behavior**
Clicking the question mark causes the tooltip to display, clicking again hides the tooltip

**Environment (please complete the following information):**
 - Laptop
 - Kubuntu 24.04
 - Firefox 148.0.2
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Additional notes**
I expect this issue may be systemic and not just affecting tooltips on the resource description page, so other areas that have tips should be tested when working on this issue and QAing.

joncameron:
👍

PR #6622, opened 2025-10-17
Replace jQuery in JS assets

Dananji:
Related issue: #6564 

Changes in this PR:
- replace jQuery syntax in the JS assets in the code with vanilla JS

Notes on  some of the files:
- **avalon_progress.js**: removed section progress-bar updates for the old structure section display with ME.js
- **crop_upload.js**: added JS to display and hide uploading in-progress while the poster data are posted to the server
- **file_browse.js**: It seems `browse-everything` still uses `jQuery` to wrap the element that is passed to it for initializing the BrowseEverything instance. Therefore we still need `jquery-rails` Gem along with `browse-everything`. I was not able to test all of its functionality locally after the changes. 
- **fix_console.js**: REMOVED. This seems to be a fix for very old browsers (earlier IE versions) that didn't have a native `console` object.
- **localist_times.js**: removed `'draw.dt'` listener, which was used for formatting `data-utc-time` values in checkouts table. This is now handled in the ReactJS component for the table.
- **dom_utils.js**: added aliases to document.* functions for picking up DOM elements using class names/tag names/ids.

PR #6757, opened 2026-03-24
Fix tooltip click event targets

Dananji:
Related issue: #6751 

This bug is a regression of the work done in https://github.com/avalonmediasystem/avalon/pull/6622. The previous work removed the `jQuery` click handler for the tooltip, instead of replacing it with a vanilla JS implementation.

The fix in this PR uses Bootstrap's `collapse` attributes `data-bs-toggle` and `data-bs-target` to show/hide the tooltip. And an additional event-listener on the tooltip 'click' event, to prevent it from bubbling up into the form-label, which focuses the input field and highlights the text inside the element.

### DIFF
--- a/app/assets/javascripts/pop_help.js
+++ b/app/assets/javascripts/pop_help.js
@@ -33,6 +33,13 @@ document.addEventListener('DOMContentLoaded', function () {
     });
   }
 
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Looks good!

Issue #6740, opened 2026-03-06
[BUG] Manager cannot create new collection


abankhele:
**Describe the bug**
A user with the Manager role is unable to create a new collection. Clicking “Create Collection” under “My Collections” triggers a Rails error page instead of loading the new collection form.

**To Reproduce**
Steps to reproduce the behavior, including the results:

1. Click on “Manage”.
2. Click “Manage Content”.
3. Under “My Collections”, click “Create Collection”.
4. A Rails error page is shown.

**Expected behavior**
The “New Collection” page should load successfully and allow the manager to create a collection.

**Screenshots/Screen Recording**
<img width="1490" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeThwRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7426800dca50836f9fde80518629b43d6ab5276d/image.png" alt="image.png" />

**Environment (please complete the following information):**
 - Device / platform: laptop
 - OS: macos
 - Browser: chrome
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

joncameron:
Easier bug or UX flow issue.

joncameron:
Still an issue; creating a new collection with user "manager@example.com" returns an Action Controller error.

joncameron:
@masaball I'm getting an error on dev for 'manager@example' when creating a collection, but it happens when submitting the 'create collection' form rather than "Create Collection" from the Manage page.  Should I open a new issue for this or do you think this is it a config thing?

<img width="1254" height="855" alt="Image" src="https://github.com/user-attachments/assets/88da04b2-9d68-4897-968d-e4f700eea3f7" />

masaball:
I'm not encountering that error locally so I lean towards it being something specific to how the dev environment or that specific user account are setup. Probably good to create a ticket for tracking but maybe on the local board would make sense?

joncameron:
https://github.iu.edu/AvalonIU/Avalon/issues/432

PR #6759, opened 2026-03-25
Allow collection managers to create collections

masaball:
Related issue: #6740

The issue arose from an ability check that we had for rendering part of the view partial. When building the new collection page there was an ability check for disabling the Unit field if a user cannot update collections. Because there is not a Unit assigned to the collection at that point, this caused an error when looking for inherited permissions on the collection when doing the ability check. Since there are other ability checks in place, we can remove this disabling because anyone who can access the new collection page at all should have minimum permissions to update a collection.


cjcolvar (approved):
👍

PR #6761, opened 2026-03-26
Bump resolv-replace to latest to ensure it can work with ruby 3+ kwargs

cjcolvar:
needed by latest `prometheus_exporter`


masaball (approved):
👍

Issue #6210, opened 2025-02-26
[BUG] Timeline Replay Streams Entire Media Object Instead of Selected Duration


abankhele:
**Describe the bug**
On mobile devices, after a timeline video finishes playing, clicking the Play button again causes the entire media object to stream instead of restricting playback to the designated timeline duration. On laptops, once the timeline video completes, clicking Play resumes playback from the next second instead of restarting the selected duration.

**To Reproduce**
Steps to reproduce the behavior, including the results:

1. Go to https://mco-staging.dlib.indiana.edu/timelines/2165
2. Click the Play button to start the video.
3. After the video finishes, click the Play button again.
4. On mobile, observe that the entire media object begins streaming instead of the defined timeline duration.
5. Drag the scrubber past the 1:00 mark and verify that playback continues beyond the intended range.

**Expected behavior**
Playback should be restricted to the selected timeline duration and should not allow streaming beyond the defined range.

**Screenshots/Screen Recording**

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3NJQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a20631ce400b0ad5b455769d99c0a3f9d8359ec3/Screen%20Recording%202025-02-26%20at%202.03.48%E2%80%AFPM.mov)

**Environment (please complete the following information):**
 - Device / platform: iphone 16 pro max, desktop
 - OS: ios, macos
 - Browser: chrome, safari
- [ ] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

joncameron:
I'll move this to the Timeliner repo.

joncameron:
Works as expected in iOS 👍

PR #6762, opened 2026-03-26
New Timeliner build

Dananji:
Issues addressed in this build:
- https://github.com/avalonmediasystem/avalon/issues/6210
- https://github.com/IUBLibTech/timeliner/issues/98
- https://github.com/IUBLibTech/timeliner/issues/65 - this does not affect Avalon's timeliner. Therefore, it needs to be tested that this is not breaking or changing anything in Avalon.


cjcolvar (approved):
👍

Issue #6742, opened 2026-03-12
Add Option to Remove Case Sensitivity from Email Addresses


joncameron:
## Description
- Copy and paste is common when entering values for special access
- Currently this is case sensitive
- GanderedGoose@gmail.com is currently distinct from ganderedgoose@gmail.com

Could be implemented as a rake task; lowercasing the value inside application logic may be better.

## Q
Is there any downside to making these values case-insensitive? They may already be handled as case-insensitive elsewhere in the application.

## Done Looks Like
- [ ] Username and email address values are not case sensitive in access control

joncameron:
👍

PR #6764, opened 2026-03-27
Make user special access case insensitive

masaball:
Related issue: #6742

### DIFF
--- a/app/models/concerns/media_object_behavior.rb
+++ b/app/models/concerns/media_object_behavior.rb
@@ -141,7 +141,7 @@ module MediaObjectBehavior
   def inherited_read_users
     return [] unless collection
     users = collection.default_read_users.to_a + collection.inherited_read_users.to_a
-    users.uniq
+    users.uniq.map(&:downcase)
   end
 
   def inherited_read_groups
--- a/app/models/concerns/media_object_behavior.rb
+++ b/app/models/concerns/media_object_behavior.rb
@@ -141,7 +141,7 @@ module MediaObjectBehavior
   def inherited_read_users
     return [] unless collection
     users = collection.default_read_users.to_a + collection.inherited_read_users.to_a
-    users.uniq.map(&:downcase)
+    users.uniq
   end
 
   def inherited_read_groups
--- a/app/models/concerns/user_normalization.rb
+++ b/app/models/concerns/user_normalization.rb
@@ -13,6 +13,9 @@
 # ---  END LICENSE_HEADER BLOCK  ---
 
 module UserNormalization
+  FIELDS = ['unit_administrators', 'collection_managers', 'default_read_users', 'inherited_read_users',
+            'inherited_edit_users', 'read_users', 'edit_users'].freeze
+
   # Method to normalize just read user fields.
   def normalize_read_users
     field = self.class == MediaObject ? :read_users : :default_read_users
@@ -22,9 +25,7 @@ module UserNormalization
   # Method to normalize all fields containing usernames/emails
   # TODO: Find consistent way to skip unchanged fields
   def normalize_user_values
-    user_methods = methods.grep(/users$/).reject { |m| m.to_s.include?('set') }
-    user_methods += [:unit_administrators, :collection_managers]
-    user_methods.each do |field|
+    FIELDS.each do |field|
       set_user_values(field)
     rescue
       # Skip any methods that do not exist in the including class e.g. :unit_administrators
--- a/app/models/concerns/user_normalization.rb
+++ b/app/models/concerns/user_normalization.rb
@@ -1,0 +1,39 @@
+# Copyright 2011-2026, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+#   under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+#   CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#   specific language governing permissions and limitations under the License.
+# ---  END LICENSE_HEADER BLOCK  ---
+
+module UserNormalization
+  # Method to normalize just read user fields.
+  def normalize_read_users
+    field = self.class == MediaObject ? :read_users : :default_read_users
+    user_array = send(field)
+    setter = "#{field}=".to_sym
+    send(setter, user_array.map(&:downcase))
+  end
+
+  # Method to normalize all fields containing usernames/emails
+  # TODO: Find consistent way to skip unchanged fields
+  def normalize_user_values
+    user_methods = methods.grep(/users$/).reject { |m| m.to_s.include?('set') }
+    user_methods += [:unit_administrators, :collection_managers]
+    user_methods.each do |field|
+      user_array = send(field)
+      setter = "#{field}=".to_sym
+      send(setter, user_array.map(&:downcase))
+    rescue
+      # Skip any methods that do not exist in the including class e.g. :unit_administrators
+      # when loaded into Admin::Collections
+      next
+    end
+  end
+end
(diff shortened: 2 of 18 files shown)


cjcolvar (commented):
I was thinking about this and I'm wondering if it might make more sense to add `before_save` callbacks on the models instead of doing the downcasing the next layer up in the controller/jobs.  Either way I think we'll probably need a migration to downcase existing `read_users` which could be expensive.  We might be able to do a solr query for uppercase characters to try and avoid migrating every collection and media object in the repository.  Alternatively we could not change how they're stored and try to find a way to downcase before querying.  But that seems like it would probably be tricky and involve overrides.  Thoughts?

lib/tasks/avalon_migrations.rake
```diff
@@ -168,5 +168,19 @@ namespace :avalon do
         ReindexJob.perform_later(collection.media_object_ids)
       end
     end
+    desc "Downcase existing special access user entries"
+    task special_access_users: :environment do
+      ids = ActiveFedora::SolrService.query('inheritable_read_access_person_ssim:/.*[A-Z].*/ OR read_access_person_ssim:/.*[A-Z].*/', fl: 'id', rows: 100_000)
```
cjcolvar, line 173: I don't think 100,000 rows is necessarily always going to be large enough.  Maybe this needs to be converted into a loop where it paginates while `start` + `rows` < `numFound`?  It could use a smaller batch size then (1000 is a default in ActiveFedora).

I like how this can be run as many times as needed until everything has been migrated.  Another idea is adding a logging line about number of rows found and number migrated in the end to help give some feedback that things worked and fully completed.  In this vein, maybe the `save!` could be changed into a `save` and the error logged if the save failed?  Then the migration could keep running and do the most it can do instead of getting hung up on a broken item.

app/models/concerns/media_object_behavior.rb
```diff
@@ -140,7 +129,7 @@ def inherited_virtual_read_groups
   def inherited_read_users
     return [] unless collection
     users = collection.default_read_users.to_a + collection.inherited_read_users.to_a
-    users.uniq
+    users.uniq.map(&:downcase)
```
cjcolvar: It would probably be safest to put the `uniq` at the end.
```suggestion
    users.map(&:downcase).uniq
```

cjcolvar (commented):
This is looking great! 

Unfortunately, I do have a couple thoughts which could expand the scope of this PR more so see if you think it would be best to handle it here or in another PR or another time (which could be never 😆):

So if usernames in special access are getting downcased then shouldn't we handle usernames across the application as downcased?  If so, then unit and collection roles as well as Admin::Group members should all be downcased.  (There might possibly be other places where usernames are getting used in playlists/timelines?)  And if all of the string references to users are being downcased then we should probably downcase the username/email in the user model too.  If those changes are made in `before_save` callbacks on the models then we probably don't need the downcases throughout the application.  We would have to migrate more objects but the ActiveRecord user objects should be fast.

masaball:
I think there is possibly too much to that discussion for the comment section here. Let's try to meet up synchronously sometime this week and hash all that out.

masaball:
Follow-up work for other parts of the application is covered in #6792, so I think this should be ready for re-review.

cjcolvar:
For my future reference, on mco-staging the rake task to migrate ActiveFedora objects would currently cover 1709 objects so much less than I feared.

cjcolvar:
This is looking good and shaping up to be fairly compact!  
Here some thoughts I had for next steps and some questions.
- Add all user permission fields to the migration query
- Maybe the migration could run `normalize_user_values` on each object identified for migation?
- I think it might be safest to have an explicit list of fields in `UserNormalization` (or in each model class that gets referenced in the mixin?)
- Maybe this same list could be used for identifying solr fields for the migration query?  That might be complicating things too much when they probably won't change often and hard-coding them in the migration is probably fine.
- If the migration is good then we probably don't need the downcasing in the controllers or model besides `UserNormalization` and `Admin::Group`.
- Do we need to ensure that ability/access controls enforcement solr queries get a downcased `user_key` or is this already happening?

Here is the list of fields I was seeing that this should be applied to.  Does this match what you were thinking?

Unit:
- unit_administrators
- collection_managers
- default_read_users
- inherited_edit_users / inherited_read_users
- edit_users / read_users

Collection: 
- collection_managers
- default_read_users
- inherited_edit_users / inherited_read_users
- edit_users / read_users

MediaObject: 
- read_users

masaball:
For listing the affected fields, are you thinking just a comment listing the expected fields? Or are you thinking to get rid of the full dynamic meta-programming approach and instead have a constant defined like `USER_FIELDS = ['read_users', etc.].freeze` that we use to build the method calls?

What I'm seeing for fields that would currently be affected:

Unit:
* unit_administrators
* collection_managers
* default_read_users
* inherited_edit_users
* edit_users/read_users/discover_users

Collection:
* collection_managers
* default_read_users
* inherited_edit_users/inherited_read_users
* edit_users/read_user/discover_users

MediaObject:
* read_users

The logic here is that MediaObject does theoretically have `inherited_read_users`, `edit_users`, and `discover_users` on it, but I think only `read_users` is actually editable from the media object edit page. `inherited_read_users` should be getting downcased before it reaches the MO level and the other permissions are controlled from the collection level, I believe. As far as shows up for me, we don't have `inherited_read_users` at the Unit level and have `discover_users` at every level, though I'm not sure that is a field that we have set up to be explicitly set.

`user_key` should be getting downcased already. It comes from our devise settings so `config.authentication_keys = [:username]` and `config.case_insensitive_keys = [:username, :email, :login]` from 'config/initializers/devise.rb' and there's a comment on the case insensitive that "These keys will be downcased upon creating or modifying a user and when used to authenticate or find a user". So if I'm reading that right, the user_key should already be getting downcased by Devise.

cjcolvar:
Good find on the devise documentation!  That means we should be good with respect to my last question.

As for affected fields, I didn't see `discover_users` used anywhere.  Can you point me to where we're using it?  Units can set "default access controls" special access which uses `inherited_read_users` so we'll need that one.

In the validation I was thinking of an explicit list of fields instead of the metaprogramming.  My thought was that it if we add a new field to one of these models it would probably be better to have to explicitly add it to a list over it possibly getting added to the list unintentionally.  Plus an explicit list acts as a form of documentation.

masaball:
`discover_users` exists as a method on each level, but I think you're right that we don't actually use it for anything so we are probably fine to skip it. And I missed that bit with Unit setting `inherited_read_users`.

app/models/concerns/user_normalization.rb
```diff
@@ -0,0 +1,44 @@
...
+  # Method to normalize all fields containing usernames/emails
+  # TODO: Find consistent way to skip unchanged fields
+  def normalize_user_values
+    FIELDS.each do |field|
+      set_user_values(field)
+    rescue
+      # Skip any methods that do not exist in the including class e.g. :unit_administrators
+      # when loaded into Admin::Collections
+      next
+    end
```
cjcolvar, line 30: I think respond_to? should avoid the need for the rescue. I think it should be a little faster too since I've read that rescuing errors is expensive.

```suggestion
    FIELDS.each do |field|
      # Skip any methods that do not exist in the including class e.g. :unit_administrators
      # when loaded into Admin::Collections
      set_user_values(field) if respond_to?(field)
    end
```

app/models/user.rb
```diff
@@ -133,7 +133,7 @@ def self.find_for_lti(auth_hash, signed_in_resource=nil)
   end
 
   def self.autocomplete(query, _id = nil)
-    self.where("username LIKE :q OR email LIKE :q", q: "%#{query}%").collect { |user|
+    self.where("LOWER(username) LIKE LOWER(:q) OR LOWER(email) LIKE LOWER(:q)", q: "%#{query}%").collect { |user|
```
cjcolvar, line 136: This makes me wonder: should we migrated User objects and downcase username and email as well as add code to force that on the User model?  I guess it isn't necessary since the model already runs `lower()` on any DB queries but wondering if it would be good for consistency.  Maybe a follow up PR if we deem it useful?

cjcolvar (dismissed):
Looks great!  Just had one small suggestion. Also a comment/question that doesn't need to be part of this PR.

app/models/user.rb
masaball: It might be worth changing for additional consistency. But that feels like a follow-up issue, if we decide to go that route.

cjcolvar (approved):

PR #6766, opened 2026-03-31
Retry Connection Refused (503) responses from solr

cjcolvar:
Using [faraday-retry middleware](https://github.com/lostisland/faraday-retry) configured in RSolr (https://github.com/rsolr/rsolr?tab=readme-ov-file#retry-503s) and setup in Blacklight (https://github.com/projectblacklight/blacklight/wiki/Solr-Configuration) and ActiveFedora (https://github.com/avalonmediasystem/avalon/blob/c6af9b362e1dc68060c7d3307415ecf350317610/config/initializers/active_fedora_env.rb#L26).

I added a test to ensure it is working within ActiveFedora.  By default it will retry 2 times.  This is configurable with the `retry_after_limit` option (but we'd have to allowlist or pass it through in Avalon's ActiveFedora `EnvironmentConfigurator`).


masaball (approved):
The code looks good as far as I can tell. I am a bit worried that rather than alleviating issues this could cause things to spiral even worse when we are being hit hard. But we can go ahead and merge and test things out.

PR #6768, opened 2026-03-31
Items inherit lending period from parent collection

### LINKED ISSUE #6749: Items Inherit CDL Values from Collection
## Description
When CDL is enabled for a collection, items should inherit CDL values in addition to the other access control attributes.

## Done Looks Like
- [ ] Items inherit CDL values are inherited from the collection 
- [ ] CDL attributes are also able to be disabled when "Disable parent permissions" is true

cjcolvar:
Resolves #6749


masaball (approved):
Looks great

Issue #6738, opened 2026-03-05
Prevent Publication of Non-accessible Items


joncameron:
## Description
Avalon currently has a "Has Transcript" and "Has Captions" attribute for records. Another is needed: is this item accessible; that is, for each of the files does it have a corresponding transcription document with which it will meet WCAG AA-level compliance? Something more sophisticated for indicating whether an item meets AV accessibility compliance required by WCAG AA.

- If the creation date for an item (the date_ingested timestamp for the media object) is past 2026-04-24 and is_accessible is false, restrict the ability to Publish.
- If a collection manager exempts the item by toggling the state of the checks a toggle to indicate that the item is not bound by normal accessibility concerns (archival footage, music, silent film), then they are allowed to Publish the item.
- The only mechanism is the binary toggle that set the is_accessible value, and there is no notes field or other additional information for collections managers to provide

## Questions
- Where should the "exempt this item" toggle live in the UI?
  - Manage Files page, above the list of sections
- Will an item be unpublished when a new file is added?
  -  No, items that are Published are not Unpublished when a new file is added. Items cannot be set to Publish if is_accessible is true, but if they are already Published they are grandfathered in

## Done Looks Like
- [x] System config boolean added: `enforce_accessibility_compliance`
 - [x] New attribute added to media objects: is_accessible.
- [x] When `enforce_accessibility_compliance` is set to true in system settings, the Publish action cannot be taken 
- [x] Administrator users can still take the Publish action on items where is_accessible is false
- [x] New attribute added to media objects: override_accessibility (probably a better name for this)
- [x] Publish action is disabled unless:
  - "Is Accessible" is true _or_
  - The item is exempted by the collection manager

joncameron:
👍

PR #6769, opened 2026-04-01
Prevent publication of accessibility non-compliant Media Objects

masaball:
Related issue: #6738 

This PR also includes a small commit to remove some unused code from the file upload step model.

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -274,8 +274,12 @@ class MediaObjectsController < ApplicationController
           error_messages += ['Failed to create media object:']+@media_object.errors.full_messages
         else
           if !!api_params[:publish]
-            @media_object.publish!('REST API')
-            @media_object.workflow.publish
+            begin
+              @media_object.publish!('REST API')
+              @media_object.workflow.publish
+            rescue Avalon::PublishingError => e
+              error_messages += ["Unable to create media object: #{e.message}"]
+            end
           else
             @media_object.publish!('')
           end
@@ -428,8 +432,12 @@ class MediaObjectsController < ApplicationController
               errors += ["#{media_object&.title} (#{id}) (missing required fields)"]
               next
             end
-            media_object.avalon_publisher = user_key.presence
-            media_object.save!
+            begin
+              media_object.publish!(user_key)
+            rescue Avalon::PublishingError => e
+              errors += ["#{media_object&.title} (#{id}) (#{e.message})"]
+              next
+            end
             success_count += 1
           when 'unpublish'
             if can? :unpublish, media_object
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -643,6 +643,10 @@ class MediaObjectsController < ApplicationController
     note = mo_parameters.delete(:note) || []
     note_type = mo_parameters.delete(:note_type) || []
     mo_parameters[:note] = note.zip(note_type).map{|a|{note: a[0],type: a[1]}}
+    # Accessibility Override. Non-managers/admin should not be able to enable, but it should
+    # stay enabled if already turned on
+    return mo_parameters if @media_object.accessibility_exempt?
+    params.delete(:override_accessibility) unless can? :override_accessibility, @media_object
 
     mo_parameters
   end
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -643,10 +643,6 @@ class MediaObjectsController < ApplicationController
     note = mo_parameters.delete(:note) || []
     note_type = mo_parameters.delete(:note_type) || []
     mo_parameters[:note] = note.zip(note_type).map{|a|{note: a[0],type: a[1]}}
-    # Accessibility Override. Non-managers/admin should not be able to enable, but it should
-    # stay enabled if already turned on
-    return mo_parameters if @media_object.accessibility_exempt?
-    params.delete(:override_accessibility) unless can? :override_accessibility, @media_object
 
     mo_parameters
   end
--- a/app/models/file_upload_step.rb
+++ b/app/models/file_upload_step.rb
@@ -36,11 +36,7 @@ class FileUploadStep < BasicStep
   end
 
   def execute context
-    deleted_sections = update_master_files context
-    context[:notice] = "Several clean up jobs have been sent out. Their statuses can be viewed by your sysadmin at #{ Settings.matterhorn.cleanup_log }" unless deleted_sections.empty?
-
-    media = MediaObject.find(context[:media_object].id)
-    context[:media_object] = media
+    update_master_files context
     context[:media_object].override_accessibility = context[:override_accessibility] == "1"
 
     context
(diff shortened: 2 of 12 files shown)


app/models/media_object.rb
```diff
@@ -340,6 +345,8 @@ def to_solr(include_child_fields: false)
       solr_doc['donor_ssim'] = self.note.collect { |n| n[:note] if n[:type] == 'acquisition' }
       solr_doc['other_identifier_ssm'] = self.other_identifier.collect { |oi| oi.to_json }
       solr_doc['related_item_url_ssm'] = self.related_item_url.collect { |r| r.to_json }
+      solr_doc['is_accessible_bsi'] = self.is_accessible?
+      solr_doc['override_accessibility_bsi'] = self.override_accessibility
```
cjcolvar: Instead of manually indexing here you should be able to specify the indexing on the property definition using:
```index.as ActiveFedora::Indexing::Descriptor.new(:boolean, :stored, :indexed)```

File: app/views/media_objects/_file_upload.html.erb [hunk 2]
```diff
@@ -15,6 +15,24 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <%= hidden_field_tag :donot_advance, true %>
 <%= hidden_field_tag :step, 'file-upload' %>
+<%# Collection manager and higher roles should be only ones able to exempt items %>
+<% if current_ability.can? :create, Admin::Collection %>
```
cjcolvar: I think it might be nice to create a new action for this instead of tying it to `:create, Admin::Collection`.  Maybe something like `can? :accessibility_exempt, @media_object`.  Then in the ability you can check that the user is a collection_manager/unit admin/administrator of this media_object.

config/settings.yml
```diff
@@ -154,3 +154,8 @@ search_pagination:
   window: 2
   left: 2
   right: 0
+# Prevent publication of items ingested after compliance date if they lack
+# captions or transcripts
+accessibility_compliance:
+  enforce: true
```
cjcolvar: Do we want this enabled by default?  I think it is fine to merge as is and figure this out later.

cjcolvar (commented):
I have a few small questions/suggestions but I think this is looking really good!

app/models/media_object.rb
masaball: Good catch. I'd looked at the predefined options/shortcuts (`:stored_sortable`, etc.) in Active Fedora and didn't see booleans, but didn't look deeper into options for declaring it manually/explicitly.

config/settings.yml
masaball: That's a good question. It might make more sense to have it disabled by default. @joncameron Thoughts?

joncameron: I agree, let's disable it by default.

app/views/media_objects/_file_upload.html.erb
```diff
@@ -15,6 +15,24 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <%= hidden_field_tag :donot_advance, true %>
 <%= hidden_field_tag :step, 'file-upload' %>
+<%# Collection manager and higher roles should be only ones able to exempt items %>
+<% if can? :override_accessibility, @media_object %>
```
cjcolvar, line 19: This guards the UI but I think it would be good to have a counterpart on the controller to guard this from being set by any user with edit privileges.  Maybe this could go in `#media_object_parameters`?

config/settings.yml
```diff
@@ -154,3 +154,8 @@ search_pagination:
   window: 2
   left: 2
   right: 0
+# Prevent publication of items ingested after compliance date if they lack
+# captions or transcripts
+accessibility_compliance:
+  enforce: false
+  compliance_date: '2026-04-26'
```
cjcolvar: I think the deadline is actually the 24th.
```suggestion
  compliance_date: '2026-04-24'
```

app/controllers/media_objects_controller.rb
```diff
@@ -635,6 +643,10 @@ def media_object_parameters
     note = mo_parameters.delete(:note) || []
     note_type = mo_parameters.delete(:note_type) || []
     mo_parameters[:note] = note.zip(note_type).map{|a|{note: a[0],type: a[1]}}
+    # Accessibility Override. Non-managers/admin should not be able to enable, but it should
+    # stay enabled if already turned on
+    return mo_parameters if @media_object.accessibility_exempt?
+    params.delete(:override_accessibility) unless can? :override_accessibility, @media_object
```
cjcolvar: Sorry, I think I led you astray with my suggestion to put this here.  `override_accessibility` isn't actually part of `params[:media_object]` so it doesn't really make sense here.  So I think it makes more sense to put this in the file upload step model.  See my comment there for a code suggestion.

app/models/file_upload_step.rb
```diff
@@ -36,11 +36,8 @@ def after_step context
   end
 
   def execute context
-    deleted_sections = update_master_files context
-    context[:notice] = "Several clean up jobs have been sent out. Their statuses can be viewed by your sysadmin at #{ Settings.matterhorn.cleanup_log }" unless deleted_sections.empty?
-
-    media = MediaObject.find(context[:media_object].id)
-    context[:media_object] = media
+    update_master_files context
+    context[:media_object].override_accessibility = context[:override_accessibility] == "1"
```
cjcolvar: I moved the ability check here and also added a check that the param is actually present to avoid removing the override when the param is not part of the update.
```suggestion
    context[:media_object].override_accessibility = context[:override_accessibility] == "1" if context[:override_accessibility].present? && context[:ability].can?(:override_accessibility, context[:media_object])
```

spec/controllers/media_objects_controller_spec.rb
```diff
@@ -2345,4 +2367,51 @@
...
+      end
+    end
+
+    context 'as editor' do
+      let(:user) { FactoryBot.create(:user) }
+      before do
+        collection = media_object.collection
+        collection.editors += [user.user_key]
+        collection.save
+        login_as(:user)
```
cjcolvar: The `login_as` here is creating a new user and not logging in as the editor.  This leads to 401 access denied responses which ends up masking the real behavior in these tests.

```suggestion
        login_user(user)
```

cjcolvar (commented):
Looking good.  My last review suggested changes in `media_object_parameters` but I think that complicated things.  I worked through it and made some suggestions which I think will be clearer.

spec/controllers/media_objects_controller_spec.rb
masaball: Gotcha. That explains some of the issues I initially had with the manager tests.

config/settings.yml
masaball: The ticket says the 26th. @joncameron What's the deadline we are supposed to be enforcing?

joncameron: It's the 24th! I'll update the ticket, thanks for the catch.

spec/controllers/media_objects_controller_spec.rb
```diff
@@ -2345,4 +2367,51 @@
...
+          patch 'update', params: { id: media_object.id, override_accessibility: '1', step: 'file-upload' }
+          expect(media_object.reload.accessibility_exempt?).to eq false
+        end
+      end
+
+      context 'with override enabled' do
+        let(:media_object) { FactoryBot.create(:media_object, override_accessibility: true) }
+
+        it 'does not disable override' do
+          patch 'update', params: { id: media_object.id, override_accessibility: '1', step: 'file-upload' }
```
cjcolvar: This looks like a copy paste typo.
```suggestion
          patch 'update', params: { id: media_object.id, override_accessibility: '0', step: 'file-upload' }
```

cjcolvar (dismissed):
Looks great!

cjcolvar (approved):

PR #6770, opened 2026-04-01
Remove remaining view and controller references to 'Apply to All Existing Items'

### LINKED ISSUE #6157: Remove
# Description
Removing the "Apply to All Existing" button and related functionality in favor of inherited permissions from collection to child media object. Avalon no longer needs to provide the option to apply collection level Special Access changes to items: we assume that a chance in either Staff Roles or Special Access should trickle down to item. Individual special access entries can still be added to the media object.

# Done Looks Like
- [ ] "Apply to All Existing Items" button removed from the "Assign special access" card on the collection admin page
- [ ] Related code and job handling removed if no longer needed

cjcolvar:
Leave background job for possible use in one-off tasks

Resolves #6157


masaball (approved):
Looks good

Issue #6674, opened 2026-01-07
Update System Groups for new inheritance paradigm


masaball:
## Description
The new setup for Units introduced a new system group: 'unit_administrator'. Even though this has been added to the list of system groups in `settings.yml`, the group is not showing up on the Admin Groups page. Additionally, with the introduction of this new group, the manager group may be redundant/superfluous.

## Current Problems
* Unit administrator is not showing up as a System group on the manage group page, even though it has been added to the settings.yml. Seems to be missing from Admin::Group table.
* Trying to remove sole collection manager at unit level throwing sole manager warning. Should remove that logic at same time as removing manager group, if that is happening.

## Questions
* Is this group setup still relevant if the only group that exists is the admin group?
  * If the manager and group_manager groups are retired, could we converge these into a single "administrators" group? The workflow could be that Admins can create Units, and can assign any username, without that user also being required to be in the 'unit_administrator' group.
* Will setting up a new application instance automatically add the 'unit_administrator' group?
* Do we need a migration for existing applications to get the 'unit_administrator' group properly added?
* Are we removing/disabling the manager group?
  * Yes 
* Do we need to switch the sole manager logic to point at Unit Admins?

joncameron:
@joncameron to create an issue for looking at the sole manager issue; can we change this so a collection could only contain inherited staff roles?

PR #6771, opened 2026-04-01
Remove manager and unit administrator groups

### LINKED ISSUE #6158: Remove "manager" System Group
# Description
With a change in permissions model, the current 'manager' group and workflow can be retired. Users with Manager membership to a Unit can be considered to be equivalent to the membership of the 'manager' group.

# Done Looks Like
- [ ] "manager" system group removed/destroyed
- [ ] manager group removed from the system groups listing at /admin/groups

cjcolvar:
Resolves #6158 and #6674

With these changes `unit_administrator`, `manager`, and `group_manager` would no longer be system groups and would not get created with new avalon instances.  The groups would continue to live on in existing avalon instances but would not have any special permissions.

### DIFF
--- a/app/controllers/admin/groups_controller.rb
+++ b/app/controllers/admin/groups_controller.rb
@@ -117,15 +117,9 @@ class Admin::GroupsController < ApplicationController
 
   def update_users
     group_name = params[:id]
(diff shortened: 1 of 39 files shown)


masaball (approved):
Looks good to me

PR #6774, opened 2026-04-06
Limit external group autocomplete results to 10 LDAP and 10 course entries

cjcolvar:
LDAP lookups are not limited and can lead to autocomplete dropdowns and are really really long.
image

This PR limits the LDAP lookup to 10 results and the course lookup to 10 results so the maximum number of results in the dropdown is 20.  We may want to limit this further.

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -39,4 +39,14 @@ describe Course do
     expect(first_media_object.read_groups).to include('sociology_101')
     expect(first_media_object.read_groups).to include('underwater_basketweaving_302')
   end
+
+  describe 'autocomplete' do
+    before do
+      10.times { Course.create(title: 'Test') }
+    end
+
+    it 'limits to 10 results' do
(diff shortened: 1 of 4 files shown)


spec/models/course_spec.rb
```diff
@@ -39,4 +39,14 @@
     expect(first_media_object.read_groups).to include('sociology_101')
     expect(first_media_object.read_groups).to include('underwater_basketweaving_302')
   end
+
+  describe 'autocomplete' do
+    before do
+      10.times { Course.create(title: 'Test') }
```
masaball:
```suggestion
      11.times { Course.create(title: 'Test') }
```
This should probably be larger than 10, because I think otherwise the test will always pass since there are only 10 results to return. Unless there is already a Course(s) that exists from earlier test setup?

masaball (commented):
Looks good overall, just a question about one of the tests.

masaball (approved):
👍

PR #6778, opened 2026-04-07
Fix unit deletion and add auth checks on target unit/collection

cjcolvar:
This PR builds out the unit deletion workflow and does some other smaller fixes:
- Implement auth checks on destroy for units/collections (`can? :update, @object`)
- Remove unit/collection being destroyed from target list
- Sort unit/collection target list
- Fix language on unit remove page

I made the permission requirement for target units to move collections to be that the current user is either admin or a unit admin.  This might be more restrictive than when deleting collections but I think matches the ability check (`can? :update, unit` is tied to `is_admin_of? unit`).  Should we expand this?

### DIFF
--- a/app/controllers/admin/units_controller.rb
+++ b/app/controllers/admin/units_controller.rb
@@ -330,10 +330,10 @@ class Admin::UnitsController < ApplicationController
     fastimage.type == :png && fastimage.size == [700, 560] # [width, height]
   end
 
-  # Returns collections for current_user
-  # @return [Collection] Collections to which current_user has manage access
+  # Returns units for current_user
+  # @return [Unit] Units to which current_user has manage access
   def get_user_units
-    # return all collections to admin
+    # return all units to admin
     if can?(:manage, Admin::Unit)
       SpeedyAF::Proxy::Admin::Unit.where("has_model_ssim: Admin\\:\\:Unit")
     else
--- a/app/controllers/admin/units_controller.rb
+++ b/app/controllers/admin/units_controller.rb
@@ -212,7 +212,7 @@ class Admin::UnitsController < ApplicationController
         @target_unit = Admin::Unit.find(params[:target_unit_id])
         # raise if user not admin of target unit
         authorize! :update, @target_unit
-        Admin::Unit.reassign_collections( @source_unit.collections, @source_unit, @target_unit )
+        Admin::Unit.reassign_collections( @source_unit.collections, @target_unit )
         target_path = admin_unit_path(@target_unit)
         @source_unit.reload
       else
--- a/app/controllers/admin/units_controller.rb
+++ b/app/controllers/admin/units_controller.rb
@@ -106,7 +106,7 @@ class Admin::UnitsController < ApplicationController
       end
       respond_to do |format|
         format.html do
-          redirect_to @unit, notice: 'unit was successfully created.'
+          redirect_to @unit, notice: 'Unit was successfully created.'
         end
         format.json do
           render json: { id: @unit.id }, status: 200
@@ -200,6 +200,7 @@ class Admin::UnitsController < ApplicationController
     @unit = Admin::Unit.find(params['id'])
     authorize! :destroy, @unit
     @objects    = @unit.collections
+    @candidates = get_user_units.reject { |u| u.id == @unit.id }.sort_by { |u| u.name.downcase }
   end
 
   # DELETE /units/1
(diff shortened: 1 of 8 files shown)


app/controllers/admin/units_controller.rb
```diff
@@ -326,4 +329,15 @@ def check_image_compliance(poster_path)
     # Size derived from width and aspect ratio from JS code, assets/javascript/crop_upload.js:60-63
     fastimage.type == :png && fastimage.size == [700, 560] # [width, height]
   end
+
+  # Returns collections for current_user
+  # @return [Collection] Collections to which current_user has manage access
+  def get_user_units
+    # return all collections to admin
```
masaball:
```suggestion
  # Returns units for current_user
  # @return [Unit] Units to which current_user has manage access
  def get_user_units
    # return all units to admin
```

app/models/admin/unit.rb
```diff
@@ -219,6 +219,13 @@ def self.autocomplete(query, _id = nil)
     search_array.map { |value| { id: value[:id], display: value[:name_ssi] } }
   end
 
+  def self.reassign_collections(collections, source_unit, target_unit)
```
masaball: We don't necessarily need to change anything here, but would it be better to not require source unit to be passed in, since it isn't used? Or at least mark it as unused `_source_unit`?

app/views/admin/units/remove.html.erb
```diff
@@ -57,14 +57,14 @@ Unless required by applicable law or agreed to in writing, software distributed
       <%= bootstrap_form_for @unit, method: :delete do |f| %>
       <% if @objects.count > 0 %>
 
-      <h2>Reassign Existing Items</h2>
+      <h2>Reassign Existing collections</h2>
```
masaball:
```suggestion
      <h2>Reassign Existing Collections</h2>
```

spec/controllers/admin_collections_controller_spec.rb
```diff
@@ -572,6 +573,36 @@
...
+      expect(controller.current_ability.can? :update, target_collection).to eq true
+      expect { delete :destroy, params: { id: collection.id, target_collection_id: target_collection.id } }.to change { Admin::Collection.count }.by(-1)
+      expect(Admin::Collection.exists?(collection.id)).to eq false
+      expect(target_collection.reload.media_objects).to include media_object
+    end
+
+    context 'when not unit admin of target unit' do
+      let!(:target_collection) { FactoryBot.create(:collection) }
+
+      it "destroys collection and reassigns items to target collection" do
```
masaball:
```suggestion
      it "does not destroy collection or reassign items" do
```
or match whats in the admin::unit tests

```suggestion
      it "redirects to restricted content page" do
```

spec/controllers/admin_units_controller_spec.rb
```diff
@@ -449,6 +450,36 @@
...
+      let!(:target_unit) { FactoryBot.create(:unit) }
+
+      it 'redirects to restricted content page' do
+        login_user unit.unit_admins.first
+        expect(controller.current_ability.can? :destroy, unit).to eq true
+        expect(controller.current_ability.can? :update, target_unit).to eq false
+        expect { delete :destroy, params: { id: unit.id, target_unit_id: target_unit.id } }.not_to change { Admin::Unit.count }
+        expect(Admin::Unit.exists?(unit.id)).to eq true
+        expect(target_unit.reload.collections).not_to include collection
+        expect(response).to have_rendered('errors/restricted_pid')
```
masaball, line 481: Is rendering the restricted content page the flow we want here? That feels like it might be obtuse to the users. Maybe we should redirect back to the original unit/collection page or the manage content page with a flash message for the error about the delete failing due to lack of permissions?

masaball (commented):
Mostly some small textual edits but do have one question about implementation details

spec/controllers/admin_units_controller_spec.rb
cjcolvar: Good question.  

This is following the pattern in the admin collections controller but maybe we want to change that behavior?  I think rendering restricted content is easy because `authorize! :update, unit` raises a `CanCan::AccessDenied` error handled in `ApplicationController` so it doesn't have to be aware of the specific context that the error is originating.  To improve this I think we'd probably need to talk about the UI flows we'd expect in each context (create/update/delete) and what the best response would be.  In theory, this scenario should be an edge case since the user should only be presented with actions they're authorized to take so I think this should only happen with the API, when manually trying requests, or in the odd case where permissions were changed in between rendering a page and an action being taken.

After writing this up I'm feeling like this is probably out of scope of this PR.  I can make a ticket to investigate improvements if you agree.

masaball: Yeah, that's feeling larger than this PR. My thinking in this case was specifically that the restricted content page message is along the lines of "You are not allowed to access this content" which could be confusing in the case of deletion, since they aren't explicitly trying to access any content.

Probably worth creating the follow up ticket to at least have the discussion on how/if situations like that should be handled differently.

masaball (approved):
👍

PR #6781, opened 2026-04-07
Supplemental file download filename uses edited label and defaults to original filename

cjcolvar:
Resolves https://github.com/samvera-labs/ramp/issues/915


masaball (approved):
Looks good!

PR #6782, opened 2026-04-08
Improve handling of unit in collection creation and editing

cjcolvar:
This PR builds on #6778 

Changes in this PR:
- Change unit input from autocomplete to drop-down select (using TomSelect to match unit/collection deletion pages)
- Only have one option for unit on /new if unit_id param present
- /new (without unit_id param) and edit collection dialog have units for which the user is unit admin
- Administrators have all units (except for /new if unit_id param present)
- Require that creating/updating user is unit admin of unit being assigned to collection
- Move get_user_units into ApplicationController and make helper method
- Skip attmpting update if unit not present or user does not have permissions
- Attempt to start unifying different approaches to error messages in json vs. html responses

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -64,7 +64,7 @@ class Admin::CollectionsController < ApplicationController
 
   # GET /collections/1
   def show
-    @candidate_units = get_user_units.sort_by { |u| u.name.downcase }
+    @candidate_units = get_user_units
 
     respond_to do |format|
       format.json { render json: @collection.to_json }
@@ -90,7 +90,7 @@ class Admin::CollectionsController < ApplicationController
       @candidate_units = [unit]
       @collection.unit = unit
     else
-      @candidate_units = get_user_units.sort_by { |u| u.name.downcase }
+      @candidate_units = get_user_units
     end
 
     respond_to do |format|
@@ -101,7 +101,7 @@ class Admin::CollectionsController < ApplicationController
 
   # GET /collections/1/edit
   def edit
-    @candidate_units = get_user_units.sort_by { |u| u.name.downcase }
+    @candidate_units = get_user_units
     respond_to do |format|
       format.js   { render json: modal_form_response(@collection) }
     end
@@ -141,7 +141,7 @@ class Admin::CollectionsController < ApplicationController
       respond_to do |format|
         format.html do
           flash.now[:error] = @collection.errors.full_messages.to_sentence
-          @candidate_units = get_user_units.sort_by { |u| u.name.downcase }
+          @candidate_units = get_user_units
           render action: 'new'
         end
         format.json do
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -190,7 +190,7 @@ class Admin::CollectionsController < ApplicationController
 
     # Update governing_policy_id if unit_id changes
     update_params = collection_params.to_h
-    if update_params['unit_id'].present? && @collection.unit.id != update_params['unit_id']
+    if update_params['unit_id'].present? && @collection.unit_id != update_params['unit_id']
       new_unit = Admin::Unit.find(update_params['unit_id']) rescue nil
       if new_unit.present? && can?(:update, new_unit)
         update_params = update_params.merge({ unit_id: new_unit.id})
@@ -206,8 +206,11 @@ class Admin::CollectionsController < ApplicationController
       end
     end
 
-    @collection.update_attributes update_params if !skip_save && update_params.present?
-    saved = @collection.save unless skip_save
+    unless skip_save
+      @collection.update_attributes update_params if update_params.present?
+      saved = @collection.save
+    end
+
     if saved
       if name_changed
         User.where(Devise.authentication_keys.first => [Avalon::RoleControls.users('administrator')].flatten).each do |admin_user|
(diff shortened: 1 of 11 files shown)


app/controllers/admin/collections_controller.rb
```diff
@@ -64,6 +64,8 @@ def index
 
   # GET /collections/1
   def show
+    @candidate_units = get_user_units.sort_by { |u| u.name.downcase }
```
masaball: Since we're calling `.sort_by` on seemingly every call to `get_user_units` should that step be pulled into that method? If you think we may have a use for the unsorted return, maybe we can add an argument like `get_user_units(sort: true)`?

app/controllers/admin/collections_controller.rb
```diff
@@ -175,10 +190,24 @@ def update
 
     # Update governing_policy_id if unit_id changes
     update_params = collection_params.to_h
-    update_params = update_params['unit_name'].present? ? update_params.merge({ unit_id: convert_unit(update_params['unit_name']) }) : update_params
-    update_params.merge!(governing_policy_id: update_params[:unit_id]) if update_params[:unit_id].present?
-    @collection.update_attributes update_params if update_params.present?
-    saved = @collection.save
+    if update_params['unit_id'].present? && @collection.unit.id != update_params['unit_id']
```
masaball:
```suggestion
    if update_params['unit_id'].present? && @collection.unit_id != update_params['unit_id']
```
Using `@collection.unit_id` should save us a fedora load.

app/controllers/admin/collections_controller.rb
```diff
@@ -175,10 +190,24 @@ def update
...
+        skip_save = true
+      else
+        not_found_msg = "Could not find unit #{update_params['unit_id']}"
+        @collection.errors.add(:unit, not_found_msg)
+        skip_save = true
+      end
+    end
+
+    @collection.update_attributes update_params if !skip_save && update_params.present?
+    saved = @collection.save unless skip_save
```
masaball: Do you think it is worth stronger guarding here? I'm pretty sure that `@collection.update_attributes` performs a save of it's own so when `update_params.present? == true` the collection will get saved twice. Maybe something like

```suggestion
    saved = @collection.update_attributes update_params if !skip_save && update_params.present?
    saved ||= @collection.save unless skip_save
```
would be sufficient?

Also, how do you feel about wrapping both lines in `unless skip_save` instead of each having that in-line? Or `if !skip_save` if that feels clearer to you.

```suggestion
    unless skip_save
      @collection.update_attributes update_params if update_params.present?
      saved = @collection.save
    end
```

app/views/admin/collections/new.html.erb
```diff
@@ -13,24 +13,50 @@ Unless required by applicable law or agreed to in writing, software distributed
   specific language governing permissions and limitations under the License.
 ---  END LICENSE_HEADER BLOCK  ---
 %>
+<% content_for :page_scripts do %>
```
masaball: Does this need to be at the start of the file? I think our `content_for :page_scripts` blocks are usually at the end of our .erb files. There may be a few outliers, but pretty sure having it after the html is our standard.

Also, how feasible would it be to have this in its own partial or asset file that we could just pull in here and in "admin/collection/_form.html.erb"? It looks like both scripts are basically the same. Since it is only in the two places, may not be worth the effort entailed in DRY-ing it up though.

spec/controllers/admin_collections_controller_spec.rb
```diff
@@ -324,100 +324,150 @@
...
-      request.headers['Avalon-Api-Key'] = 'secret_token'
+    context 'as administrator' do
+      before(:each) do
+        ApiToken.create token: 'secret_token', username: administrator.username, email: administrator.email
+        request.headers['Avalon-Api-Key'] = 'secret_token'
+      end
+
+      it "should notify administrators" do
+        login_as(:administrator) #otherwise, there are no administrators to mail
+        # mock_email = double('email')
```
masaball: I see that this was just copy pasted, but would it be worth it to clean out these bits of commented out stuff?

masaball (commented):
Have a few suggestions for improvements, but this is looking pretty good so far

spec/controllers/admin_collections_controller_spec.rb
cjcolvar: Totally.  I was getting lazy.

app/views/admin/collections/new.html.erb
cjcolvar: :+1: I copy-pasted from `app/views/admin/units/remove.html.erb` so I can try DRY-ing it up for all of them with a local param for the element id.

app/controllers/admin/collections_controller.rb
cjcolvar: I started down a path like that then back off once I figured out that there are a number of different uses of this method which makes things tricky.  There is the collection information modal form which I think depends on the `update_attributes` but the rest of the collection edit (really `show` view) that does access control probably relies on the `save`.  Not to mention API updates.  I put in the `skip_save` because I was trying to have the ability check failing not persist anything while still responding with a JSON 422 with error messages (and not a 401 from an `authorize!` or a 200 with no errors).  I was thinking that it is technically possible to have other edits (like description) save while the unit update fail but decided that was too complicated and led to weirdness with the JSON responses as they are now.  

This is all to say that I think the double save were there before so I'd like to leave both the `update_attributes` and `save` there for now and create a ticket to try and untangle this later.  (And the long explanation above can be some documentation for the ticket.)

I can wrap the two lines with the `skip_save` to avoid repeating the guard.

masaball (approved):
Looks great!

PR #6783, opened 2026-04-09
Ensure current unit is in list of candidate units

cjcolvar:
Fixes a case where a manager of a collection can change the unit of their collection but current unit would not appear in list if they were not a unit admin of that unit.


masaball (approved):
Looks good to me

Issue #6775, opened 2026-04-06
Update Manage Content Dashboard


joncameron:
## Description
Carrying on from #6693, refine the layout proposed and implement for the Manage/admin page.

## Notes
- Unit carousel can be a table to match the collections table in mockup
- MCO has around 60 units, so a table feels like a better match
- The overview data with icons at the top could be removed
- No need to update current mockup; can start making changes directly in Avalon
- Need to tidy up forms that direct to /admin/collection or collection index routes

## Questions
- How will we place/represent the Create Collection and Create Unit buttons?
  - Keep them in the current position of mockup

## Done Looks Like
- [ ] Initial implementation of new layout - making changes directly to code

joncameron:
The mockup may not be 1:1 in code because of our use of Bootstrap; could still use React bootstrap code from Figma.

There's already a table component for other pages in Avalon; we can re-use that for the manage content page (+1). Figuring out styling will be more complicated, because the code skeleton will work but won't match exactly with our styling currently in place.

Create Collection in Manage Content page: leave Unit undetermined
Create Collection within the Unit Admin page: autopopulate "create collection" form with name of Unit

Dananji:
The `Dashboard` component fetches units and collections for the current user from a JSON endpoint and renders them as two independent sections as before.

In the new UI,
- to edit a unit/collection - click the edit icon in the respective data row
- to view a unit/collection at `/units/id` or `/collections/id` - click the unit/collection name
- to see the filtered collections in a unit in Blacklight - click the number of collections link in the respective data row in units table
- to see the filtered items/unpublished items in a collection in Blacklight interface - click the number of items/unpulished text in the respective data row in collections table
- to view all units/collections - click 'View All' link at the bottom of each table, takes the user to `/admin/units`/`/admin/collections`

I didn't change the `Cancel` button path from the create form for unit/collection that takes the user back to `/admin/units`/`/admin/collections` page respectively as part of this work because it is not a part of this UI. Was this needing to be changed as per the notes? 
> Need to tidy up forms that direct to /admin/collection or collection index routes

@joncameron I kept the administrators column in units table as both the original design in Avalon and the Figma units carousel had this information. If this was intended to be removed please let me know. I can make that change.

joncameron:
Looks amazing! I created #6795 for any further changes or tweaks.

PR #6785, opened 2026-04-14
Manage content UI

Dananji:
Related issue: #6775

Changes in this PR:
- adds JSON endpoint to the dashboard index  that fetches units and collections for the current user
- adds `Dashboard` React component that fetches from the JSON endpoint and renders `Units` and `Collections` sections
- adds `UnitsTable` and `CollectionsTable` components backed by the existing `GenericTable` infrastructure, with column definitions, sort config, search, pagination, and edit/delete action buttons

units-table
collections-table

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -64,7 +64,12 @@ class Admin::DashboardController < ApplicationController
   def index
     respond_to do |format|
       format.html
-      # TODO: Figure out a json response?
+      format.json do
+        render json: {
+          units: @units.map { |u| u.as_json(nil) },
+          collections: @collections.map { |c| c.as_json(nil) }
+        }
+      end
     end
   end
 end
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -66,8 +66,14 @@ class Admin::DashboardController < ApplicationController
       format.html
       format.json do
         render json: {
-          units: @units.map { |u| u.as_json(nil) },
-          collections: @collections.map { |c| c.as_json(nil) }
+          units: @units.map do |u|
+            hash = u.as_json(nil)
+            can?(:destroy, u) ? hash.merge(remove_url: remove_admin_unit_path(u.id)) : hash
+          end,
+          collections: @collections.map do |c|
+            hash = c.as_json(nil)
+            can?(:destroy, c) ? hash.merge(remove_url: remove_admin_collection_path(c.id)) : hash
+          end
         }
       end
     end
--- a/app/javascript/components/dashboard/CollectionsTable.jsx
+++ b/app/javascript/components/dashboard/CollectionsTable.jsx
@@ -60,11 +60,11 @@ const CollectionsTable = ({ data, helpers }) => {
       unit: row.unit || '',
       collection_url: `/collections/${row.id}`,
       edit_url: `/admin/collections/${row.id}`,
-      remove_url: `/admin/collections/${row.id}/remove`,
+      // Collection remove_url = '' if the user doesn't have permission to destroy it
+      remove_url: row.remove_url || '',
       search_url: `/catalog?f[collection_ssim][]=${encodeURIComponent(row.name || '')}`,
       unpublished_url: `/catalog?f[collection_ssim][]=${encodeURIComponent(row.name || '')}&f[workflow_published_sim][]=Unpublished`,
       items: row.object_count?.total || 0,
-      total_items: row.object_count?.total || 0,
       unpublished_items: row.object_count?.unpublished || 0,
       description: row.description || '',
     }),
@@ -82,7 +82,7 @@ const CollectionsTable = ({ data, helpers }) => {
           return (
             <div>
               <a href={item.search_url} className="fw-bold text-decoration-none text-dark">
-                {item.total_items} {item.items === 1 ? 'item' : 'items'}
+                {item.items} {item.items === 1 ? 'item' : 'items'}
               </a>
               {item.unpublished_items > 0 && (
                 <div>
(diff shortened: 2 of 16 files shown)


app/javascript/components/dashboard/CollectionsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+          return (
+            <a href={item.collection_url} className="fw-bold text-decoration-none text-dark" data-testid="collection-name-table">
+              {item.name}
+            </a>
+          );
+        case 'items':
+          return (
+            <div>
+              <a href={item.search_url} className="fw-bold text-decoration-none text-dark">
+                {item.total_items} {item.items === 1 ? 'item' : 'items'}
```
cjcolvar: Could this be changed to `items` to match the approach in `UnitsTable`?  If so then I think `total_items` can be removed from the data mapping above (also like `UnitsTable`).
```suggestion
                {item.items} {item.items === 1 ? 'item' : 'items'}
```

app/javascript/components/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,175 @@
...
+ * Shared edit/delete action button pair used by dashboard table rows
+ */
+const DashboardActionButtons = ({ editUrl, removeUrl, editTitle, deleteTitle, deleteTestId }) => (
+  <div className="dashboard-action-icons">
+    <a href={editUrl} className="dashboard-action-edit btn btn-outline" title={editTitle}>
+      <i className="fa-regular fa-pen-to-square" />
+    </a>
+    <a href={removeUrl} className="dashboard-action-delete btn btn-danger" title={deleteTitle} data-testid={deleteTestId}>
+      <i className="fa-regular fa-trash-can" />
+    </a>
```
cjcolvar: What do you think about including the textual button labels to be consistent with playlists/timelines?

```suggestion
    <a href={editUrl} className="dashboard-action-edit btn btn-outline" title={editTitle}>
      <i className="fa-regular fa-pen-to-square" /> Edit
    </a>
    <a href={removeUrl} className="dashboard-action-delete btn btn-danger" title={deleteTitle} data-testid={deleteTestId}>
      <i className="fa-regular fa-trash-can" /> Delete
    </a>
```

app/javascript/components/dashboard/CollectionsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+    columns: [
+      { key: 'name', label: 'Name', sortable: true, dataType: 'string', width: '22%' },
+      { key: 'items', label: 'Items', sortable: true, dataType: 'number', width: '12%' },
+      { key: 'unit', label: 'Unit', sortable: true, dataType: 'string', width: '18%' },
+      { key: 'description', label: 'Description', sortable: false, width: '40%' },
+      { key: 'actions', label: '', sortable: false, width: '8%' },
+    ],
+
+    // View all link for collections
+    viewAllUrl: '/admin/collections',
```
cjcolvar, line 54: I was thinking that maybe this dashboard would replace the `admin/collections` and `admin/units` index routes and wouldn't link to them.  @joncameron Is this what you were thinking?

app/javascript/components/dashboard/CollectionsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+    // Parsing function to extract data from back-end
+    parseDataRow: (row) => ({
+      id: row.id,
+      name: row.name || '',
+      unit: row.unit || '',
+      collection_url: `/collections/${row.id}`,
+      edit_url: `/admin/collections/${row.id}`,
+      remove_url: `/admin/collections/${row.id}/remove`,
```
cjcolvar: Unfortunately, the delete action is only available to certain users and not all users who have edit permissions. In app/views/modules/_admin_table.html.erb, that I think this is replacing, we do a can?(:destroy, item) ability check for each collection/unit.

Given this I think we'd probably have to pass this information through the JSON somehow. Not sure if it would be cleaner to have a top-level key like deletable_collections/deletable_units (with an array of ids) or an attribute on each collection/unit like remove_url. I'm currently leaning towards the latter? I'll add a code suggestion in the dashboard controller.

```suggestion
      remove_url: ${row.remove_url} || '',
```

app/controllers/admin/dashboard_controller.rb
```diff
@@ -64,7 +64,12 @@ def load_and_authorize_units
   def index
     respond_to do |format|
       format.html
-      # TODO: Figure out a json response?
+      format.json do
+        render json: {
+          units: @units.map { |u| u.as_json(nil) },
+          collections: @collections.map { |c| c.as_json(nil) }
```
cjcolvar: I think this should work for adding the remove_url attribute to each unit/collection:
```suggestion
          units: @units.map do |u|
            hash = u.as_json(nil)
            hash.merge(remove_url: remove_admin_unit_path(u)) if can(:destroy, u)
          end,
          collections: @collections.map do |c|
            hash = c.as_json(nil)
            hash.merge(remove_url: remove_admin_collection_path(c)) if can(:destroy, c)
          end
```

cjcolvar (commented):
I'm excited to see the improved admin dashboard and the standardization of tables across end-user and cataloging views!  I see there are some failing tests because of the changed views but they can probably wait until things solidify.

Dananji:
Still working on fixing the failing tests

app/javascript/components/dashboard/CollectionsTable.jsx
Dananji: Absolutely! 👍

app/javascript/components/dashboard/Dashboard.jsx
Dananji: Yes, that sounds good!!

Dananji:
After adding `remove_url` to the server response for each unit/collection depending on user's permissions to destroy a unit/collection, the delete button does not get populated for a user who is added as a depositor at unit-level as follows;

 unit-depositor-view

cjcolvar (approved):
Looks great!  This is fine to merge now because my other question about removing the collections and units index routes can be handled in a different PR if we decide to do that.

PR #6798, opened 2026-04-20
Set a unit_id on the unsaved collection instance for easier ability checking

### LINKED ISSUE #6796: Error Creating Collection as Unit Admin
## Description
As a Unit Admin but not Administrator, selecting "Create Collection" on the Manage Content page pops an error. This does not happen when creating a collection using the button on the Unit admin page (that seems to work fine).

## To Reproduce
- On avalon-dev,
- Switch to a Unit Admin account (ex: "joncameron")
- Go to the Manage Content page
- Select "Create Collection"
- A NoMethodError page returns

Image

cjcolvar:
Resolves #6796


masaball (approved):
Looks great

Issue #6795, opened 2026-04-20
Manage Content Dashboard Tweaks


joncameron:
## Description

Remove margin from bottom of table:

```
.dashboard-table-container table.generic-table {
    margin-bottom: 1rem; # remove or set to 0rem
}
```

Change Edit button to View, Change hyperlinked collection name target to collection admin page

Remove Unit Administrators column from Units table

"View all": can this fully expand table, rather than going to /admin/units or /admin/collections?

"Create Unit" can also be btn-outline class instead of btn-primary

Move headings and create buttons into tables

cjcolvar:
>Move headings and create buttons into tables

How about removing the border instead of moving these inside the border?  I think this might look more consistent matching the playlist, timeline, and encode dashboard.

>"View all": can this fully expand table, rather than going to /admin/units or /admin/collections?

Should we replace this with an "all" option in the per page drop down? Or leave the "View all" which would effectively do this?  Or remove it entirely since we don't have an analogue in other tables in avalon?

joncameron:
> "View all": can this fully expand table, rather than going to /admin/units or /admin/collections?

My preference would be to have both the "all" in the per page drop down and "View all", and the "View all" select action changing the current number of entries to "all".

joncameron:
Dropping the top border could be a good approach to save a bit of space. Here's what I was thinking for moving the headers down:
<img width="1349" height="509" alt="Image" src="https://github.com/user-attachments/assets/1a5f0a55-4fb9-49f7-bcb6-1dcc526a2b14" />

<img width="1341" height="507" alt="Image" src="https://github.com/user-attachments/assets/bbf7fce4-0c94-48d6-a288-c8dd2dd888bd" />

joncameron:
- Add underlines to links on hover

joncameron:
- Remove "View all" text and link
- Add item limit value in table to local storage (persistent for users across pages and reloads)
- Current cap of 50 is good

joncameron:
- Unit default number of entries: 5, 10, 25, 50
- Collection default number of entries: 10, 25, 50, 100

joncameron:
Working great!

PR #6799, opened 2026-04-20
/admin/units and /admin/collections index routes subsumed by admin dashboard

cjcolvar:
Follow up to #6785; related to #6795


Dananji (approved):
👍

Issue #5997, opened 2024-08-20
Add check to SpeedyAF proxies to check that solr doc has_model_ssim matches expected model given proxy class name


cjcolvar:
Occasionally a request is made to the incorrect route for a given object (e.g.` /media_objects/<masterfile-id>`).  This used to raise an `ActiveFedora::ModelMismatch` but now raises a NoMethodFound error when using SpeedyAF proxies.  ActiveFedora has a check that the model class returned matches the model class requested.  Avalon should implement something similar in SpeedyAF to catch these errors and raise them.

See https://app.honeybadger.io/projects/54117/faults/110462479

cjcolvar:
Resolved by https://github.com/samvera-labs/speedy_af/pull/21

joncameron:
On avalon-dev I now receive a redirect to home and error flash message with text "Requested resource type does not match type of #{id}."

PR #6806, opened 2026-04-22
Add catch for new SpeedyAF::ModelMismatch error

masaball:
Related issue: #5997


cjcolvar (approved):
Looks great!  Good work catching the various places the error needed to be handled.

PR #6808, opened 2026-04-22
Make admin collection and unit page titles a little clearer

cjcolvar:
Was "Manage Units"; now...
image
Was "Manage Content"; now...
image


masaball (approved):
Looks great!

Issue #6767, opened 2026-03-31
Handle publishing during Media Object creation via API


masaball:
## Description
Since we are adding checks for accessibility compliance that prevent publication, how should we handle the publishing action when creating via the API? Currently, there is just a `publish: true` flag that can be passed during API creation. With the default publishing enforcement behavior being implemented, any API creation post compliance date (2026-04-26) with the publish flag will be blocked.

## Possible Implementations
* Enforce accessibility compliance for API creation
* Bypass enforcement for API creation for all users (keep current behavior)
* Bypass enforcement for API creation for certain users (Admins, Collection managers?)
* Provide an additional flag for explicitly overriding enforcement
* Provide an additional flag for certain users

## Questions
* Are there other implementations that might be better?
* Are there any concerns/considerations we need to keep in mind when choosing a way forward?

## Additional Notes
* API creation is typically going to be limited to power users who should be able to be trusted to determine if it is okay to publish an item without captions/transcripts

masaball:
@joncameron I've confirmed that adding `override_accessibility: true` to the creation payload works for publishing via the API. So the JSON just needs to include `publish: true, override_accessibility: true`. If only publish is provided, the user will receive an error about publication not being allowed and the media object will not be created. I am not sure how we want to phrase things in the documentation, but it should just be updating the example payload and adding an entry for "override_accessibility" [in this section](https://samvera.atlassian.net/wiki/spaces/AVALON/pages/1957954917/REST+API#Create-new-media-object)

joncameron:
Tested and REST API docs updated.

PR #6814, opened 2026-04-23
Prevent publication of API media object without masterfile

masaball:
This is a tie in to #6767. It turned out that we already had everything in place for that ticket and did not need to make any code changes, we just need to update documentation. 

However, I did find an edge case where `avalon_publisher` can be passed in as part of the metadata payload for object creation. When that is passed in and no master files are provided, none of the regular API publishing workflow is touched and so the empty object gets published. This does not seem like behavior we want, so I made a change to strip that metadata field when performing the 'create' action.


cjcolvar (approved):
~This makes sense but I think we may have to leave the old behavior so that intercom push can create published items without losing the `avalon_publisher` field.~

This makes sense and since the `avalon_publisher` field gets automatically set to `REST API` when published there isn't any risk of losing information sent via intercom push or the API.  I could foresee us possibly wanting to change this behavior to allow retaining this bit of information but for now this totally works.

Issue #6804, opened 2026-04-21
Accessibility Compliance UI Changes


joncameron:
## Description
- "Exempt this item" check box appears in the view page, to the right of existing admin buttons
- Publish button appears even when the item is not exempt
- Selecting the Publish button displays a popover menu if the item is no exempt, prompting the user to exempt and publish or cancel the action

- The check box updates the status of value without a page refresh
  - Possible but could take more time to figure out, we won't do this right now.
  - 

## Done Looks Like
- [ ] Text updated for displaying on the item view page rather than in a panel in Manage Files
- [ ] "Exempt this item" check box appears in the view page, to the right of existing admin buttons
- [ ] Publish button appears even when the item is not exempt
- [ ] Selecting the Publish button displays a popover menu if the item is no exempt, prompting the user to exempt and publish or cancel the action

Dananji:
The related PR moves the accessibility compliance exemption check to the item view page from manage files page. 

The behavior of the checkbox on item view page;
- Publish button is created with a confirmation popover with an exempt and publish action
- When the checkbox is checked/unchecked, media object's `override_accessibility` attribute in the server is updated via an AJAX request. This also updates the status of the publish button so that, publish action works as usual.
- After publishing an item with exemption the checkbox is displayed checked on the page to permitted users

New publish button behavior with the checkbox;
- If the user clicks the publish button when exempt checkbox is checked, the publish action works as usual
- If the user clicks the publish button when exempt checkbox is unchecked, a popover is displayed to either exempt and publish or cancel the publish action

After publication, the checkbox remains on the page like before on the manage files page. Only the checkbox behavior is changed with the AJAX request format.

joncameron:
The new workflow works great and has received positive feedback from all of the collection staff users I've previewed it to.

PR #6815, opened 2026-04-24
A11y compliance UI

Dananji:
Related issue: #6804 

Changes in this PR:
- add `toggle_accessibility_exempt` `PUT` endpoint on `MediaObjectsController` to set `override_accessibility` via AJAX
- move the a11y exempt checkbox from the `Manage Files` workflow step to the item view page, which toggles the media object state for a11y exemption via AJAX with no page reload
- change the Publish button so that, a popover is shown when publishing without exempting a non-accessible item
- generalize the `shown.bs.popover` focus handler and `keydown` focus trap in `button_confirmation.js` to cover all popovers


cjcolvar (commented):
We have the `update_status` endpoint but I'm hesitant to make another endpoint for a single attribute instead of using `update`.  @masaball [mentioned](https://github.com/avalonmediasystem/avalon/issues/6767#issuecomment-4306867860) that passing `override_accessibility` into a `create` call worked so I'm wondering if the same could be done here with 'update'.  Similarly I would hope that the 'exempt and publish' action could be done in one request to the server with the existing endpoints (maybe by having `update_status` accept `override_accessibility`?).  Do you think that makes sense and is doable?

app/controllers/media_objects_controller.rb
```diff
@@ -701,4 +718,10 @@ def master_files_params
   def api_params
     params.permit(:collection_id, :publish, :import_bib_record, :replace_masterfiles)
   end
+
+  def apply_accessibility_override(media_object)
+    return unless params[:override_accessibility].present? && can?(:override_accessibility, media_object)
+    media_object.override_accessibility = params[:override_accessibility] == '1'
+    media_object.save(validate: false)
```
cjcolvar: Can this `save` be moved out of this method into `update_status` in the `status.blank?` block?  The other code branch (with a `status`) does a save so it doesn't need the save here.

app/views/media_objects/_item_view.html.erb
```diff
@@ -28,14 +28,24 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+      <%= render 'administrative_links' %>
+      <% if current_ability.can? :edit, @media_object %>
+        <% if lending_enabled?(@media_object) %>
+          <%= render 'thumbnail' if can_stream %>
+        <% else %>
+          <%= render 'thumbnail' %>
+        <% end %>
       <% end %>
+    </div>
+    <% if can? :override_accessibility, @media_object %>
```
cjcolvar: Should the checkbox only show if accessibility enforcement is enabled?
```suggestion
    <% if Settings.accessibility_compliance.enabled && can?(:override_accessibility, @media_object) %>
```

app/views/media_objects/_administrative_links.html.erb
```diff
@@ -22,7 +22,15 @@ Unless required by applicable law or agreed to in writing, software distributed
     <%= link_to 'Unpublish', update_status_media_object_path(@media_object, status:'unpublish'), method: :put, class: 'btn btn-primary', data: {testid: 'media-object-unpublish-btn'} %>
     <% end %>
     <% else %>
-    <%= link_to 'Publish', update_status_media_object_path(@media_object, status:'publish'), method: :put, class: 'btn btn-primary', data: {testid: 'media-object-publish-btn'} %>
+    <button class="btn btn-primary" id="publish-btn"
+      data-publish-url="<%= update_status_media_object_path(@media_object, status: 'publish') %>"
+      data-testid="media-object-publish-btn"
+      <% if can?(:override_accessibility, @media_object) %>
```
cjcolvar: Should this also only show if accessibility compliance is enabled?

cjcolvar (commented):
Sorry for being slow to finish reviewing.  What do you think about having the checkbox post to `update` instead of `update_status` to keep the endpoints scoped semantically.  The publish button and popover can still post to `update_status`.  I'm thinking that there are probably other cases (like API requests) where we'll want to handle the `override_accessibility` attribute.  Maybe this would as easy as putting a call to `apply_accessibility_override` in `update_media_object`?

app/views/media_objects/_item_view.html.erb
Dananji: Yes 👍

Dananji:
@cjcolvar As far as I understand the following changes are requested. It'd be great if you could clarify. Thanks!
1. Move `save` out of `apply_accessibility_override` to avoid duplicated saves
2. Add `apply_accessibility_override` to the `update_media_object` method. Adding this just before the media object validation code in the method seems good place to me, unless there is a better place.
3. Keep the `apply_accessibility_override` call in the `update_status` for the `status.blank?` branch
4. Guard the display of both toggle and pop-over with `Settings.accessibility_compliance.enforce` in the view

cjcolvar (approved):
Thanks for the revisions.  This looks great!

Issue #6081, opened 2024-10-22
Objects and Supplemental Fiels Controllers Makes Unneccessary Fedora Requests


joncameron:

joncameron:
Supplemental files controller change should be simple; may be more complicated with the objects controller.

joncameron:
👍

PR #6816, opened 2026-04-28
Use SpeedyAF proxies for non-save actions

masaball:
Related issue: #6081

Fedora requests were being made on all action types for supplemental file controller and the show action of the objects controller. This commit adds a `fetch_proxy` method so we can retrieve object proxies for any controller actions that do not require a save, reducing unnecessary calls to fedora.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -263,6 +263,12 @@ class ApplicationController < ActionController::Base
     obj || GlobalID::Locator.locate(id)
   end
 
+  def fetch_proxy(id)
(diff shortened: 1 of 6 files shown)


cjcolvar (commented):
This looks great!  If possible can you add some tests to the controller actions that were changed to ensure that it doesn't read from fedora?  That would make sure we didn't regress by accident in the future.

cjcolvar (commented):
Looks good!  Do you want to make this ready for review?  I think it would be nice to include in this release.

cjcolvar (approved):
👍

PR #6817, opened 2026-04-28
Fix and test issues with disable inheritance

cjcolvar:
Inherited non-admin roles and restricting visibility from non-private to private would return true in `is_exclusively_inherited_from_parent?` leading to inherited managers/editors/depositors being able to discover and edit but not view items as well as end users being able to discover but not view items.

The problem with restricting visibility from non-private to private with disabled inheritance is that non-private visibility is encoded using a read_group while private is the absence of a read_group.  While this isn't wrong in this case the current logic wasn't able to handle cases where a user was explicitly added to an item's special access because the group exclusive inheritance check would always return true.

For example if the parent collection is `public` and the item is `private` then:
- `(@user_groups & media_object.read_groups).empty?` is true  because `media_object.read_groups` is empty when private
- `!(@user_groups & media_object.inherited_read_groups).empty?` is true because `@user_groups` will always include `public` (and `registered` when logged in) and the collection includes `public`

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -211,7 +211,7 @@ class Ability
       # end
 
       cannot :read, [MediaObject, SpeedyAF::Proxy::MediaObject] do |media_object|
-        media_object.disable_inheritance? && is_exclusively_inherited_from_parent?(media_object)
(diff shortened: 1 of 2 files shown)


masaball (approved):
Good catch! Those are some gnarly permissions checks but looks good to me.

Issue #6811, opened 2026-04-22
500 When Updating Collection Name


joncameron:
## Description
POST to admin/collections/#{id}.json returns 500 when the collection name contains a colon character (':').

## To reproduce
- On avalon-dev:
- Go to a collection admin page
- Select "Edit Collection Info"
- Enter a string with a colon character. ex: 'Spiders of the Market: Ghanaian Trickster Performance in a Web of Neoliberalism'
- Select "Update"
- The button switches to an endless "Saving..." state

## Error Message
RSolr::Error::Http in Admin::CollectionsController#update

RSolr::Error::Http - 400 Bad Request
Error: org.apache.solr.search.SyntaxError: Cannot parse 'has_model_ssim:"Admin::Collection" AND name_uniq_si:spidersofthemarket:ghanaiantricksterperformanceinawebofneoliberalism': Encountered " ":" ": "" at line 1, column 70.
Was expecting one of:

joncameron:
👍

PR #6818, opened 2026-04-29
Escape solr query in uniqueness validator

masaball:
Related issue: #6811


cjcolvar (commented):
This looks great!  Can you please add a test so we don't accidentally regress on this in the future?

cjcolvar (approved):
Looks good!

PR #6823, opened 2026-04-30
Set minimum date in end-date picker if a begin date exists

Dananji:
When selecting a date range using begin and end date-pickers, in access controls the end date picker always shows up from today's date onward. So, when the user has already selected a future begin date they have to navigate the end date picker a date past the begin date to select a valid end date. Setting the selected begin date as the minimum date in the end date picker allows the user to easily select an end date.


masaball (approved):
👍

PR #6824, opened 2026-04-30
Fix CDL message styling and bug in access controls

Dananji:
Changes in this PR:
- guard against `nil` for `@media_object.lending_period` when parent permissions is disabled in item's access controls with CDL turned ON at the collection level
- center the CDL message in audio containers

Before: 
before
After:
after

### DIFF
--- a/app/views/media_objects/_access_control.html.erb
+++ b/app/views/media_objects/_access_control.html.erb
@@ -41,7 +41,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <div class="mb-3">
         <% if @media_object.disable_inheritance? %>
         <%=  render partial: "modules/tooltip", locals: { form: vid, field: :lending_period, tooltip: t("access_control.#{:lending_period}"), options: {display_label: (t("access_control.#{:lending_period}label")+'*').html_safe} } %><br />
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks great!

app/views/media_objects/_access_control.html.erb
```diff
@@ -41,7 +41,7 @@ Unless required by applicable law or agreed to in writing, software distributed
       <div class="mb-3">
         <% if @media_object.disable_inheritance? %>
         <%=  render partial: "modules/tooltip", locals: { form: vid, field: :lending_period, tooltip: t("access_control.#{:lending_period}"), options: {display_label: (t("access_control.#{:lending_period}label")+'*').html_safe} } %><br />
-        <% d, h = (@media_object.lending_period/3600).divmod(24) %>
+        <% d, h = ((@media_object.lending_period || 0)/3600).divmod(24) %>
```
cjcolvar, line 44: If `lending_period` is nil then it looks like this would make a lending period of 0 days/hours instead of using the default period.  I think `active_lending_period` should fix this.
```suggestion
        <% d, h = (@media_object.active_lending_period/3600).divmod(24) %>
```

Issue #6820, opened 2026-04-30
Batch Entries Fail When Caption Information is Present


joncameron:
## Description
In an example batch manifest containing a row with caption file location, label and language, the entry fails and is unable to be processed. The completion email reports a "no implicit conversion of StringIO into String" error:

> The following rows of your spreadsheet failed:
> Row	Error
> 5	no implicit conversion of StringIO into String

## To Reproduce
- Copy the files from this example batch into a collection folder for an Avalon 8.1 instance: https://drive.google.com/file/d/1zBkpx7fi0lTINPwnaUImb4EbA4VqT1yi/view?usp=drive_link
- Eventually an "Errors Present" email will be sent indicating the failure of Row 5 (containing Lunchroom Manners and the columns for caption file info)
- The media object will be created but there will be no media file attached

joncameron:
👍

PR #6826, opened 2026-05-01
Refactor SupplementalFile#attach_file

masaball:
Related issue: #6820

A bug was introduced in commit 079b20e when the `io` option was added to the #attach_file method. This involved us passing the wrong type to the method in the case of batch processing, causing batches containing captions/transcripts to error out.

This commit refactors the `io` mode to utilize a FileLocator instance with filename provided. This should allow us to be more intentional, clear, and consistent with how the method is used.

### DIFF
--- a/app/models/supplemental_file.rb
+++ b/app/models/supplemental_file.rb
@@ -34,10 +34,11 @@ class SupplementalFile < ApplicationRecord
   after_destroy_commit :remove_from_index
   before_save :default_label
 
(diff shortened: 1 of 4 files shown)


cjcolvar (commented):
This looks good to me.

I'm thinking of a follow up ticket to change `attach_file` so that it doesn't use the `io` kwarg anymore but does class matching on `new_file` to determine which path it takes.  This could also allow for an else clause that would raise an ArgumentError for unhandled types.

cjcolvar (approved):

Issue #6812, opened 2026-04-22
Support Exemption Status in Batch Manifest


joncameron:
## Description
Adding the new exemption value to batch manifests.

## Done Looks Like
- [x] Exemption status added as a possible property for media objects in batch manifests
- [ ] Exemption status added as a column for example batch manifest
- [ ] Added to documentation on the wiki

joncameron:
👍

PR #6827, opened 2026-05-01
Allow accessibility exemption in batch manifest

masaball:
Related issue: #6812

### DIFF
--- a/lib/avalon/batch/manifest.rb
+++ b/lib/avalon/batch/manifest.rb
@@ -146,7 +146,6 @@ module Avalon
                 end
                 content.last[f] = f == :skip_transcoding ? Avalon::Batch.true_field?(values[i]) : values[i]
               else
-                f == :override_accessibility ? Avalon::Batch.true_field?(values[i]) : values[i]
                 fields[f] << values[i]
               end
             end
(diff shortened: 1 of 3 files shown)


lib/avalon/batch/manifest.rb
```diff
@@ -145,6 +146,7 @@ def create_entries!
                 end
                 content.last[f] = f == :skip_transcoding ? Avalon::Batch.true_field?(values[i]) : values[i]
               else
+                f == :override_accessibility ? Avalon::Batch.true_field?(values[i]) : values[i]
                 fields[f] << values[i]
```
cjcolvar: It looks like these two lines need to be combined somehow since `Avalon::Batch.true_field?` doesn't modify the value in place.  Line 147 directly stores the result.
```suggestion
                fields[f] << f == :override_accessibility ? Avalon::Batch.true_field?(values[i]) : values[i]
```

cjcolvar (commented):
Looking good overall.  I had one suggestion and I think it would be good to add a test or two to handle the cases:
- when the exemption is set ensure it doesn't throw the error
- when the exemption isn't set ensure it throws the error

cjcolvar (approved):
Looks great!

PR #6832, opened 2026-05-05
Use @samvera/rampv5.1.0 in the package.json for 8.2

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/941


cjcolvar (approved):
Looks good to me!

Issue #6750, opened 2026-03-23
Avalon 8.2 Release


joncameron:
## Description
Avalon 8.2 includes
- Units as Objects
- Permissions inheritance changes
- Accessibility workflows
- Ramp Updates and Upgrades

joncameron:
- [x] Final dependency upgrade for vulnerabilities
- [x] Build new docker image for hls streaming and put in docker-compose.yml files
- [x] Need to rebuild images due to copy fail vulnerability
- [x] Example batch manifest (in progress)
- [x] Release testing (Jon) (in progress)
- [x] Documentation (upgrade, known problems, configuration?) (all need to be updated)
- [x] Delete old searches as part of upgrade instructions? (drop table + recreate it?) (optional)
- [x] API documentation - need to add routes for units?
- [x] Figure out how to do token encryption migration in terraform
- [x] Upgrade demo site
- [x] Other administrative tasks for release

PR #6833, opened 2026-05-05
Tag SME for avalon 8.2 release

Dananji:
Related issue: #6750

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "react-bootstrap": "^2.10.9",
     "react-dom": "^19.1.0",
     "react-on-rails": "^14.2.1",
-    "react-structural-metadata-editor": "https://github.com/avalonmediasystem/react-structural-metadata-editor.git#b12c4fb",
+    "react-structural-metadata-editor": "https://github.com/avalonmediasystem/react-structural-metadata-editor#avalon-8.2",
     "sass": "^1.93.2",
     "sass-loader": "^13.3.2",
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
Does SME bundle any dependencies that we should worry about dependabot warnings or do those all get resolved in Avalon's package.json/yarn.lock?

Dananji:
SME's dependencies are not isolated, so they are resolved in Avalon's `yarn.lock` file. These get resolved as dependencies of SME, unless Avalon installs them directly via its `pacakage.json` as direct dependencies. 

Therefore, any dependency bundled into SME will trigger dependabot warnings if there are any vulnerabilities. In the current state there's one dependency `uuid` with a known vulnerability out of the dependencies brought in by SME.

If any of these dependencies get flagged by dependabot in Avalon, then we'll have to use `resolutions` in Avalon's `package.json` to override the vulnerable versions.

Do you think it's best to resolve existing ones in SME and bring a patched version (tag: avalon-8.2.1) in place of this?

Dananji:
New PR to bump dependencies in SME: https://github.com/avalonmediasystem/react-structural-metadata-editor/pull/279

cjcolvar:
Thanks for explaining how this works!  I think we can handle resolutions in avalon so I don't think there is a need for an 8.2.1 release right now.

cjcolvar (dismissed):
👍

cjcolvar (approved):

Issue #6024, opened 2024-09-06
Playlist editing/duplicating sometimes causes DB deadlock


cjcolvar:
Duplicating or changing order on a playlist sometimes causes a DB deadlock error

See Honeybadger reports:

PR #6835, opened 2026-05-06
Reduce database calls for Playlist#duplicate and Playlist#update_multiple

masaball:
Related issue: #6024

Major changes:

* Remove `.move_to_bottom` calls. 'acts_as_list' default appends newly
  created items to the end of the list rendering the explicit call
redundant.

* Refactor `PlaylistItem#duplicate!` to have an optional argument
  `to_playlist`. Default behavior will assign the item to the currently
active playlist. Providing `to_playlist` allows assigning the item to an
arbitrary user defined playlist.

* Refactor the `Playlist#duplicate` and `Playlist#update_multiple`
  methods to use the new argument on `PlaylistItem#duplicate!`. This
allows us to rely entirely on the logic in `PlaylistItem#duplicate!` and
avoid having to reassign and resave items. This should reduce a good bit
of the overhead on these methods and hopefully avoid deadlocks.

This PR also includes a commit to clean up the linting in the playlist controller and playlist controller spec.

Follow-up work: Look at the regular Playlist#update method. There may be improvements that could be made there (might be a good candidate for a retry functionality?).

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -133,10 +133,7 @@ class PlaylistsController < ApplicationController
       # copy items
       old_playlist.items.each do |item|
         next if item.clip.master_file.nil?
-        copy_item = item.duplicate!
-        copy_item.playlist_id  = @playlist.id
-        copy_item.save!
-        copy_item.move_to_bottom
+        item.duplicate!(to_playlist: @playlist)
       end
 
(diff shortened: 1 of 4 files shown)


cjcolvar (approved):
Looks great!  Love how this clean things up.  Seems like this code hasn't been touched in years so it's great for it finally get some attention.

PR #6837, opened 2026-05-06
Fix AJAX save for master file form in file-upload step

Dananji:
This PR restores the AJAX save in the master file form in the file upload step. I think this broke during the `jQuery` removal work, which made the save button click to reload the entire page.


masaball (approved):
👍

PR #6838, opened 2026-05-11
Cypress refinements

cjcolvar:
This PR has a couple small fixes as well so it would be good to get it merged for 8.2.

### DIFF
--- a/app/assets/javascripts/add-playlist-option.js
+++ b/app/assets/javascripts/add-playlist-option.js
@@ -102,7 +102,17 @@ this.add_new_playlist_option = function () {
             // Set "Add new playlist" order 0 to make it appear on top of playlists
             option.$order = (option.text === addnew) ? 0 : 1;
           });
-        },
+          // Set ARIA attributes for tom-select controls flagged by SiteImprove
+          const dropdownList = query('#post_playlist_id-ts-dropdown');
+          if (dropdownList) {
+            dropdownList.setAttribute('aria-label', 'list of playlists');
+          }
+          const playlistsCombobox = query('#post_playlist_id-ts-control');
+          if (playlistsCombobox) playlistsCombobox.setAttribute('aria-labelledby', 'post_playlist_id');
+          // Add data-testid attributes for testing
+          if (this.control_input) this.control_input.setAttribute('data-testid', 'media-object-playlist-search-input');
+          if (this.control) this.control.setAttribute('data-testid', 'media-object-playlist-dropdown-control');
+	},
         allowEmptyOption: false,
         hideSelected: false,
         closeAfterSelect: true,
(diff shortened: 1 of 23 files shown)


Dananji (commented):
Wow! This is a lot of work... Code-wise, it looks great as far as I can tell 🎉 

I tried running this locally using the interactive mode, and got the following error;
```
index-h9zRIGOb.js:96625 TypeError: Cypress.expose is not a function

Because this error occurred during a `before each` hook we are skipping all of the remaining tests.
    at Context.eval (webpack:///../../node_modules/@bahmutov/cy-grep/src/support.js:36:0)
```
I'm not sure whether this is a local issue of my dev environment. What I did,
- checked out the branch, ran yarn install because of the cypress version bump
- ran cypress in interactive mode
- once the tests open in the selected browser clicked on a spec file, which gave me this error

cjcolvar:
@Dananji Thanks for the review!  Sorry for not splitting this into smaller commits/PRs.  

I think you're seeing the error due to an old version of cypress.  Try running `npx cypress --version` and make sure that you have 15.10+.

Dananji:
@cjcolvar No worries! It seems it would add a lot more work to break it down to multiple PRs because of the shared test setup code.

After updating cypress I was able to run the tests 🎉 

But I ran into the following issue,
```
index-CyEuMe1-.js:1625 AssertionError: Timed out retrying after 10000ms: Expected to find element: `[data-testid='collection-name-table']`, but never found it.

Because this error occurred during a `after all` hook we are skipping the remaining tests in the current suite: `Collections Test`
    at CollectionPage.navigateToCollection (webpack:///./pageObjects/collectionPage.js:12:7)
    at CollectionPage.deleteCollectionByName (webpack:///./pageObjects/collectionPage.js:222:9)
    at Context.eval (webpack:///./integration/collection_spec.js:70:19)
```

This happens for spec files that go through the admin collection page to create an item for the tests. In my local Avalon instance I only see the automated units not the collections.
The automated collection creation doesn't seem to happen. In the interactive mode, I see the test setup creating the unit and go straight to searching for the collection which is non-existent.
Screenshot 2026-05-12 at 11 52 24 AM

The other tests that doesn't require this setup work great 🚀

cjcolvar:
I'm not sure what is going wrong now.  Maybe we could pair up sometime and I can watch it run?  My goal is to eventually make these tests more reliable and less flaky in all of our dev environments but in this PR I'm just hoping to improve them.

Dananji:
Yes, sounds great! Do you think we can merge this for now and work on the collection setup fix in a different PR? 

I think the goal of making them less flaky and getting them to work in an overall-sense is achieved in the current work!

Dananji (approved):
🚀

PR #6842, opened 2026-05-12
Unit migration improvements

cjcolvar:
Remove TODO because we'll be testing it and fixing anything as part of release.
Skip reindexing collection members because it isn't really needed.


masaball (approved):
👍

Issue #6831, opened 2026-05-04
Add Unit Breadcrumb Navigation to Admin Pages


joncameron:
## Description
Adding links to parent Units when viewing Collection pages.

cjcolvar:
#6843 looks great!

I think it would be nice to be consistent and add breadcrumbs:
- Admin units page: `Manage Content / Archives Unit`
- Item edit page: extend the breadcrumbs `Manage Content / Archives Unit / Default Collection / Untitled item`

joncameron:
💯

PR #6843, opened 2026-05-12
Breadcrumbs for admin collection pages

Dananji:
Related issue: #6831 

Screenshot 2026-05-12 at 2 25 59 PM

### DIFF
--- a/app/views/admin/collections/show.html.erb
+++ b/app/views/admin/collections/show.html.erb
@@ -15,6 +15,11 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <% @page_title = t('collections.show.title', :collection_name => @collection.name, :application_name => application_name) %>
 
+<ol class="breadcrumb">
+  <li style="padding-bottom: 0px" class="breadcrumb-item"><%= link_to 'Manage Content', admin_dashboard_path %></li>
+  <li style="padding-bottom: 0px" class="breadcrumb-item"><%= link_to @collection.unit.name, admin_unit_path(@collection.unit) %></li>
+  <li style="padding-bottom: 0px" class="breadcrumb-item active"><%= @collection.name %></li>
+</ol>
 <div class="page-title-wrapper">
   <h1 class="page-title">Manage Collection</h1>
 </div>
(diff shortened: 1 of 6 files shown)


app/views/admin/collections/show.html.erb
```diff
@@ -15,6 +15,11 @@ Unless required by applicable law or agreed to in writing, software distributed
 %>
 <% @page_title = t('collections.show.title', :collection_name => @collection.name, :application_name => application_name) %>
 
+<ol class="breadcrumb">
+  <li style="padding-bottom: 0px" class="breadcrumb-item"><%= link_to 'Manage Content', admin_dashboard_path %></li>
```
cjcolvar: Does it make sense to add `padding-bottom: 0px` to `.breadcrumb-item` or is that class used in other places that still wants to use a bottom padding?

app/views/admin/collections/show.html.erb
```diff
@@ -15,6 +15,11 @@ Unless required by applicable law or agreed to in writing, software distributed
...
+  <li style="padding-bottom: 0px" class="breadcrumb-item"><%= link_to @collection.unit.name, admin_unit_path(@collection.unit) %></li>
```
cjcolvar: Based upon a user's privileges they might not be able to go to the unit page.
```suggestion
  <li style="padding-bottom: 0px" class="breadcrumb-item"><%= can?(:read, @collection.unit) ? link_to( @collection.unit.name, admin_unit_path(@collection.unit)) : @collection.unit.name %></li>
```

cjcolvar (commented):
I love the breadcrumbs here!

When playing with this I noticed an issue that needs some thought.
If a user has a role in a collection but not a unit then the user isn't allowed to access the unit page.  We could either make that level of the breadcrumbs be text and not a link OR we could allow users with privileges under the umbrella of a unit to get a read-only view of a unit.  We could do the former for this PR and  then write up a ticket to think about it more.

app/views/admin/collections/show.html.erb
Dananji: Good catch! 

The breadcrumbs in media object edit page don't have the inline style. But removing the `padding-bottom: 0px` doesn't seem to change anything, so I guess the inline styling can be removed :)

Dananji:
Admin unit page:
unit-breadcrumbs

Admin collection page for non-admin user (with unlinked unit name): 
user-collection-breadcrumbs

Media object edit page for admin:
admin-collection-breadcrumbs

Media object edit page for a non-admin user (with unlinked unit name):
user-item-edit-breadcrumbs

cjcolvar (approved):
Looks great!

PR #6846, opened 2026-05-13
Build assets in production node environment for prod images

cjcolvar:
This is needed now with the new JS build environments to keep from getting jsxDev included in the assets which fails to load in the prod rails environment leading to all react components breaking.


Dananji (approved):
🚀

PR #6847, opened 2026-05-14
Fixes for 8.2 RC1

cjcolvar:


package.json
```diff
@@ -45,18 +49,13 @@
...
   },
   "devDependencies": {
-    "@babel/plugin-transform-runtime": "7",
-    "@babel/preset-env": "^7.29.5",
     "@bahmutov/cy-grep": "^3.0.4",
-    "babel-loader": "^10.0.0",
     "cypress": "^15.14.1",
-    "cypress-file-upload": "^5.0",
-    "html-webpack-plugin": "^5.3.2",
-    "prop-types": "^15.7.2",
```
masaball, line 58: This wasn't caught in the PR to dev, but was `prop-types` intentionally removed? Doesn't look like it made it up into the regular dependencies but was deleted here.

cjcolvar: When I was moving things I noticed that `prop-types` was already in `dependencies` so this one in `devDependencies` was a duplicate.

masaball (approved):
👍

Issue #5239, opened 2023-06-21
Investigate changes with IIIF Auth 2.0 and update auth service as necessary


masaball:
The IIIF Authorization flow spec has been updated to version 2.0 as of June 2, 2023: https://iiif.io/api/auth/2.0. We should see what has changed in the new specification and consider updating our AuthService constructor to match the most recent spec.

masaball:
Follow-up tickets: #6849, samvera-labs/ramp#320

PR #6848, opened 2026-05-14
Initial baseline changes for IIIF auth 2.0

masaball:
Related issue: #5239

TODO:
* Make auth probe its own endpoint instead of piggybacking on hls_manifest? Auth 2.0 probes are supposed to be GET requests, not HEAD requests so we would need to find another way to differentiate. May be easiest to use a separate method/endpoint.
* ~~Make auth service changes in the playlist workflow (playlists_controller, iiif_playlist models, etc)~~
* ~~Add auth 2 context info to top level manifest context~~
* ~~Fix and add spec tests~~
* Find a iiif auth 2 client to test against and test against Ramp once it has been updated for iiif auth 2.
* Discussion on whether we want to make any changes to the headers and messages of the auth service.
* We use I18n translations for the headers and messages. Should we be using { "en": ["message"] } for these entries? It should be within spec to have the language marked "none", maybe that is better than specifying a specific language. Or maybe we can pull the default locale information?

There are a couple minor changes to `iiif_manifest` that may be necessary. The `auth_service` method expects a simple hash so passing in an array of hashes results in `[[{service}, {service}]]`. Adjusting how the gem serializes `auth_service` may be desirable, or we may need to find a way to flatten the service definition. Additionally, we need to add a value to the top level manifest context. This is currently done in Avalon but it may be desirable to add a way to pass the value to `iiif_manifest` (or maybe there's already a way and I just missed it).

### DIFF
--- a/app/models/iiif_playlist_canvas_presenter.rb
+++ b/app/models/iiif_playlist_canvas_presenter.rb
@@ -235,7 +235,7 @@ class IiifPlaylistCanvasPresenter
     def auth_service(quality)
       [
         {
-          "context": "http://iiif.io/api/auth/1/context.json",
+          "@context": "http://iiif.io/api/auth/1/context.json",
           "@id": Rails.application.routes.url_helpers.new_user_session_url(login_popup: 1),
           "@type": "AuthCookieService1",
           "confirmLabel": I18n.t('iiif.auth.confirmLabel'),
--- a/app/models/iiif_playlist_canvas_presenter.rb
+++ b/app/models/iiif_playlist_canvas_presenter.rb
@@ -233,36 +233,69 @@ class IiifPlaylistCanvasPresenter
     end
 
     def auth_service(quality)
-      {
-        "context": "http://iiif.io/api/auth/1/context.json",
-        "@id": Rails.application.routes.url_helpers.new_user_session_url(login_popup: 1),
-        "@type": "AuthCookieService1",
-        "confirmLabel": I18n.t('iiif.auth.confirmLabel'),
-        "description": I18n.t('iiif.auth.description'),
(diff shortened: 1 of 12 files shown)


app/models/iiif_playlist_canvas_presenter.rb
```diff
@@ -233,36 +233,69 @@ def marker_attributes(marker)
...
-            "@id": Rails.application.routes.url_helpers.destroy_user_session_url,
-            "@type": "AuthLogoutService1",
-            "label": I18n.t('iiif.auth.logoutLabel'),
-            "profile": "http://iiif.io/api/auth/1/logout"
-          }
-        ]
-      }
+      [
+        {
+          "context": "http://iiif.io/api/auth/1/context.json",
```
cjcolvar:
```suggestion
          "@context": "http://iiif.io/api/auth/1/context.json",
```

cjcolvar (dismissed):
Great work on this so far!

I agree that it would probably be best to move the auth 2 context changes into `iiif_manifest`  along with fixing the bug with multiple auth services but that could be follow up work and doesn't need to be in this PR.

One thing I'm wondering is if we need to continue to support auth v1 and if it would simplify things to only support auth v2.  There probably aren't many clients that support v2 but I'm not sure if any clients ever worked with our auth v1.  Timeliner at one point worked with our auth v1 but we never really used the full flow since we embed timeliner in avalon.  So I'm not sure if it makes sense to retain auth v1 for timeliner or just upgrade it to auth v2.  Maybe we can move forward with this PR as-is as proof of concept and revisit it?

@joncameron @Dananji Do you have insight into what makes sense given what you know about the community and current ecosystem?

cjcolvar (approved):

PR #6850, opened 2026-05-14
Allow for enabling temprorary encryption config properties for migration using ENVVAR

cjcolvar:
In Avalon 8.2 API tokens are encrypted in the DB (#6289).  By default ActiveRecord disallows reading from existing records with unencrypted values.  In order to make migrating existing API tokens easier this PR wraps config properties that enable reading unencrypted values with an ENVVAR.  This way `config/application.rb` doesn't need to be edited, which is harder in a docker environment.  The migration then looks like:
```
ACTIVE_RECORD_ENCRYPTION_MIGRATION=true bundle exec rails r 'ApiToken.all.each(&:encrypt)'
```
After the migration is run then the special temporary properties are not needed.


masaball (approved):
Nice!

Issue #5037, opened 2023-01-20
Improve Response When Non-successful API Request is Sent


joncameron:

joncameron:
👍

PR #6853, opened 2026-05-20
Improve error handling for routing errors

masaball:
Related issue: #5037

After a lot of testing, it was determined that error handling for most/all API routes was already fairly robust. The most likely cause of the error reporting issue was users making their request with the wrong HTTP method, e.x. Media Object update is required to be a PUT. Users making the request via PATCH would encounter a 500 error for the incorrect route. This commit adds handling for this situation by creating a new route that catches all requests that don't match an existing route and directing them to specific error handling to provide a 404 page or JSON error, depending on request format. This is necessary because routing errors occur in middleware which cannot be rescued via standard `rescue_from` error handling.

Follow-up work could be to audit our API endpoints and see if there are any it might make sense to add, such as a PATCH option for API updates of Media Objects.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -274,4 +274,11 @@ Rails.application.routes.draw do
       end
     end
   end
+
+  # Routing errors are triggered in the middleware stack so can't be handled by
+  # a regular rescue_from block. Constraint is necessary to not break retrieval
+  # of things from active storage.
+  match '*unmatched', to: 'application#handle_routing_error', via: :all, constraints: lambda { |req|
+    req.path.exclude? 'rails/active_storage'
+  }
 end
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -274,4 +274,11 @@
       end
     end
   end
+
+  # Routing errors are triggered in the middleware stack so can't be handled by
+  # a regular rescue_from block. Constraint is necessary to not break retrieval
+  # of things from active storage.
+  match '*unmatched', to: 'application#handle_routing_error', via: :all, constraints: lambda { |req|
+    req.path.exclude? 'rails/active_storage'
```
cjcolvar, line 282: I did some manual testing and only found one place that appears to be blocked and needs excluding.
```suggestion
    req.path.exclude? 'rails/active_storage'
    req.path.exclude? 'persona/users'
```

cjcolvar (dismissed):
I have to admit that at first this approach concerned me but it appears that it is an accepted solution and it works as far as I can tell.  We'll have to be careful if we add gems/engines that do dynamic routing to be sure to add them to the exceptions.  (I think we may still need to test omniauth strategies but this may be easier on a testing server.)

cjcolvar (approved):

Issue #6013, opened 2024-08-30
User Inputs for Searching Can Results in Query Errors


joncameron:
# Description
There are two main areas where we see 500 errors in the application from bad URIs:
- Item view pages can have lots of junk
- Catalog search

Is extra handling an option so users get an "not found" page rather than a 500?

cjcolvar:
See https://app.honeybadger.io/projects/54117/faults/104411152/01J6AH2VTFA9B23ZB9P0W5229A#notice-summary for an example of a search that doesn't get the user input sanitized enough and ends up breaking the solr query.

joncameron:
These cases could also be caught in the ApplicationController so that users get redirected to the home page across the board when making bad requests.

cjcolvar:
Here is another example: https://app.honeybadger.io/projects/54117/faults/113131909

joncameron:
👍

PR #6858, opened 2026-05-28
Add catch for blacklight search errors

masaball:
Related issue: #6013

This commit also includes a small fix for the transcript search. In certain (rare) circumstances, there can be matches returned by the search that have a blank `transcripts_tsim`. This caused the processing step to break. We should be able to safely skip these blank entries and keep the search from bombing.


cjcolvar (approved):
This looks good.  Would be nice to figure out what exactly is wrong with the searches but this covers the cases we haven't fixed (or discovered) which is nice for the users. 🙂

PR #6862, opened 2026-05-28
Run batch ingest notification job every minute

### LINKED ISSUE #6821: Increase Frequency of Batch Email Notification Job
## Description
The current Sidekiq setup lets the batch email notification run every 15 minutes. It would nice to lower this to 1 or 2 minutes; could this also be something other than a cron job? Can it just run as actions complete or be sent as a normal queued job to resolve as the system is able?

This should be able to be run every minute, along with the batch scan. If it takes too long, we can open another issue to examine performance.

## Done Looks Like
- [ ] Lower time value for recurrence of the job (1 minute)
- [ ] Uniqueness check on this job so it doesn't start if a previous job is still running

cjcolvar:
Only needed to change the cron configuration as the job already has a uniqueness lock.

Resolves #6821


masaball (approved):
👍

Issue #6154, opened 2024-12-18
Add Physical Format to Descriptive Metadata


joncameron:
# Description
Physical format can be defined out of band in Avalon and show up in facets and in the interface, but there is no current capability for users to add or edit this field. Avalon should be able to accommodate physical carrier information for items, particularly for media which is not born-digital.

Could use the controlled vocabularies in PBCore: https://pbcore.org/elements/instantiationphysical

joncameron:
For IU Avalon where this field is already present, there will need to be an extra migration step to map existing values to the controlled vocabulary put into place.

PR #6864, opened 2026-06-01
Add physical format to descriptive metadata

masaball:
Related issue: #6154


cjcolvar (commented):
I'm a bit confused about the goals of the original ticket.  Is the aim to have an item level physical format field or a way to edit the section level physical description fields?  In MCO we display a facet for the latter because we got that for MDPI items and gather them up when indexing the item.  We don't currently have a way to edit that sort of technical metadata of a section.  In avalon core we have an item level editable physical description but it is textual and not a controlled vocabulary.  Adding a field for physical format at the item level might duplicate information at the section level.  But if it is already described at the item level in the catalog bib import can't import it into a section.  FWIW the pbcore instantiationPhysical property mentioned in the ticket maps to the ebucore hasFormat property currently used for section physical description.

Maybe we could get together with @joncameron and talk this out?

PR #6865, opened 2026-06-02
Explicitly enable S3 auth

cjcolvar:
Required by new build of container image (see https://github.com/avalonmediasystem/avalon-docker/pull/104)


masaball (approved):
👍

Issue #5988, opened 2024-08-13
Don't Destructure IP Ranges in Index


joncameron:
# Description
IP address ranges in CIDR notation can be used in access control for Avalon media objects; however, in the Solr index these are destructured and each individual address within the range is stored for the item. This results in a huge DRY problem where the index can become huge from the sheer number of addresses listed. 

# Done Looks Like
- [ ] Solr index values represent range rather than destructuring/splatting each member of the range into an individual entry

elynema:
Likely will need to change the query to make this work.

PR #6866, opened 2026-06-03
Expand IP address ranges on retrieval, not storage

masaball:
Related issue: #5988


cjcolvar:
Let's hold this for now until we know more about how to handle the implementation for gated discovery and potential costs of reindexing.

PR #6867, opened 2026-06-10
Remove unused and no longer relevant method

### LINKED ISSUE #5999: Removing a member from the managers group takes a long time
Removing a member from the managers group takes a long time when there are a large number of collections (~3.5 minutes in the worst case in IU production).  This is because each collection is fetched from Fedora to check if the user is the sole manager.  A solr query could make this O(1) instead of O(n).

See https://github.com/avalonmediasystem/avalon/blob/main/app/controllers/admin/groups_controller.rb#L142-L148

cjcolvar:
`check_for_sole_managers` is not used and is no longer useful because the managers group no longer has special privileges and does not affect any units or collections.

Resolves #5999


masaball (approved):
👍

PR #6868, opened 2026-06-10
Remove configurations for old unused services

### LINKED ISSUE #5365: Code Climate reporting fails when using Rerun failed tests in CircleCI
# Description
Rerun failed tests is a new feature in CircleCI which allows us to rerun only the tests that failed a previous build.  This is useful to determine if a test is flakey or an actual failure.  Avalon's CircleCI config is setup to enable this feature for rspec tests but the steps to aggregate the Code Climate reporting from the parallel builds fail because some of the parallel builds don't have any results. (See https://app.circleci.com/pipelines/github/avalonmediasystem/avalon/4352/workflows/acaa6da0-092f-4175-877d-0104940bb347/jobs/6312/parallel-runs/0?filterBy=ALL)  We can either fix this scenario or avoid it by seeing if we can set the parallelism to 1 when doing Rerun failed tests.  Also need to ensure that coverage reporting includes the original run and not just the rerun.

Read https://circleci.com/docs/rerun-failed-tests/#known-limitations for helpful hints and example solution configurations

# Done Looks Like
- [ ] Fix or avoid failures due to parallel build
- [ ] Ensure that coverage reporting includes both the original run and the rerun

cjcolvar:
Continues the work done in https://github.com/avalonmediasystem/avalon/pull/6459
Resolves #5365


masaball (approved):
👍

Issue #5992, opened 2024-08-16
Roll active_annotations gem into core Avalon


masaball:
With the circleci suite being busted for active_annotations, would it be worth the effort to move the code from the gem into core Avalon? Bringing it into Avalon would reduce overhead needed in upgrades because it would be one less separate piece we have to keep up to date.

As far as we know Avalon is the only application using the gem.

elynema:
This gem was created for Avalon, not the broader Samvera community. If anyone happens to be using it within the Avalon community, there is the older published version they can continue to use.

cjcolvar:
#6871 has been merged! :tada:
How about we archive the independent git repo after the next release of avalon?

joncameron:
👍

PR #6871, opened 2026-06-15
Roll active_annotations into core Avalon

masaball:
Related issue: #5992


cjcolvar (approved):
Looks good!  I did some minimal testing with an existing playlist and it worked!

PR #6875, opened 2026-06-17
Reduce number of queries as well as optimize queries used for gated discovery

cjcolvar:
Avalon 8.2 changes for permission inheritance do not scale for the number of units and collections in MCO.  On mco-staging non-admin searches timed out at 3 minutes.  The first change bringing the query down to 12s was to switch from the many joins in `policy_clauses` to a single join with a `terms` query.  Switching from that to a follow up query in `policies_with_access` to resolve units to collections brought the query down to 3s.  The rest of the changes in this PR remove small queries (~10ms each) and condense some queries to possibly save ~200ms.  Terms queries were also added in other places to reduce the number of boolean clauses.  

Further visual cleanup of the query could be to use [local params](https://solr.apache.org/guide/solr/latest/query-guide/local-params.html) to avoid having to reference the user groups multiple times.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -30,14 +30,14 @@ class SearchBuilder < Blacklight::SearchBuilder
   end
 
   def only_published_items(_permission_types = discovery_permissions, _ability = current_ability)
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good as far as I can tell. Clever to use the terms component to cut down the amount of clauses.

Issue #6388, opened 2025-06-24
Don't return all fields for every solr query


elynema:
# Description
With 7.8, we are including more binary objects in the Solr index. However, there are plenty of times in Avalon when we query Solr, but don't need all the data that lives there. Reducing the amount of data that is returned for queries could help improve Solr performance. Note that we are already limiting data returned in the query for Blacklight search results, but we potentially are not in other places.

Chris has already completed some of this work to rework queries; this ticket is to update the other untouched areas of the application to update to the new pattern.

# Done Looks Like

- [ ] review solr queries throughout Avalon to see what fields are being returned
- [ ] apply new pattern used for solr queries to any areas of Avalon where the old one is still used

joncameron:
👍

PR #6878, opened 2026-06-18
Add fl to solr query in MediaObject

masaball:
Related issue: #6388

After checking every explicit solr query I could find in the application, it appears that all of them already had 'fl' clauses or used other ways to limit the return such as being a facet query. This commit adds an 'fl' to the one place I found that still needed one.


cjcolvar (approved):
Nice catch!

PR #6885, opened 2026-06-22
Bump certain dependencies

cjcolvar:
We'll also need to update devise, view_components/blacklight, and puma (v7 may have changed the name of callbacks used in our custom prometheus setups) but those will take a little more care.


masaball (approved):
👍

PR #6886, opened 2026-06-23
Gated discovery speedup part deux

cjcolvar:
Continues #6875 
Related to #6729 

TODO:
- ~Fix: admin/collection show page is visible to inherited read users~
- ~write tests for above~
- ~write tests for admin/dashboard json for non-admin users -> nothing more, nothing less in unit/collection lists~
- ~Fix: `Reifying Admin::Unit because collection_managers called from /srv/services/avalon_mco-stage-8130/avalon/releases/20260622214422/app/models/concerns/admin_unit_behavior.rb:22:in 'AdminUnitBehavior#managers'`~

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -35,6 +35,7 @@ class Admin::CollectionsController < ApplicationController
       end
       builder.user = user
     end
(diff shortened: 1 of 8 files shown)


masaball (approved):

Issue #6000, opened 2024-08-20
Explore caching for improved performance


elynema:
# Description

This may be useful for loading thumbnail images, generating manifests, etc.

joncameron:
Large investigation; timeboxed to a 2 for an investigation. Caching is in place on manifests, but there are other areas to look at. Playlist manifests, for example, are not currently cached.

masaball:
After investigation, there is no discernible improvement to load times from caching thumbnails or the thumbnail component. Specifically, caching the thumbnail component directly does not preserve the data loaded, rather it only caches generated html so the link to the thumbnail route is still queried and has to load.

Caching the thumbnail itself did not appreciably speed up the loading either. The frontend still has to query the backend and wait for the response and this still triggers after parts of the page have loaded. Having said that, we may want to do this caching anyways. @cjcolvar recently found that `master_file#get_frame` was a major component to slow down on MCO, so even though it has no direct effect on thumbnail loading speed it may help mitigate slowdown of the application as a whole by limiting how many times that endpoint has to load the data from Solr or Fedora. If we do that work, we will probably want to add a `cache_key_with_version` method to SpeedyAF::Base to mimic standard Rails behavior so we don't have to reify the master file to generate a cache key.

As for other areas, nothing majorly stood out to me as potentially benefiting from caching.

masaball:
Closing for implementation ticket #6892, can create new caching tickets if/when other areas stand out as potentially benefiting.

PR #6887, opened 2026-06-25
Add caching to playlist iiif manifest

masaball:
Part of #6000

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -214,37 +214,41 @@ class PlaylistsController < ApplicationController
   def manifest
     authorize! :read, @playlist
 
-    # Fetch all master files related to the playlist items in a single SpeedyAF::Base.where
-    master_file_ids = @playlist.clips.collect(&:master_file_id)
-    master_files = []
-    master_files = SpeedyAF::Proxy::MasterFile.where("id:#{master_file_ids.join(' id:')}", load_reflections: true) if master_file_ids.present?
-    media_objects = master_files.collect(&:media_object).uniq(&:id)
+    cached_manifest = Rails.cache.fetch([@playlist.cache_key_with_version, 'iiif_playlist_manifest'], expires_in: 1.week) do
+      # Fetch all master files related to the playlist items in a single SpeedyAF::Base.where
+      master_file_ids = @playlist.clips.collect(&:master_file_id)
+      master_files = []
+      master_files = SpeedyAF::Proxy::MasterFile.where("id:#{master_file_ids.join(' id:')}", load_reflections: true) if master_file_ids.present?
+      media_objects = master_files.collect(&:media_object).uniq(&:id)
 
-    # This small optimization relies on the assumption that can? :read, master_file is the same as can? :read, master_file.media_object
-    # This only optimizes the case where multiple playlist items come from the same media object
-    cannot_read_hash = {}
-    media_objects.each { |mo| cannot_read_hash[mo.id] = cannot?(:read, mo) }
+      # This small optimization relies on the assumption that can? :read, master_file is the same as can? :read, master_file.media_object
+      # This only optimizes the case where multiple playlist items come from the same media object
+      cannot_read_hash = {}
+      media_objects.each { |mo| cannot_read_hash[mo.id] = cannot?(:read, mo) }
 
-    # Condense secure_streams into single call using master_files
-    stream_info_hash = secure_stream_infos(master_files, media_objects)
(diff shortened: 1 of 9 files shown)


cjcolvar (commented):
The cache key handles when the playlist is updated but can it handle when referenced media objects get metadata updates since some basic metadata is included in the manifest?  

I was looking back at the media object manifest caching and had the same question about children updates like section supplemental file label updates.  Testing on avalon-dev it appears that the manifest is getting these changes but I can't tell if it is actually getting cached (and having a cache hit) in the first place.  Maybe we could add or turn on cache logging for debugging?

This will be really neat to have alongside the media object manifest caching for reducing page load to playable!

app/models/playlist.rb
```diff
@@ -107,6 +107,21 @@ def clips_with_section_proxies
...
 
+  def parent_updated?
+    media_objects.any? { |mo| mo.timestamp.to_time > self.updated_at.to_time } ||
+      master_files.any? { |mf| mf.timestamp.to_time > self.updated_at.to_time }
+  end
+
+  def master_files
+    # Fetch all master files related to the playlist items in a single SpeedyAF::Base.where
+    master_file_ids = clips.collect(&:master_file_id)
+    @master_files ||= master_file_ids.present? ? SpeedyAF::Proxy::MasterFile.where("id:#{master_file_ids.join(' id:')}", load_reflections: true) : []
```
cjcolvar: I think the call to `clips` here will perform an SQL query each time `#master_files` is called.  How about refactoring this to either change the memoization from a single line into a block or put a guarded return before these lines?
```suggestion
    @master_files ||= begin
      master_file_ids = clips.collect(&:master_file_id)
      master_file_ids.present? ? SpeedyAF::Proxy::MasterFile.where("id:#{master_file_ids.join(' id:')}", load_reflections: true) : []
    end
```
```suggestion
    return @master_files unless @master_files.nil?
    master_file_ids = clips.collect(&:master_file_id)
    @master_files = master_file_ids.present? ? SpeedyAF::Proxy::MasterFile.where("id:#{master_file_ids.join(' id:')}", load_reflections: true) : []
```

app/controllers/playlists_controller.rb
```diff
@@ -214,37 +214,40 @@ def destroy
...
-    # Condense secure_streams into single call using master_files
-    stream_info_hash = secure_stream_infos(master_files, media_objects)
-
-    canvas_presenters = @playlist.items.collect.with_index do |item, i|
-      master_file = master_files.find { |mf| mf.id == item.clip.master_file_id }
-      cannot_read_item = master_file.nil? || cannot_read_hash[master_file.media_object_id]
-      position = i + 1
-      IiifPlaylistCanvasPresenter.new(playlist_item: item, stream_info: stream_info_hash[master_file&.id], cannot_read_item: cannot_read_item, position: position, master_file: master_file)
-    end
+    @playlist.touch if @playlist.parent_updated?
```
cjcolvar: If I'm reading this correctly the SpeedyAF fetches of the media objects and master files inside the cache block are now memoized in the playlist model but also have to be called here via `parent_updated?` to determine if the cache needs to be invalidated.  So I think the net gain of caching will be the JSON object construction done by `iiif_manifest` (and the serialization done in the `responds_to` block which in theory could get moved into the caching block).  I'm wondering if it would make sense to move cache invalidation from a changing of the key to an explicit eviction done whenever a media object or master file is saved.  This might look like an after save callback calling a scoped query on Playlist to find relevant playlists then computing their cache key and evicting them from the cache.  This approach is more complex and ties the models together more but could probably save the solr calls.  Thoughts?  I'm happy to talk this out synchronously.

app/models/playlist.rb
```diff
@@ -127,5 +125,11 @@ class << self
     def default_folder_name
       I18n.translate(:'playlists.default_playlist_name')
     end
+
+    # MediaObject is not directly associated with playlist/clips.
+    # Use class method for scoping instead of ActiveRecord scope.
+    def contains_media_object(id)
+      all.select { |playlist| playlist.media_objects.map(&:id).include?(id) }
```
cjcolvar, line 132: This makes sense but it still feels like it will be a performance issue.  How about opening a ticket to address this then this PR could move forward without being blocked?

cjcolvar (dismissed):
Loving this!  It feels pretty clean and that it will be a nice speed up! 🚀

app/models/playlist.rb
masaball: Follow up ticket created #6898

cjcolvar (approved):

Issue #6792, opened 2026-04-17
Downcase User Values


joncameron:
## Description
Beyond the entries in special access for a media object, there are other areas in the application where it would make sense to downcase values before storing them. Usernames, email addresses and groups(possibly?) should all be downcased before in case mixed-case input is provided.

joncameron:
👍

PR #6891, opened 2026-07-01
Downcase user values in all Staff Role and Special Access fields

masaball:
Related issue: #6792

For some reason some fields do not consistently show up in the `changed`
array, so we have to run the normalize against all user fields on every
save. This should not have much performance impact in most
circumstances, but an item with a lot of assigned roles and/or special
access may see some slow down on saves.

As far as I could tell, playlists and timelines retrieve user values via
the user_id, so downcasing is handled by the retrieval methods in the
User model.

This branch is built off the existing special access downcase branch, so putting it in draft until that prior work is merged.

### DIFF
--- a/app/models/admin/collection.rb
+++ b/app/models/admin/collection.rb
@@ -23,6 +23,7 @@ class Admin::Collection < ActiveFedora::Base
   include Identifier
   include MigrationTarget
   include AdminCollectionBehavior
+  include UserNormalization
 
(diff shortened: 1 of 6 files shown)


cjcolvar:
Looks like the failing tests are issues with the test groups being passed user objects instead of username strings in `media_objects_controller_spec.rb` on lines 1278 and 1335.

cjcolvar:
What do you think about changing this PR so it targets the other PR's branch?  Then we could merge this and do a final review with it all together.

cjcolvar (approved):
Looks great!  Love the refactor cleanup!

Issue #6069, opened 2024-10-07
Improve or remove `lsof` file checking


masaball:
In our batch process we [call lsof](https://github.com/avalonmediasystem/avalon/blob/37cfbea682640796d9d33f4069c8a773537f581e/lib/avalon/batch.rb#L28-50) to make sure the files are fully uploaded to the dropbox before being ingested. This method is also utilized in the regular file upload flow so that only files that are fully processed will be shown in the dropbox listing.

This accounts for as much as 30-40% of the load time when rendering the file-upload-step page. Improving the efficiency of the check, moving the check to happen when opening dropbox instead of at initial page load, or removing the file check entirely would net us a significantly lower load time.

elynema:
If we remove the lsof check on batch ingest in general, files could be truncated when ingested or could error on encoding. If there is an error on encoding, then that will go out in a report to the user post-ingest. If the file is truncated, there's no error to report. However, we *think* the moving the original file to ArchiverSpool to get sent to SDA happens post-encoding, so by that time it is much more likely that the entire file has uploaded and the full version is sent to tape for safe storage.

joncameron:
🎉

PR #6895, opened 2026-07-08
Remove lsof check from file upload step

masaball:
Related issue: #6069

As far as could be determined, it looks like we do not actually use the output of the dropbox lsof check during the file upload step. Dropbox is handled by Browse Everything, which has its own file gathering in place.


cjcolvar (approved):
This will be amazing if testing reveals we haven't needed it for a long time! ✨ 🤞

PR #6899, opened 2026-07-10
Respond to external valid stream requests

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/978

Respond to valid external stream requests with the requested HLS manifest instead of a JSON. This allows IIIF auth 2.0 flow to authorize and stream restricted-access media from outside of Avalon using media object's IIIF Manifest.

Note: to make use of this change, IIIF Manifests for restricted-access media objects need be allowed to read without authorization. 
To test this locally in Ramp demo site I commented the authorization behind the `/manifest` request for media object as follows:
```
def manifest
    @media_object = SpeedyAF::Proxy::MediaObject.find(params[:id])
    # authorize! :read, @media_object
    .....
```
in `controllers/media_objects_controller.rb` [file](https://github.com/avalonmediasystem/avalon/blob/main/app/controllers/media_objects_controller.rb#L509).

However, I didn't add this change to this PR because this seems to be something that needs a team discussion :) (and probably there's a more elegant way to handle this!).

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -250,7 +250,15 @@ class MasterFilesController < ApplicationController
       end
     elsif auth_token.present?
       return render json: iiif_auth_probe_resp(success: false), status: :unauthorized unless StreamToken.valid_token?(auth_token, @master_file.id) || can?(:read, @master_file)
-      render json: iiif_auth_probe_resp(success: true), status: :ok
+      if request.headers['Accept']&.include?('application/json')
+        render json: iiif_auth_probe_resp(success: true), status: :ok
+      else
+        @hls_streams = if quality == "auto"
+                       gather_hls_streams(@master_file)
+                     else
+                       hls_stream(@master_file, quality)
+                     end
+      end
     else
       return head :unauthorized if cannot?(:read, @master_file)
       @hls_streams = if quality == "auto"


cjcolvar (commented):
While this might work I think it skips a part of the auth flow by not receiving and handling the probe response.  I think the spec isn't very clear (or I missed something) but I think the flow at this point should be:
1. Ramp requests the probe service with auth token (currently https://avalon-dev.dlib.indiana.edu/master_files/0g354f22c/auto.m3u8 with Authorization-Bearer header)
2. Ramp gets probe JSON response (`iiif_auth_probe_resp`)
3. Ramp requests the Choice Body item `id` with auth token (currently https://avalon-dev.dlib.indiana.edu/master_files/0g354f22c/auto.m3u8 with Authorization-Bearer header)
4. Ramp gets m3u8 manifest for streaming

The problem seems to come with this flow and avalon using the same url for the probe service and m3u8 stream manifest.  Avalon needs something to distinguish a probe request from a stream request in order to determine the correct response to send.  This could be creating a different probe endpoint (e.g. `<stream_url>/probe`) or maybe just adding a query param (e.g. `?probe=true`) to the `hls_manifest` route.

Does this seem right or did I get it wrong?  Thoughts on different route vs query param?

Dananji:
You're right!

Using just the `Accept: application/json` header to distinguish probe vs stream requests only works because, that happens to be what Ramp sends today. This could be different in another IIIF viewer. Also, the current assumption is not explicit so, this could break in the future even for Ramp if things were to change on Ramp side. And eventually the auth flow could stop working entirely in the future.

I agree we need an explicit way for Avalon to tell the two requests apart. I don't have a strong preference between a separate probe endpoint vs a query param on the existing route, both sounds good to me! 

I think, we can pick whichever fits Avalon's current IIIF service related routing conventions. Or we could further discuss the best approach if there are other concerns I'm missing.

@masaball, since you did the original IIIF Auth 2.0 work, curious if you have thoughts about this too, a separate route vs query param or anything about this flow we should be thinking about?

masaball:
Yeah, that all sounds right to my understanding. I was wondering if we would need to divide the probe logic when I was working on the Avalon side.

I think my preference would be for having a separate endpoint for the probe requests, so that way the endpoints are single responsibility. We can clean up the `hls_manifest` method and hopefully keeps things more maintainable.

cjcolvar:
That makes sense to me.  I doubt anyone is using avalon IIIF manifests outside of avalon with IIIF auth but we'll need to make note of when we remove support for Auth v1 from the `hls_manifest` route because that will break old manifests.

Dananji:
Closing in favor of #6903

Issue #6893, opened 2026-07-02
[BUG] Forced captions do not properly save other attributes


masaball:
**Describe the bug**
Captions that are marked as "forced" do not successfully save when trying to update other attributes.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Go to manage files page for any media object with child master file(s)
2. Upload a caption to a section if one is not already extant
3. Open the edit menu for a caption file
4. Check the forced check box
5. Save the caption file
6. Reopen the edit menu
7. Change any other attribute (update the label, check "treat as transcript", etc.)
8. Hit save, see "Successfully saved" message popup.
9. Reload page
10. Reopen edit menu
11. Observe that attributes have not been updated

**Expected behavior**
Caption attributes are successfully updated regardless of what other attributes are enabled/disabled.

joncameron:
Similar as the forced caption routing bug which was fixed a few weeks ago. Conditional handling for the 'forced' tag is breaking update for the other tag types.

Attention to refactoring/DRY in this too.

joncameron:
👍

PR #6900, opened 2026-07-10
Fix improper 'forced' validation check

masaball:
Related issue: #6893

We validate that only one caption file on a master file may have the 'forced' attribute at a time. The validation to enforce this behavior was improperly setup such that the validation would raise the error for any save on a caption marked 'forced'.

This commit also includes a small change to the `SupplementalFilesController#handle_error` method because updates to supplemental files are handled via javascript. The order we had the if...else block in resulted in every action generating a success message, regardless of whether it succeeded or not. The edit page should now reflect when an error has occurred while trying to save supplemental files.

### DIFF
--- a/app/controllers/supplemental_files_controller.rb
+++ b/app/controllers/supplemental_files_controller.rb
@@ -43,7 +43,6 @@ class SupplementalFilesController < ApplicationController
       raise Avalon::BadRequest, "Missing required Content-type headers" unless request.headers["Content-Type"] == 'application/json'
     end
     raise Avalon::BadRequest, "Missing required parameters" unless validate_params
-    raise Avalon::BadRequest, "Forced attribute is already assigned to another caption. Ensure no other captions are forced before setting attribute." unless validate_forced
 
     @supplemental_file = SupplementalFile.new(**metadata_from_params)
 
@@ -107,7 +106,6 @@ class SupplementalFilesController < ApplicationController
       raise Avalon::BadRequest, "Missing required Accept headers" unless request.headers["Accept"] == 'application/json'
     end
     raise Avalon::BadRequest, "Missing required parameters" unless validate_params
-    raise Avalon::BadRequest, "Forced attribute is already assigned to another caption. Ensure no other captions are forced before setting attribute." unless validate_forced
 
     find_supplemental_file
 
@@ -211,11 +209,11 @@ class SupplementalFilesController < ApplicationController
 
 
     def handle_error(message:, status:)
-      if request.format == :json || request.headers['Avalon-Api-Key'].present?
-        render json: { errors: message }, status: status
-      else
+      if request.format == :html
         flash[:error] = message
         redirect_to edit_structure_path
+      else
+        render json: { errors: message}, status: status
       end
     end
 
@@ -262,14 +260,6 @@ class SupplementalFilesController < ApplicationController
           @supplemental_file.tags -= [tag]
         end
       end
-    end
-
-    def validate_forced
-      return true unless params["forced_#{params[:id]}".to_sym]
-      forced_files = @object.supplemental_files(tag: 'forced')
-      return false if forced_files.length.positive?
-
-      true
     end
 
     def metadata_from_params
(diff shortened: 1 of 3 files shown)


app/controllers/supplemental_files_controller.rb
```diff
@@ -211,12 +209,10 @@ def find_supplemental_file
 
 
     def handle_error(message:, status:)
-      if request.format == :json || request.headers['Avalon-Api-Key'].present?
-        render json: { errors: message }, status: status
-      else
-        flash[:error] = message
-        redirect_to edit_structure_path
-      end
+      return render json: { errors: message}, status: status unless request.format == :html && request.headers['Avalon-Api-Key'].blank?
```
cjcolvar, line 212: Good catch!

There seems to have been a bit of mismatch between the logic here checking the request format and the API header versus the update action where it checks the request format and the Accept header.   I think it would be nice to be consistent between the two.  I'm not sure which one makes the most sense though...maybe API header?  Seeing as how this method is used by other actions I want to make sure that changes don't break any other assumptions.

app/models/supplemental_file.rb
```diff
@@ -33,6 +34,7 @@ class SupplementalFile < ApplicationRecord
   after_update_commit :update_index, prepend: true
   after_destroy_commit :remove_from_index
   before_save :default_label
+  before_save :validate_forced, if: -> { tags.include?('forced') }
```
cjcolvar, line 37: This feels a lot better than the controller level enforcement!

cjcolvar (commented):
Looks great!  I forgot that we made the supplemental files API weird which led to this tricky scenario with error messages.  After rediscovering that I had one question / concern.

app/controllers/supplemental_files_controller.rb
masaball: The update method does check the API header:
```
if metadata_upload?
  raise Avalon::BadRequest, "Incorrect request format. Use HTML if updating attached file." if attachment
  raise Avalon::BadRequest, "Missing required Content-type headers" unless request.headers["Content-Type"] == 'application/json'
elsif request.headers['Avalon-Api-Key'].present?
  raise Avalon::BadRequest, "Incorrect request format. Use JSON if updating metadata." unless attachment
  raise Avalon::BadRequest, "Missing required Accept headers" unless request.headers["Accept"] == 'application/json'
end
```

Or maybe I'm misunderstanding your concern there?

cjcolvar: I thought the change you made to the `handle_error` method is covering the error case when a supplemental files API HTML request is made but JSON is returned (indicated by the Accept request header).  That the JS request to update made by the browser has an error and the HTML response (setting the flash message that doesn't get rendered) is returned which isn't acted upon by the JS so it thinks it is a successful request.  Maybe I misunderstood the problem and your change?

masaball: The logic for the error handling is that JSON should be returned unless it is specifically an HTML request from the browser. 

The original issue with the JS was that all non-JSON format requests were returning the redirect, which counted as a success to our JS function since it is a 302 status. So I tried flipping the logic around, so that it would do the redirect for only HTML requests and the JSON would have the proper 4xx status to get the JS to generate a "failed" message. This caused the automated tests to fail because API HTML requests should be returning the JSON error, not the HTML redirect. The idea being that API requests will be through a CLI and so JSON responses are clearer.

So if I'm thinking through the implemented logic correctly:
* HTML without API key falls through to the redirect
* HTML with API key returns JSON error message
* All non-HTML requests return JSON

cjcolvar: That makes sense.  A change in behavior but one that seems okay to make.  I'm still confused as to if the error response should be JSON for HTML requests that have a JSON Accept header (whether it is an API request or not) since the successful response is JSON.  But I guess the old code didn't handle this case either so it's fine to handle this in a separate issue/PR.

cjcolvar (approved):

Issue #6898, opened 2026-07-09
Improve efficiency of Playlist.contains_media_object


masaball:
Follow-up to #6887

Media Objects are not directly referenced in any of the Annotation or Playlist database columns, making it cumbersome to set up an ActiveRecord scope since those utilize SQL queries.

The approach taken in #6887 was to add a class method which iterates through all playlists and checks if the provided media object id is included in the `Playlist#media_objects` array. While the `Playlist#master_files` and `Playlist#media_objects` methods both use SpeedyAF::Proxies, the initial `Playlist.all` could be very unperformant as every playlist gets initialized from the DB. Additionally, a very large playlist could theoretically hit solr quite hard loading the `Playlist#master_files` array.

We should reexamine this approach to determine if there is a single SQL or maybe Solr query or another more efficient approach that we could use to gather all Playlists containing clips from a specific media object.

masaball:
Is there a meta-programming approach where we could load the MediaObject then dynamically generate a series of `.or` calls? So result of the method would be something like:

```
mo = SpeedyAF::Proxy::MediaObject.find(mo_id)
sections = mo.sections
dynamic_sql = Playlist.contains_master_file(sections.first.id)
    .or(Playlist.contains_master_file(sections.second.id)).or(...etc,etc,etc)
```

edit: Would `.reduce` work? 
```
mo = SpeedyAF::Proxy::MediaObject.find(id)
sections = mo.section_ids
query = sections.reduce(Playlist.none) do |r, s|
  r.or(Playlist.contains_master_file(s))
end
```

joncameron:
👍

PR #6905, opened 2026-07-14
Refactor `contains_media_object`

masaball:
Related issue: #6898

Improvements:
* Use `after_update` callback on media object since association with playlist will not exist for newly created media objects
* Use media object in Playlist scope. The object is loaded into memory already because it is being updated. Don't waste bandwidth loading the proxy.
* Use `reduce` to iterate over sections ids and generate a single SQL statement to pull playlists containing the media object's child master files.
* Use `.touch_all` to update all the playlists in a single SQL action instead of iterating through each individually.

### DIFF
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -22,6 +22,11 @@ class Playlist < ActiveRecord::Base
   end
   scope :with_tag, ->(tag_filter) { where("tags LIKE ?", "%\n- #{tag_filter}\n%") }
   scope :contains_master_file, ->(master_file_id) { joins(:clips).merge(AvalonClip.from_master_file(master_file_id)).distinct }
+  scope :contains_media_object, lambda { |media_object|
+    sections = media_object.section_ids || []
+    # Initialize from blank ActiveRecord relation to avoid unnecessary loading into memory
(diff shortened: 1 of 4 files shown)


app/models/playlist.rb
```diff
@@ -22,6 +22,11 @@ class Playlist < ActiveRecord::Base
   end
   scope :with_tag, ->(tag_filter) { where("tags LIKE ?", "%\n- #{tag_filter}\n%") }
   scope :contains_master_file, ->(master_file_id) { joins(:clips).merge(AvalonClip.from_master_file(master_file_id)).distinct }
+  scope :contains_media_object, lambda { |media_object|
+    sections = media_object.section_ids || []
+    # Initialize from blank ActiveRecord relation to avoid unnecessary loading into memory
+    sections.reduce(Playlist.none) { |r, s| r.or(Playlist.contains_master_file(s)) }
```
cjcolvar, line 28: I've never really used `reduce` before.  Pairing it with the scope is next level!

cjcolvar (commented):
This looks amazing! ✨

cjcolvar (approved):

PR #6906, opened 2026-07-14
Use modern, more semantic language

cjcolvar:
This is overdue and brings us closer to the [Samvera contributing community guidelines](https://github.com/samvera/hyrax/blob/main/.github/CONTRIBUTING.md#language).


masaball (commented):
Found one more usage of whitelist: https://github.com/avalonmediasystem/avalon/blob/6072cf35bdd794f882b595cbe9bca5cc5549346f/app/controllers/application_controller.rb#L99

masaball (approved):

Issue #6889, opened 2026-06-30
Add IIIF Auth 2 to timeline manifests


masaball:
## Description
When implementing IIIF Auth 2.0, it was missed that timelines also produce a manifest. We should add an auth 2 service to the timeline manifests.

### Questions
- Are there any other manifests produced by Avalon? 
  - Currently: Media objects, playlists and timelines
  - Both v1 and v2 auth service are in place for our manifests
- Will Timeliner need to be updated at all to accommodate the changed manifest output? 
- Can manifest generation be abstracted so there doesn't need to be separate code for creation and parsing of manifests across the application?

Initial manifest generation for Timeliner is on the Avalon side, so adding the new auth service to the Timeliner manifest should be just on the Avalon side.

## Done Looks Like
- [ ] IIIF Auth 2 service written into manifests delivered to Timeliner
- [ ] Auth v1 service removed; does it break anything if so?
- [ ] Issue created for Timeliner for compatibility (does anyone have Timeliner manifests from Avalon?)

PR #6912, opened 2026-07-16
Add auth 2 to timeline manifests

masaball:
Related issue: #6889

This commit breaks iiif auth out into its own concern for better maintainability. It also brings in the most recent mainline iiif_manifest gem to clean up our implementation of auth 2.

There may be a migration needed? I think timeline manifests are generated at timeline creation/update rather than being generated on the fly like the rest of our manifests.


cjcolvar:
@masaball Sorry, I was slow reviewing this and it has conflicts now.  Could you please rebase this PR and resolve the conflicts?

app/models/concerns/iiif_auth_service.rb
```diff
@@ -0,0 +1,86 @@
...
+          "label": I18n.t('iiif.auth.logoutLabel'),
+          "profile": "http://iiif.io/api/auth/1/logout"
+        }
+      ]
+    }
+  end
+
+  def auth2_service(quality = 'auto')
+    {
+      "id": probe_url(quality),
```
cjcolvar: This probably needs to be adjusted now given #6918, right?

cjcolvar (commented):
This looks really good! I like the refactor pulling things out to a mixin.  It really DRYs things up!

AFAICT timeliner only handles auth 1 and shouldn't mind if auth 2 is present as well so this PR seems like a good first step.  Like you said in the parent issue it would be nice to transition to auth 2 and remove auth 1 from avalon so maybe the steps could be:

1. Add auth 2 to timeline manifests (this PR 🙂)
2. Add auth 2 support to timeliner
3. Move to auth 2 for avalon timeline manifest generation
4. Migrate existing timeline manifests from auth 1 to auth 2
5. Remove auth 1 support from timeliner

Steps 4 and 5 don't necessarily have to happen.  It might be good for timeliner to keep support for auth 1 for use outside of avalon.  If so there wouldn't be a strong case for migrating existing manifests except for standardization.

cjcolvar (approved):
Looks great!

PR #6914, opened 2026-07-20
Users cannot read items with public visibility if inheriting private visibility

### LINKED ISSUE #6913: [BUG] Item retains stale public read access after unchecking "Disable parent permissions"
**Describe the bug**

When "Disable parent permissions" is unchecked on an item after having been checked, the item's read access group retains the `public` visibility setting from when inheritance was disabled, even though the item is now supposed to inherit access from its parent collection.

As a result, if the parent collection has "Item Access" set to "Collection staff only" or "Logged in users only," the item's access-control page correctly displays the inherited (restricted) access setting, but the item can still be streamed by anonymous public users who have the item's direct URL, because the stale `public` value remains in the item's own read access group and continues to be consulted by the access control logic.

Note that this issue may not surface through normal browsing, since the item may not be discoverable via search/collection browse under a restricted collection. However, the item is still directly accessible via its URL (e.g., shared via a link, handle, or bookmark), so the restricted streaming access is effectively bypassed for anyone with the URL.

**To Reproduce**

1. Create a collection and set its "Item Access" to "Collection staff only" (or "Logged in users only").
2. Create and publish an item within that collection.
3. On the item's "Edit Media Object > Access Control" page, check "Disable parent permissions."
4. Under "Item Access," select "Available to the general public" and save.
5. Verify the item is now streamable to anonymous public users.
6. Uncheck "Disable parent permissions" (without changing the "Item Access" radio selection back from public) and save.
7. Copy the item's direct URL.
8. In a private/incognito browser window (not logged in), go to the item's direct URL.
9. Attempt to stream the item.
10. See error: the item is still streamable to the anonymous public user, even though the access-control page shows the item is now inheriting the collection's restricted "Item Access" setting.

**Expected behavior**

After unchecking "Disable parent permissions," the item should fully inherit the collection's access settings. Since the collection's "Item Access" is "Collection staff only," the item should **not** be streamable to anonymous public users, even when accessed directly via its URL.

**Screenshots/Screen Recording**

N/A — will attach if needed.

- Device / platform: laptop
- OS: N/A
- Browser: N/A
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Release:**

8.2

**Additional notes**

The root cause is that when "Disable parent permissions" is unchecked, the item's read_groups field is not cleared. The item-level UI and access-control page correctly reflect the intended inherited state, but the underlying virtual_read_groups — which is what the access control logic actually consults — continues to pull in the stale read_groups values, evaluating the item as if it still had an active override even though it is supposed to be fully inheriting from the collection.

This is likely to have gone unnoticed because the affected item is often not discoverable through normal search/browse once the collection restricts access — but the item remains fully accessible to anyone with its direct URL, so the restriction is only cosmetic in this scenario.

Two possible approaches to fix this were identified, and either would resolve the underlying issue:

1. **Skip stale visibility groups when not overriding:** Item-level read groups (`public`, `restricted`/`registered`, `private`) should not be consulted when the item does not have "Disable parent permissions" checked. Other item-level special access grants (e.g., IP-based or group-based access) should still be consulted regardless of override status. This addresses the issue at the point where access is evaluated, so it also covers any other code paths that might leave stale values in place, while preserving the item-level access setting if we want to revert to it.

2. **Clear stale state on save:** Clear the item-level visibility read groups (`public`/`registered`) when "Disable parent permissions" is unchecked, so the stored state matches the displayed inherited state at the moment inheritance is re-enabled. This addresses the issue at the point where the item is saved.

We'd welcome input from maintainers on which approach best fits the existing access control architecture, or if a combination of both is preferred.

cjcolvar:
Resolves #6913

The issue here is that a media object can have a locally set visibility that is different than its parent.  We want to save that value even if it is inactive because inheritance is not disabled so that when it is disabled the user will restore the previous value.  The media object's visibility is stored in `read_groups` which is used in `test_read` (provided by `blacklight-access_controls`) to determine access.  Instead of overriding this I chose to add to the `cannot` clause that already exists to also override the access granted by default.  This PR blocks access if disable inheritance is not set and the user's groups don't overlap with the groups inherited from the collection, which includes visibility, and the special access groups which are still active.  (This is effectively a clone of `test_read` with a changed set of groups to test specific to this case.)

I also added more tests specific to this case.  Are there any more that should be added?

See #6913 for steps to reproduce.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -388,13 +388,13 @@ class Ability
   end
 
   def is_exclusively_inherited_from_parent?(media_object)
-    # User isn't in item's read users or item's read groups (not explicitly granted access to item) AND
-    # User isn't in item's read users and is inherited from parent (exclusively inherited user) OR
-    # User isn't in item's read groups and is in a read group inherited from parent (exclusively inherited group)
-    # Short form: NOT explicitly granted access to item AND (explicitly inherited user OR explicitly inherited group)
-    !(@user.in?(media_object.read_users) || !(@user_groups & media_object.read_groups).empty?) &&
-    ((!@user.in?(media_object.read_users) && @user.in?(media_object.inherited_read_users)) ||
-      ((@user_groups & media_object.read_groups).empty? && !(@user_groups & media_object.inherited_read_groups).empty?))
+    # NOT granted access from item and its leases AND granted access from policies
+    lease_read_users = media_object.leases.collect { |l| read_users_from_policy(l.id) }.flatten
+    lease_read_groups = media_object.leases.collect { |l| read_groups_from_policy(l.id) }.flatten
+    media_object_read_users = read_users(media_object.id) + lease_read_users
+    media_object_read_groups = read_groups(media_object.id) + lease_read_groups
+    access_from_media_object = !(user_groups & media_object_read_groups).empty? || media_object_read_users.include?(current_user.user_key)
+    !access_from_media_object && test_read_from_policy(media_object.id)
   end
 
   def is_member_of_any_collection?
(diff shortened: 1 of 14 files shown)


cjcolvar:
When adding more tests to exercise the `cannot :read, media_object` block that enforces disable inheritance I found some more cases which were broken that required reworking the block.  Additionally IP range permissions granted at the collection and unit level required changes to indexing.  We probably need to add a migration (or at least a code snippet in the upgrade instructions) to make sure that any existing in the wild get reindexed.

masaball (approved):

PR #6917, opened 2026-07-23
Unwrap Devise initializer from Rails.application.reloader.to_prepare

cjcolvar:
Devise has an engine initializer that runs after load_config_initializers (see https://github.com/heartcombo/devise/blob/372b295fe6f63b4af3269f5dcd51a18c0bc2016c/lib/devise/rails.rb#L28-L38). It appears that the reloader.to_prepare runs after this initializer so none of the omniauth strategies get fully setup (strategy and strategy_class set on the omniauth_config and the strategy class pushed onto the middleware stack).  Attempts to login result in an infinite request loop and for SAML requests to the metadata endpoint return 404.

Also needed is a configuration for adding CSRF protection to omniauth calls to avoid authenticity errors when starting the login process.

Thanks to UCLA for finding this issue and patiently working through pinpointing it and testing.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,354 +1,352 @@
-Rails.application.reloader.to_prepare do
-  # Use this hook to configure devise mailer, warden hooks and so forth.
-  # Many of these configuration options can be set straight in your model.
-  Devise.setup do |config|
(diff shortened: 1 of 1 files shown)


masaball (approved):
👍

PR #6919, opened 2026-07-24
Update rubocop file

masaball:
Update the rubocop config to point to current ruby/rails versions and change cop name to match more recent versions of rubocop.

While we no longer have linting setup in our CI pipeline, this will facilitate having a baseline working rubocop config available for developers who want to run linting locally.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,8 +5,8 @@ inherit_gem:
   bixby: bixby_default.yml
 
 AllCops:
-  TargetRubyVersion: 2.4
-  TargetRailsVersion: 5.1
+  TargetRubyVersion: 4.0
+  TargetRailsVersion: 8.0
   DisplayCopNames: true
   Exclude:
     - 'db/**/*'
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Do you think we should add rubocop back into our CI / PR checks? We'd probably have a big todos file but might be worth bringing back in.  Blacklight still uses it and a recent PR to update rubocop with some automatic style fixes applied revealed some new things about ruby / rails that I didn't know...so rubocop could help keep us up to date.  But definitely is another hoop to jump through which takes time and slows things down.

masaball:
I think it would probably be worthwhile to at least look into. While it is another hoop, it is beneficial for catching small mistakes and showing new/better coding patterns. And would help keep things consistent. But also, I've just been running it locally for the last year or 2 and that's been working for me so if we don't want to enforce it globally (at least for now), that's fine too.

If we do reimplement it, we will definitely want to go through the legwork of disabling certain cops either overall or for specific methods to minimize false positives. Like, I know offhand `MediaObjectsController#update_media_object` is in breach of at least 3 or 4 different cops related to length and complexity.

PR #6920, opened 2026-07-27
Consider inheritance when determining if auth service needs to be included on the canvas

cjcolvar:
Related to #6914 

This is another instance where we were relying on `media_object.visibility` and didn't add checking of `disable_inheritance?`.  This PR resolves this by using a new `active_visibility` convenience method which checks `disabled_inheritance?` before providing the correct visibility.

@masaball With this change do you think we need to consider parent collection edits in manifest cache expiration?

### DIFF
--- a/app/controllers/media_objects_controller.rb
+++ b/app/controllers/media_objects_controller.rb
@@ -511,7 +511,7 @@ class MediaObjectsController < ApplicationController
     authorize! :read, @media_object
 
     # TODO: Add a rake task to clear expired items from the cache?
-    cached_manifest = Rails.cache.fetch([@media_object.cache_key_with_version, 'iiif_manifest'], expires_in: 1.week) do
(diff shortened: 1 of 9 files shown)


masaball (dismissed):
These changes look good. Good catch on moving the visibility stuff to the behavior concern. 

Yeah, we probably will need to have something setup to make sure cache keys get invalidated when permissions on the collection change.

app/controllers/media_objects_controller.rb
```diff
@@ -511,7 +511,7 @@ def manifest
     authorize! :read, @media_object
 
     # TODO: Add a rake task to clear expired items from the cache?
-    cached_manifest = Rails.cache.fetch([@media_object.cache_key_with_version, 'iiif_manifest'], expires_in: 1.week) do
+    cached_manifest = Rails.cache.fetch([@media_object.cache_key_with_version, @media_object.active_visibility, 'iiif_manifest'], expires_in: 1.week) do
```
masaball, line 514: Good solution. I was expecting we would need something like we're doing with the playlists but this is a lot cleaner.

masaball (approved):

PR #6921, opened 2026-07-27
Avoid caching stream tokens by introducing proxy endpoint for non HLS streams

### LINKED ISSUE #6915: IIIF manifests for mp3 stream urls are cached longer than stream tokens
## Description
As reported by Kyle on slack, IIIF manifests with progressive download mp3s are cached with the stream url that includes the token.  This causes problems because the stream token will expire (20 min TTL) before the manifest is expired (1 week TTL) from the cache leading to the inability to stream if the stream is not already cached in the browser.  This also has the potential to leak stream tokens from one user to another which could lead to the same inability to stream issue where one user requests the IIIF manifest and effectively blocks other users from streaming for that item for a week.  It is not a security issue because it doesn't change the fact that only users that can read a media object will be able to get the manifest.

## Done Looks Like
-  [ ] make an endpoint to proxy the mp3 stream url with the token so manifests don't include a live token.
-  [ ] write tests and everything else that comes with adding a new route

cjcolvar:
Resolves #6915

### DIFF
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -266,9 +266,13 @@ class MasterFilesController < ApplicationController
 
   def stream
     return head :unauthorized if cannot?(:read, @master_file)
+    stream = file_stream(@master_file, params[:quality])
 
-    stream = mp3_stream(@master_file)
-    redirect_to(stream)
+    if stream[:mimetype] == 'application/x-mpegURL'
+      redirect_to hls_manifest_master_file_url
+    else
+      redirect_to(stream[:url], allow_other_host: true)
+    end
   end
 
   def structure
@@ -442,11 +446,10 @@ protected
     hls_stream
   end
 
-  def mp3_stream(master_file)
+  def file_stream(master_file, quality)
     stream_info = secure_streams(master_file.stream_details, master_file.media_object_id)
-    mp3_stream = stream_info[:stream_hls].first[:url]
-    unnest_wowza_stream(mp3_stream) if Settings.streaming.server.to_sym == :wowza
-    mp3_stream
+    file_stream = stream_info[:stream_hls].find { |stream| stream[:quality] == quality }
+    file_stream
   end
 
   def unnest_wowza_stream(stream)
--- a/app/controllers/master_files_controller.rb
+++ b/app/controllers/master_files_controller.rb
@@ -268,7 +268,9 @@ class MasterFilesController < ApplicationController
     return head :unauthorized if cannot?(:read, @master_file)
     stream = file_stream(@master_file, params[:quality])
 
-    if stream[:mimetype] == 'application/x-mpegURL'
+    if stream.nil?
+      return head :not_found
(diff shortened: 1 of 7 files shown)


config/url_handlers.yml
```diff
@@ -3,6 +3,7 @@ generic:
   http:
     video: <%=http_base%>/<%=path%>/<%=filename%>.<%=extension%>.m3u8
     audio: <%=http_base%>/<%=path%>/<%=filename%>.<%=extension%>.m3u8
+    audio_mp3: <%=http_base%>/<%=path%>/<%=filename%>.<%=extension%>
```
cjcolvar, line 6: Had to copy this here from the nginx adapter to generic adapter for managed mp3 derivatives.

cjcolvar:
@masaball Any other tests that you think we should add?  Maybe a test (and implementation) to handle when a requested quality derivative isn't found?

cjcolvar:
Thinking about this last case some more, I'm thinking that the response precedence should be requested quality, first derivative, 404 if none found.  @masaball What do you think?

spec/routing/master_files_routing_spec.rb
```diff
@@ -19,5 +19,13 @@
     it "routes to #move" do
       expect(:post => "/master_files/abc1234/move").to route_to("master_files#move", id: 'abc1234')
     end
+
+    it "routes to stream" do
```
masaball:
```suggestion
    it "routes to #hls_manifest" do
```

masaball (commented):
Yeah, it definitely makes sense to have handling for if a quality isn't found. Your suggestion of ordering sounds good to me. Requested, first deriv, 404 👍

app/controllers/master_files_controller.rb
```diff
@@ -264,6 +264,19 @@ def hls_manifest
     end
   end
 
+  def stream
+    return head :unauthorized if cannot?(:read, @master_file)
+    stream = file_stream(@master_file, params[:quality])
+
+    if stream.nil?
+      return head :not_found
```
masaball, line 272: Probably an edge case, but do we need an additional clause in case `stream[:url]` is nil? I think `redirect_to` throws an error for a nil argument
```suggestion
    if stream.nil? || stream[:url].nil?
      return head :not_found
```

Though maybe it would be better to leave as is, because a derivative missing the stream url probably means we have bigger problems that should be getting logged/reported...

masaball (dismissed):
Looks good. Just had one thought, but probably not worth adding?

app/controllers/master_files_controller.rb
cjcolvar: Good catch!

Might be good to handle this now while we're thinking of it.  A 404 not found will be broken to a user and visible that way probably about as much as a 500 would be in the logs.  Ramp may provide an error with a link to report it even.

I'm going to move it down to a new `elsif` below the m3u8 redirect since that doesn't really depend on the `stream[:url]`.

masaball (approved):

PR #6922, opened 2026-07-28
Only render per_page widget if logged in and enforce per_page limit by changing blacklight config in before action

cjcolvar:
~Proof of concept which appears to work locally.~

I've pulled this into mco-staging so will be deployed to production with 8.2.1.

This doesn't block any requests but it should hopefully reduce the complexity of solr requests as well as parallel requests for thumbnails on search results.  Overall I'm hoping this will reduce spikes of traffic and ease the burden on solr.


masaball (dismissed):
Sorry for missing this. Looks good!

PR #6925, opened 2026-07-29
Various enhancements and fixes from Northwestern's Avalon implementation

mbklein:
This PR contains multiple small fixes we've applied on top of several versions of Avalon while applying our own branding and additional in-house features to the base version. They include:

- A fix for a situation where poster generation would raise an exception if `master_file.file_location` was `nil` (e.g., when the MasterFile management strategy is set to `delete`)
- A fix for streaming URLs with spaces in them (e.g., when a MasterFile is ingested with spaces in the filename and goes through the whole pipeline like that). (This _might_ only apply to AWS CloudFront streaming, but shouldn't break any others)
- A fix to allow the MediaPlayer component to send credentials (cookies and/or auth headers) with streaming requests
- A `CatalogController` `before_action` to validate sort params before allowing the request to continue
- A fix to prevent HTML encoding of URLs inside dynamically generated HLS manifests
- An enhancement to allow `WaveformJob` to generate waveforms from an AWS S3 source by streaming it instead of downloading the entire file to temp storage
- A fix for batch ingest that makes sure that the ingest object's `@current_package` instance variable always gets set correctly

I left them as separate commits to make them easier to cherry pick, but I didn't want to submit 7 separate tiny PRs.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -22,6 +22,7 @@ class CatalogController < ApplicationController
   include BlacklightHelperReloadFix
 
   # These before_actions apply the hydra access controls
+  before_action :block_invalid_sort_params, only: :index
   before_action :enforce_show_permissions, only: :show
   before_action :load_home_page_collections, only: :index, if: proc { helpers.current_page? root_path }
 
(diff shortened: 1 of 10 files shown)


joncameron:
We'll review and pull in after 8.2.1 release.

mbklein:
I just realized that I based this PR against `main` instead of `develop` and I can't change it now. I can either close it and recreate it, or someone with admin access to the repo can probably change the base.

PR #6927, opened 2026-07-30
Support static variable bitrate HLS manifests

mbklein:
When we migrated our older Avalon instance to v7, we had a lot of preexisting variable bitrate manifests (with Derivatives) already in the system. This PR allows Avalon to recognize when there's an existing `auto` quality derivative and redirect to its streaming URL in response to a request for an `auto` quality HLS manifest, or fall back to building one on the fly if there isn't one. It also makes two small changes to other behaviors when requesting single quality manifests:

- Instead of building a one-off manifest containing the requested quality, it simply redirects to it
- If the requested quality doesn't exist, it returns a `404 Not Found` instead of delivering an empty manifest


cjcolvar (approved):
This looks good to me as far as I can tell!  Thanks for the contribution!

Issue #4401, opened 2021-02-17
Unsafe Inlining


joncameron:
### Description 
There are some directives with unsafe inlining that are problematic from a security standpoint. Where are these, and can we move away from unsafe inlining?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src
https://content-security-policy.com/unsafe-inline/

### User Report
I’ve discovered I cannot remove the ‘unsafe-inline’ keyword from the script-src and style-src directives without breaking Avalon.  Has this been updated in a later version of Avalon, or are there plans to separate scripts and styles from the code?

### Notes
Should be straightforward to add these headers to Avalon itself.

### Done Looks Like
- CSP headers changed to more secure default for Avalon
- Tickets created to fix any errors or broken pages as discovered

joncameron:
Theoretically still relevant. Approaches could be pulling inline scripts out of the HTML and into asset files (significant amount of work), or research in other ways to force headers from Avalon to remove any security issues.

joncameron:
👍

PR #6928, opened 2026-08-03
Provide a default CSP policy and add nonces to our JS

masaball:
Related issue: #4401

Avalon previously has required implementers to keep script and style directives in CSPs set to `unsafe-inline`. This commit adds a nonce value to all Javascript to allow at least script-src to be locked down tighter. Unfortunately, there are A LOT of inline styles throughout the codebase and some of our dependencies so locking down style-src is still not tenable at this moment. This work should provide better baseline security and provides implementers a basic default configuration that they can tighten or loosen as needed.

Manual testing performed in most recent Firefox and Chrome, currently untested in Safari

Tested working with this policy:
* File, Dropbox, and sharepoint uploading
* Derivative download from Minio
* Media Object, Playlist, Timeline, and Embed playback
* API requests
* Catalog index and search

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -68,7 +68,7 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     if login_popup?
       flash[:success] = nil
-      render inline: '<html><head><script>window.close();</script></head><body></body><html>'.html_safe
+      render inline: "<html><head><script nonce='#{content_security_policy_nonce}'>window.close();</script></head><body></body><html>".html_safe
     else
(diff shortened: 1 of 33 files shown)


cjcolvar (approved):
This looks good as far as I can tell.  Nice job on figuring this all out!  I'm learning about CSP through this PR.

As a follow up what do you think about using `<%= javascript_tag nonce: true do -%>` instead of having explicit script tags using `nonce="<%= content_security_policy_nonce %>"`?  This seems to be the approach used by the [rails guide](https://edgeguides.rubyonrails.org/security.html#adding-a-nonce).

masaball:
Yeah, I think that would be a worthwhile follow-up. Would be a little cleaner and more inline with Rails convention. I'll create a follow-up ticket.

PR #6934, opened 2026-08-04
Scope stream route to avoid conflict with other endpoints

### LINKED ISSUE #6930: Item page transcript search is non-functional
Transcript search seems to be completely broken on Avalon-dev. When attempting to search all that is returned is the no results message and there are JSON syntax errors in the console.

cjcolvar:
Resolves #6930 
Related to #6921 

This PR proposes fixing the routing conflict by making the stream endpoint be namespaced (e.g. `/master_file/<id>/high` would become `/master_file/<id>/stream/high`).  This does make it different from the existing hls manifest endpoint (e.g. `master_file/<id>/high.m3u8`).  I couldn't think of a different change that would be clear, flexible, and still retain the proximity to the hls manifest endpoint.

This route change should be safe to make since the stream route was added recently and hasn't made it into a release yet.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -175,7 +175,7 @@ Rails.application.routes.draw do
       get :captions
       get :waveform
       match ':quality.m3u8', to: 'master_files#hls_manifest', via: [:get], as: :hls_manifest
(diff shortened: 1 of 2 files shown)


masaball (approved):
Looks good

PR #6937, opened 2026-08-08
Fixes JS loading issue by modifying what get passed to babel-loader

cjcolvar:


masaball:
Would it be better to merge this to develop and then bring it forward into staging?

cjcolvar:
@masaball Good point. I'll do that.

PR #6945, opened 2026-08-12
Allow setting of solr timeouts via Settings or ENV

cjcolvar:
Also allow setting solr url at a Settings path similar to fedora


masaball (approved):
Looks good

Issue #6929, opened 2026-08-03
Remove inline styles


masaball:
Follow-up to #4401

Currently, Avalon has a lot of styles that are defined via HTML attribute. In order to facilitate more secure Content Security Protocols and containing all styling in a single directory, these inline styles should be removed from the HTML and added to an appropriate CSS file. This possibly includes pulling styling out of some javascript files that generate and inject HTML.

We should also test for if any of our dependencies are causing problems and try to find a way to mitigate it. Webpack supposedly can inject nonce values into packages as it bundles them, so hopefully we can just leverage built-ins.

PR #6946, opened 2026-08-12
Handle inline styles so we can tighten CSP style directive

masaball:
Related issue: #6929

Most important changes:
* Switch to `css-mini-extract-plugin` for (s)css loading. The default `style-loader` creates inline style elements, while the extract plugin extracts the css into a file that we can then pull in via `stylesheet_tag nonce: true`.
* Added some  hashes to the style directive. This is because Video.js dynamically creates a handful of style elements. Providing hashes for those elements seems more secure than other options but will potentially have some amount of upkeep associated with it when performing Video.js upgrades. If the hashes end up being inconsistent in future testing, we may be forced to leave `style-src` on `unsafe_inline`.


app/assets/stylesheets/avalon.scss
```diff
@@ -1313,3 +1326,207 @@ h5.card-title {
 .document-metadata {
   margin-bottom: 3rem;
 }
+
+/* De-inlining styles */
+// Styling for variations import option
+.varitions_import_form {
```
Dananji: This selector seems to be misspelled. The view `app/views/playlists/index.html.erb` has the class `variations_import_form`.

app/assets/stylesheets/avalon.scss
```diff
@@ -1313,3 +1326,207 @@ h5.card-title {
...
+  }
+
+  .title_original,
+  .start_time_original,
+  .end_time_original,
+  .comment_original {
+    display: none;
+  }
+
+  .clip_start, .clip-end {
```
Dananji: The `clip_start` selector doesn't match with the class name set in the view `app/views/playlists/_edit_form.html.erb`, which is `clip-start`.

app/assets/stylesheets/avalon.scss
```diff
@@ -1313,3 +1326,207 @@ h5.card-title {
...
+
+.new-link-template {
+  width: 15em;
+}
+
+.playlist-tags {
+  width: 100%;
+}
+
+.timelines-tags {
```
Dananji: The `timelines-tags` selector doesn't match with the class name set in the view template `app/views/timelines/_tag_form.html.erb`, which is `timeline-tags`.

app/views/playlists/_edit_form.html.erb
```diff
@@ -60,7 +60,7 @@ Unless required by applicable law or agreed to in writing, software distributed
             <ul class="dropdown-menu dropdown-menu-right">
               <% @playlists.each do |p| %>
               <li class="dropdown-item">
-                <a data-id="<%= p.id %>" style="text-decoration:underline; cursor:pointer"
+                <a data-id="<%= p.id %>"
```
Dananji, line 63: Nice! 🚀

app/views/bookmarks/update_access_control.html.erb
```diff
@@ -42,8 +42,7 @@ Unless required by applicable law or agreed to in writing, software distributed
               <p><%= t('blacklight.bookmarks.no_bookmarks') %></p>
             </div>
           <% else %>
-            <div class="alert" style="display: none;">
-            </div>
+            <div class="alert bookmark"/>
```
Dananji: The self-closing syntax doesn't apply to `<div>` elements because HTML5 doesn't treat it as a void element. So, it needs the explicit `</div>` at the end.
This renders an empty access-control modal in bookmarks.

Dananji (commented):
Kudos to undertaking this huge task! It all looks great 🎉 

I looked through almost all the pages in the app where changes were made, and the the code changes and it all looked good to me.

I noticed a couple of minor typos in the new CSS selectors, and a syntax error in HTML5 in the code.

The only major thing I came across is the timeline show page. The timeliner iframe on the page rendered fully un-styled for me. Could this be due to the changes in the Webpack config?

Image

app/assets/stylesheets/avalon.scss
masaball: Good catch! Ruby prefers snake case so it's what my hands automatically go to type.

app/views/bookmarks/update_access_control.html.erb
masaball: Right. I'd noticed the pattern elsewhere and meant to go back to here to check things but with all the other changes it slipped through.

masaball:
The timeliner page issues look to be from the CSP changes. `app/javascript/iiif-timeliner.js`  has a couple places that are setting an element's style directive using `Array.prototype.push()` or I guess is using `.push()` to build an array of html attributes that are then added to the element in an unsafe way? Although the code setting these styles appears to come from an inlined version of MediaElement.js, so the unminified code was maybe doing something else and that's just how it compiled. End result is the same though. 

These were missed because I was focused on the `views` directory and stuff generated by our javascript should have theoretically been fine, but I guess that file is doing things in a disallowed way.

@Dananji do you have any thoughts on what the best approach here may be? Looking at the console, it seems untenable to attempt allowing unsafe-hashes like we're going to attempt with Video.js, there are just too many of them. Trying to adjust the minified code in place seems like it could be finicky.

Dananji:
@masaball Yes I agree! Both unsafe-hashes and adjusting the minified code seem untenable. 

Since the error in the console suggests using either a hash or a nonce. I tried using a nonce, following the same approaches in this PR. And was not able to re-use the same exact code. However, I came across something else while doing some more digging into CSP and JS libraries/frameworks.

I came across posts in forums where people were pointing to MUI v3's use of JSS (JavaScript Style Sheets) which doesn't necessarily work well together with CSP.
And found a [guide](https://v3.mui.com/css-in-js/advanced/#what-is-csp-and-why-is-it-useful) for MUI v3 (the version Timeliner uses), which provides a solution. 
Since the server-side rendering doesn't apply here and the initial setup is already handled in your work in this PR, adding the `<meta>` to the `app/views/timelines/timeliner.html.erb` view template resolved the console error.
I added the meta tag to the `<head>` in the view template as follows re-using the already available nonce,
```
<meta property="csp-nonce" content="<%= content_security_policy_nonce %>" />
```
This seemed to work, and eliminated the inline style CSP violations.

To fix the broken styles, I added a reference to the new webpack-generated stylesheet with the `_webpack` suffix in the same view template,
```
<%= stylesheet_link_tag "iiif-timeliner_webpack", "data-turbo-track": "reload", nonce: true %>
```
And this fixed the styling rendering issue.

#### Questions and concerns with these solutions,
- As far as I understand, the nonce value in the Rails app is generated per session? If that's the case, the suggested soltuion will be re-using the same nonce value across multiple requests. And this doesn't sound very secure. Maybe the nonce value used to in the `<meta>` needs to be generated per request using the dynamic approach suggested in the MUI guide?
- With the `styleseet_link_tag` in place, can we remove the `import "./iiif-timeliner-styles.css";` from `app/javascript/iiif-timeliner.js` entirely?

masaball:
> With the styleseet_link_tag in place, can we remove the import "./iiif-timeliner-styles.css"; from app/javascript/iiif-timeliner.js entirely?

No, it can't be removed. The way webpack processes the files it does not actually know that the css file is there to be parsed and bundled unless the import statement is included.

> Maybe the nonce value used to in the <meta> needs to be generated per request using the dynamic approach suggested in the MUI guide?

I was going off [the rails guide](https://guides.rubyonrails.org/security.html#adding-a-nonce) which cautions about potential conflicts with GET caching and recommends using the session id if that is a worry. But looking at things, I do not think we are using the type of caching it could be problematic with, so I think changing the nonce to be generated per request would be fine for us and definitely more secure.

PR #6947, opened 2026-08-12
Avoid encoding with html entities when using rails truncate helper

cjcolvar:
Steps to Test:
- Create an object with an alternative title, a description, and a  with an apostrophe.
- Search for the object and make sure that the alternative title in the result does not have an html entity (`&#39;`)
- Go to the item view page and make sure both the alternative title and description do not have an html entity (`&#39;`)
- Also check with special characters in filenames on the Manage Files page as well as the Structure page


masaball (approved):
Good catch! Glad it was an easy fix. Encoding stuff can be a nightmare.

Issue #6936, opened 2026-08-06
Convert html `<script>` tags to `javascript_tag` blocks


masaball:
## Description
Follow-up to #6928

Rails provides a view helper for inlining javascript which we have not historically been making use of. Changing our convention would allow us to clean up our view code a little by using `<%= javascript_tag nonce: true do %>` instead of `<script nonce="<%= content_security_policy_nonce %>">`. This would also simplify what is needed to create new inline scripts. Additionally, there is a config option `Rails.application.config.content_security_policy_nonce_auto = true` that would potentially allow us to skip including the nonce argument entirely, allowing us to just use `<%= javascript_tag do %>. 

## Done Looks Like
- [x] All html `<script>` elements in our code converted to `javascript_tag`
- [x] Test that inline javascript is being generated and works as expected
- [ ] Test and make a decision on using Rails auto nonce config option

joncameron:
👍

PR #6949, opened 2026-08-12
Convert script elements to javascript_tag helpers

masaball:
Related issue: #6936

Utilizing the built in Rails helpers provides a much easier and more concise interface for providing nonce values to all of our inline scripts.


cjcolvar (approved):
Looks good!

PR #6950, opened 2026-08-12
Fix more issues with permission inheritance

### LINKED ISSUE #6943: [BUG] More issues with inherited visibility and non-active stored visibility
**Describe the bug**
If an item in a collection with default visibility of private has disabled inheritance and been set to public then inheritance reenabled it will have a stored visibility of public and an active_visibility of private.  This (or similar mismatch scenarios) leads to incorrect information being displayed:
- 'Item access' admin facet
- 'Access Restrictions' on item view page
- Collection browse shows collections with no items available to users

**To Reproduce**
1. Login as an admin
2. Create a collection with default visibility of private
3. Create an item
4. Disable inheritance, set to public visibility and save the item
5. Re-enable inheritance and save the item
6. Go to Preview
7. Notice it says the item is available to the public in 'Access Restrictions'
8. Search for the item
9. Notice it appears as public in the item access admin facet count
10. Log out
11. Go to Collections
12. Find the collection created in step 2 and click into it
13. Notice it does not display any items

**Expected behavior**

'Item access' facet and 'Access Restrictions' on item view page both report the active visibility.  Collection browse does not show collections with no accessible items and collection landing pages with no accessible items shows Restricted Content view with login button if applicable.

**Environment (please complete the following information):** - Device / platform [e.g. laptop, iphone pro 12] - OS: [e.g. iOS]- Browser [e.g. chrome version 90.0.4430.93] [X] avalon-dev
- [X] avalon-staging
- [X] mco-staging
- [ ] mco-production
- [?] demo

**Release:**
Introduced in Avalon 8.2

cjcolvar:
Resolves #6943 

This PR has three commits for each of the issues in #6943.
- Show page / manifest text change is a straightforward switch to `active_visibility` in `#access_text`
- Item access facet required adding joins to the queries used by the query facet
- Collection browsing and landing pages were fixed by pulling in all of the changes in SearchBuilder by subclassing it and simplifying to only have the important parts: only_wanted_models changed to Admin::Collection and all of the gated discovery from SearchBuilder wrapped in a join to find collections


Dananji:
Since I couldn't understand the code changes 100%, I tested the changes according to the ticket :)

I followed the steps in the ticket and I was able to see the expected behaviors 🎉 
- 'Item access' facet and 'Access Restrictions' on item view page both reported the active visibility (as an admin)
- Collection browse did not show collections with no accessible items (as a normal user)
- Collection landing page with no accessible items showed the Restricted Content view (as a normal user)

However, I noticed an issue (I deemed) regarding the browse page. After following steps 1 through 4 in the ticket, I logged out and visited the Browse page. And I was not able to see the item that was created and set to be public in the previous steps in the Browse page.
Here are the exact steps I followed:
1. Log in as an admin
2. Create a collection with the default visibility set to private (Collection staff only)
3. Create an item
4. Disable inheritance, set to public visibility and save the item
5. Log out
6. Navigate to the Browse page

Initially, I thought this might be expected behavior since the parent collection was still set to private. So, I updated the collection's visibility to public, repeated steps 5 and 6 above, and the item was still not visible on the Browse page. And the collection did not appear in the Collection browse page.

Dananji (approved):
Wait, I did not publish the item! Apologies!

The changes work 🎉 🚀

PR #6952, opened 2026-08-18
Item access facet fixes

cjcolvar:
Fixes issues found with #6950 

- Fix facet counts: Existing items will not have a value for disable_inheritance so change queries to NOT true instead of explicit false.
- Condense two join queries into one using terms query parser for better performance.
- Do not include facets in solr requests that do not pass if/unless tests for better performance.


Dananji (approved):
🎉

Issue #6031, opened 2024-09-12
Add LTI 1.3 Strategy to Omniauth


joncameron:
# Description
Continuing from the investigation into adding LTI 1.3 support to Avalon in #5980, the decision was made to add an Omniauth strategy to support the currently supported version of the API standard. Avalon implementers can then continue with the existing LTI 1.1 strategy or switch to the new one for LTI 1.3.

This is not a requirement for implementation in 8.0 as support for LTI 1.1 is expected to continue in the foreseeable future.

# Done Looks Like
- [ ] Add a new LTI 1.3 strategy that can be configured alongside the existing :lti strategy
- [ ] Documentation adding for configuring the new LTI 1.3 strategy in an Avalon instance

joncameron:
@masaball to add some details to Done Looks Like based on implementation notes. There are still some details to fill out regarding adding LTI 1.3 support.

joncameron:
Could specify in the auth provider setup which strategy to use, 1.1 or 1.3—we could provide that, or do a fallback approach where LTI 1.1 would be used if 1.3 responses were not available.

To start we could focus on 1.3 as the new default, and then add handling for 1.1 support if needed.

cjcolvar:
Possibly helpful resources:
https://github.com/1EdTech/ltibootcamp
https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.lti_launch_overview
https://blackboard.github.io/lti/tutorials/implementation-guide

From a quick glance at our old tickets and some of these resources I think it is probably best to start developing a new omniauth gem to be configured as another option in avalon.  Unfortunately it doesn't look like `ims-lti`or any other gem supports the LTI 1.3 launch flow so we'll probably have to code it all ourselves.

PR #6954, opened 2026-08-18
Add LTI 1.3 support

mohideen:
## Summary

Wires LTI 1.3 into Avalon's existing Devise/OmniAuth stack alongside the existing LTI 1.1 integration. A single `:lti` provider now supports either protocol, selected by a new `protocol:` config key (default `'1.1'` for backward compatibility with existing installs). Built and verified against a real local Moodle instance, not just specs — that process surfaced two boot-order bugs and a security issue that are included here.

## What changed

**Gemfile / provider wiring** (`config/initializers/devise.rb`)
`omniauth-lti13` added alongside the existing `omniauth-lti` (1.1) gem — currently `git:`-sourced from a personal fork (`mohideen/omniauth-lti13`) pending an org repo; swap before merging for real. `devise.rb` picks `OmniAuth::Strategies::Lti13` vs `::Lti` at registration time based on `protocol:`, passing `strategy_class:` explicitly on both branches rather than relying on Devise's implicit name-based lookup (which becomes ambiguous once two `Lti*`-named strategies are loaded). The 1.1-only `consumers:` merge (from `lti.yml`) is skipped for 1.3, whose platform registration lives entirely in `settings.yml` instead — the two config schemas don't overlap.

**Two boot-order bugs, found live against Moodle, not caught by specs:**
- `devise.rb` wrapped provider registration in `Rails.application.reloader.to_prepare`, which runs *after* Devise's own middleware-insertion initializer — so the `:lti` strategy was correctly registered in `Devise.omniauth_configs` but never actually landed in the request-serving middleware stack. Launches silently fell through to a generic form-render action instead of being intercepted. Unwrapped (matches upstream fix `afa7eb51c`, already on `origin/staging`).
- `OmniAuth::Strategy#request_path`/`#callback_path` memoize themselves from `OmniAuth.config.path_prefix` on whichever request first reaches them; that value is otherwise only set correctly as a side effect of Devise drawing routes, which are lazy-loaded — so literally the first request a fresh process ever served got stuck with the wrong path for its whole lifetime, while every later request worked fine (which is why it passed in specs but failed on a live cold boot). Fixed by setting `path_prefix` directly instead of depending on route-draw timing.

**Controller** (`app/controllers/users/omniauth_callbacks_controller.rb`)
`after_omniauth_failure_path_for` branches on the failed strategy to find the right "return to LMS" param (1.1's `launch_presentation_return_url` vs 1.3's `target_link_uri`). Originally redirected there unvalidated — an **open redirect**: an attacker could POST a registered `iss` with an off-host return URL and a launch engineered to fail, sending an unauthenticated victim off-site with the OmniAuth failure message (which can include the registered `client_id`/issuer) attached as a query param. Fixed with the same same-host guard `ApplicationController#find_redirect_url` already uses (VOV-5662); also rescues `Addressable::URI::InvalidURIError` on malformed input (was an uncaught 500) in both places. Extracted into a private `lti_failure_return_uri` for readability.

`find_user` no longer appends a `nil` LTI context id to `virtual_groups` (the context claim is optional — platform-level launches outside a course omit it).

**Model** (`app/models/user.rb`)
`find_for_lti` now raises `Avalon::MissingUserEmail` (handled like the existing `MissingUserId`, with an operator-facing flash message) when the LMS omits the email claim — both Moodle and Canvas support withholding it for privacy. Previously this silently produced an unpersisted `User` and bounced the user with zero explanation.

**Cross-cutting fixes**
- **CSRF**: LTI's third-party-initiated login arrives as a token-less cross-site POST from the LMS. `OmniAuth.config.request_validation_phase` (a single global callable, no per-strategy option in the gem) is overridden to skip the default check only for `:lti`, folded into `devise.rb` next to its other `OmniAuth.config` globals.
- **SameSite**: investigated and found Avalon's session cookie (via `activerecord-session_store`) already gets `SameSite=None` correctly regardless of `cookies_same_site_protection` — that Rails setting only governs `ActionDispatch::Cookies`-managed cookies and the built-in `:cookie_store`, neither in play here. A proposed fix that did nothing was caught and reverted before merge; a regression test (Chrome UA over HTTPS, matching production conditions) stands in its place.

**`config/settings.yml`**: adds a documented, commented LTI 1.3 example block matching the gem's actual `OmniAuth::Lti13::Platform` schema (`deployment_ids` — plural, quoted strings; unquoted YAML integers silently fail to match the token's string claim).

## Tests

61 examples across `spec/models/user_spec.rb`, `spec/controllers/omniauth_callbacks_controller_spec.rb`, `spec/config/omniauth_spec.rb`, `spec/config/omniauth_middleware_spec.rb`, `spec/requests/samesite_cookie_spec.rb`. Every new assertion added across two rounds of review was individually verified to actually catch a regression — introduced the underlying bug, confirmed the relevant spec failed, restored the fix.

## Manually verified end-to-end

Full launch flow against a local Moodle instance (login → JWKS fetch/token verify → deployment match → callback → catalog redirect filtered by course group), including cold-boot behavior on freshly recreated containers.

See #6031


mohideen:
The implementation currently pulls the gem from my personal Github: https://github.com/mohideen/omniauth-lti13. We can change it once we have the Gem in the correct Github account.

cjcolvar (commented):
I did a quick look through the commits but will need some more time to dig into a few particular changes since there is a lot in this PR.

I switched the target branch of this PR to `develop` (from `main`) but that left the PR with conflicts.  After reading through the changes I noticed that some of the fixes were already put in `develop` (specifically the devise initializer unwrapping).   The `develop` branch also has an updated `sqlite3` dependency so it will probably clear up the circleci build issue.  Can you please rebase this PR on `develop`?

mohideen:
@cjcolvar, I have rebased the PR against the `develop` branch, and the tests as passing now.

PR #6955, opened 2026-08-20
Faster solr queries

cjcolvar:
To address slowness due to large join queries


masaball (approved):
Looks good as far as I can tell. Really interesting that the graph query syntax has the `to...from` clauses opposite to the join queries.

cjcolvar:
I think it allows reversing the direction of the relationship so that it can start with a smaller set of documents (matching collections) to then traverse and grow the set instead of starting with a really large set (all media objects) and then narrow.  Or at least that's how I've been thinking about it.

PR #6959, opened 2026-08-24
Authorize GET hls_manifest via Bearer auth token in request

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/978

In the current implementation, the `hls_manifest` GET requests are authenticated using CanCan. And as far as I understand, this is not available for requests generated from outside of Avalon.

Therefore, to allow external IIIF clients to use IIIF auth flow to request and access restricted resources, the `hls_manifest` endpoint in Avalon should facilitate GET requests to be authenticated via the token method. 

Suggested changes in this PR:
- use auth token to authorize GET `hls_manifest` requests
- add `/iiif_auth_probe/*` endpoint to CORS allowlist

Relevant work in Ramp: https://github.com/samvera-labs/ramp/pull/986


cjcolvar (dismissed):
I think this change means that an avalon session cookie isn't required anymore for GET requests but I think that is okay since it matches the same requirements of a valid token as the streaming server.  Thanks for fixing this!

Dananji:
Closing in favor of https://github.com/avalonmediasystem/avalon/pull/6971

PR #6962, opened 2026-08-27
Redirect master_files#transcript to supplemental_files#show

cjcolvar:
It appears that the transcript endpoint isn't referenced anymore and is purely for backwards compatibility. Instead of maintaining two implementations turn the route into a redirect.

I'm hoping to get this into 8.2.1 if possible.

### DIFF
--- a/spec/requests/master_files_spec.rb
+++ b/spec/requests/master_files_spec.rb
@@ -1,0 +1,51 @@
+# Copyright 2011-2026, The Trustees of Indiana University and Northwestern
+#   University.  Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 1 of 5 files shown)


masaball (dismissed):
Looks good to me. Definitely agree implementing a redirect is better than keeping two implementations floating around.

masaball (approved):
👍

spec/requests/master_files_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+require 'rails_helper'
+
+describe '/master_files/' do
+  describe "#transcript" do
+    let(:user) { FactoryBot.create(:administrator) }
+    let(:master_file) { FactoryBot.create(:master_file, supplemental_files: [supplemental_file]) }
+    let(:supplemental_file) { FactoryBot.create(:supplemental_file, :with_transcript_file, :with_transcript_tag, label: 'transcript') }
+
+    before do
+      allow(Settings.supplemental_files).to receive(:proxy).and_return(true)
```
cjcolvar, line 24: The active_storage routes were returning 404 so instead of trying to solve that problem setting `Settings.supplemental_files.proxy = true` avoids those routes.  This also more closely matches how these tests were running in the MasterFilesController specs before.

PR #6965, opened 2026-09-02
[8.2.1 RC3] Restrict browse-everything routes to users who are members of a collection or unit

cjcolvar:


masaball (approved):
Looks good!

PR #6966, opened 2026-09-03
Fix underscoring of keys in MediaConvert presets

mbklein:
The ActiveSupport `String#underscore` method turns `M3u8Settings` into `m3u8_settings`, while `Aws::MediaConvert::Client#create_preset` expects `m3u_8_settings`. The same holds true for other instances where a lowercase letter is followed by a digit - the SDK expects `vc_3_settings`, `vp_8_settings`, etc., but the default underscoring changes them into `vc3_settings` and `vp8_settings`.

My temporary workaround was to change the key to `M3u_8Settings` in my preset JSON, but this fix will make sure Avalon can convert an unmodified preset export into a valid SDK call.


cjcolvar (approved):
Interesting.  I guess we never hit this with the simple presets that avalon ships.

Issue #6960, opened 2026-08-27
[BUG] Infinite loop in hls manifest request cycle


masaball:
**Describe the bug**
There is an infinite loop occurring when trying to request the `auto.m3u8` hls manifest

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Open the browser developer tools
2. Navigate to the network tab in the dev tools
3. Go to any media object page
4. Select `auto` from the quality menu on the player 
5. Observe in the network tab that the `auto.m3u8` request is made 10-20 times before the browser terminates the loop
6. Ramp falls back to the `high.m3u8` manifest and playback can occur

**Expected behavior**
`auto.m3u8` request is only made once and redirects to the appropriate quality manifest

**Screenshots/Screen Recording**
<img width="962" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHZhRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--19c741cdf089e255155161c51edc3b34163ad6e6/Screenshot_20260827_100133.png" alt="Screenshot_20260827_100133.png" />

**Environment (please complete the following information):**
 - Laptop
 - OS: Ubuntu
 - Browser: Firefox 154.0
- [x] avalon-dev
- [ ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

**Additional notes**
This looks to be related to [these changes](https://github.com/avalonmediasystem/avalon/commit/15a322b4f57c48f7e2d71eb4e7ab61d2557474b5)

PR #6978, opened 2026-09-17
Fix infinite loop in hls manifest request

masaball:
Related issue: #6960

In #6927 support was added for static `auto` derivatives/manifests to be served instead of generating the manifest on the fly. These changes caused an infinite loop in the hls manifest request cycle when multiple derivatives are present. This is because the [`MasterFileBehavior#stream_details`](https://github.com/avalonmediasystem/avalon/blob/05a663ecf344de050320efd52dbb658d880234ea/app/models/concerns/master_file_behavior.rb#L34) method adds an entry for `auto` quality when there is more than one derivative. Thus `hls_streams(..., quality: 'auto')` would not return nil and subsequently not go through the fall back route that dynamically generates the manifest. Without the `auto` derivative present, redirecting to the `auto.m3u8` endpoint would request the `hls_manifest` method again resulting in an infinite redirect loop.

Adding a guard clause that checks for the existence of an `auto` quality derivative should hopefully resolve the infinite loop while maintaining the ability to serve static variable bitrate manifests.


mbklein (approved):
Works well with our data.

cjcolvar (approved):
Looks good to me!

Issue #6944, opened 2026-08-11
[BUG] Collection items endpoint does not scale even when using pagination


cjcolvar:
**Describe the bug**
When the number of items in a collection increases so does the number of fedora requests when returning results for `collections/<id>/items.json`. Even with a `per_page` of 1 a media object will get fetched a number of times equal to the number of items in collection divided by 1000.

[Reported by Kyle Ratliff on slack](https://samvera.slack.com/archives/C1C3C4F5L/p1786393619376979)

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. Login in as an administrator
2. Go to '/admin/collections/<id>/items.json' for a collection with a number of items
3. See it take a _long_ time

**Expected behavior**
The JSON endpoint does not call fedora and is performant.

joncameron:
Other paginated API routes are probably suffering the same performance issues; we may need to redo the pagination and endpoint as part of this to make sure it's performant.

PR #6980, opened 2026-09-22
Use solr for collection and unit's `items` endpoints

masaball:
Related issue: #6944


spec/controllers/admin_collections_controller_spec.rb
```diff
@@ -238,6 +238,17 @@
...
     end
+
+    context 'read from solr' do
+      it 'should not read from fedora' do
+        collection = FactoryBot.create(:collection, items: 5)
+        WebMock.reset_executed_requests!
+        get :index
+        get :items, params: { id: collection.id, format: 'json', per_page: '2' }
+        expect(a_request(:post, /#{ActiveFedora.solr.conn.uri.to_s}/)).to have_been_made.at_least_once
+        expect(a_request(:post, /#{ActiveFedora.solr.conn.uri.to_s}.*\?fl=id&q=/)).not_to have_been_made
```
cjcolvar: I tested this out locally and it looks like the MediaObject proxy is reifying due to `avalon_resource_type` being called.  Since the proxies in the test don't have any sections this value doesn't get set and stored in solr leading to the reification.  If we default avalon_resource_type to nil in `config/initializers/presenter_config.rb` then I think it won't reify and this test can be changed to the stricter version used in the admin units controller spec.

This helped chase down another edge case where reification is happening so it is really helpful in our quest to avoid calling fedora from the front end!

cjcolvar (commented):
This looks good!  I would have expected that the switch to speedyaf proxies would have caused problems with `paginate` so glad that it was easy and straight-forward!

